PR #165, opened 2015-01-11
add checks to prevent panic when parsing zone text.

xofyarg:
Empty text or comments in zone format will cause panics. Also add a description.

### DIFF
--- a/zscan.go
+++ b/zscan.go
@@ -102,12 +102,13 @@ type Token struct {
 	Comment string      // a potential comment positioned after the RR and on the same line
 }
 
-// NewRR reads the RR contained in the string s. Only the first RR is returned.
-// The class defaults to IN and TTL defaults to 3600. The full zone file
-// syntax like $TTL, $ORIGIN, etc. is supported.
-// All fields of the returned RR are set, except RR.Header().Rdlength which is set to 0.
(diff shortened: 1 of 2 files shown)


zscan.go
```diff
@@ -102,11 +102,15 @@ type Token struct {
...
+// NewRR reads the RR contained in the string s. Only the first RR is
+// returned. If s contains no RR, return nil with no error. The class
+// defaults to IN and TTL defaults to 3600. The full zone file syntax
+// like $TTL, $ORIGIN, etc. is supported. All fields of the returned
+// RR are set, except RR.Header().Rdlength which is set to 0.
 func NewRR(s string) (RR, error) {
+	if s == "" {
+		s = "\n"
+	}
 	if s[len(s)-1] != '\n' { // We need a closing newline
```
miekg: Why not add `len(s) > 0 && s[len(s)-1] != '\n' { ...` here?

miekg:
thanks for spotting this. Can you also add a little test case?

zscan.go
xofyarg: It should be. I was thinking just 'add' something to your code. My fault.

xofyarg:
The commit has been modified, some test cases have been added.

PR #168, opened 2015-01-14
client: c.exchange: receive multiple messages and combine into one

### LINKED ISSUE #167: c.Exchange does not support long responses like AXFR
Hi there, cool library! I'm using it to build a "reverse proxy" that redirects requests to one server or another depending on the domain asked.
It works well, but I'm having an issue with AXFR: it looks like [c.Exchange](https://github.com/miekg/dns/blob/master/client.go#L156) only reads one message, not the whole thing?

To reproduce, sample proxy code:
```
package main

import (
        "log"

        "github.com/miekg/dns"
)

func main() {
        server := &dns.Server{Addr: ":53", Net: "tcp"}
        dns.HandleFunc(".", proxy)
        log.Fatal(server.ListenAndServe())
}

func proxy(w dns.ResponseWriter, req *dns.Msg) {
        c := &dns.Client{Net: "tcp"}
        resp, _, err := c.Exchange(req, "37.187.31.39:53")
        if err != nil {
                log.Println(w.RemoteAddr(), err)
                dns.HandleFailed(w, req)
                return
        }
        w.WriteMsg(resp)
}
```
Update IP with an existing DNS server that allows AXFR for a domain, then:
```
$ dig axfr stalkr.net @localhost +tcp
; <<>> DiG 9.9.5-6-Debian <<>> AXFR stalkr.net @localhost +tcp
;; global options: +cmd
stalkr.net.             3600    IN      SOA     ns0.stalkr.net. hostmaster.stalkr.net. 2015001401 3600 1800 604800 86400
stalkr.net.             3600    IN      RRSIG   SOA 8 2 3600 20150122000000 20150108000000 61206 stalkr.net. rhCGJDljnE9DVU2eVYxp7+JKxArEkJ567ydX0zDXOPS5KnTKfscO+qT1 2TDzm8Dy8jHHXTJKjdj34k4fRB7NtpByIXcTiv7g/Kf2mf33PCOBy8ti /QbHBH0ugLht6Onn+1XV4aCWWDBwL6xeLTF+7Q4UtayM50/p4YI1gV6I D6M=
;; communications error to 37.187.31.39#53: end of file
```

I'm explicitly using tcp to rule out a possible issue with udp/edns0/max message size.

StalkR:
We combine the answers into a single *Msg because `c.Exchange` API returns a single *Msg. For exactness with the server, it would be better if we could return the list of *Msg received.

Alternatively `*Client` could expose the underlying `*Conn` and the caller could call `co.ReadMsg` itself if it wants more than the first message replied by `c.Exchange`. Unfortunately, `c.exchange` [closes](https://github.com/miekg/dns/blob/master/client.go#L142) the conn after use, so we can't.

I did not modify the other `Exchange` functions, I don't know if you wish to do so. This can be done in a separate change anyway.

Fixes #167

### DIFF
--- a/client.go
+++ b/client.go
@@ -154,7 +154,18 @@ func (c *Client) exchange(m *Msg, a string) (r *Msg, rtt time.Duration, err erro
 		return nil, 0, err
 	}
 	r, err = co.ReadMsg()
-	return r, co.rtt, err
+	rtt = co.rtt
+	for {
+		p, err := co.ReadMsg()
+		if err != nil {
+			break
+		}
+		r.Answer = append(r.Answer, p.Answer...)
+		r.Ns = append(r.Ns, p.Ns...)
+		r.Extra = append(r.Extra, p.Extra...)
+		rtt += co.rtt
+	}
+	return r, rtt, err
 }
 
 // ReadMsg reads a message from the connection co.


miekg:
```
co := &dns.Conn{Conn: c} // c is your net.Conn
co.WriteMsg(m)
in, _  := co.ReadMsg()
co.Close()
```

And re-create a dns.Msg after the fact?

If you need this for AXFR, then this is not the right approach, but I'll comment on that issue.

StalkR:
I don't understand what you suggest - I'm using a `dns.Client` which doesn't expose `net.Conn`?

PR #171, opened 2015-01-15
xfr: Transfer.Out use a new Msg every time

StalkR:
I thought about your comment, indeed there seems to be a simpler fix. I don't know if that's what you had in mind, but it solves the issue I observed.

### DIFF
--- a/xfr.go
+++ b/xfr.go
@@ -160,12 +160,11 @@ func (t *Transfer) inIxfr(id uint16, c chan *Envelope) {
 // The server is responsible for sending the correct sequence of RRs through the
 // channel ch.
 func (t *Transfer) Out(w ResponseWriter, q *Msg, ch chan *Envelope) error {
-	r := new(Msg)
-	// Compress?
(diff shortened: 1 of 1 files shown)


xfr.go
```diff
@@ -160,20 +160,18 @@ func (t *Transfer) inIxfr(id uint16, c chan *Envelope) {
 // The server is responsible for sending the correct sequence of RRs through the
 // channel ch.
 func (t *Transfer) Out(w ResponseWriter, q *Msg, ch chan *Envelope) error {
-	r := new(Msg)
-	// Compress?
-	r.SetReply(q)
-	r.Authoritative = true
-
+	w.TsigTimersOnly(true)
```
miekg: This needs to happen after the first set the RRs have been written, because only then they become used. (have to reread the spec and actually test for it with tsig to see if I'm right here). Just keep it where is was.

StalkR:
Done, I left it where it was.

PR #172, opened 2015-01-15
minor clean ups: use time.Second instead of 1e9

StalkR:
Also do not reference `ns` in comments: it is an implementation detail of `time.Duration`, users should use time constants (`time.Nanosecond`, `time.Microsecond`, `time.Millisecond`, `time.Second`, etc.)

Just aesthetic :)

### DIFF
--- a/client.go
+++ b/client.go
@@ -9,7 +9,7 @@ import (
 	"time"
 )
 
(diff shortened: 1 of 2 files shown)


miekg:
agreed

PR #173, opened 2015-01-23
PrivateKey refactor

FiloSottile:
This is a pretty big one. What we want is to make signatures with your library, but now necessarily passing a private key, just a signing function.

How I tentatively implemented that is by making PrivateKey the following interface 

```go
type PrivateKey interface {
	Sign(hashed []byte, alg uint8) ([]byte, error)
	String(alg uint8) string
}
```

The usual {rsa|dsa|ecdsa}.PrivateKey are now wrapped in a PrivateKey, and Sign and String are implemented for them. But if you want to define your own PrivateKey that uses for example PKCS#11 for Sign, you can and then use it with RRSIG.Sign().

What I like of this is that it moves the signing and exporting code and switches out of the logic.

I should not be breaking any interface, and all code that did not inspect PrivateKey should keep working.

I also fixed a couple critical bugs in RSA key export.


miekg:
Looks good. I will go over it in some more detail.

Good you split of the moves from keygen.go -> dnssec_keygen.go and friends in a separate PR?

FiloSottile:
Just split the PR into two commits, the first is a minimized diff for review: https://github.com/FiloSottile/dns/commit/e9faa971b320961a208a7ab53c7805d4f280f809

The second is strictly code moved from place to place.

The resulting changeset is the same.

PR #179, opened 2015-02-05
Used shared backing array for Msg.Copy

DanielMorsing:
If you have a system with large amounts of copies, these slice
allocations start stacking up. Use a shared slice and then subslice
them with a cap limit so that append works properly.

Also, add a benchmark and test for Msg.Copy

Benchcmp:
benchmark         old ns/op     new ns/op     delta
BenchmarkCopy     1880          1672          -11.06%

benchmark         old allocs     new allocs     delta
BenchmarkCopy     13             11             -15.38%

benchmark         old bytes     new bytes     delta
BenchmarkCopy     528           528           +0.00%

### DIFF
--- a/dns_test.go
+++ b/dns_test.go
@@ -510,6 +510,48 @@ func TestCopy(t *testing.T) {
 	}
 }
 
+func TestMsgCopy(t *testing.T) {
+	m := new(Msg)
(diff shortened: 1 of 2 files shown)


miekg:
Thanks!

PR #186, opened 2015-02-24
Convert tests from being t.Log(..) then t.Fail() to just t.Error(...)

michaelharo:
Convert tests from being t.Log(..) then t.Fail() to just t.Error(...) as t.Error(...) does both and makes it more clear which messages are errors vs information log messages.

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -21,24 +21,20 @@ func TestClientSync(t *testing.T) {
 	c := new(Client)
 	r, _, e := c.Exchange(m, addrstr)
 	if e != nil {
-		t.Logf("failed to exchange: %s", e.Error())
-		t.Fail()
+		t.Errorf("failed to exchange: %s", e.Error())
 	}
 	if r != nil && r.Rcode != RcodeSuccess {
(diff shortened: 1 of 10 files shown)


client_test.go
```diff
@@ -21,24 +21,20 @@ func TestClientSync(t *testing.T) {
...
-		t.Logf("failed to exchange: %s", e.Error())
-		t.Fail()
+		t.Errorf("failed to exchange: %s", e.Error())
 	}
 	if r != nil && r.Rcode != RcodeSuccess {
-		t.Log("failed to get an valid answer")
-		t.Fail()
-		t.Logf("%v\n", r)
+		t.Error("failed to get an valid answer")
+		t.Errorf("%v\n", r)
```
miekg: Should this not just be on `Errorf` instead of an `Error` *and* `Errorf`?

michaelharo: There are a lot of cases where log lines could be combined or restructured.  I opted to leave that out of scope for this round of changes.

miekg:
Ack.

Issue #180, opened 2015-02-06
TSig check on second DNS message fails (TCP)


flebarbe:
I'm issuing an IXFR query using the lib and a TCP connection.
In my case, the server answers with several DNS messages (about 5000 RR total, each DNS message has about 1300 RR). TSigVerify returns ok on the first DNS message. Starting with the second DNS message, it fails because the received MAC and the computed one do not match.

This is implemented in tsig.go.
I had a look at RFC 2845 and the related section "4.4. TSIG on TCP connection".
Looking at the code, I'm not sure the correct elements are included in the byte stream that is hashed. Has anyone ever had success with this ?

miekg:
There could be bugs lurking here... Yes. Can I publicly AXFR from your
server to test?

miekg:
Ah, IXFR and TCP...Hmmm.

flebarbe:
The server is public, but the key I use is not. :)

andrewtj:
Had a quick look at this. Changing `(t *Transfer) ReadMsg()` to update `t.tsigRequestMAC` with `ts.MAC` following the call to `TsigVerify` allows a successful multi-envelope AXFR from a BIND test install. I'd need to revisit the RFC and find out what the behaviour of servers in the wild is before calling this a fix though as if memory serves a TSIG on every envelope isn't required (just first and last) with intermediary ones optionally being concatenated.

miekg:
Cool.

tsig is required on each envelop except that you only check the running MAC, i.e. not the full message (AFAIK)

miekg:
Can this be closed now?

flebarbe:
yes.

PR #187, opened 2015-02-24
Fix TSig check on second DNS message (TCP) #180

flebarbe:
The next chunk needs to use the previous MAC
Using this fix, I can successfully verify the signatures of not
only the first but the subsequent envelopes as well.

Patch was proposed by andrewtj in a comment. Kudos !


miekg:
Awesome! :-)

PR #188, opened 2015-02-24
If the Matching Type is 1 or 2 we need to return a hash.

fobser:


miekg:
But am I reading https://tools.ietf.org/html/rfc6698#section-2.1.3 wrong then?

fobser:
probably :)
e.g.:
_443._tcp.nl-ams-as3333.anchors.atlas.ripe.net. 21600 IN TLSA 1 0 1 88422D55424CA8F6F74E165016851CB195FC0919F82F8762574A4D71 868964E9

the selector is 0 == Full certificate and the Matching Type is 1, so sha256 hash over the selected content

PR #190, opened 2015-02-26
More test clean up

michaelharo:
Remove trailing \n from t.Log and t.Error messages as it's unnecessary.

In some instances, combine multiple t.Error()s into one

To provide more consistency across the tests, rename e to err and use %v as the format arg for errors.

Replace Logf and Errorf with Log and Error when it made sense.  For example t.Errorf("%v", err) to t.Error(err)

Run gofmt -s over the code to simplify it.


miekg:
Could you fixed the go1.3 compile failure?

The gofmt -s makes sense, but could you separate those out in another PR?

michaelharo:
gofmt changes reverted

PR #193, opened 2015-03-17
RRSIG.Sign: remove duplicate before signing records as per RFC

FiloSottile:
Signatures computed on duplicate records are actually considered BOGUS.


FiloSottile:
Ah, to address your concerns regarding letting people who don't like beer commit ;)

image

miekg:
That's *holding* a beer. But fair enough: merged.

DanielMorsing:
I can confirm that the beer was drunk.

DanielMorsing:
but not filippo, it was only one beer.

miekg:
+1 :-)

PR #194, opened 2015-03-17
Add support for custom EDNS options in GoDNS.

reyzell:
A handful of EDNS options have been standardized, and they each have a type defined in GoDNS.  However there is currently no way a development team can use GoDNS with internally defined options, or with new options that may be proposed in the future.

This change solves the problem by giving users an EDNS0_CUSTOM type to allow clients to send, and servers to receive, custom EDNS options.

See client_test.go for example usage.

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -1,6 +1,7 @@
 package dns
 
 import (
+	"strconv"
 	"testing"
 	"time"
 )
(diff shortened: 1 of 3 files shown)


miekg:
Looks OK. But I would like to rename some of the name and use the reserved values for this (http://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#dns-parameters-11)
```
65001-65534	Reserved for Local/Experimental Use
```
So EDNS0_CUSTOM -> EDNS0_LOCAL 
And defined an option code EDNS0LOCAL that starts at 65001 (with some docs)

reyzell:
> And defined an option code EDNS0LOCAL that starts at 65001 (with some docs)

It's not clear to me how the library or consumers will use the option code EDNS0LOCAL.

Perhaps you're thinking that if the option code is in the range 65001-65534 then unpackStructValue() should unpack it as a EDNS0_LOCAL type?  Or are you thinking that the EDNS0LOCAL code of 65001 will serve only as documentation for users, and that the library will unpack any unknown option (regardless of code) as EDNS0_LOCAL?

The second choice makes sense to me.  I'm working with systems that are sending custom options with codes less than 65000.  The second choice also allows consumers to read options defined in the future without upgrading their version of this DNS library.

miekg:
[ Quoting <notifications@github.com> in "Re: [dns] Add support for custom ED..." ]
>> And defined an option code EDNS0LOCAL that starts at 65001 (with some docs)
>
>It's not clear to me how the library or consumers will use the option code EDNS0LOCAL.
>
>Perhaps you're thinking that if the option code is in the range 65001-65534
>then unpackStructValue() should unpack it as a EDNS0_LOCAL type?  Or are you
>thinking that the EDNS0LOCAL code of 65001 will serve only as documentation for
>users, and that the library will unpack any unknown option (regardless of code)
>as EDNS0_LOCAL?

Yes, I would like to see some documentation that you are free to use whatever
you like, but *officially* these are the typecodes you should use. So indeed:
just docs.

>The second choice makes sense to me.  I'm working with systems that are sending
>custom options with codes less than 65000.  The second choice also allows
>consumers to read options defined in the future without upgrading their version
>of this DNS library.

Yeah, sure. As above just as documentation, you should be free to use whatever
code point you like.

/Miek

--
Miek Gieben

reyzell:
> Yes, I would like to see some documentation that you are free to use whatever
> you like, but *officially* these are the typecodes you should use. So indeed:
> just docs.

Makes sense, I've applied the feedback.

PR #197, opened 2015-03-26
Refactor DNSSEC to use crypto.{PrivateKey,Signer}

FiloSottile:
This will allow RRSIG.Sign to use generic crypto.Signer implementations.

This is a interface breaking change, even if the required changes are most
likely just type asserions from crypto.PrivateKey to the underlying type or
crypto.Signer.

### DIFF
--- a/dnssec.go
+++ b/dnssec.go
@@ -6,14 +6,14 @@ import (
 	"crypto/dsa"
 	"crypto/ecdsa"
 	"crypto/elliptic"
-	"crypto/md5"
+	_ "crypto/md5"
+	"crypto/rand"
 	"crypto/rsa"
-	"crypto/sha1"
-	"crypto/sha256"
-	"crypto/sha512"
+	_ "crypto/sha1"
+	_ "crypto/sha256"
+	_ "crypto/sha512"
+	"encoding/asn1"
 	"encoding/hex"
-	"hash"
-	"io"
 	"math/big"
 	"sort"
 	"strings"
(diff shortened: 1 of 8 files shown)


miekg:
It deletes more than it add, so I like it. I have some nits, but it does not build with go 1.3? Which is annoying...

FiloSottile:
Umh, yeah `crypto.Signer` is new in 1.4.

I'd say that using stdlib interfaces is worth the restriction, but what about a nodnssec tag?

FiloSottile:
Added a go13 tag that disables Sign and sig0

miekg:
I won't kill 1.3 support atm. When 1.5 comes out, I'm fine with merging this (and dropping Go 1.3)

FiloSottile:
What if I swap things and make it compile by default on 1.3 without Sign and with a tag for 1.4-only Sign?

FiloSottile:
Just realized I'm an idiot. Go supports version tags.

So I can make it compile cleanly on all versions, with RRSIG.Sign (and Sig0) only on >= 1.4.

miekg:
Just to be clear: does this mean Sign() won't work at all for go < 1.4 ? I.e. not DNSSEC?

FiloSottile:
Yes, Sign() would simply not be compiled for < 1.4 but Verify and everything else would still work.

miekg:
Hmmm, I see. Then I rather wait a bit and merge this when 1.5 is out (and dropping 1.3 support at the same time).

FiloSottile:
Alright, I'm going to merge this on cloudflare/dns and leave it open here for later.

I edited it anyway to use the built-in version tags. Works with the current Travis config out of the box.

miekg:
[ Quoting <notifications@github.com> in "Re: [dns] Refactor DNSSEC to use cr..." ]
>Alright, I'm going to merge this on cloudflare/dns and leave it open here for later.

Ack.

>I edited it anyway to use the built-in version tags. Works with the current Travis config out of the box.

/Miek

--
Miek Gieben

miekg:
Think it is about time to get this in... Can you rebase/fix/tweak this a little? Thanks!

FiloSottile:
Done. This is the vanilla change, without reordering the files, which can be a future PR.

Issue #206, opened 2015-04-30
Algorithm name mismatch when generating/reading private keys


roosmaa:
When using algorithm 7 (RSASHA1-NSEC3-SHA1) reading the generated private key doesn't work. This is due to two places having different string values:
- [dnssec_keyscan.go:53](https://github.com/miekg/dns/blob/master/dnssec_keyscan.go#L53) (RSASHA1NSEC3SHA1)
- [dnssec.go:610](https://github.com/miekg/dns/blob/master/dnssec.go#L610) (RSASHA1-NSEC3-SHA1)

I'd submit a PR fixing it, but I don't know which format is the correct one that BIND9 produces.

miekg:
[ Quoting <notifications@github.com> in "[dns] Algorithm name mismatch when ..." ]
>When using algorithm 7 (RSASHA1-NSEC3-SHA1) reading the generated private key doesn't work. This is due to two places having different string values:
>- [dnssec_keyscan.go:53](https://github.com/miekg/dns/blob/master/dnssec_keyscan.go#L53) (RSASHA1NSEC3SHA1)
>- [dnssec.go:610](https://github.com/miekg/dns/blob/master/dnssec.go#L610) (RSASHA1-NSEC3-SHA1)
>
>I'd submit a PR fixing it, but I don't know which format is the correct one that BIND9 produces.

The mnemonic
http://www.iana.org/assignments/dns-sec-alg-numbers/dns-sec-alg-numbers.xhtml
use the one with dashes: RSASHA1-NSEC3-SHA1 show that should be used there as
well. (Note: not tested with bind)

/Miek

--
Miek Gieben

roosmaa:
Did some digging in BIND9 source. Seems there is a large mismatch between the algorithm names it uses and the ones in the iana.org website.

From `dst_parse.c`:

```c
	fprintf(fp, "%s %d ", ALGORITHM_STR, dst_key_alg(key));

	/* XXXVIX this switch statement is too sparse to gen a jump table. */
	switch (dst_key_alg(key)) {
	case DST_ALG_RSAMD5:
		fprintf(fp, "(RSA)\n");
		break;
	case DST_ALG_DH:
		fprintf(fp, "(DH)\n");
		break;
	case DST_ALG_DSA:
		fprintf(fp, "(DSA)\n");
		break;
	case DST_ALG_RSASHA1:
		fprintf(fp, "(RSASHA1)\n");
		break;
	case DST_ALG_NSEC3RSASHA1:
[...]

PR #207, opened 2015-04-30
Use algorithm number to determine private key type.

roosmaa:
Proposed fix for #206

### DIFF
--- a/dnssec_keyscan.go
+++ b/dnssec_keyscan.go
@@ -6,6 +6,7 @@ import (
 	"crypto/rsa"
 	"io"
 	"math/big"
+	"strconv"
 	"strings"
 )
 
@@ -34,8 +35,12 @@ func (k *DNSKEY) ReadPrivateKey(q io.Reader, file string) (PrivateKey, error) {
 		return nil, ErrPrivKey
 	}
 	// TODO(mg): check if the pubkey matches the private key
-	switch m["algorithm"] {
-	case "3 (DSA)":
+	algo, err := strconv.Atoi(strings.SplitN(m["algorithm"], " ", 2)[0])
+	if err != nil {
+		return nil, ErrPrivKey
+	}
+	switch uint8(algo) {
+	case DSA:
 		priv, e := readPrivateKeyDSA(m)
 		if e != nil {
 			return nil, e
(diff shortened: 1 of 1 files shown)


dnssec_keyscan.go
```diff
@@ -46,15 +51,15 @@ func (k *DNSKEY) ReadPrivateKey(q io.Reader, file string) (PrivateKey, error) {
 		}
 		priv.PublicKey = *pub
 		return (*DSAPrivateKey)(priv), e
-	case "1 (RSAMD5)":
+	case RSAMD5:
 		fallthrough
-	case "5 (RSASHA1)":
+	case RSASHA1:
 		fallthrough
-	case "7 (RSASHA1NSEC3SHA1)":
```
miekg: Why not add the different spellings here?

roosmaa: Other cases would have to be updated as well to match BIND9 names then. This approach is more error proof as it doesn't care what the human readable version of the string says. This is how BIND9 parses it's own files as well (from what I saw in code, didn't test).

miekg: [ Quoting <notifications@github.com> in "Re: [dns] Use algorithm number to d..." ]
>>  		fallthrough
>> -	case "7 (RSASHA1NSEC3SHA1)":
>
>Other cases would have to be updated as well to match BIND9 names then. This approach is more error proof as it doesn't care what the human readable version of the string says. This is how BIND9 parses it's own files as well (from what I saw in code, didn't test).

I can see that this is somewhat cleaner, but how does it solve the
RSA-SHA1-NSEC-SHA1 vs RSASHA1NSECSHA1 thing?

roosmaa: The `RSASHA1NSECSHA1` was clearly a typo and was supposed to be `RSA-SHA1-NSEC-SHA1`. Could've just fixed the typo and be done with it. However, as we discovered, BIND9 uses completely different human readable algorithm names (ie instead of `RSA-SHA1-NSEC-SHA1` it uses `NSEC3RSASHA1`, instead of `RSAMD5 ` it uses `RSA`, etc).

BIND9 also doesn't care about the human readable version of the algorithm when it's parsing the files (at least if I read the code correctly). So in order to fix reading of private keys generated by this library and BIND9 we should also ignore the human readable version of the algorithm. Which this PR does.

*Note: BIND9 should've been able to read private keys generated by this lib regardless of this PR since it didn't care about the human readable names of algos.*

miekg: [ Quoting <notifications@github.com> in "Re: [dns] Use algorithm number to d..." ]
>>  		fallthrough
>> -	case "7 (RSASHA1NSEC3SHA1)":
>
>The `RSASHA1NSECSHA1` was clearly a typo and was supposed to be `RSA-SHA1-NSEC-SHA1`. Could've just fixed the typo and be done with it. However, as we discovered, BIND9 uses completely different human readable algorithm names (ie instead of `RSA-SHA1-NSEC-SHA1` it uses `NSEC3RSASHA1`, instead of `RSAMD5 ` it uses `RSA`, etc).
>
>BIND9 also doesn't care about the human readable version of the algorithm when it's parsing the files (at least if I read the code correctly). So in order to fix reading of private keys generated by this library and BIND9 we should also ignore the human readable version of the algorithm. Which this PR does.
>
>*Note: BIND9 should've been able to read private keys generated by this lib regardless of this PR since it didn't care about the human readable names of algos.*

Ah, sorry. We only looking at the algorithm number *only* in this PR. Which
should solve this.

PR #208, opened 2015-05-06
Check that the query ID matches the answer ID.

michaelharo:
Reduce some code duplication by making Exchange() use Client.Exchange().

When performing an Exchange if the query ID does not match the answer ID
return an error.  Also add a test for this condition.

### DIFF
--- a/client.go
+++ b/client.go
@@ -46,9 +46,30 @@ type Client struct {
 //	co.Close()
 //
 func Exchange(m *Msg, a string) (r *Msg, err error) {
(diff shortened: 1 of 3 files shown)


miekg:
Thanks. the query id fix is good.

I think I want to keep exchange as a light weight alternative to client.Exchange(). client.Exchange does a bunch of things (and allocates a lot more things) that only make sense if you can access the actual client structure.

michaelharo:
Updated.

PR #209, opened 2015-05-06
Cleanup Client.exchange

michaelharo:
Make Client's exchange function easier to read by moving timeout logic
into separate functions.

Start the timers closer to where they're used so that time from other
logic doesn't impact the deadlines.

### DIFF
--- a/client.go
+++ b/client.go
@@ -122,31 +122,39 @@ func (c *Client) Exchange(m *Msg, a string) (r *Msg, rtt time.Duration, err erro
 	return r, rtt, nil
 }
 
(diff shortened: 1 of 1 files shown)


miekg:
This actually adds more code than I deletes... not 100% in favor. I do agree with the fact of placing the deadline setting closer to the actual call.

michaelharo:
I wasn't trying to reduce the number of lines of code.  The goal was to make it more readable by moving logic into more focused functions.

miekg:
Ack. It is more readable.

PR #210, opened 2015-05-07
TLSA records currently return nil after parsing

asergeyev:
This is just heads up. Currently TLSA records are not parsed and give nil RR. I expect to fix this soon. 

This PR could be closed if I'll be unable to do so after creating it :)


asergeyev:
Sure it breaks tests, that was intentional in commit :)

miekg:
The RR with lenght enormous overflows the maxTok limit, hence the crash (although there are length checks of course). Btw is this a legal TLSA record?

The other error is more interesting, playing with your commit in the tlsa branch.

miekg:
See PR #211, which has the fix. Note that I had to change the error var `e` to `e2` (i.e. a new var) to make it work, which is strange.

Identified a few other issues. Your test is now working. Please have a look before I merge.

asergeyev:
Oh, cool, thanks, anything I can help with instead? :)

I generated records using: https://www.huque.com/bin/gen_tlsa 
and cert I pulled with openssl from example.org. Bind has very large variety of TLSA recs it can chew, almost looks like it does not care much about actual RDATA being valid.

I have bunches of TLSA that I can run thru it today and check for more errors but since they do not belong to me I'd try to make comparable examples if I'd find them.

miekg:
[ Quoting <notifications@github.com> in "Re: [dns] TLSA records currently re..." ]
>Oh, cool, thanks, anything I can help with instead? :)
>
>I generated records using: https://www.huque.com/bin/gen_tlsa
>and cert I pulled with openssl from example.org. Bind has very large variety of TLSA recs it can chew, almost looks like it does not care much about actual RDATA being valid.
>
>I have bunches of TLSA that I can run thru it today and check for more errors but since they do not belong to me I'd try to make comparable examples if I'd find them.

Thanks. Getting larger TLSA parsed means splitting them up in 2048 char chunks.
2K chars per token seems like a reasonable limit. Not sure how bind deals with
this though. It also has repercussions on how large your response DNS message
will get.

/Miek

--
Miek Gieben

asergeyev:
Yeah most things are split in real zonefiles, much shorter than 2048 per line.

PR #211, opened 2015-05-07
Fix TLSA parsing

miekg:
Fix an overflow and report back a proper error when we hit that limit.


asergeyev:
I will try this for real zones and report how it goes.

asergeyev:
This fix works for all TLSA records I got. Thank you.

miekg:
[ Quoting <notifications@github.com> in "Re: [dns] Fix TLSA parsing (#211)..." ]
>This fix works for all TLSA records I got. Thank you.

Cool. Merging

/Miek

--
Miek Gieben

PR #212, opened 2015-05-07
Changes for rarely-used RR types

asergeyev:
this will continue, comments are welcome but I noticed that some odd types are not quite supported fully (HINFO, SSHFP) and some, have updated RFCs that need lib changes (NSAP).

I expect to commit more to this PR.


miekg:
Ok.

asergeyev:
I think I'm done adding stuff to PR at this point. Let me know what you think please and where I should update my edits.

BTW, I also can add support for APL rrtype but it's slightly messy one, would require new dns:"xxx" encoding for variable number of bytes from IP :(

miekg:
I won't NACK new record types, but APL is messy indeed (one of the reasons it is still out of this library)

Issue #166, opened 2015-01-13
A records are stored as 16 bytes


miekg:
While working on the format branch I noticed that rr.A's are stored as 16 bytes. Even if you call To4() on it. Seems to be a limitation in the net package, but it is wasteful none the less.

asergeyev:
Seem like adding copy of net.parseIPv4 with edited return and use it in setA and may be other places. 

I use int IPs for v4 but I added modified version that seem to do what you need, feel free to use it, repo is public domain code:  https://github.com/asergeyev/go-toys/tree/master/loadip (so far I'd ask to not import from there, that's unstable place)

michaelharo:
I'm not sure I understand the problem.  When you call To4() on a net.IP you get back a 4 byte slice (or nil).  I confirmed this by looking at http://play.golang.org/p/98OA0dt7X8.

asergeyev:
To4 just creates new slice without actually releasing underlying structure of 16. 
See http://blog.golang.org/go-slices-usage-and-internals  FindDigits/CopyDigits example. And copying each IP to new slice is somewhat wasteful so I suggested solution with new parser for v4 that could replace ParseIP(string).To4() combination.

miekg:
I welcome a pr :-) don't have much time to work on this specific issue
myself.

asergeyev:
Do you want to have some test that would actually show that we won some memory overall?

miekg:
[ Quoting <notifications@github.com> in "Re: [dns] A records are stored as 1..." ]
>Do you want to have some test that would actually show that we won some memory overall?

That would be really awesome

asergeyev:
will do, in other PR though... we can merge test first, measure and merge fix, measure again then.

miekg:
[ Quoting <notifications@github.com> in "Re: [dns] A records are stored as 1..." ]
>will do, in other PR though... we can merge test first, measure and merge fix, measure again then.

fair enough, this doesn't have to be in the test. But some numbers would be
nice.

/Miek

--
Miek Gieben

asergeyev:
// see PR, test converted to benchmark that shows no memory difference at high counts of iterations.

PR #213, opened 2015-05-08
Adding IPv4 specific loading code

asergeyev:
Follow up on #166 ... I don't know some internals so please feel free to disagree with this way of plugging it in.

### DIFF
--- a/zscan.go
+++ b/zscan.go
@@ -3,6 +3,7 @@ package dns
 import (
 	"io"
 	"log"
+	"net"
 	"os"
 	"strconv"
 	"strings"
@@ -961,4 +962,45 @@ func stringToNodeID(l lex) (uint64, *ParseError) {
 		return 0, &ParseError{l.token, "bad NID/L64 NodeID/Locator64", l}
 	}
 	return u, nil
+}
+
+// parse IPv4 address to four-byte slice under net.IP to save memory (see #166)
+func stringToIPv4(l lex) (net.IP, *ParseError) {
+	var (
+		ip  uint64
+		oct uint32
+		b   byte
+		num byte
+	)
+
+	// not checking len upfront, error will be obvious in soon anyway
+	for _, b = range []byte(l.token) {
+		switch {
+		case b == '.':
+			if num++; num > 3 {
+				return nil, &ParseError{l.token, "bad IP address", l}
+			}
+			ip = ip<<8 + uint64(oct)
+			if ip > 0xffffffff {
+				return nil, &ParseError{l.token, "bad IP address", l}
+			}
+			oct = 0
+		case b >= '0' && b <= '9':
+			oct = oct*10 + uint32(b-'0')
+			if oct > 255 {
+				return nil, &ParseError{l.token, "bad IP address", l}
+			}
+		default:
+			return nil, &ParseError{l.token, "bad IP address", l}
+		}
+	}
+	if num < 3 {
+		return nil, &ParseError{l.token, "bad IP address", l}
+	}
+	ip = ip<<8 + uint64(oct)
+	if ip > 0xffffffff {
+		return nil, &ParseError{l.token, "bad IP address", l}
+	}
+
+	return net.IP([]byte{byte(ip >> 24), byte(ip >> 16), byte(ip >> 8), byte(ip)}), nil
 }
--- a/zscan.go
+++ b/zscan.go
@@ -969,12 +969,11 @@ func stringToIPv4(l lex) (net.IP, *ParseError) {
 	var (
 		ip  uint64
 		oct uint32
-		b   rune
 		num byte
 	)
 
 	// not checking len upfront, error will be obvious in soon anyway
-	for _, b = range l.token {
+	for _, b := range l.token {
 		switch {
 		case b == '.':
 			if num++; num > 3 {
@@ -1001,6 +1000,5 @@ func stringToIPv4(l lex) (net.IP, *ParseError) {
 	if ip > 0xffffffff {
 		return nil, &ParseError{l.token, "bad IP address", l}
 	}
-
 	return net.IP([]byte{byte(ip >> 24), byte(ip >> 16), byte(ip >> 8), byte(ip)}), nil
 }
--- a/zscan.go
+++ b/zscan.go
@@ -969,12 +969,12 @@ func stringToIPv4(l lex) (net.IP, *ParseError) {
 	var (
 		ip  uint64
 		oct uint32
-		b   byte
+		b   rune
 		num byte
 	)
 
 	// not checking len upfront, error will be obvious in soon anyway
-	for _, b = range []byte(l.token) {
+	for _, b = range l.token {
 		switch {
 		case b == '.':
 			if num++; num > 3 {
--- a/zscan_rr.go
+++ b/zscan_rr.go
@@ -136,7 +136,7 @@ func endingToTxtSlice(c chan lex, errstr, f string) ([]string, *ParseError, stri
 	return s, nil, l.comment
 }
 
-func setA(h RR_Header, c chan lex, o, f string) (_ RR, e *ParseError, _ string) {
+func setA(h RR_Header, c chan lex, o, f string) (RR, *ParseError, string) {
 	rr := new(A)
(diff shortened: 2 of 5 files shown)


asergeyev:
found another place, fixing

asergeyev:
I think now it should be doing what's intended in #166

zscan_rr.go
```diff
@@ -136,7 +136,7 @@ func endingToTxtSlice(c chan lex, errstr, f string) ([]string, *ParseError, stri
 	return s, nil, l.comment
 }
 
-func setA(h RR_Header, c chan lex, o, f string) (RR, *ParseError, string) {
+func setA(h RR_Header, c chan lex, o, f string) (_ RR, e *ParseError, _ string) {
```
miekg: Could you not use named params here?

zscan.go
```diff
@@ -962,3 +963,44 @@ func stringToNodeID(l lex) (uint64, *ParseError) {
...
+func stringToIPv4(l lex) (net.IP, *ParseError) {
+	var (
+		ip  uint64
+		oct uint32
+		b   byte
+		num byte
+	)
+
+	// not checking len upfront, error will be obvious in soon anyway
+	for _, b = range []byte(l.token) {
```
miekg: why does this need to be converted to []byte? We could use the string as-is, no? And convert a single value to a byte?

zscan.go
```diff
@@ -962,3 +963,44 @@ func stringToNodeID(l lex) (uint64, *ParseError) {
...
+		switch {
+		case b == '.':
+			if num++; num > 3 {
```
miekg: why not use the counter you get back from `range`?

zscan.go
```diff
@@ -962,3 +963,44 @@ func stringToNodeID(l lex) (uint64, *ParseError) {
...
+		case b >= '0' && b <= '9':
+			oct = oct*10 + uint32(b-'0')
+			if oct > 255 {
+				return nil, &ParseError{l.token, "bad IP address", l}
+			}
+		default:
+			return nil, &ParseError{l.token, "bad IP address", l}
+		}
+	}
+	if num < 3 {
```
miekg: ah you need `num` here. OK.

zscan_rr.go
asergeyev: That's lazy.. I should just reverse that and add var e *ParseError few lines lower. 
Named params just not going to be used, rr has to have type A and comment will also stay unused.

zscan.go
asergeyev: that's for full octets, not for individual characters

asergeyev:
Thanks, I'll get fixes for []byte() after mem test.

asergeyev:
Well... Fixes for []byte and other code review notes separated from memory test. I decided to put things in same branch and test with hack that swaps parser to old function...

Looks like many small slices end up allocating with padding to fit more data later....  I see buckets size 8 and 16 go up when test is ran

asergeyev:
Hard to call memprofiler accurate, but when I crank LENGTH from 0 to 1000 it seem there is some extractable numbers 

HeapObjects growth always smaller on new ip parser
HeapInUse always bigger on new ip parser
TotalAlloc bigger on old parser but closer to new one if tests swapped in order

On million A recs to parse new one wins on all numbers all the time regardless of where it placed... But IMO those wins are relatively small.

asergeyev:
After changing this code to be a benchmark instead:

		$ go test -v -run=nothing -bench=BenchmarkZoneParseWith
		PASS
		BenchmarkZoneParseWithStringToIPv4	   10000	    127196 ns/op
		--- BENCH: BenchmarkZoneParseWithStringToIPv4
			mem_test.go:57: Average alloc growth per iteration: 111520 (1 parsing rounds)
			mem_test.go:57: Average alloc growth per iteration: 98664 (2 parsing rounds)
			mem_test.go:57: Average alloc growth per iteration: 91289 (5 parsing rounds)
			mem_test.go:57: Average alloc growth per iteration: 87604 (20 parsing rounds)
			mem_test.go:57: Average alloc growth per iteration: 86621 (100 parsing rounds)
			mem_test.go:57: Average alloc growth per iteration: 86457 (300 parsing rounds)
			mem_test.go:57: Average alloc growth per iteration: 86400 (1000 parsing rounds)
			mem_test.go:57: Average alloc growth per iteration: 86384 (3000 parsing rounds)
			mem_test.go:57: Average alloc growth per iteration: 86378 (10000 parsing rounds)
		BenchmarkZoneParseWithNetParseIP	   10000	    124895 ns/op
		--- BENCH: BenchmarkZoneParseWithNetParseIP
			mem_test.go:57: Average alloc growth per iteration: 111488 (1 parsing rounds)
			mem_test.go:57: Average alloc growth per iteration: 98928 (2 parsing rounds)
			mem_test.go:57: Average alloc growth per iteration: 91398 (5 parsing rounds)
			mem_test.go:57: Average alloc growth per iteration: 87631 (20 parsing rounds)
			mem_test.go:57: Average alloc growth per iteration: 86627 (100 parsing rounds)
			mem_test.go:57: Average alloc growth per iteration: 86459 (300 parsing rounds)
			mem_test.go:57: Average alloc growth per iteration: 86401 (1000 parsing rounds)
			mem_test.go:57: Average alloc growth per iteration: 86384 (3000 parsing rounds)
			mem_test.go:57: Average alloc growth per iteration: 86378 (10000 parsing rounds)
		ok  	github.com/miekg/dns	9.897s

miekg:
[ Quoting <notifications@github.com> in "Re: [dns] Adding IPv4 specific load..." ]
>Hard to call memprofiler accurate, but when I crank LENGTH from 0 to 1000 it seem there is some extractable numbers
>
>HeapObjects growth always smaller on new ip parser
>HeapInUse always bigger on new ip parser
>TotalAlloc bigger on old parser but closer to new one if tests swapped in order
>
>On million A recs to parse new one wins on all numbers all the time regardless of where it placed... But IMO those wins are relatively small.

So, not worth merging in your opinion?

We could still do a net.Parse().To4() copy out the bytes and destroy (or let gc
take care of the unused value).

asergeyev:
I think it's worth merging but I know that fact that I wrote that code plays a role in making my opinion...
* functionally we do not change much, code is still readeable
* we do not do anything that makes code slower
* overall memory consumption may go down (I'd speculate that there is possibility that in certain conditions go would choose 4-byte objects to store our 4-byte slices, we just do not see this happen in tests)

Copying bytes would waste another 8 or 16 bytes (whichever bucket go would choose for new make()), it's unlikely to have just 4 byte impact. Also many memmove would slow thigns down. Also, it seem to me that even calling To4 when you don't need to is worse than just keeping 16 bytes, new slice eats new 12 bytes off stack to save cap/len values (reason why in last version of test I removed that, also matches current state of the library)

Against the merge is the fact that **there is nothing wrong I can see with current solution**. Most users expect there net.IP in structs, we could not get to revolutionize too much.

miekg:
[ Quoting <notifications@github.com> in "Re: [dns] Adding IPv4 specific load..." ]
>I think it's worth merging but I know that fact that I wrote that code plays a role in making my opinion...
>* functionally we do not change much, code is still readeable
>* we do not do anything that makes code slower
>* overall memory consumption may go down (I'd speculate that there is possibility that in certain conditions go would choose 4-byte objects to store our 4-byte slices, we just do not see this happen in tests)
>
>Copying bytes would waste another 8 or 16 bytes (whichever bucket go would choose for new make()), it's unlikely to have just 4 byte impact. Also many memmove would slow thigns down. Also, it seem to me that even calling To4 when you don't need to is worse than just keeping 16 bytes, new slice eats new 12 bytes off stack to save cap/len values (reason why in last version of test I removed that, also matches current state of the library)
>
>Against the merge is the fact that **there is nothing wrong I can see with current solution**. Most users expect there net.IP in structs, we could not get to revolutionize too much.

That last point is intersting... I'm leaning against closing this issue and
*not* merging this PR...

asergeyev:
Sure :) I actually agree.

asergeyev:
You might want to close #166.

miekg:
Thanks!

miekg:
done

PR #215, opened 2015-06-04
Check that the RRs passed to Sign and Verify form a valid RFC2181 RRSet

aaronlehmann:
Add a sanity check used by RRSig's Sign and Verify functions making sure
that the records they operate on form a valid RRSet (same name, type,
and class).

Add a unit test TestInvalidRRSet that calls RRSig's Sign and Verify
methods with invalid RRSets, and makes sure the correct error is
returned.

### DIFF
--- a/dnssec.go
+++ b/dnssec.go
@@ -205,11 +205,39 @@ func (d *DS) ToCDS() *CDS {
 	return c
 }
 
+// isValidRRSet checks if a set of RRs is a valid RRset as defined by RFC 2181.
+// This means the RRs need to have the same type, name, and class. Returns true
+// if the RR set is valid, otherwise false.
+func isValidRRSet(rrset []RR) bool {
+	if len(rrset) == 0 {
+		return false
+	}
+	if len(rrset) == 1 {
+		return true
+	}
+	rrHeader := rrset[0].Header()
+	rrType := rrHeader.Rrtype
+	rrClass := rrHeader.Class
+	rrName := rrHeader.Name
+
+	for _, rr := range rrset[1:] {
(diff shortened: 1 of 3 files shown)


dnssec.go
```diff
@@ -221,6 +249,10 @@ func (rr *RRSIG) Sign(k PrivateKey, rrset []RR) error {
 		return ErrKey
 	}
 
+	if !isValidRRSet(rrset) {
```
miekg: If you know for sure when you call Sign that you have a proper RRset this test does not add anything. I rather still leave this out of Sign itself.

dnssec.go
```diff
@@ -205,11 +205,39 @@ func (d *DS) ToCDS() *CDS {
 	return c
 }
 
+// isValidRRSet checks if a set of RRs is a valid RRset as defined by RFC 2181.
+// This means the RRs need to have the same type, name, and class. Returns true
+// if the RR set is valid, otherwise false.
+func isValidRRSet(rrset []RR) bool {
```
miekg: I think we should move this to `default.go` and name it `IsRRSet`

miekg:
.. and sorry for the late reply, this PR slipped by me.

aaronlehmann:
Thanks for the review. I added some commits to make these changes.

miekg:
[ Quoting <notifications@github.com> in "Re: [dns] Check that the RRs passed..." ]
>Thanks for the review. I added some commits to make these changes.

cool!

FiloSottile:
Bump

miekg:
sorry, busy busy busy :-)

PR #219, opened 2015-06-16
Fix CAA parsing

rolandshoemaker:
Enables direct parsing of CAA records instead of parsing them as RFC3597 records. This allows the [Boulder](https://github.com/letsencrypt/boulder) developers to clear CAA record parsing code out of our code base.

I tested this against *almost* all the CAA records I could find in the wild ([relevant gist](https://gist.github.com/rolandshoemaker/4911f0c39acb25d68e80)) and everything seems to be working correctly...

### DIFF
--- a/msg.go
+++ b/msg.go
@@ -100,7 +100,7 @@ var TypeToString = map[uint16]string{
 	TypeANY:        "ANY", // Meta RR
 	TypeATMA:       "ATMA",
 	TypeAXFR:       "AXFR", // Meta RR
-	TypeCAA:        "TYPE257",
+	TypeCAA:        "CAA",
 	TypeCDNSKEY:    "CDNSKEY",
 	TypeCDS:        "CDS",
 	TypeCERT:       "CERT",
@@ -543,6 +543,36 @@ func packTxtString(s string, msg []byte, offset int, tmp []byte) (int, error) {
 	return offset, nil
 }
 
+func packOctetString(s string, msg []byte, offset int, tmp []byte) (int, error) {
+	if offset >= len(msg) {
+		return offset, ErrBuf
+	}
+	bs := tmp[:len(s)]
+	copy(bs, s)
+	for i := 0; i < len(bs); i++ {
+		if len(msg) <= offset {
+			return offset, ErrBuf
+		}
+		if bs[i] == '\\' {
+			i++
+			if i == len(bs) {
+				break
+			}
+			// check for \DDD
+			if i+2 < len(bs) && isDigit(bs[i]) && isDigit(bs[i+1]) && isDigit(bs[i+2]) {
+				msg[offset] = dddToByte(bs[i:])
+				i += 2
+			} else {
+				msg[offset] = bs[i]
+			}
+		} else {
+			msg[offset] = bs[i]
+		}
+		offset++
+	}
+	return offset, nil
+}
+
 func unpackTxt(msg []byte, offset, rdend int) ([]string, int, error) {
 	var err error
 	var ss []string
@@ -890,6 +920,12 @@ func packStructValue(val reflect.Value, msg []byte, off int, compression map[str
 				// length of string. String is RAW (not encoded in hex, nor base64)
 				copy(msg[off:off+len(s)], s)
 				off += len(s)
+			case `dns:"octet"`:
+				var varstrTmp []byte
+				off, err = packOctetString(fv.String(), msg, off, varstrTmp)
+				if err != nil {
+					return lenmsg, err
+				}
 			case `dns:"txt"`:
 				fallthrough
 			case "":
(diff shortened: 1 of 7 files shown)


msg.go
```diff
@@ -100,7 +100,7 @@ var TypeToString = map[uint16]string{
 	TypeANY:        "ANY", // Meta RR
 	TypeATMA:       "ATMA",
 	TypeAXFR:       "AXFR", // Meta RR
-	TypeCAA:        "CAA",
+	TypeCAA:        "TYPE257",
```
miekg: This can just stay "CAA" or have to a reason to make it `TYPE257` ?

types.go
```diff
@@ -1527,25 +1527,25 @@ func (rr *EUI64) copy() RR           { return &EUI64{*rr.Hdr.copyHeader(), rr.Ad
...
 func (rr *EUI64) len() int           { return rr.Hdr.len() + 8 }
 
-// Support in incomplete - just handle it as unknown record
-/*
 type CAA struct {
 	Hdr   RR_Header
 	Flag  uint8
 	Tag   string
-	Value string `dns:"octet"`
+	Value string `dns:"hex"`
```
miekg: Why is tagged the `dns:"hex"`?

miekg:
Think this is a good start, but I'm not sure the value's type should be hex, but just a string or something new, because there is no length to be prefixed.

msg.go
asergeyev: This makes go-dns print TYPE257 everywhere it outputs CAA record as string and also messes up parsing of record when CAA used, will only understand TYPE257 (since StringToType that tokenizer uses is a reverse of TypeToString).

asergeyev:
I'd add 2 cents that it needs some added testcases in types_test.go

msg.go
rolandshoemaker: Whoops, mean't to switch this back before pushing, will revert to `CAA`.

types.go
rolandshoemaker: I originally just decoded this directly to a string but this noticed in section 4.1 of [RFC6844](https://tools.ietf.org/html/rfc6844)

```
Value:  A sequence of octets representing the property value.
   Property values are encoded as binary values and MAY employ sub-
   formats.
```

90% of the records I've found have binary encoded ASCII content in `Value` but a few (e.g. for domains `madtech.nl, codewiz.org, www.isoc.lu` etc) contain non-ASCII binary data. My thinking was that since the RFC explicitly defines it to be binary encoded (and possibly containing sub-formats) it would be better to just return a `hex` string and let the user decide what to do with it based on the `Tag` contents.

I'd be happy to switch this back to a `string` (or perhaps a `byte` slice?) if you think that would be more appropriate.

miekg:
Yep, some test cases would be nice. The wireformat should be OK, when using the hex tag, except I don't think it should be hex, example from the RFC include:
```
.       CAA 0 issue "ca.example.net; account=230123"
```
Let's check what BIND does :)
```
/ # dig CAA madtech.nl +short
2 auth "0>09\006\010+\006\001\004\001\214y\002\003\001\006\009`\134H\001e\003\004\002\001\004 y\209\012\221r\220\156Q\218\150\150{\166\245:\231\182%\157:\133\179}\1923r\238\151\255\128q\145\002\001\000"
```
So this is parsing it just as a string, but the downside of using just a string is that it will prefix it with the length of the string in the wiredata, something the RFC says should not happen as it is the last field in the data, hence my use of octect to check for this, but a better name would be preferred.

zscan_rr.go
```diff
@@ -2170,10 +2171,59 @@ func setIPSECKEY(h RR_Header, c chan lex, o, f string) (RR, *ParseError, string)
 	return rr, nil, c1
 }
 
+func setCAA(h RR_Header, c chan lex, o, f string) (RR, *ParseError, string) {
+	rr := new(CAA)
+	rr.Hdr = h
+	l := <-c
+	if l.token != "\\#" {
```
miekg: you are parsing this is an unknown RR (see `setRFC3597`). This should be able to parse
`0 issue "ca.example.net; account=230123"` and `128 tbs "Unknown"`

rolandshoemaker:
I've switched to parsing the presentation format in `setCAA` and `CAA.String` and added a parsing test, I also switched back to the `octect` tag because I couldn't come up with anything... better (`var-str`..?). I'm not completely confident about the wiredata packing I've implemented so any comments would be especially appreciated.

msg.go
```diff
@@ -890,6 +920,12 @@ func packStructValue(val reflect.Value, msg []byte, off int, compression map[str
 				// length of string. String is RAW (not encoded in hex, nor base64)
 				copy(msg[off:off+len(s)], s)
 				off += len(s)
+			case `dns:"octet"`:
+				var varstrTmp []byte
```
miekg: don't need to repeat 'var' here.

zscan_rr.go
```diff
@@ -2170,10 +2170,46 @@ func setIPSECKEY(h RR_Header, c chan lex, o, f string) (RR, *ParseError, string)
...
+		return nil, &ParseError{f, "bad CAA Tag", l}, ""
+	}
+	rr.Tag = l.token
+
+	<-c // zBlank
+	s, e, c1 := endingToTxtSlice(c, "bad CAA Value", f)
+	if e != nil {
+		return nil, e, ""
+	}
+	if len(s) > 1 {
```
miekg: I don't understand this if-elseif-else tree?

miekg:
Looks good. There is some code duplication, but I will clean that up in another PR.

zscan_rr.go
rolandshoemaker: It is my understanding that the presentation format should have a single quoted string, so if `endingToTxtSlice` returns more than one an error should be thrown. I've removed the `else if` as I didn't realize that `endingToTxtSlice` would return an empty string if it encountered `""` in the record.

miekg: Ack.

Issue #222, opened 2015-06-17
Lazy message reading in dns.Client


asergeyev:
Currently ReadMsg parses wire message fully before it returns (and reads next one). When you just need an answer very fast it's not quite ideal.

I can suggest small experiment with lazy reader that would return message header and blob to parse later...   Would that be interesting to have?   IMO it's not quite universal, only should come handy when reading lo-o-ots of messages from fast DNS cache that's close on network and does not give time for few reader threads to idle...

miekg:
sgtm. we'll later see how this fits in.

asergeyev:
Well... Added patch and PR...  Let me know what you think.... Somewhat for my needs (getting that wire Header is what I want to have as soon as possible and not worry about buffer allocation)

I wonder if ReadMsg could actually be changed now to ReadMsgBytes followed by Msg.Unpack and followed by Tsig validation.  Should be effectively same as it is now but will cut duplicated code between msg.Unpack and ReadMsg.

miekg:
[ Quoting <notifications@github.com> in "Re: [dns] Lazy message reading in d..." ]
>Well... Added patch and PR...  Let me know what you think.... Somewhat for my needs (getting that wire Header is what I want to have as soon as possible and not worry about buffer allocation)

Yes, that makes sense.

>I wonder if ReadMsg could actually be changed now to ReadMsgBytes followed by Msg.Unpack and followed by Tsig validation.  Should be effectively same as it is now but will cut duplicated code between msg.Unpack and ReadMsg.

As commented in the PR, yes something like this should be wired up in the lib
itself. 99% of stuff hitting a server is crap, getting rid of it early and fast
is good.

Thanks!

PR #224, opened 2015-06-24
Added function for lazy message reading per #222

asergeyev:


miekg:
As I think is my normal review mode, i will go over each change in detail and then (hopefully) see the bigger picture and moan about that :-)

client.go
```diff
@@ -201,14 +201,16 @@ func (co *Conn) ReadMsg() (*Msg, error) {
 		}
 	}
 	n, err := co.Read(p)
-	if err != nil && n == 0 {
+	if err != nil {
 		return nil, err
+	} else if n < 12 {
```
miekg: Think 12 is better as a constant, headerSize ?

client.go
```diff
@@ -201,14 +201,16 @@ func (co *Conn) ReadMsg() (*Msg, error) {
...
+		return nil, ErrShortRead
 	}
+
 	p = p[:n]
 	if err := m.Unpack(p); err != nil {
 		return nil, err
 	}
-	co.rtt = time.Since(co.t)
```
miekg: why this removal?

client.go
```diff
@@ -219,6 +221,40 @@ func (co *Conn) ReadMsg() (*Msg, error) {
 	return m, err
 }
 
+// ReadMsgBytes reads a message bytes from the connection. Parses and fills
```
miekg: drop the "message bytes"? This does not read right.

client.go
```diff
@@ -219,6 +221,40 @@ func (co *Conn) ReadMsg() (*Msg, error) {
...
+// dns message wire header (passing nil would skip header parsing) and
+// returns message bytes to process them later.
+//
+// Note that this function would not be able to report TSIG error or
+// check it got actual DNS payload.
+func (co *Conn) ReadMsgBytes(hdr *Header) ([]byte, error) {
+	var p []byte
+	if _, ok := co.Conn.(*net.TCPConn); ok {
+		p = make([]byte, MaxMsgSize)
```
miekg: Why such a large buffer if we only need to read the header?

client.go
```diff
@@ -219,6 +221,40 @@ func (co *Conn) ReadMsg() (*Msg, error) {
...
+			p = make([]byte, co.UDPSize)
+		} else {
+			p = make([]byte, MinMsgSize)
+		}
+	}
+
+	n, err := co.Read(p)
+	if err != nil {
+		return nil, err
+	} else if n < 12 {
```
miekg: 12 -> constant

client.go
```diff
@@ -219,6 +221,40 @@ func (co *Conn) ReadMsg() (*Msg, error) {
...
+		return nil, ErrShortRead
+	}
+
+	p = p[:n]
+	if hdr != nil {
```
miekg: what is the use-case for giving this function a hdr=nil?

client.go
```diff
@@ -259,6 +295,8 @@ func (co *Conn) Read(p []byte) (n int, err error) {
 	if err != nil {
 		return n, err
 	}
+
+	co.rtt = time.Since(co.t)
```
miekg: ah this is where the time.Since moves to. Ok

client_test.go
```diff
@@ -32,7 +32,7 @@ func TestClientSync(t *testing.T) {
 	if err != nil {
 		t.Errorf("failed to exchange: %v", err)
 	}
-	if r != nil && r.Rcode != RcodeSuccess {
+	if r == nil || r.Rcode != RcodeSuccess {
```
miekg: geh, yes.

client.go
```diff
@@ -219,6 +221,40 @@ func (co *Conn) ReadMsg() (*Msg, error) {
 	return m, err
 }
 
+// ReadMsgBytes reads a message bytes from the connection. Parses and fills
+// dns message wire header (passing nil would skip header parsing) and
+// returns message bytes to process them later.
+//
+// Note that this function would not be able to report TSIG error or
+// check it got actual DNS payload.
+func (co *Conn) ReadMsgBytes(hdr *Header) ([]byte, error) {
```
miekg: ReadHeader? as a function name?

miekg:
Looks pretty good. Actually hooking it up in the lib itself is another PR (which is really fine)

client.go
asergeyev: Any reason to choose private vs making it public?

client.go
asergeyev: It basically repeats Read but with allocating buffer for message bytes... Hm... Just because it's easier than creating new error about it I guess...

client.go
asergeyev: Also wanted to time it as close as possible to reading a packet....

client.go
asergeyev: Technically it reads whole message... not just header...

asergeyev:
Should I continue in this one to reuse Msg.Unpack  properly or finish this one and open new later?

client.go
asergeyev: ok

client.go
asergeyev: I'll think about this one...

client.go
miekg: [ Quoting <notifications@github.com> in "Re: [dns] Added function for lazy m..." ]
>>  		return nil, err
>> +	} else if n < 12 {
>
>Any reason to choose private vs making it public?

Could as well be public.

client.go
miekg: Which is worth while, but I think the network latency is so huge that we can incur a function call in Code. But fair enough.

client.go
miekg: Sorry, I don't understand that remark. At most this function read 12 bytes from a message?

miekg:
Fixing this up first and then a new PR is fine. Does  this function need to be exported though?

asergeyev:
> Does this function need to be exported though?

Yeah... it's good function for convenience of those who write clients. Sometimes you only need packets with answers and without this function you'd go and parse all fields in authority/additional even if your client does not care about them.

miekg:
[ Quoting <notifications@github.com> in "Re: [dns] Added function for lazy m..." ]
>> Does this function need to be exported though?
>
>Yeah... it's good function for convenience of those who write clients. Sometimes you only need packets with answers and without this function you'd go and parse all fields in authority/additional even if your client does not care about them.

Ok. Works for me.

/Miek

--
Miek Gieben

asergeyev:
Well... Here is what I got...

I still followed with changing in one PR, since there is not much changed as of now. Logic is still same.  I'm not sure there is any good in pursuing creation of a function that's not reserving such large buffers... In TCP it's probably doable, but in UDP things should be loaded in one read call so it's really not possible to change...

miekg:
[ Quoting <notifications@github.com> in "Re: [dns] Added function for lazy m..." ]
>Well... Here is what I got...
>
>I still followed with changing in one PR, since there is not much changed as of now. Logic is still same.  I'm not sure there is any good in pursuing creation of a function that's not reserving such large buffers... In TCP it's probably doable, but in UDP things should be loaded in one read call so it's really not possible to change...

Ah yeah. I see what you mean... you can't do a partial read and then
read-some-more when you're ready for it... Part of the appeal of such
a function it that it would be nice if you *could*. Not sure how though...

asergeyev:
Last patch shows how I see it could be done for TCP. 64K is very large buffer I agree. dns.Conn.Read becomes less useful here since it expects to read everything in one pre-allocated buffer in TCP.

client.go
```diff
@@ -189,34 +189,100 @@ func (c *Client) exchange(m *Msg, a string) (r *Msg, rtt time.Duration, err erro
...
+	return m, err
+}
+
+// ReadMsgBytes reads DNS packet, parses and fills wire header (passing nil
+// would cause skipping that). Returns message in byte format to parse with
+// Msg.Unpack later on.
+//
+// Note that this function would not be able to report TSIG error or
+// check it got actual DNS payload.
+func (co *Conn) ReadMsgBytes(hdr *Header) ([]byte, error) {
```
miekg: So... What if we would really makes this only read 12 (or 14 in case of tcp) bytes and let it return the Conn, the bytes read and an error? Maybe the buffer should also by given to the function, so it is really without any internal allocations? So if you get an error you just discard and write FormErr (maybe from within the lib already). If thinks look OK we can parse the bytes to a real header and assemble the packet...

asergeyev: Once you read from UDP rest is discarded. You could not really read again there so you end up buffering and faking ability to read which is slow. 

http://golang.org/src/net/udpsock_posix.go?s=1547:1620#L53
http://golang.org/src/net/fd_unix.go#L256

then from rcvfrom(2):

    All three routines return the length of the message on successful completion. If a message 
    is too long to fit in the supplied buffer, excess bytes may be discarded depending on the 
    type of socket the message is received from.

miekg: [ Quoting <notifications@github.com> in "Re: [dns] Added function for lazy m..." ]
>> +// Note that this function would not be able to report TSIG error or
>> +// check it got actual DNS payload.
>> +func (co *Conn) ReadMsgBytes(hdr *Header) ([]byte, error) {
>
>Once you read from UDP rest is discarded. You could not really read again there so you end up buffering and faking ability to read which is slow.
>
>http://golang.org/src/net/udpsock_posix.go?s=1547:1620#L53
>http://golang.org/src/net/fd_unix.go#L256
>
>then from rcvfrom(2):
>
>    All three routines return the length of the message on successful completion. If a message
>    is too long to fit in the supplied buffer, excess bytes may be discarded depending on the
>    type of socket the message is received from.
>

Grrrr :( So what would be the win then, this is the typical space/time trade
off? So the allocation of memory still happens, the only thing that you shortcut
is the potentially slowish parsing of the rest of the packet. I rather spend
cycles on making that faster, or fail ealier.

/Miek

--
Miek Gieben

asergeyev:
I'll experiment with profiler a bit and let you know. IMO net lib has limits by itself plus fact that dns lib built around strings... Allocs for message reading are probably not worst thing there....

asergeyev:
I created new clientbench branch and put some test code there... As expected, UDP is close, almost no difference but TCP is faster in new implementation (not reserving 64K each time):

	BenchmarkNewUDPClient	   10000	    342285 ns/op
	BenchmarkOldUDPClient	    5000	    347112 ns/op
	BenchmarkNewTCPClient	   10000	    132508 ns/op
	BenchmarkOldTCPClient	   10000	    202532 ns/op

With pprof and web -cum BenchmarkName it looks that most CPU time was still on syscalls... and pack/unpack StructValue;  I decided to stub all of those, wrote small fake net.Conn for that... (second commit in that branch)... I also changed UDPSize to 4096 and run it with benchtime=10s to make sure I'll have small CPU expenses visible (paranoid Alex).

Here is what it looks like for "web -cum ReadMsg" only on new client test:

     https://www.dropbox.com/s/3iw1o23ytnksvie/pprof0002.svg?dl=1

You can see there is almost no time spent where you think we can win something.

miekg:
[ Quoting <notifications@github.com> in "Re: [dns] Added function for lazy m..." ]
>I created new clientbench branch and put some test code there... As expected, UDP is close, almost no difference but TCP is faster in new implementation (not reserving 64K each time):
>
>	BenchmarkNewUDPClient	   10000	    342285 ns/op
>	BenchmarkOldUDPClient	    5000	    347112 ns/op
>	BenchmarkNewTCPClient	   10000	    132508 ns/op
>	BenchmarkOldTCPClient	   10000	    202532 ns/op
>
>With pprof and web -cum BenchmarkName it looks that most CPU time was still on syscalls... and pack/unpack StructValue;  I decided to stub all of those, wrote small fake net.Conn for that... (second commit in that branch)... I also changed UDPSize to 4096 and run it with benchtime=10s to make sure I'll have small CPU expenses visible (paranoid Alex).
>
>Here is what it looks like for "web -cum ReadMsg" only on new client test:
>
>     https://www.dropbox.com/s/3iw1o23ytnksvie/pprof0002.svg?dl=1
>
>You can see there is almost no time spent where you think we can win something.

Ack. Yep, that points to something different.

asergeyev:
But convinience that I was asking for in #222 is already there plus small bonus for TCP... Maybe you can merge this one? (I understand if not...)

as of performance, could be separate effort to look at different places but the need of reserving smallish structures with new(...) (or &Name{}) and work with strings is what could be hard to deal with... 
And, since your lib is not just wire client but also works with text representation a lot I do not think you should change the approach.

miekg:
[ Quoting <notifications@github.com> in "Re: [dns] Added function for lazy m..." ]
>But convinience that I was asking for in #222 is already there plus small bonus for TCP... Maybe you can merge this one? (I understand if not...)

Yes, I will. But I first want to take another look.

>as of performance, could be separate effort to look at different places but the
>need of reserving smallish structures with new(...) (or &Name{}) and work with
>strings is what could be hard to deal with...  And, since your lib is not just
>wire client but also works with text representation a lot I do not think you
>should change the approach.

No, indeed. Having just strings make it easy to work with.

/Miek

--
Miek Gieben

client.go
```diff
@@ -189,34 +189,100 @@ func (c *Client) exchange(m *Msg, a string) (r *Msg, rtt time.Duration, err erro
...
+		if _, ok := co.TsigSecret[t.Hdr.Name]; !ok {
+			return m, ErrSecret
+		}
+		// Need to work on the original message p, as that was used to calculate the tsig.
+		err = TsigVerify(p, co.TsigSecret[t.Hdr.Name], co.tsigRequestMAC, false)
+	}
+	return m, err
+}
+
+// ReadMsgBytes reads DNS packet, parses and fills wire header (passing nil
```
miekg: ReadMsgBytes reads a DNS message, parses it and populates hdr (when hdr is not nil)

client.go
```diff
@@ -189,34 +189,100 @@ func (c *Client) exchange(m *Msg, a string) (r *Msg, rtt time.Duration, err erro
...
+// would cause skipping that). Returns message in byte format to parse with
```
miekg: It return the rest of the message as a byte slice to pasrse with Msg.Unpack later on.

client.go
```diff
@@ -189,34 +189,100 @@ func (c *Client) exchange(m *Msg, a string) (r *Msg, rtt time.Duration, err erro
...
+// Msg.Unpack later on.
+//
+// Note that this function would not be able to report TSIG error or
```
miekg: Note that error handling on the message body is not possible as only the header is parsed.

client.go
miekg: ReadMsgBytes -> ReadMsgHeader as a function name.

client.go
```diff
@@ -189,34 +189,100 @@ func (c *Client) exchange(m *Msg, a string) (r *Msg, rtt time.Duration, err erro
...
+// check it got actual DNS payload.
+func (co *Conn) ReadMsgBytes(hdr *Header) ([]byte, error) {
+	var (
+		p   []byte
+		n   int
+		err error
+	)
+
+	if t, ok := co.Conn.(*net.TCPConn); ok {
+		// we got two byte header to know how much to receive...
```
miekg: // First two bytes specify the length of the entire message.

client.go
```diff
@@ -189,34 +189,100 @@ func (c *Client) exchange(m *Msg, a string) (r *Msg, rtt time.Duration, err erro
...
 			p = make([]byte, MinMsgSize)
 		}
+		n, err = co.Read(p)
 	}
-	n, err := co.Read(p)
-	if err != nil && n == 0 {
+
+	if err != nil {
 		return nil, err
+	} else if n < _HBytes {
```
miekg: _Hbytes is not an idiomatic name.

client.go
```diff
@@ -189,34 +189,100 @@ func (c *Client) exchange(m *Msg, a string) (r *Msg, rtt time.Duration, err erro
...
+		}
 	}
-	co.rtt = time.Since(co.t)
-	if t := m.IsTsig(); t != nil {
-		if _, ok := co.TsigSecret[t.Hdr.Name]; !ok {
-			return m, ErrSecret
+	return p, err
+}
+
+// tcpMsgLen - helper func to read first two bytes of stream as uint16 packet length
```
miekg: .. is a helper function to read ...
(and add a closing dot)

client.go
```diff
@@ -189,34 +189,100 @@ func (c *Client) exchange(m *Msg, a string) (r *Msg, rtt time.Duration, err erro
...
+func tcpMsgLen(t *net.TCPConn) (int, error) {
+	p := [2]byte{0, 0}
+	n, err := t.Read(p[:])
```
miekg: `p[:]` is older, just `p` should suffice.

client.go
```diff
@@ -189,34 +189,100 @@ func (c *Client) exchange(m *Msg, a string) (r *Msg, rtt time.Duration, err erro
...
+	if err != nil {
+		return 0, err
+	} else if n != 2 {
```
miekg: drop the else if and create of new if statement.

client.go
```diff
@@ -189,34 +189,100 @@ func (c *Client) exchange(m *Msg, a string) (r *Msg, rtt time.Duration, err erro
...
+		return 0, ErrShortRead
+	}
+	l, _ := unpackUint16(p[:], 0)
```
miekg: p[:] => p

client.go
```diff
@@ -189,34 +189,100 @@ func (c *Client) exchange(m *Msg, a string) (r *Msg, rtt time.Duration, err erro
...
+	if l == 0 {
+		return 0, ErrShortRead
+	}
+	return int(l), nil
+}
+
+// tcpRead - calls TCPConn.Read enough times to fill allocated buffer
```
miekg: drop the dash and add a closing .

client.go
```diff
@@ -189,34 +189,100 @@ func (c *Client) exchange(m *Msg, a string) (r *Msg, rtt time.Duration, err erro
...
+	if err != nil {
+		return 0, err
+	} else if n != 2 {
+		return 0, ErrShortRead
+	}
+	l, _ := unpackUint16(p[:], 0)
+	if l == 0 {
+		return 0, ErrShortRead
+	}
+	return int(l), nil
```
miekg: isn't `l` already an int?

client.go
asergeyev: _DNSHeaderSize?

client.go
asergeyev: I wanted to do static array, not slice here to force it being on stack. To turn it into a slice you need [:]... So I can do that probably anyway by removing 2 from [2] up top... I truly don't see any performance effect.

asergeyev:
Thanks for careful review. Done with all, just did not commit yet, want your take on couple of my notes above.

client.go
asergeyev: Nevermind, I decided to do it.

asergeyev:
Thanks for careful review... I decided not to wait on your comments, I think I followed most of your recommendations here.

miekg:
Ack, two naming nits. But it fix those shortly.

Issue #227, opened 2015-07-13
Detect invalid ACE?


rafaeljusto:
Hi,

I would like to know if in this library is there any way to detect an invalid domain ACE format when using the idn package? (FromPunycode)

For example, the FQDN **xn--7warambrecmdctfmhlrtrmbpyjdccaaahkrgl4chlip9bclcgddbb4ooioa.bnnhg** according to [Verisign](http://mct.verisign-grs.com/convertServlet?input=xn--7warambrecmdctfmhlrtrmbpyjdccaaahkrgl4chlip9bclcgddbb4ooioa.bnnhg) fails due to `Error:: Disallowed/Unassigned 3A3`, but the FromPunycode returns `σειράτάξησυπουργείωνΣύνθεσηυπουργικούσυμβουλίουουουο.bnnhg`.

```go
package main

import (
	"github.com/miekg/dns/idn"
)

func main() {
	output := idn.FromPunycode("xn--7warambrecmdctfmhlrtrmbpyjdccaaahkrgl4chlip9bclcgddbb4ooioa.bnnhg")
	println(output)
}
```

Best regards,
Rafael Justo

rafaeljusto:
I found a way to check for errors, converting the unicode back to ACE format and comparing with the initial ACE format. Will lose some performance to convert back to ACE, but it's a solution. =)

```go
package main

import (
	"github.com/miekg/dns/idn"
)

func main() {
	aces := []string{
		"xn--7warambrecmdctfmhlrtrmbpyjdccaaahkrgl4chlip9bclcgddbb4ooioa.bnnhg",
		"xn--a-bcdefghi-k4amdkp8de4at4dxdxcgrz4k6b.com.br",
	}

	for _, ace := range aces {
		unicode := idn.FromPunycode(ace)

		if idn.ToPunycode(unicode) != ace {
			println("invalid ACE -->", ace)
		} else {
			println("valid ACE   -->", ace)
		}
	}
}
```

Thanks!

Best regards,
Rafael Justo

rafaeljusto:
Looking closely to the RFCs, looks like the algorithm is accepting invalid code points, according to [RFC 5892 (page 4)](https://tools.ietf.org/html/rfc5892#page-4).

```
o  DISALLOWED: Those that should clearly not be included in IDNs.
    Code points with this property value are not permitted in IDNs.
```

[...]

PR #229, opened 2015-07-16
change function return behaviour on error

rafaeljusto:
See #227

### DIFF
--- a/idn/punycode.go
+++ b/idn/punycode.go
@@ -145,7 +145,7 @@ func tfunc(k, bias rune) rune {
 
 // encode transforms Unicode input bytes (that represent DNS label) into
 // punycode bytestream. This function would return nil if there's an invalid
-// character in the label
+// character in the label.
 func encode(input []byte) []byte {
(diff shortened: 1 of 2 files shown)


idn/punycode.go
```diff
@@ -139,14 +143,16 @@ func tfunc(k, bias rune) rune {
 	return k - bias
 }
 
-// encode transforms Unicode input bytes (that represent DNS label) into punycode bytestream
+// encode transforms Unicode input bytes (that represent DNS label) into
+// punycode bytestream. This function would return nil if there's an invalid
+// character in the label
```
miekg: label -> label.

miekg:
@asergeyev think this looks ok (as did the other PR). What do you think?

idn/punycode.go
rafaeljusto: done!

asergeyev:
Good to me now. Before it was just feeling irrational to return unicode, nothing was conceptually wrong with it.

PR #231, opened 2015-07-20
Correct CAA len method and octet packing buffer

rolandshoemaker:
This fixes two bugs in my original CAA implementation, one in the record length calculation and another where the packing buffer was of size 0 and would always cause a out of bounds slice panic, and adds a test case that verifies the fix.


msg.go
```diff
@@ -921,7 +921,7 @@ func packStructValue(val reflect.Value, msg []byte, off int, compression map[str
 				copy(msg[off:off+len(s)], s)
 				off += len(s)
 			case `dns:"octet"`:
-				bytesTmp := make([]byte, 0)
+				bytesTmp := make([]byte, 256*4+1)
```
miekg: This is some average size that makes it more efficient?

miekg:
Looks good.

msg.go
michaelharo: I'm unclear what 256*4+1 is.  I see it's used more than once in this file.  Should it be a constant?

miekg: it is the size we expect before we need to enlarge it. What the average binary size for CAA here?

miekg: And yes a constant would be better.

rolandshoemaker: I actually meant to add a comment asking about this, I just borrowed the buffer size here from the size used for TXT record contents. CAA value bodies will *generally* contain a domain name, but can contain arbitrary data, from a quick scan at the RFC they don't mention a strict limit just saying

```
The length of the value field is specified implicitly as the remaining length of the enclosing Resource Record data field.
``` 

It might be best to set the buffer size to 255, which I believe is the maximum size for a single string in a TXT or SPF record, since from at least what I've seen the value should generally take up far less.

miekg: 256 sgtm, the actual limit is 64K-ish bytes.

PR #236, opened 2015-08-04
add interfaces to allow packet-level inspection for pre/post processing

jdef:
I've been thinking of adding support for "dnstap" to mesos-dns. Having direct access to the DNS data frames would be *much* more efficient (and accurate) than packing/unpacking Msg objects.

This PR defines additional, optional fields to Server that allow callers to optionally inject pre/post processing for DNS data frames.

### DIFF
--- a/server.go
+++ b/server.go
@@ -47,6 +47,7 @@ type response struct {
 	tcp            *net.TCPConn      // i/o connection if TCP was used
 	udpSession     *SessionUDP       // oob data to get egress interface right
 	remoteAddr     net.Addr          // address of the client
+	writer         Writer            // writer to output the raw DNS bits
 }
 
 // ServeMux is an DNS request multiplexer. It matches the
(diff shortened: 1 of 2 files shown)


jdef:
xref https://github.com/mesosphere/mesos-dns/pull/212

miekg:
Yes, nice. Adding some small comments.

server.go
```diff
@@ -225,6 +251,10 @@ type Server struct {
 	Unsafe bool
 	// If NotifyStartedFunc is set is is called, once the server has started listening.
 	NotifyStartedFunc func()
+	// ReaderBuilder is optional, allows customization of the process that reads DNS frames
+	ReaderBuilder ReaderBuilder
```
miekg: Don't really like the 'Builder' part of the name, would ReaderFunc be more appropriate? And the comment needs a closing dot.

server.go
```diff
@@ -197,6 +198,31 @@ func HandleFunc(pattern string, handler func(ResponseWriter, *Msg)) {
 	DefaultServeMux.HandleFunc(pattern, handler)
 }
 
+type Writer interface {
+	io.Writer
+}
+
+type Reader interface {
```
miekg: Some docs strings here?

miekg:
..and maybe a small Example function that shows a manipulation, like set the message id to zero (or whatever)?

jdef:
@miekg I think I've addressed your concerns, let me know if there's anything else you'd like to see in this PR.

I went with DecorateReader/Writer instead of Reader/WriterFunc because the later seems to more closely align with the func-as-an-interface-adapter pattern (like Handler/HandlerFunc) which is definitely not the case here. 'Decorate' seems pretty explicit as to the intent.

server.go
```diff
@@ -197,6 +198,43 @@ func HandleFunc(pattern string, handler func(ResponseWriter, *Msg)) {
 	DefaultServeMux.HandleFunc(pattern, handler)
 }
 
+// Writer writes DNS data frames; each call to Write should send an entire frame.
```
miekg: I would just use DNS message as I use this term throughout the lib. Or maybe (raw) DNS message to distinguish them from the (parsed) DNS message.

miekg:
Looks good. Nit about using the word 'frame' which is not a term used in DNS (a lot).

jdef:
updated docs, PTAL

server.go
```diff
@@ -197,6 +198,43 @@ func HandleFunc(pattern string, handler func(ResponseWriter, *Msg)) {
...
+	return dr.readUDP(conn, timeout)
+}
+
+// DecorateReader is a decorator hook for extending or supplanting the functionality of a Reader.
+// Implementations should never return a nil Reader.
+type DecorateReader func(Reader) Reader
+
+// DecorateWriter is a decorator hook for extending or supplanting the functionality of a Writer.
+// Implementations should never return a nil Writer.
+type DecorateWriter func(Writer) Writer
```
tsenart: Nice :-)

PR #240, opened 2015-08-07
Adding DNS debugger android app

sajal:
I wrote an experimental Android app to make DNS queries. It uses this library behind the scenes. Additional info : http://www.sajalkayan.com/post/android-apps-golang.html


miekg:
Thanks!

PR #242, opened 2015-08-10
Do simple prechecks to avoid heap allocations

ErikDubbelboer:
In one of our programs we do a `ToPunycode()` on all URLs to make sure they are properly encoded. 99% of the URLs don't need any encoding but I noticed we are still seeing a lot of heap allocations being done in `punycode.go`. This pull request adds some checks to `ToPunycode()` and `FromPunycode()` to prevent unneeded heap allocations.

`needpuny()` doesn't make any heap allocations as demonstrated by: http://play.golang.org/p/F3u7F4ZILD

### DIFF
--- a/idn/punycode.go
+++ b/idn/punycode.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"strings"
 	"unicode"
+	"unicode/utf8"
 
 	"github.com/miekg/dns"
 )
@@ -27,9 +28,15 @@ const (
 )
 
 // ToPunycode converts unicode domain names to DNS-appropriate punycode names.
-// This function would return an empty string result for domain names with
+// This function will return an empty string result for domain names with
 // invalid unicode strings. This function expects domain names in lowercase.
 func ToPunycode(s string) string {
+	// Early check to see if encoding is needed.
+	// This will prevent making heap allocations when not needed.
+	if !needToPunycode(s) {
+		return s
+	}
+
 	tokens := dns.SplitDomainName(s)
 	switch {
 	case s == "":
@@ -51,7 +58,14 @@ func ToPunycode(s string) string {
 }
 
 // FromPunycode returns unicode domain name from provided punycode string.
+// This function expects punycode strings in lowercase.
 func FromPunycode(s string) string {
+	// Early check to see if decoding is needed.
+	// This will prevent making heap allocations when not needed.
+	if !needFromPunycode(s) {
+		return s
+	}
+
 	tokens := dns.SplitDomainName(s)
 	switch {
 	case s == "":
@@ -141,6 +155,51 @@ func tfunc(k, bias rune) rune {
 		return _MAX
 	}
 	return k - bias
+}
+
+// needToPunycode returns true for strings that require punycode encoding
+// (contain unicode characters).
+func needToPunycode(s string) bool {
+	// This function is very similar to bytes.Runes. We don't use bytes.Runes
+	// because it makes a heap allocation that's not needed here.
+	for i := 0; len(s) > 0; i++ {
+		r, l := utf8.DecodeRuneInString(s)
+		if r > 0x7f {
+			return true
+		}
+		s = s[l:]
+	}
+	return false
+}
+
+// needFromPunycode returns true for strings that require punycode decoding.
+func needFromPunycode(s string) bool {
+	if s == "." {
+		return false
+	}
+
+	off := 0
+	end := false
+	pl := len(_PREFIX)
+	sl := len(s)
+
+	// If s starts with _PREFIX.
+	if sl > pl && s[off:off+pl] == _PREFIX {
+		return true
+	}
+
+	for {
+		// Find the part after the next ".".
+		off, end = dns.NextLabel(s, off)
+		if end {
+			return false
+		}
+		// If this parts starts with _PREFIX.
+		if sl-off > pl && s[off:off+pl] == _PREFIX {
+			return true
+		}
+	}
+	panic("dns: not reached")
 }
 
 // encode transforms Unicode input bytes (that represent DNS label) into
--- a/idn/punycode_test.go
+++ b/idn/punycode_test.go
@@ -8,9 +8,9 @@ import (
 var testcases = [][2]string{
 	{"", ""},
 	{"a", "a"},
-	{"A-B", "a-b"},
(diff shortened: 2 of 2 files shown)


idn/punycode.go
```diff
@@ -52,6 +59,12 @@ func ToPunycode(s string) string {
 
 // FromPunycode returns unicode domain name from provided punycode string.
 func FromPunycode(s string) string {
+	// Early check to see if decoding is needed.
+	// This will prevent making heap allocations when not needed.
+	if !strings.Contains(s, _PREFIX) {
```
miekg: strings.HasPrefix?

idn/punycode.go
```diff
@@ -30,6 +31,12 @@ const (
 // This function would return an empty string result for domain names with
 // invalid unicode strings. This function expects domain names in lowercase.
 func ToPunycode(s string) string {
+	// Early check to see if encoding is needed.
+	// This will prevent making heap allocations when not needed.
+	if !needpuny(s) {
+		return strings.ToLower(s)
```
miekg: Why `ToLower` here? The doc already mentions that `s` should be lowercased.

miekg:
Some nits, looks good.

idn/punycode.go
ErikDubbelboer: When `s` is not lowercased the result will still be lowercased. Without the ToLower the tests fail. I suggest to modify the tests and document this behavior better.

idn/punycode.go
ErikDubbelboer: That won't work since it can be in any of the labels. For example `www.xn--z-0ub.com`.

miekg: [ Quoting <notifications@github.com> in "Re: [dns] Do simple prechecks to av..." ]
>> @@ -52,6 +59,12 @@ func ToPunycode(s string) string {
>>
>>  // FromPunycode returns unicode domain name from provided punycode string.
>>  func FromPunycode(s string) string {
>> +	// Early check to see if decoding is needed.
>> +	// This will prevent making heap allocations when not needed.
>> +	if !strings.Contains(s, _PREFIX) {
>
>That won't work since it can be in any of the labels. For example `www.xn--z-0ub.com`.

But I can have miekxn--miek as a label. If you want to do this properly you need
to Split and check if one of the labels start with the _PREFIX.

/Miek

--
Miek Gieben

idn/punycode.go
miekg: [ Quoting <notifications@github.com> in "Re: [dns] Do simple prechecks to av..." ]
>> @@ -30,6 +31,12 @@ const (
>>  // This function would return an empty string result for domain names with
>>  // invalid unicode strings. This function expects domain names in lowercase.
>>  func ToPunycode(s string) string {
>> +	// Early check to see if encoding is needed.
>> +	// This will prevent making heap allocations when not needed.
>> +	if !needpuny(s) {
>> +		return strings.ToLower(s)
>
>When `s` is not lowercased the result will still be lowercased. Without the ToLower the tests fail. I suggest to modify the tests and document this behavior better.

Fair enough.

How can the docs be better?

/Miek

--
Miek Gieben

idn/punycode.go
ErikDubbelboer: But doing a split already allocates memory on the heap which is what we want to prevent in most cases with this check. That's why I'm doing the Contains which might return true for some cases where we don't need to decode anything. But that is ok cause then the other decode code just won't do anything. And it's is only in a very small number of cases.

I guess I could write a quick loop with `dns.NextLabel` to check with `strings.HasPrefix` for `_PREFIX`. This would be more correct but it would also be slower.

Which option do you prefer?

idn/punycode.go
ErikDubbelboer: I would suggest removing the ToLower and documenting that upper case characters might be lowercased in the result.

idn/punycode.go
miekg: [ Quoting <notifications@github.com> in "Re: [dns] Do simple prechecks to av..." ]
>> @@ -52,6 +59,12 @@ func ToPunycode(s string) string {
>>
>>  // FromPunycode returns unicode domain name from provided punycode string.
>>  func FromPunycode(s string) string {
>> +	// Early check to see if decoding is needed.
>> +	// This will prevent making heap allocations when not needed.
>> +	if !strings.Contains(s, _PREFIX) {
>
>But doing a split already allocates memory on the heap which is what we want to
>prevent in most cases with this check. That's why I'm doing the Contains which
>might return true for some cases where we don't need to decode anything. But
>that is ok cause then the other decode code just won't do anything. And it's is
>only in a very small number of cases.

True, but it is very wrong in those cases.

>I guess I could write a quick loop with `dns.NextLabel` to check with
>`strings.HasPrefix` for `_PREFIX`. This would be more correct but it would also
>be slower.
>
>Which option do you prefer?

I think the loop with NextLabel is the right thing to do.

/Miek

--
Miek Gieben

ErikDubbelboer: It's not very wrong. It will do exactly the same thing as it does now for those cases.

The NextLabel solution would be more correct but I guess also quite a bit slower. So the question is if you want to slow down the general case just to handle those special cases a bit better.

miekg: How much slower would it be?

The thing that irks me a bit is that this is library code and it shoud "do the right thing"(tm).

ErikDubbelboer:
I did some tests and Contains is faster than a similar function with NextLabel. Only the solutions with NextLabel doesn't require a strings.ToLower to be correct which makes it overall faster than the Contains solution.

I have updated the pull request with this solution.

idn/punycode.go
```diff
@@ -143,6 +157,46 @@ func tfunc(k, bias rune) rune {
...
+		r, l := utf8.DecodeRuneInString(s)
+		if r > 0x7f {
+			return true
+		}
+		s = s[l:]
+	}
+	return false
+}
+
+func needFromPuny(s string) bool {
```
miekg: Could add a little bit of function doc here and fmt this code?

miekg:
The travis failure is my fault. For the rest this looks good.

ErikDubbelboer:
I added a docstring and cleaned up the code a bit. It's basically the same as [dns.Split()](https://github.com/miekg/dns/blob/0fe10e90c4a2cb63d270c21e5090a99ab383fc9c/labels.go#L107).

PR #245, opened 2015-08-12
fix some data races in Server

### LINKED ISSUE #244: race conditions detected on Go 1.5rc1
The letsencrypt/boulder project has found some race conditions in the dns library on Go 1.5rc1. They appear to be real.

Here's a Travis CI build that exemplifies the issue: https://travis-ci.org/jmhodges/boulder/builds/75167987 (Search for "DATA RACE" in the build logs but beware that other non-dns races were found, as well.).

The issue appears to be a race when Server.Shutdown and Server.ListenAndServe are called in different goroutines. It appears there are a few fields on srv that are being written to without a mutex. Specifically, Listener, and PacketConn. I've not gone looking for other obvious races on srv fields. ListenAndServe/Shutdown have not yet been run in multiple goroutines in the miekg/dns test suite.

jmhodges:
This avoids some issues with Server.PacketConn and Server.Listener being accessed outside of their locks.

Fixes #244.

### DIFF
--- a/server.go
+++ b/server.go
@@ -288,7 +288,6 @@ func (srv *Server) ListenAndServe() error {
 	}
 	srv.stopUDP, srv.stopTCP = make(chan bool), make(chan bool)
 	srv.started = true
-	srv.lock.Unlock()
 	addr := srv.Addr
 	if addr == "" {
 		addr = ":domain"
@@ -307,6 +306,7 @@ func (srv *Server) ListenAndServe() error {
 			return e
 		}
 		srv.Listener = l
+		srv.lock.Unlock()
 		return srv.serveTCP(l)
 	case "udp", "udp4", "udp6":
 		a, e := net.ResolveUDPAddr(srv.Net, addr)
@@ -321,8 +321,10 @@ func (srv *Server) ListenAndServe() error {
 			return e
 		}
 		srv.PacketConn = l
+		srv.lock.Unlock()
 		return srv.serveUDP(l)
 	}
+	srv.lock.Unlock()
 	return &Error{err: "bad network"}
 }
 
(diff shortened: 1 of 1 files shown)


server.go
```diff
@@ -288,7 +288,6 @@ func (srv *Server) ListenAndServe() error {
 	}
 	srv.stopUDP, srv.stopTCP = make(chan bool), make(chan bool)
 	srv.started = true
-	srv.lock.Unlock()
```
miekg: just make it `defer srv.lock.Unlock()` and remove the `Unlock`s futher down

jmhodges: Can't do that because it would leave it locked for the entirety of the serve lifetimes.

server.go
```diff
@@ -367,7 +371,6 @@ func (srv *Server) Shutdown() error {
 		return &Error{err: "server not started"}
 	}
 	srv.started = false
-	srv.lock.Unlock()
```
miekg: Same here, just `defer srv.lock.Unlock()` after we acquired the lock. If you do that I think the rest of the function can actually stay the same.

server.go
jmhodges: (That is to say, serveTCP and serveUDP never return.)

miekg: ah, yes. :-(

server.go
```diff
@@ -336,20 +338,22 @@ func (srv *Server) ActivateAndServe() error {
...
 	srv.lock.Unlock()
-	if srv.PacketConn != nil {
+	if pConn != nil {
 		if srv.UDPSize == 0 {
 			srv.UDPSize = MinMsgSize
 		}
-		if t, ok := srv.PacketConn.(*net.UDPConn); ok {
+		if t, ok := pConn.(*net.UDPConn); ok {
 			if e := setUDPSocketOptions(t); e != nil {
 				return e
```
michaelharo: Do you need to Unlock when returning error?

jmhodges: Nope, because the lock is for the fields on srv while UDPConn and Listener are expected to be thread-safe.

server.go
jmhodges: Ah, I see you merged this in. I was totally willing to do this one but was going to mention being a little worried about future accidents because the scope of data the lock would cover would become more broad.

But, hey, cool with me either way.

miekg: [ Quoting <notifications@github.com> in "Re: [dns] fix some data races in Se..." ]
>> @@ -367,7 +371,6 @@ func (srv *Server) Shutdown() error {
>>  		return &Error{err: "server not started"}
>>  	}
>>  	srv.started = false
>> -	srv.lock.Unlock()
>
>Ah, I see you merged this in. I was totally willing to do this one but was going to mention being a little worried about future accidents because the scope of data the lock would cover would become more broad.
>
>But, hey, cool with me either way.

Yeah, just looking after the merged code. Going to change it a little, to do use
`defer`, just makes it nicer and leaves the thing unlocked after the method
returns.

/Miek

--
Miek Gieben

PR #247, opened 2015-08-16
added proxy support for TCP DNS query

hugozhu:
To support socks5 proxy while performing DNS queries via TCP (eg. 8.8.8.8, 8.8.4.4)
the proxy can be ssh -D to ensure security.

### DIFF
--- a/client.go
+++ b/client.go
@@ -4,10 +4,14 @@ package dns
 
 import (
 	"bytes"
+	"errors"
 	"golang.org/x/net/proxy"
 	"io"
+	"log"
 	"net"
 	"net/url"
+	"strconv"
+	"strings"
 	"time"
 )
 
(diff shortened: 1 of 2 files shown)


miekg:
Thanks, but a couple of things. I want the dns package to only depend on std library packages and I think this patch adds some functionally that is easily replicated outside Go dns.

hugozhu:
My script is at: https://github.com/hugozhu/dns_proxy  It seems to me I can't add proxy support to DNS client outside Go dans.

miekg:
But you can use the dns with any `conn` you give it:
```
       // Create a proxy-conn, called c
       co := &dns.Conn{Conn: c} // c is your net.Conn
        co.WriteMsg(m)
        in, err  := co.ReadMsg()
        co.Close()
```
Which is your `Exchange` function, you might need to lift out some things from Exchange, but that's it. There is even `ExchangeConn` but that one is slated for removal.

miekg:
Closing this, 'cause I think this should be possible to to outside the lib. If not, please proof me wrong and reopen a PR.

PR #248, opened 2015-08-19
Travis: -> new architecture; - short mode; + race detector

FiloSottile:
If sudo is not needed, Travis will run the builds in a faster containerized architecture.

Removed "-short", added "-race". Wall time loss seems reasonable.


FiloSottile:
There seems to be a test in `client_test.go` that depends on a wall time race.

Issue #225, opened 2015-06-26
Create a msg.Deduplicate function/method


miekg:
Copy/improve SkyDNS' dedup <https://github.com/skynetservices/skydns/blob/0078e9f5139e334e75b6f16f1faa4c938ed79954/server/server.go#L875> function over in this lib.

miekg:
Another being tested/implemented in SkyDNS: short CNAMEs before A/AAAA in a message. This is not to break stub-resolvers like glibc.

miekg:
Dedup should work on a `[]dns.RR`. Should it also sort CNAMEs before A records? If so `Cleanup` might be a better name (or at least something different than Dedup). It also ties in with #223 because dedupping makes the msg smaller.

In general I think we want:
* `Dedup([]RR)`
* `SomethingSomething(dns.Msg)` that does dedup as one of the steps.

miekg:
Adding some commments from @asergeyev in here

* ideally work on wireformat and get the TTL out from there, dropping the whole Sprintf (for now at least)
* check the TTLs and keep the longest value when dedupping
* Look at CNAMEs as well, maybe DNAME and friends and filter those too
* Normalize the domain name before comparing, because of DNS case preservation crap
* A Go `map` might not be the best thing for this.

michaelharo:
It looks like this post has been changed since it arrived in email.  Re TTL longest vs shortest values...

http://tools.ietf.org/html/rfc2181#section-5.2

Should an authoritative source send such a malformed RRSet, the
   client should treat the RRs for all purposes as if all TTLs in the
   RRSet had been set to the value of the lowest TTL in the RRSet.  In
   no case may a server send an RRSet with TTLs not all equal.

miekg:
[ Quoting <notifications@github.com> in "Re: [dns] Create a msg.Deduplicate ..." ]
>It looks like this post has been changed since it arrived in email.  Re TTL longest vs shortest values...

Yes, I edited after I saved. Github sends immediately.

>http://tools.ietf.org/html/rfc2181#section-5.2
>
>Should an authoritative source send such a malformed RRSet, the
>   client should treat the RRs for all purposes as if all TTLs in the
>   RRSet had been set to the value of the lowest TTL in the RRSet.  In
>   no case may a server send an RRSet with TTLs not all equal.

Yep, indeed we want lowest.

/Miek

--
Miek Gieben

miekg:
[...]

Issue #223, opened 2015-06-18
Add a ShrinkToSize


miekg:
> @miekg maybe something to determine the appropriate payload size given the  
> transport and clients message, and a packer that shrinks the Additional section
> if it can or drops the body and truncates otherwise (doing the right thing by  
> Opt, RRSets, TSIG, etc).                                                       

Yes, ShrinkToSize() or somehting., I'll open an issue.

miekg:
I'm playing around with a  `Fit` function in https://github.com/skynetservices/skydns/blob/caching/server/msg.go . That will probably move to go-dns at some point.

miekg:
A more generic `dns.Fit(size int) error` function would

* Outright kill the additional section, except in case of delegation when in-baliwick glue is needed.
* Dedup what is left
* more ??
* should it set the tc bit? If so, this function won't return an error. A message will always Fit

PR #253, opened 2015-08-24
Add Dedup function - WIP do not merge

miekg:
Add a function to dedup an set of RRs. For this to work nicely I needed
something to format an RR with a Sprintf like function. So add that too.
For the Sprintf to work a rdataString() for each RR would be nice: add
that too. The %r modifier in Sprintf allows you to access this.

Don't really want to add RdataString() as an exported function to each
RR, as this add a whole load of extra methods. So just a Sprintf
function.

Also add Dedup function that uses all this.

### DIFF
--- a/dns.go
+++ b/dns.go
@@ -30,6 +30,9 @@ type RR interface {
 	Header() *RR_Header
 	// String returns the text representation of the resource record.
 	String() string
+
+	// rdataString returns the text representation of the rdata of record.
+//	rdataString() string
 	// copy returns a copy of the RR
 	copy() RR
 	// len returns the length (in octets) of the uncompressed RR in wire format.
--- a/format.go
+++ b/format.go
@@ -94,3 +94,48 @@ func Field(r RR, i int) string {
 	}
 	return ""
 }
+
+// Sprintf formats according to the format specifier and returns the
+// resulting string. The following verbs are understood by this
+// function:
+//
+//	%n	the ownername
+//	%c	the class
+//	%t	the type
+//	%T	the ttl
+//	%r	the rdata
+//	%%	literal percent sign
+//
+// In the future modifiers and more verbs might be added.
+func Sprintf(format string, r RR) string {
+	s := ""
+	verb := false
+	for _, f := range format {
+		if f == '%' {
+			verb = true
+			continue
+		}
+		if verb {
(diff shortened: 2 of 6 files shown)


miekg:
@asergeyev I could use a second pair of eyes on this. I'm working on #223 and #225. I think these changes in this PR make sense *and* are usable for users of this lib, specifically Sprintf() and maybe rdataString too? Anyway I you could take a small look? That would be really cool! thanks

asergeyev:
Why not to dedup using wire format?

Few comments would be in lined into code...

asergeyev:
"OK with comments"...

asergeyev:
To comment my point about dedupping wire format....

    PackStruct(rr)

gives set of bytes, where you have TTL in 4 bytes starting at len(name)+6 (where name has final dot and +7 without it)  Idea is to pack RR, then set TTL to 0,0,0,0 and then use result as key for comparisons....

miekg:
Re: about using wiredata: that makes sense, but you cannot use a map then as keys can't be sliced types.

asergeyev:
You can convert buffer to string and then reuse buffer to wirepack next rr.
But I'm not pushing on it, it's just another approach.

miekg:
That's true and memory wise, it will only cost the extra allocation for the []byte. I like it.

asergeyev:
It would be ideal at some way of evolving go-dns to have dns/wire sublib that operates on bytes only... No strings or structs yet :)) But it sounds scary and time-costly to get into such thing...

miekg:
Yeah, but most of the time, having such things as strings and structs is just so easy :-)

asergeyev:
I agree... for a general purpose lib anyway...

If you make some decision on how you want to do it (with pack/ or with rdatastring) and need another review just mention it please... I'll hold off from code comments until then.

I really do not have a good opinion about adding/keeping rdataString method and making it public. On one side I found those things useful in past (Net::DNS life and so on), but I also do not see it being a necessity, everything I've done with go-dns so far was easy to write without having separate Stringer for rdata... and once you make it public it's hard to pull back :)

miekg:
Agreed :-)
I'm going with your suggestion to use the binary data, will have something ready in the next our to tmrw :)

PR #254, opened 2015-08-24
Dedup rrset - WIP do not merge.

miekg:
New approach: work on the binary data and normalize that. Few fixes:
like looking the TTL of the records and a few new problems, like
randomizing the return value.


miekg:
@asergeyev something like this... Although the memory req. are prolly way over the top... Pondering a string based approach again, but then writing my own compare function, NormalizedRRCompare or something, that disregard character casing, and just skips the TTL.

asergeyev:
Close this?

PR #255, opened 2015-08-24
Add Dedup function - take 4 - wip don't merge

miekg:
Add function that dedups a list of RRs. Work on strings, which
adds garbage, but seems to be the least intrusive and takes the
last amount of memory.

Some fmt changes snook in as well.

### DIFF
--- a/sanitize.go
+++ b/sanitize.go
@@ -117,6 +117,8 @@ func normalizedString(r RR) (string, int) {
 // needsDeletion checks if the RR is masked by either a CNAME or a DNAME.
 // If so it return true.
 func needsDeletion(r RR, s string, cname, dname []string) bool {
+	// TODO(miek): fix.
+	// This is too broad, but we have to be care full not to delete ourselves.
 	if r.Header().Rrtype == TypeCNAME || r.Header().Rrtype == TypeDNAME {
 		return false
 	}
--- a/sanitize.go
+++ b/sanitize.go
@@ -1,10 +1,11 @@
 package dns
 
 // Dedup removes identical RRs from rrs. It preserves the original ordering.
+// The lowest TTL of any duplicates is used in the remaining one.
 // TODO(miek): CNAME, DNAME 'n stuff.
 func Dedup(rrs []RR) []RR {
 	m := make(map[string]RR)
-	keys := []string{}
+	keys := make([]string, 0, len(rrs))
 
 	for _, r := range rrs {
 		key := normalizedString(r)
@@ -19,8 +20,7 @@ func Dedup(rrs []RR) []RR {
 		m[key] = r
 	}
 	// If the length of the result map equals the amount of RRs we got,
-	// it means they were all different. We can then just return the original
-	// rrset.
+	// it means they were all different. We can then just return the original rrset.
 	if len(m) == len(rrs) {
 		return rrs
 	}
--- a/sanitize.go
+++ b/sanitize.go
@@ -8,28 +8,28 @@ package dns
 // if it finds a: a.miek.nl. CNAME foo, all other RRs with the ownername a.miek.nl.
 // will be removed. When a DNAME is found all RRs with an ownername below that of
 // the DNAME will be removed.
+// Note that the class of the CNAME/DNAME is *not* taken into account. TODO(miek)?
 func Dedup(rrs []RR) []RR {
 	m := make(map[string]RR)
 	keys := make([]string, 0, len(rrs))
-	var cname map[nameClass]bool
-	var dname map[nameClass]bool
 
-	for _, r := range rrs {
+	/*
+		make separate step that remove cname, dname
 		switch r.Header().Rrtype {
 		case TypeCNAME:
-			if cname == nil {
-				cname = make(map[nameClass]bool)
-			}
-			cname[nameClass{r.Header().Name, r.Header().Class}] = true
+			cname = append(cname, strings.ToLower(r.Header().Name))
 		case TypeDNAME:
-			if dname == nil {
-				dname = make(map[nameClass]bool)
-			}
-			dname[nameClass{r.Header().Name, r.Header().Class}] = true
+			dname = append(dname, strings.ToLower(r.Header().Name))
 		default:
-			if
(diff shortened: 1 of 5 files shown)


miekg:
Ok, @asergeyev this works and seems to be the simplest solution thus far. Not that different than iterating over a byte slice, but less memory in the normal case (I think).
(thanks for pushing me to try alternatives btw :) )

sanitize.go
```diff
@@ -0,0 +1,76 @@
+package dns
+
+// Dedup removes identical RRs from rrs. It preserves the original ordering.
+// TODO(miek): CNAME, DNAME 'n stuff.
+func Dedup(rrs []RR) []RR {
+	m := make(map[string]RR)
+	keys := []string{}
```
asergeyev: make([]string, 0, len(rrs))  maybe?

asergeyev:
I might be tired, do not see anything that worries me, this looks good.

miekg:
[ Quoting <notifications@github.com> in "Re: [dns] Add Dedup function - take..." ]
>I might be tired, do not see anything that worries me, this looks good.

Thanks, applied your comment.

/Miek

--
Miek Gieben

miekg:
@asergeyev soo, this was very much non-trivial. Just finished the tests; this finally looks to be working OK. I'm also adding a benchmark, because I think it can somewhat be optimized.

sanitize.go
```diff
@@ -0,0 +1,143 @@
+package dns
+
+import "strings"
+
+// Dedup removes identical RRs from rrs. It preserves the original ordering.
+// The lowest TTL of any duplicates is used in the remaining one.
+//
+// TODO(miek): This function will be extended to also look for CNAMEs and DNAMEs.
```
asergeyev: I wonder if this should be optional... or in another function now... Not sure where this would be used.

sanitize.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	return ret
+}
+
+// normalizedString returns a normalized string from r. The TTL
+// is removed and the domain name is lowercased. The returned integer
+// is the index where the domain name ends + 1.
+func normalizedString(r RR) (string, int) {
+	// A string Go DNS makes has: domainname<TAB>TTL<TAB>...
```
asergeyev: maybe comment that DomainName<TAB>TTL<TAB>CLASS<TAB>TYPE<TAB>RDATA is turned into  lowercasename<TAB>CLASS<TAB>TYPE....

sanitize.go
miekg: Yeah. Have the same feeling...it is also expensive to check for these.

sanitize.go
```diff
@@ -0,0 +1,143 @@
...
+			if m[key].Header().Ttl > r.Header().Ttl {
+				m[key].Header().Ttl = r.Header().Ttl
+			}
+			continue
+		}
+
+		if r.Header().Rrtype == TypeCNAME {
+			// we do end+3 here, so we capture the full domain name *and*
+			// the class field which mnemonic is always two chars.
+			cname = append(cname, key[:end+3])
```
asergeyev: "always two letters" does not hold since there is ANY class :(

miekg: True. But it will still work, because there is no overlap. But indeed needs
commenting.

sanitize_test.go
```diff
@@ -0,0 +1,122 @@
...
+// BenchMark test as well TODO(miek)
+
+func TestNormalizedString(t *testing.T) {
+	tests := map[RR]string{
+		newRR(t, "mIEk.Nl. 3600 IN A 127.0.0.1"):     "miek.nl.\tIN\tA\t127.0.0.1",
+		newRR(t, "m\\ iek.nL. 3600 IN A 127.0.0.1"):  "m\\ iek.nl.\tIN\tA\t127.0.0.1",
+		newRR(t, "m\\\tIeK.nl. 3600 in A 127.0.0.1"): "m\\tiek.nl.\tIN\tA\t127.0.0.1",
+	}
+	for tc, expected := range tests {
+		a1, _ := normalizedString(tc)
```
asergeyev: maybe worth to check that end value points to space...

sanitize_test.go
```diff
@@ -0,0 +1,122 @@
...
+			if o.String() != expected[i] {
+				t.Fatalf("expected %v, got %v", expected[i], o.String())
+			}
+		}
+	}
+}
+
+func TestDedupWithCNAMEDNAME(t *testing.T) {
+	testcases := map[[4]RR][]string{
+		[...]RR{
```
asergeyev: in 1.5 you don't really need [...]RR part but that would not compile in 1.4/1.3 if you remove it...  :(

sanitize.go
```diff
@@ -0,0 +1,143 @@
...
+// is the index where the domain name ends + 1.
+func normalizedString(r RR) (string, int) {
+	// A string Go DNS makes has: domainname<TAB>TTL<TAB>...
+	b := []byte(r.String())
+
+	// find the first non-escaped tab, then another, so we capture
+	// where the TTL lives.
+	esc := false
+	ttlStart, ttlEnd := 0, 0
+	for i, c := range b {
```
asergeyev: just btw...   if you do plain for loop on i here, you can add "&& ttlEnd == 0" into condition...  and code below could be in a single big switch without  continue/break statements

sanitize.go
asergeyev: ah. good...  I wonder if end is  len(rr.Header().Name) or no

asergeyev:
Well... most important point is that sometimes you don't want to dedup CNAME/DNAME stuff... e.g. when recursive sends packet it really appends everything and client looks at most specific record in an answer...  So that's why I'm thinking dedup and remove records masked by aliases if needed....

asergeyev:
I also wonder if you can leave this as wip after you done for a day or so to play with.

miekg:
Thanks for your comments. I'm travelling atm. I'm removing the cname/dname stuff (but you can pull it out from the git history).
There is not rush, although I want to use this function in SkyDNS - to cleanup some stuff there.

Issue #250, opened 2015-08-21
fix failing test: TestSingleInflight


miekg:
```
github.com/miekg/dns
github.com/miekg/dns/idn
20.15s$ go test -race -bench=.
--- FAIL: TestSingleSingleInflight-2 (0.01s)
	client_test.go:200: all rtts should be equal.  got 678575 want 756598
	client_test.go:200: all rtts should be equal.  got 699434 want 756598
	client_test.go:200: all rtts should be equal.  got 699434 want 756598
	client_test.go:200: all rtts should be equal.  got 1252951 want 756598
	client_test.go:200: all rtts should be equal.  got 1252951 want 756598
	client_test.go:200: all rtts should be equal.  got 1252951 want 756598
	client_test.go:200: all rtts should be equal.  got 1252951 want 756598
	client_test.go:200: all rtts should be equal.  got 580212 want 756598
	client_test.go:200: all rtts should be equal.  got 580212 want 756598
```

miekg:
Can't reproduce locally (with 1.4 nor 1.5)

asergeyev:
TestSingleSingleInflight not TestSingleSingleInflight, right?

You're testing INT64 time values there, they all carry variability from goroutine scheduling in golang core, you have decent chances to hit outlier in first element and therefore have error all the way thru the test.

IMO testing RTT being certain number is not helpful at all... VMs are noisy and even on solid 100% dedicated for this test server you can see NTPD shifting time during the test and it'll fail anyway. 

If you strongly desire to keep this test it's possible to see that test is generating set of 100+ values where:
* none are 0
* on histogram, majority of values hit same bucket as the median

IMO it's still all overkill...   Premise of the test is that there is no extra time spent on things except the latency between server and client but, since your server is process-inlined and latency is measured in nano and micro-seconds it's unlikely to have that information derived from RTT. You'd always pick go runtime and system noise.

miekg:
Agreed.

asergeyev:
Remove this test or change what it's checking?

miekg:
[ Quoting <notifications@github.com> in "Re: [dns] fix failing test: TestSin..." ]
>Remove this test or change what it's checking?

Remove (didn't get around to doing that)

asergeyev:
in case if you want to test c.SingleInflight feature we could design/discuss possible replacement.

So far the feature actually has:

    // This adds a bunch of garbage, TODO(miek).

attached to code.. maybe something to look into?

[...]

PR #261, opened 2015-09-17
Slightly updated singleinflight code and added test...

asergeyev:
Just to discuss these changes in #250, not for merging yet.

### DIFF
--- a/client.go
+++ b/client.go
@@ -116,16 +116,9 @@ func (c *Client) Exchange(m *Msg, a string) (r *Msg, rtt time.Duration, err erro
 	if cl1, ok := ClassToString[m.Question[0].Qclass]; ok {
 		cl = cl1
 	}
-	r, rtt, err, shared := c.group.Do(m.Question[0].Name+t+cl, func() (*Msg, time.Duration, error) {
+	return c.group.Do(m.Question[0].Name+t+cl, func() (*Msg, time.Duration, error) {
 		return c.exchange(m, a)
 	})
-	if err != nil {
-		return r, rtt, err
-	}
-	if shared {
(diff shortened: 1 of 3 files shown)


client.go
```diff
@@ -116,16 +116,9 @@ func (c *Client) Exchange(m *Msg, a string) (r *Msg, rtt time.Duration, err erro
...
 	}
-	r, rtt, err, shared := c.group.Do(m.Question[0].Name+t+cl, func() (*Msg, time.Duration, error) {
+	return c.group.Do(m.Question[0].Name+t+cl, func() (*Msg, time.Duration, error) {
 		return c.exchange(m, a)
 	})
-	if err != nil {
-		return r, rtt, err
-	}
-	if shared {
-		return r.Copy(), rtt, nil
```
miekg: I think this .Copy() is quite important...

singleinflight.go
```diff
@@ -30,28 +32,42 @@ type singleflight struct {
...
+	if ok { // wait in line and copy the result
+		c.dup = true
 		g.Unlock()
-		c.wg.Wait()
-		return c.val, c.rtt, c.err, true
+		c.Wait()
+		if c.val == nil {
+			return c.val, c.rtt, c.err
+		}
+		return c.val.Copy(), c.rtt, c.err
```
miekg: ah, you copy here...

miekg:
What is it you actually improving here?

asergeyev:
I wasn't really planning to change singleinflight (at least yet), I was just poking at things and it could be rolled back.. Focus was on creating a  test... but now I see that it failed anyway and wondering why it happens...

PR #265, opened 2015-09-23
Allow supplying a Conn in Transfer struct

### LINKED ISSUE #264: Setting the source address of an AXFR
Hello! I hope an issue is the right avenue for this request/question. If not feel free to point me in the right direction and shame me appropriately.

I'd like to essentially do this:
https://www.socketloop.com/tutorials/golang-set-up-source-ip-address-before-making-http-request

But for AXFRs. There is a use case where you might want to send your AXFR (or regular query) out of a specific interface or IP on a box that has permission to poke through the firewall.

I believe this can be accomplished by setting `LocalAddr` on the `Dialer` that eventually gets used in `Dial`, or `DialTimeout`: http://golang.org/src/net/dial.go?s=5630:5708#L42

I'm trying to figure out the best way to do this in this library. Perhaps you can already do it, and I just don't know how :smile: 

What I've come up with:

Add a `DialTimeoutSource` in `client.go` that looked something like:
```go
// DialTimeoutSource acts like Dial but takes a timeout and a source.
func DialTimeoutSource(network, address string, timeout time.Duration, source Addr) (conn *Conn, err error) {
	conn = new(Conn)
	d := net.Dialer{Timeout: timeout, LocalAddr: source}
	conn.Conn, err = d.Dial(network, address)
	if err != nil {
		return nil, err
	}
	return conn, nil
}
```
and make use of that in `func (t *Transfer) In` in `xfr.go`, by appending a `Source` field to the `Transfer` struct. 

Something like:
```go
// A Transfer defines parameters that are used during a zone transfer.
type Transfer struct {
	*Conn
....
	tsigTimersOnly bool
	Source         *Addr            // A source IP for the transfer
}

// Think we need to away to stop the transfer

// In performs an incoming transfer with the server in a.
func (t *Transfer) In(q *Msg, a string) (env chan *Envelope, err error) {
	timeout := dnsTimeout
	source := ""
	if t.DialTimeout != 0 {
		timeout = t.DialTimeout
	}
	if t.Source != nil {
		source = t.Source
	}
	t.Conn, err = DialTimeoutSource("tcp", a, timeout, source)
```

This doesn't solve the problem generally for other types of queries. I'll probably end up making this type of change for my personal use, and I'd be happy to submit a pull request. But if there's something better, more general, or cleaner that someone more familiar with the library could suggest, I'd be happy to implement that as well.

I'd love to know any ideas anyone has, or if I should do something different, or just sod off :rocket: 

Thank you @miekg or anyone else that has ideas!

TimSimmons:
- This allows a user to call `Dial` before they call `Transfer.In`,
thus specifying the attributes of a [Dialer](https://golang.org/src/net/dial.go?s=5630:5708#L17)
before their AXFR

This is what it looks like to use:
```go
d := net.Dialer{LocalAddr: transfer_source}
c, _ := d.Dial("tcp", *master)
dnscon := &dns.Conn{Conn:c}
transfer = &dns.Transfer{Conn: dnscon}
channel, err := transfer.In(message, *master)
```
closes #264

### DIFF
--- a/xfr.go
+++ b/xfr.go
@@ -23,14 +23,26 @@ type Transfer struct {
 // Think we need to away to stop the transfer
 
 // In performs an incoming transfer with the server in a.
+// If you would like to set the source IP, or some other attribute
+// of a Dialer for a Transfer, you can do so by specifying the attributes
+// in the Transfer.Conn:
+//
+//	d := net.Dialer{LocalAddr: transfer_source}
+//	con, _ := d.Dial("tcp", *master)
+//	dnscon := &dns.Conn{Conn:con}
+//	transfer = &dns.Transfer{Conn: dnscon}
+//	channel, err := transfer.In(message, *master)
+//
 func (t *Transfer) In(q *Msg, a string) (env chan *Envelope, err error) {
 	timeout := dnsTimeout
 	if t.DialTimeout != 0 {
 		timeout = t.DialTimeout
 	}
-	t.Conn, err = DialTimeout("tcp", a, timeout)
-	if err != nil {
-		return nil, err
+	if t.Conn == nil {
+		t.Conn, err = DialTimeout("tcp", a, timeout)
+		if err != nil {
+			return nil, err
+		}
 	}
 	if err := t.WriteMsg(q); err != nil {
 		return nil, err


miekg:
Looks good. Can you add the PR description (the code in there) as documentation to the Tranfer function. I.e. something like "if you want to set the source address uses for this transfer you can use the following: ... "
Otherwise this might be really hard to discover.

TimSimmons:
Done @miekg thank you for your prompt feedback :+1:

miekg:
Thanks. I will slightly rework the doc (i.e. just put the err vars in there, so force people cut&pasting this to actually handle those errors) :)

TimSimmons:
Thank you!

miekg:
+1

PR #270, opened 2015-10-05
Fix (*HINFO).len() and HINFO canonicalization in rawSignatureData

FiloSottile:
Those len() functions should really be autogenerated

### DIFF
--- a/dnssec.go
+++ b/dnssec.go
@@ -638,6 +638,9 @@ func rawSignatureData(rrset []RR, s *RRSIG) (buf []byte, err error) {
 			x.Target = strings.ToLower(x.Target)
 		case *DNAME:
 			x.Target = strings.ToLower(x.Target)
+		case *HINFO:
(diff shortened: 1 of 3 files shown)


andrewtj:
[RFC 6840 - Clarifications and Implementation Notes for DNS Security (DNSSEC)](https://tools.ietf.org/html/rfc6840#section-5.1):

    Section 6.2 of [RFC4034] also erroneously lists HINFO as a record
    that needs conversion to lowercase, and twice at that.  Since HINFO
    records contain no domain names, they are not subject to case
    conversion.

miekg:
rolled forward and removed the HINFO downcasing.

FiloSottile:
Yeah, assumed too fast from reading the comment. This should probably be automated from the struct tags. Thanks for patching.

10 minutes after my commit @ogud was already sending me this https://www.rfc-editor.org/errata_search.php?rfc=4034 ^^

PR #272, opened 2015-10-06
client: SetReadDeadline after write, before read

tsenart:
The current code sets the read deadline at the same time as the write
deadline. If the write nearly times out but doesn't, the read timeout
can fire before the read happens within the specified deadline.


miekg:
thanks!

PR #274, opened 2015-10-07
Generate programmatically per-type code, including .len() and .copy()

FiloSottile:
image

Killed 536 lines of repetitive code, replaced with 281 lines of generation code.

This process apparently fixes a bug in `NSAPPTR.len()`, after a similar one was found in `HINFO.len()`, and bugs in `MB.copy()` and `CNAME.copy()` (using `sprintName` for copying), `IPSECKEY.copy()` (missed `copyIP`) and `OPT.copy()` (partially, `Options` was not copied as slice; `EDNS0` objects themselves are still pointers).

This should also make it easier to make changes to these functions, and check their correctness.

Generate the code by running "go generate".

Note for the future: wether records fit in their `len()` is a good fuzzing target.


FiloSottile:
Also now all lengths/copy methods are defined right next to the type, and appear in code right next to the field name, instead of being aggregated numbers in a oneliner.

FiloSottile:
Needs a bit of comments/documentation in the generation file before being merged.

types.go
```diff
@@ -1544,9 +1224,6 @@ type CAA struct {
 	Value string `dns:"octet"`
 }
 
-func (rr *CAA) Header() *RR_Header { return &rr.Hdr }
-func (rr *CAA) copy() RR           { return &CAA{*rr.Hdr.copyHeader(), rr.Flag, rr.Tag, rr.Value} }
-func (rr *CAA) len() int           { return rr.Hdr.len() + 2 + len(rr.Tag) + len(rr.Value) }
 func (rr *CAA) String() string {
```
miekg: huh? Still here?

miekg:
Very nice. Thanks!

miekg:
Is there a convention on how to call generated fiales? Don't really like `_auto`, but something like `_generated` looks too much like types_generated

davecheney:
The (weak) convention in the stdlib is to prefix the file with `z`, ie, `zsyscall_linux.arm64.go`

FiloSottile:
Yeah, I didn't like _auto either. Moved to ztypes, thanks @davecheney.

Added a couple comments, should be good to review/merge.

I was going to do also `String()` but then I noticed it was 6am (east-bound jetlag is the worst).

miekg:
[ Quoting <notifications@github.com> in "Re: [dns] Generate programmatically..." ]
>Yeah, I didn't like _auto either. Moved to ztypes, thanks @davecheney.
>
>Added a couple comments, should be good to review/merge.

Ack.

>I was going to do also `String()` but then I noticed it was 6am (east-bound jetlag is the worst).

you're getting old :)

/Miek

--
Miek Gieben

FiloSottile:
DNM, I used the old static analysis packages.

https://groups.google.com/forum/#!topic/golang-dev/V0n9dBpPSIA

FiloSottile:
Fixed to use the new hot cool stuff.

miekg:
here goes nothing!

andrewtj:
@FiloSottile have you got your eye on generating code for anything else?

FiloSottile:
Probably String, but no ETA, so tell me if you want to put a mutex on it :)

PR #277, opened 2015-10-16
Expose TypeToRR

FiloSottile:


miekg:
Not against it per se, but why? (think any extension of the api needs some discussion)

FiloSottile:
Sure. I have a data model that we can imagine like this:

```go
type Records struct {
    RRType uint16
    TTL uint32
    Values []json.RawMessage
}
```

Once I decode a Records object, I know the objects I want to decode the Values into, but I have no way to instantiate them without TypeToRR.

With TypeToRR I just use the table and then Unmarshal the Values[] into the RR.

PR #281, opened 2015-10-30
Instead of removing all RRs on Truncated, attempt to unpack

fastest963:
This Pull Request fixes the `TODO(miek): this isn't the best strategy!` regarding `Truncated` messages received. Instead of just returning no RRs it's now attempting to do as many RRs as possible until it gets an error then returning `ErrTruncated` along with the partial response. This allows someone to naively just check `err != nil` if they didn't know about truncated messages or someone could check if `err == ErrTruncated` additionally and proceed with the truncated response if they're okay with a truncated response.

More realistically, I don't think servers will be returning partial responses (like how skydns removes the `Extra` first), but it supports that. If a non-partial response is returned (that was truncated) then no error is returned. I can make it still return a `ErrTruncated` but since there was no error unpacking the message, I didn't think it made sense to return an error.

### DIFF
--- a/msg.go
+++ b/msg.go
@@ -54,6 +54,9 @@ var (
 	ErrSoa error = &Error{err: "no SOA"}
 	// ErrTime indicates a timing error in TSIG authentication.
 	ErrTime error = &Error{err: "bad time"}
+	// ErrTruncated indicates that we failed to unpack a truncated message.
+	// We unpacked as much as we had so Msg can still be used, if desired.
+	ErrTruncated error = &Error{err: "failed to unpack truncated message"}
 )
 
 // Id, by default, returns a 16 bits random number to be used as a
@@ -1238,8 +1241,8 @@ func unpackStructValue(val reflect.Value, msg []byte, off int) (off1 int, err er
 						continue
 					}
 				}
-				if off == lenmsg {
-					// zero rdata foo, OK for dyn. updates
+				if off == lenmsg && int(val.FieldByName("Hdr").FieldByName("Rdlength").Uint()) == 0 {
+					// zero rdata is ok for dyn updates, but only if rdlength is 0
 					break
 				}
 				s, off, err = UnpackDomainName(msg, off)
@@ -1394,6 +1397,32 @@ func UnpackRR(msg []byte, off int) (rr RR, off1 int, err error) {
 		return &h, end, &Error{err: "bad rdlength"}
 	}
 	return rr, off, err
+}
+
+// unpackRRslice unpacks msg[off:] into an []RR.
+// If we cannot unpack the whole array, then it will return nil
+func unpackRRslice(l int, msg []byte, off int) (dst1 []RR, off1 int, err error) {
+	var r RR
+	// Optimistically make dst be the length that was sent
+	dst := make([]RR, 0, l)
+	for i := 0; i < l; i++ {
+		off1 := off
+		r, off, err = UnpackRR(msg, off)
+		if err != nil {
+			off = len(msg)
+			break
+		}
+		// If offset does not increase anymore, l is a lie
+		if off1 == off {
+			l = i
+			break
+		}
+		dst = append(dst, r)
+	}
+	if err != nil && off == len(msg) {
+		dst = nil
+	}
+	return dst, off, err
 }
 
 // Reverse a map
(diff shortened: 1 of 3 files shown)


miekg:
Nice!

> If a non-partial response is returned (that was truncated) then no error is returned. I can make it still  return a ErrTruncated but since there was no error unpacking the message, I didn't think it made sense to return an error.

So if a server sends a truncated response, but it just happens to fall within an RRset boundery no error is returned? If think in that case we should still return ErrTruncated...

fastest963:
I've updated the PR and made it so it will return an `ErrTruncated` no matter if it ran into an error during RRSets or not. The only exception to this is if it failed to unpack the questions, if any errors occur during unpacking questions, it returns that error.

msg.go
```diff
@@ -1396,6 +1399,32 @@ func UnpackRR(msg []byte, off int) (rr RR, off1 int, err error) {
 	return rr, off, err
 }
 
+// UnpackRRArr unpacks msg[off:] into an []RR.
+// If we cannot unpack the whole array, then it will return nil
+func UnpackRRArr(l int, msg []byte, off int) (dst1 []RR, off1 int, err error) {
```
miekg: unpackRRslice? Also don't think this needs to be exported.

fastest963: Gotcha, I only exported it because `UnpackRR` was exported.

msg.go
```diff
@@ -1594,84 +1623,51 @@ func (dns *Msg) Unpack(msg []byte) (err error) {
...
-		}
-		dns.Extra = append(dns.Extra, r)
+	// The header counts might have been wrong so we need to update it
+	dh.Nscount = uint16(len(dns.Ns))
+	if err == nil {
+		dns.Extra, off, err = UnpackRRArr(int(dh.Arcount), msg, off)
 	}
+	// The header counts might have been wrong so we need to update it
+	dh.Arcount = uint16(len(dns.Extra))
 	if off != len(msg) {
```
miekg: this looks really weird now, but I'll change that after I merge.

msg.go
```diff
@@ -1594,84 +1623,51 @@ func (dns *Msg) Unpack(msg []byte) (err error) {
...
 		// TODO(miek) make this an error?
 		// use PackOpt to let people tell how detailed the error reporting should be?
 		// println("dns: extra bytes in dns packet", off, "<", len(msg))
+	} else if dns.Truncated {
+		// Whether we ran into a an error or not, we want to return that it
+		// was truncated
+		if err != nil {
+			println("dns: error unpacking truncated message", err.Error())
```
miekg: nobody will read this `println`, if this needs to be returned to the user somehow, we another err. Don't think we should take that route, so just drop this.

fastest963:
Okay I just updated the PR and addressed your comments.

bboreham:
I'm not so sure that erroring on a correctly-decoded response that the server declared to be truncated is correct.  Suppose someone is using DNS for load-spreading, and has many machines behind the same name; do you want users of your library to receive an error purely dependent on the value of 'many' compared to the size of their UDP buffer?

miekg:
But if you don't make it an error you can get the situation where a response happens to be correctly decoded (because the last RR just ended on byte 512 for example), while in other cases the parsing would fail and you would return an error.

I don't understand your example.

bboreham:
OK, maybe I misunderstand the behaviour of DNS servers; does the `TC` bit mean the response was arbitrarily truncated at the limit of the buffer, and never that the server truncated the response to return a whole number of RRs that fit in the buffer?  RFC1035 doesn't seem to nail it either way.

miekg:
I'm not sure actually. Think it is left to the server implementation which could do either. 

This is one of those times when (DNS) people do the "Let's see what BIND does"

rade:
From [RFC2181](https://tools.ietf.org/html/rfc2181#section-9)
> Where TC is set, the partial RRSet that would not completely fit may be left in the response.  When a DNS client receives a reply with TC set, it should *ignore that response*, and query again, using a mechanism, such as a TCP connection, that will permit larger replies.

So clients should not be looking at the contents when TC is set. No idea whether that reflects reality.

miekg:
A partial RRset means the packet should be completely valid. Dunno... think the current approach is most sensible @bboreham. Willing to drop the if truncated then err, but I think that is more complex than just always returning `ErrTruncated`.

rade:
> A partial RRset means the packet should be completely valid.

says who?

miekg:
I think I read that in "the partial RRSet", i.e the RRset is partial, but the RRs in the set are complete.

PR #283, opened 2015-10-30
Only re-try AcceptTCP() if the error is temporary

### LINKED ISSUE #282: Shutdown() of TCP server may cause infinite spin
`Shutdown` closes the listener, which causes `AcceptTCP()` to return an error "use of closed network connection".  However the [code around that](https://github.com/miekg/dns/blob/master/server.go#L437) is keen to keep trying:

````
	for {
		rw, e := l.AcceptTCP()
		if e != nil {
			continue
		}
````

It looks like this change came in with #271.

bboreham:
If I understand correctly what you might have meant here.

Fixes #282

### DIFF
--- a/server.go
+++ b/server.go
@@ -434,7 +434,10 @@ func (srv *Server) serveTCP(l *net.TCPListener) error {
 	for {
 		rw, e := l.AcceptTCP()
 		if e != nil {
-			continue
+			if neterr, ok := e.(net.Error); ok && neterr.Temporary() {
+				continue
+			}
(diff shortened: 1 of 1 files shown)


miekg:
What would be a non-temporary error? I.e. can you kill a running server by sending crap on the tcp socket?

bboreham:
ENOBUFS and EMFILE seem like good examples. But errClosing is the one specifically generated when you close the listener in Shutdown.

PR #285, opened 2015-11-18
More efficient A record (un)packing

cbranch:
The generic struct pack and unpack routines make heavy use of reflection to correctly translate between bytes and Go objects.

Given most questions and answers are A records, writing a specific A record packer/unpacker makes the fast path quicker and generates significantly less garbage.

The pack and unpack routines for specific data types were extracted from the generic routine so that they may be shared with the optimised routines.

Results of `go test -bench \(Unp\|P\)ackRRA`

Benchmark | Iterations | ns/op | B/op | allocs/op
--------- | ---------- | ----- | ---- | ---------
BenchmarkPackRRA (before) | 2000000 |       688 ns/op |     112 B/op |    7 allocs/op
BenchmarkPackRRA (after)  | 20000000 | 85.9 ns/op | 0 B/op   | 0 allocs/op
BenchmarkUnpackRRA (before) | 1000000 |      1443 ns/op |     272 B/op |     12 allocs/op
BenchmarkUnpackRRA (after) | 5000000  | 290 ns/op  | 128 B/op | 4 allocs/op

### DIFF
--- a/msg.go
+++ b/msg.go
@@ -1008,14 +1008,12 @@ func unpackStructValue(val reflect.Value, msg []byte, off int) (off1 int, err er
 						continue
 					}
 				}
-				if off == lenmsg {
-					break // dyn. update
+				var a net.IP
+				off, err = unpackIPv4(&a, msg, off)
+				if err != nil {
+					return lenmsg, err
 				}
-				if off+net.IPv4len > lenmsg {
-					return lenmsg, &Error{err: "overflow unpacking a"}
-				}
-				fv.Set(reflect.ValueOf(net.IPv4(msg[off], msg[off+1], msg[off+2], msg[off+3])))
-				off += net.IPv4len
+				fv.Set(reflect.ValueOf(a))
 			case `dns:"aaaa"`:
 				if val.Type().String() == "dns.IPSECKEY" {
 					// Field(2) is GatewayType, must be 2
@@ -1143,11 +1141,11 @@ func unpackStructValue(val reflect.Value, msg []byte, off int) (off1 int, err er
 				return lenmsg, err
 			}
 			if val.Type().Field(i).Name == "Hdr" {
-				lenrd := off + int(val.FieldByName("Hdr").FieldByName("Rdlength").Uint())
-				if lenrd > lenmsg {
-					return lenmsg, &Error{err: "overflowing header size"}
+				rdlength := uint16(val.FieldByName("Hdr").FieldByName("Rdlength").Uint())
+				msg, err = truncateMsgFromRdlength(msg, off, rdlength)
+				if err != nil {
+					return lenmsg, err
 				}
-				msg = msg[:lenrd]
 				lenmsg = len(msg)
 			}
 		case reflect.Uint8:
@@ -1160,24 +1158,19 @@ func unpackStructValue(val reflect.Value, msg []byte, off int) (off1 int, err er
 			fv.SetUint(uint64(uint8(msg[off])))
 			off++
 		case reflect.Uint16:
-			if off == lenmsg {
-				break
-			}
 			var i uint16
-			if off+2 > lenmsg {
-				return lenmsg, &Error{err: "overflow unpacking uint16"}
+			i, off, err = unpackStructUint16(msg, off, lenmsg)
+			if err != nil {
+				return lenmsg, err
 			}
-			i, off = unpackUint16(msg, off)
 			fv.SetUint(uint64(i))
 		case reflect.Uint32:
-			if off == lenmsg {
-				break
+			var i uint32
+			i, off, err = unpackStructUint32(msg, off, lenmsg)
+			if err != nil {
+				return lenmsg, err
 			}
-			if off+4 > lenmsg {
-				return lenmsg, &Error{err: "overflow unpacking uint32"}
-			}
-			fv.SetUint(uint64(uint32(msg[off])<<24 | uint32(msg[off+1])<<16 | uint32(msg[off+2])<<8 | uint32(msg[off+3])))
-			off += 4
+			fv.SetUint(uint64(i))
 		case reflect.Uint64:
 			if off == lenmsg {
 				break
@@ -1331,6 +1324,10 @@ func unpackUint16(msg []byte, off int) (uint16, int) {
 	return uint16(msg[off])<<8 | uint16(msg[off+1]), off + 2
 }
 
+func unpackUint32(msg []byte, off int) (uint32, int) {
+	return uint32(uint64(uint32(msg[off])<<24 | uint32(msg[off+1])<<16 | uint32(msg[off+2])<<8 | uint32(msg[off+3]))), off + 4
+}
+
 func packUint16(i uint16) (byte, byte) {
 	return byte(i >> 8), byte(i)
 }
@@ -1359,6 +1356,87 @@ func fromBase64(s []byte) (buf []byte, err error) {
 	return
 }
 
+// Unpack a uint16 from a struct, computing the new offset and handling errors
+func unpackStructUint16(msg []byte, off int, lenmsg int) (i uint16, off1 int, err error) {
+	if off == lenmsg {
+		return 0, off, nil
+	}
+	if off+2 > lenmsg {
+		return 0, lenmsg, &Error{err: "overflow unpacking uint16"}
+	}
+	i, off = unpackUint16(msg, off)
+	return i, off, nil
+}
+
+// Unpack a uint32 from a struct, computing the new offset and handling errors
+func unpackStructUint32(msg []byte, off int, lenmsg int) (i uint32, off1 int, err error) {
+	if off == lenmsg {
+		return 0, off, nil
+	}
+	if off+4 > lenmsg {
+		return 0, lenmsg, &Error{err: "overflow unpacking uint32"}
+	}
+	i, off = unpackUint32(msg, off)
+	return i, off, nil
+}
+
+// Truncates the byte slice in msg to match the expected length of the RR.
+// Returns an error if the slice is smaller than the expected size.
+func truncateMsgFromRdlength(msg []byte, off int, rdlength uint16) (truncmsg []byte, err error) {
+	lenmsg := len(msg)
+	lenrd := off + int(rdlength)
+	if lenrd > lenmsg {
+		return msg, &Error{err: "overflowing header size"}
+	}
+	msg = msg[:lenrd]
+	return msg, nil
+}
+
+// Unpack an RR header, returning the offset to the end of the header and a
+// re-sliced msg according to the expected length of the RR.
+func unpackHeader(hdr *RR_Header, msg []byte, off int) (off1 int, truncmsg []byte, err error) {
+	lenmsg := len(msg)
+	if off == lenmsg {
+		return off, msg, nil
+	}
+	hdr.Name, off, err = UnpackDomainName(msg, off)
+	if err != nil {
+		return lenmsg, msg, err
+	}
+	hdr.Rrtype, off, err = unpackStructUint16(msg, off, lenmsg)
+	if err != nil {
+		return lenmsg, msg, err
+	}
+	hdr.Class, off, err = unpackStructUint16(msg, off, lenmsg)
+	if err != nil {
+		return lenmsg, msg, err
+	}
+	hdr.Ttl, off, err = unpackStructUint32(msg, off, lenmsg)
+	if err != nil {
+		return lenmsg, msg, err
+	}
+	hdr.Rdlength, off, err = unpackStructUint16(msg, off, lenmsg)
+	if err != nil {
+		return lenmsg, msg, err
+	}
+	msg, err = truncateMsgFromRdlength(msg, off, hdr.Rdlength)
+	return off, msg, nil
+}
+
+// Unpacks the IP stored in an A record.
+func unpackIPv4(a *net.IP, msg []byte, off int) (off1 int, err error) {
+	lenmsg := len(msg)
+	if off == lenmsg {
+		return off, nil // dyn. update
+	}
+	if off+net.IPv4len > lenmsg {
+		return lenmsg, &Error{err: "overflow unpacking a"}
+	}
+	*a = net.IPv4(msg[off], msg[off+1], msg[off+2], msg[off+3])
+	off += net.IPv4len
+	return off, nil
+}
+
 // PackRR packs a resource record rr into msg[off:].
 // See PackDomainName for documentation about the compression.
 func PackRR(rr RR, msg []byte, off int, compression map[string]int, compress bool) (off1 int, err error) {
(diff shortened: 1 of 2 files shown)


msg.go
```diff
@@ -1359,14 +1333,191 @@ func fromBase64(s []byte) (buf []byte, err error) {
...
+	}
+	off, err = packStructUint16(hdr.Rdlength, msg, off, lenmsg)
+	if err != nil {
+		return lenmsg, err
+	}
+	return off, nil
+}
+
+// Packs the IP stored in an A record.
+func packA(a net.IP, msg []byte, off int) (off1 int, err error) {
```
miekg: Longer term I feel go dns will be moving to a reflection-less impl. so merging this makes sense. Does it also make sense to make these pack functions part of the RR interface? So they become methods on the type?

cbranch:
PackA (accepting type A) makes total sense. The IPv4 packer currently named packA could be packIPv4 instead.

miekg:
Yep. Trick question: can these pack functions be auto generated? Just like
the other methods on rr types?

cbranch:
They can be autogenerated, except:

# the generator looks only at structs beginning with RR_Header. It would have to change to those structs implementing the RR interface, so we can generate RR_Header too.
# for simplicity and consistency, the compression data needed by PackDomainName should probably be stored in a 'packerContext' accepted by all Pack methods, instead of handling the cases where compression data either is required or not.
# from experience, debugging a code generator is harder than debugging code; it's better to get a single 'expected case' done by hand before you try doing the rest :)

+1 to the idea, save it for a rainy day!

miekg:
True. Can you rework this pr be a method on the A rr? (I'm on holiday so
not really responsive ATM)

cbranch:
Sure!

cbranch:
Is there anything else that needs to be done to this PR?

miekg:
[ Quoting <notifications@github.com> in "Re: [dns] More efficient A record (..." ]
>Is there anything else that needs to be done to this PR?

Dunno... didn't have to time to recheck it. Will do that this evening (UTC).

Thanks for you work!

/Miek

--
Miek Gieben

miekg:
Just to re-iterate on this. A long time ago I've added a branch that did away with all the reflection and used a per-record Pack/Unpack function. End result: not faster, so I've abandoned the effort. (don't know if this still lives somewhere in the git history).

Also a second concern is that this reflection-based (un)packer has been battle tested on the internet and has seem numerous updates. It is somewhat wasteful to put it aside and use a brand new code.

I'm pondering keeping this as a branch of some sorts, or something... dunno yet. Real life benchmarks would help, i.e 2 servers and something as plain of dnsperf (or something?)

jfesler:
"Real life benchmarks would help, i.e 2 servers and something as plain of dnsperf (or something?)"

I spent some time on this today.  I've got both golang benchmark data, and dnsperf data.

As I outline what I have here I want to share my testing set up.   In some ways optimal, and others not.

## host information

Hypervisor:
- 2x: Intel(R) Xeon(R) CPU E5-2620 v3 @ 2.40GHz,  HT enabled.
- Translation: 12 cores total
- Translation: 24 "cpu" (hyper threading is enabled on this host), each claiming 1239.375 mhz.

Traffic Generator:
- VM with 4 "CPUs" allocated
- Happens to be my actual name server, which is usually idle.
- Running `yes "whoami.miek.nl. A" | dnsperf -s HOSTNAME -p 8100  -l 120`
Traffic Sync
- VM with 4 "CPUs" allocated
- Happens to be a general purpose server, where I do some development.  Also usually idle.
- Running variants of Miek's reflect
Network:
- kvm to kvm, same host - no physical interface.



Ideally, I'd take the time to carve out more dedicated VMs; but .. I'm stealing time where I can to get this done.  Take it all with a grain of salt.  All load tests are done for 2 minutes each, with several minute separation; and 3 tests total for each scenario.  Hopefully this should give some view of host repeatability.

All code is at https://github.com/jfesler/benchmark_golang_dns .

## dnsperf
###  test information

I've taken Miek's "reflect" (from github.com/miekg/exdns) as a starting point as something relatively simple, straightforward, and also reflective of what someone might want to do for simple projects. 

I'm using today's version of github.com/miekg/dns and github.com/miekg/exdns/reflect .

4 scenarios are tested:

scenario | description
------------|----------------
reflect_t1_stock  | stock dns, stock reflect, port 8100
reflect_t1p_stock | patched dns, stock reflect, port 8101
reflect_t2_notxt | stock dns, reflect has "TXT" removed (and related variables), port 8200
reflect_t2p_notxt | patched dns, reflect has "TXT" removed (and related variables), port 8200

###  data

scenario | qps | average
-----|-----|-----
reflect_t1_stock | 72761, 73253, 72528 | 72847
reflect_t1p_stock | 76906, 78539, 78909 | 78118
reflect_t2_notxt | 84428, 84679, 84618 | 84575
reflect_t2p_notxt | 90532, 93696, 91484 | 91904

### Observation
reflect_t1p_stock \> reflect_t1_stock by **7.2%** in terms of queries per second.
reflect_t2p_notxt \> reflect_t2_notxt by **8.7%** in terms of queries per second.


## golang benchmarks

I wanted to take what @cbranch did, and go a lot further.

I'm testing not just RR pack/unpack, but also message packing/unpacking.  Why benchmark this?  I feel this gives a better understanding of what people actually will do with the library.  Load seem stuff into the structure, and either WriteMsg  (or PackMsg and Write).   How much does improving a single RR help the overall message time?

I'm doing all benchmarks with both the patched and stock version of Miek's DNS library compiled in; and the code selects which is in use.

### Chris's original benchmarks for RR

We'll start off with Chris's benchmarks on my hardware.  The names are slightly different but the intent is the same.  My numbers vary with Chris's, presumably due to hardware differences.

```
BenchmarkRRPackAStock-4        	10000000	      1395 ns/op	     112 B/op	       7 allocs/op
BenchmarkRRPackAPatched-4      	100000000	       107 ns/op	       0 B/op	       0 allocs/op
BenchmarkRRUnpackAStock-4      	 5000000	      2557 ns/op	     272 B/op	      12 allocs/op
BenchmarkRRUnpackAPatched-4    	30000000	       594 ns/op	     128 B/op	       4 allocs/op
```

And IPv6 (AAAA) added in as well.   Almost 40% of the requests to my DNS servers, are for AAAA records - making this record type pretty important.

```
BenchmarkRRPackAAAAStock-4     	 1000000	      1435 ns/op	     112 B/op	       7 allocs/op
BenchmarkRRPackAAAAPatched-4   	10000000	       117 ns/op	       0 B/op	       0 allocs/op
BenchmarkRRUnpackAAAAStock-4   	  500000	      2757 ns/op	     272 B/op	      12 allocs/op
BenchmarkRRUnpackAAAAPatched-4 	 2000000	       627 ns/op	     128 B/op	       4 allocs/op
```

### Message Packing Benchmarks

```
BenchmarkMsgPackAStock-4       	 5000000	      3624 ns/op	     336 B/op	      18 allocs/op
BenchmarkMsgPackAPatched-4     	10000000	      2380 ns/op	     224 B/op	      11 allocs/op
BenchmarkMsgPackAAAAStock-4    	 5000000	      3898 ns/op	     352 B/op	      18 allocs/op
BenchmarkMsgPackAAAAPatched-4  	 5000000	      2397 ns/op	     224 B/op	      11 allocs/op
BenchmarkMsgPackTXTStock-4     	 3000000	      4770 ns/op	    1536 B/op	      20 allocs/op
BenchmarkMsgPackTXTPatched-4   	 3000000	      4763 ns/op	    1536 B/op	      20 allocs/op
```



### Message Unpacking Benchmarks

```
BenchmarkMsgUnpackAStock-4     	 3000000	      4701 ns/op	     400 B/op	      18 allocs/op
BenchmarkMsgUnpackAPatched-4   	 5000000	      2537 ns/op	     256 B/op	      10 allocs/op
BenchmarkMsgUnpackAAAAStock-4  	 3000000	      4488 ns/op	     400 B/op	      18 allocs/op
BenchmarkMsgUnpackAAAAPatched-4	 5000000	      2506 ns/op	     256 B/op	      10 allocs/op
BenchmarkMsgUnpackTXTStock-4   	 3000000	      5097 ns/op	     432 B/op	      20 allocs/op
BenchmarkMsgUnpackTXTPatched-4 	 3000000	      4898 ns/op	     416 B/op	      19 allocs/op
```


### Observation

Even at a per-message level, the numbers are quite impressive.   Between a third to a half of the CPU cost in "Pack" and "Unpack" can be saved.  How much this affects a given app, is debatable - which is why I include data for dnsperf + reflect for consideration.

Type | Pack Savings | Unpack Savings
------- | ----------------- | ---------------------
A | 34.3% | 46.0%
AAAA | 38.5% | 44.2%
TXT | - | 3.9%

miekg:
Very nice! Thanks.

I double check this PR today and see how we can gradually move away from reflection.

msg.go
```diff
@@ -1359,14 +1333,187 @@ func fromBase64(s []byte) (buf []byte, err error) {
...
+	if off+net.IPv4len > lenmsg {
+		return lenmsg, &Error{err: "overflow packing a"}
+	}
+	switch len(a) {
+	case net.IPv6len:
+		msg[off] = a[12]
+		msg[off+1] = a[13]
+		msg[off+2] = a[14]
+		msg[off+3] = a[15]
+		off += net.IPv4len
```
miekg: this should IPv6len, saw some other ipv4/ipv6 mixups else where. Still holding off on merging, as I am going to autogenerate most of it (hopefully)

miekg:
So... I've slowly started to incorporate some of this the no-reflect branch, where:
`msg_types.go`: contains all pack/unpack functions for all types we have (just started, only does A and AAAA)
`msg_util.go`: helper functions, some from this PR.

It should no be *too* hard (but still a lot of work). But my plan would be to write a `msg_generate.go` (in the same vain as `types_generate.go` that generates a `zmsg.go` when building.

miekg:
Ok, my `no-reflect` branch contains more bits (except actual packing/unpacking), but with it we can disable reflection on a per type basis and the methods are generated from `msg_generate.go`.
Stuff still doesn't work, but now it is just more filling in the blanks (I hope)

miekg:
@cbranch are you willing to take it from that `no-reflect` branch? I liberally copied from your PR anyway. As said, I think my pieces are in place to make go dns reflectionless and generate packers/unpackers.

Side note: maybe `unpack()` should be in the RR interface as well, just as `pack` is right now.

cbranch:
Sure thing. I'd like to get an improved benchmark for the different record types too, before starting - A records just happened to be the fast path for our (proprietary) application.

miekg:
[ Quoting <notifications@github.com> in "Re: [dns] More efficient A record (..." ]
>Sure thing. I'd like to get an improved benchmark for the different record types too, before starting - A records just happened to be the fast path for our (proprietary) application.

Cool. With this in place we can do a staged rollout: A and AAAA first and then
slowly add other records..

/Miek

--
Miek Gieben

miekg:
any luck with this?
Otherwise I am going to pick this up this/next week

cbranch:
Other issues took priority last month. Did you make any progress? If not, I'll start this week.

miekg:
No, me neither, tackled a bunch of smaller issues.

miekg:
I'm going to give this some love, this week + weekend (and next week, depending how far I will get)

yannchristensen:
Hi, I'm very interested in reducing GC pressure caused by the unpack routines. I was happy to see that there is a branch already under way. Is it functional enough to test out? Anything I can do to help move it along?  -cheers

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] More efficient A re..." ]
>Hi, I'm very interested in reducing GC pressure caused by the unpack routines. I was happy to see that there is a branch already under way. Is it functional enough to test out? Anything I can do to help move it along?  -cheers

Cool. Well, yes you can help ;-) I think most of the changes are mostly
mechanical.

1. msg_generate needs to actually emit code
2. the pack/unpack function needs to be called from msg.go when doing the
   packing/unpacking. This needs to happen in a way that we can flip types
   on. The original PR has code for this.

of course, this is all still not that trivial. I won't have any time for this
until Sunday.

/Miek

--
Miek Gieben

miekg:
So, I've copy-pasted most of this branch into #356, which is a work-in-progress. Right now it compiles :) But does nothing. The next steps are:
1. Generate pack/unpack for A and AAAA and check how mechanical this can be. For good measure maybe also generate them for MX (as that is a more complex record)
2. Tweak PackRR/UnpackRR to short circuit the reflection for these types
3. fuzz this setup a bit
4. Move more types
5. Kill off the `packStructValue` and `unpackStructValue`

miekg:
So 1. has been done. 2. Should not be that hard. And the rest is just more of the same.

miekg:
2. Has be done. 
3. Tests now panic :)

4. and 5. are really easy now :)

miekg:
The only failing test is `TestNoRdataUnpack`, probably related to some added length checks in the generated code.
I think I'm starting to get happy how this is looking.

miekg:
a bit stumped on why this test fails. I do think the length checks:
```
	if off+net.IPv6len > lenmsg {
		return nil, lenmsg, &Error{err: "overflow unpacking aaaa"}
	}
```
should be `off+net.IPv6len-1` because only the next packer/unpacker hits to new offset, but even with that change that test keeps failing.

@andrewtj can you double check my logic here? And the overall PR (if you have time). It's #356

andrewtj:
@miekg I can't get to it tonight but I'll put it on the list for tomorrow.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] More efficient A re..." ]
>@miekg I can't get to it tonight but I'll put it on the list for tomorrow.

np and thanks!

andrewtj:
Well, I'm kind of lost :)

I don't follow the logic of `dynamicUpdate()`: It gets called with offset and `len(msg)`, so won't that only be true when the zero-length rdata is the last RR? Could the length from `RR_Header` be checked in the RR unpackers instead?

The subjects in `TestNoRdataUnpack()` that fail, L32 and MX, both seem to fail because the packed rrdata isn't empty. If you hack the test to truncate the data after rdlen and set rdlen to zero, the test passes. Both L32 and MX start with a u16, and that seems to be what gets packed as the rdata is two zero bytes.

So I guess the old unpack style was more forgiving, although I'm not confident about that as I wrote a quick test to see if any other empty structs produce non-empty payloads, and I got failures from 4 *other* RRs but not those two. So I've got something twisted.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] More efficient A re..." ]
>Well, I'm kind of lost :)
>
>I don't follow the logic of `dynamicUpdate()`: It gets called with offset and `len(msg)`, so won't that only be true when the zero-length rdata is the last RR? Could the length from `RR_Header` be checked in the RR unpackers instead?

Good point.

>The subjects in `TestNoRdataUnpack()` that fail, L32 and MX, both seem to fail because the packed rrdata isn't empty. If you hack the test to truncate the data after rdlen and set rdlen to zero, the test passes. Both L32 and MX start with a u16, and that seems to be what gets packed as the rdata is two zero bytes.
>
>So I guess the old unpack style was more forgiving, although I'm not confident about that as I wrote a quick test to see if any other empty structs produce non-empty payloads, and I got failures from 4 *other* RRs but not those two. So I've got something twisted.

Thanks for looking at it! I will poke some more.

/Miek

--
Miek Gieben

miekg:
Fixed by properly checking the rdlength before packing/unpacking.

Hilariously it now fails on:
```
% go test
--- FAIL: TestRFC6605P256 (0.00s)
	dnssec_test.go:568: failure to validate the spec RRSIG: dns: bad signature
--- FAIL: TestRFC6605P384 (0.05s)
	dnssec_test.go:644: failure to validate the spec RRSIG: dns: bad signature
FAIL
exit status 1
FAIL	github.com/miekg/dns	12.216s
```
Which uses the same code, but quite funny because it is at the other side of the DNS spectrum :)

miekg:
Ah, this happens because we read the RR from a string (which does not set the rdlenght - long story). so The isdynamicUpdate thing it hit and the last bit of he RR is not encoded.

miekg:
Ok, so we always packed more (for some RRs) in Go dns, as there is no sane way to signal "nordata", i.e. in the case of mx the preference is 0 which is valid, so it gets packed. The solution is to check for "off == len(msg)" after each *UN*pack and then just return what we have packed so far. Done this, tests are happy again.

miekg:
Getting back to my list:

3. fuzz this setup a bit
4. Move more types
5. Kill off the packStructValue and unpackStructValue

fuzzing is in progress and got some results :(
More types have been moved
Killing of packStructValue and unpackStructValue will have to wait a bit.

IOW: I'm almost ready to merge #356 , @cbranch OK to close this PR?

PR #287, opened 2015-11-26
Fix for issue 286 - panic after closing ResponseWriter in server handler

### LINKED ISSUE #286: panic: runtime error: invalid memory address or nil pointer dereference
This panic happens when call w.Close() in a tcp dns server.

[signal 0xb code=0x1 addr=0x0 pc=0x73a807]

goroutine 378753 [running]:
github.com/miekg/dns.(*Server).readTCP(0xc82146fc70, 0x0, 0x1dcd65000, 0x0, 0x0, 0x0, 0x0, 0x0)
        /root/fastcat/src/github.com/miekg/dns/server.go:562 +0x5d7
github.com/miekg/dns.(*defaultReader).ReadTCP(0xc82017e6d8, 0x0, 0x1dcd65000, 0x0, 0x0, 0x0, 0x0, 0x0)
        /root/fastcat/src/github.com/miekg/dns/server.go:226 +0x5c
github.com/miekg/dns.(*Server).serve(0xc82146fc70, 0x7f9d658d9a48, 0xc8214a74d0, 0x7f9d658d9a20, 0xc82017fa68, 0xc82220caa0, 0xc, 0xc, 0x0, 0x0, ...)
        /root/fastcat/src/github.com/miekg/dns/server.go:552 +0x450
created by github.com/miekg/dns.(*Server).serveTCP
        /root/fastcat/src/github.com/miekg/dns/server.go:450 +0x3e9

https://github.com/miekg/dns/blob/master/server.go#L683
https://github.com/miekg/dns/blob/master/server.go#L555

andrewtj:
I think the test is okay. I'm not that familiar with `Server{}` but I'm pretty sure this is the only place a check is needed to fix #286.

### DIFF
--- a/server_test.go
+++ b/server_test.go
@@ -2,6 +2,7 @@ package dns
 
 import (
 	"fmt"
+	"io"
 	"net"
 	"runtime"
 	"sync"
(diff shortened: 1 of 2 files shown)


miekg:
ah, cool was just about to take a look myself. Looks legit. Except that the test has a data race.
https://travis-ci.org/miekg/dns/jobs/93314264#L580

server_test.go
```diff
@@ -389,6 +390,46 @@ func TestShutdownTCP(t *testing.T) {
...
+			if tries == 3 {
+				return
+			}
+			time.Sleep(time.Second / 10)
+			tries += 1
+			goto exchange
+		}
+	}()
+	server.ActivateAndServe()
+	if !h_triggered {
```
miekg: `h_triggerd` is causing the datarace, (also style nit of using an underscore)

PR #294, opened 2015-12-18
Expose methods to determine the connection type in the ResponseWriter

rguliyev:
Expose methods to determine the connection type in the ResponseWriter to be able to set the Truncate bit for replies bigger than 512 bytes for UDP, e.g.:
```
func handleDnsRequest(w dns.ResponseWriter, r *dns.Msg) {
        m := new(dns.Msg)
        m.SetReply(r)
        m.Compress = false

       // Code to generate the response...

        answerSize := m.Len()
        if answerSize > 512 && w.IsUDP() {
                m.Truncated = true
                m.Answer = nil
        }
        w.WriteMsg(m)
}
```

### DIFF
--- a/server.go
+++ b/server.go
@@ -38,6 +38,10 @@ type ResponseWriter interface {
 	// Hijack lets the caller take over the connection.
 	// After a call to Hijack(), the DNS package will not do anything with the connection.
 	Hijack()
+	// Returns true if it is a UDP connection
+	IsUDP() bool
+	// Returns true if it is a TCP connection
+	IsTCP() bool
 }
 
 type response struct {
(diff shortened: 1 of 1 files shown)


andrewtj:
I can see how this could be useful. With (D)TLS and HTTP(S) being looked into by various IETF WGs maybe an `IsVC() bool` (virtual circuit) method would be a future-proof addition/substitute, @miekg?

miekg:
Why can't something like this being used? (from https://github.com/miekg/exdns/blob/master/reflect/reflect.go#L67)
```
	if ip, ok := w.RemoteAddr().(*net.UDPAddr); ok {
		str = "Port: " + strconv.Itoa(ip.Port) + " (udp)"
		a = ip.IP
		v4 = a.To4() != nil
	}
	if ip, ok := w.RemoteAddr().(*net.TCPAddr); ok {
		str = "Port: " + strconv.Itoa(ip.Port) + " (tcp)"
		a = ip.IP
		v4 = a.To4() != nil
	}
```

andrewtj:
That didn't occur to me :blush: Determining payload sizes and how to handle XFRs were the uses I had in mind when I said this could be useful. In general for the former you'd want to check EDNS payload size and for the later you'd likely want to do more than just check that the transport is appropriate. So they're probably not as useful as I'd initially thought.

rguliyev:
That would work, thank!

PR #300, opened 2016-01-07
DNS over TLS

### LINKED ISSUE #297: Is it possible to support DNSCrypt
[DNSCrypt](https://dnscrypt.org/) protocol is widely used in China, hopefully, dns package could support this protocol.

rafaeljusto:
Developed DNS over TLS support as described in draft:
http://tools.ietf.org/html/draft-ietf-dprive-dns-over-tls-02

Tested with Unbound 1.4.22 with the following extra configuration:
```yaml
server:
  interface: 127.0.0.1@853
  interface: 127.0.0.1@53
  access-control: 127.0.0.0/8 allow
  verbosity: 4
  ssl-service-key: "/etc/unbound/dev.key"
  ssl-service-pem: "/etc/unbound/dev.cert"
  ssl-port: 853
```

I left the fallback strategy from TLS to normal DNS for who is using the client, as it will need to change the port from 853 to 53.

See #297

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ Send pull request if you want to be listed here.
 * EDNS0, NSID;
 * AXFR/IXFR;
 * TSIG, SIG(0);
-* DNS over TLS;
+* DNS over TLS: optional encrypted connection between client and server;
 * DNS name compression;
 * Depends only on the standard library.
 
--- a/README.md
+++ b/README.md
@@ -62,6 +62,7 @@ Send pull request if you want to be listed here.
 * EDNS0, NSID;
 * AXFR/IXFR;
 * TSIG, SIG(0);
+* DNS over TLS;
 * DNS name compression;
 * Depends only on the standard library.
 
@@ -136,6 +137,7 @@ Example programs can be found in the `github.com/miekg/exdns` repository.
 * 7314 - DNS (EDNS) EXPIRE Option
 * 7553 - URI record
 * xxxx - EDNS0 DNS Update Lease (draft)
+* yyyy - DNS over TLS: Initiation and Performance Considerations (draft)
 
 ## Loosely based upon
 
--- a/client.go
+++ b/client.go
@@ -4,6 +4,7 @@ package dns
 
 import (
 	"bytes"
+	"crypto/tls"
 	"io"
 	"net"
 	"time"
@@ -26,6 +27,7 @@ type Conn struct {
 type Client struct {
 	Net            string            // if "tcp" a TCP query will be initiated, otherwise an UDP one (default is "" for UDP)
 	UDPSize        uint16            // minimum receive buffer for UDP messages
+	TLS            bool              // enables TLS connection (port 853)
 	DialTimeout    time.Duration     // net.DialTimeout, defaults to 2 seconds
 	ReadTimeout    time.Duration     // net.Conn.SetReadTimeout value for connections, defaults to 2 seconds
 	WriteTimeout   time.Duration     // net.Conn.SetWriteTimeout value for connections, defaults to 2 seconds
@@ -152,11 +154,18 @@ func (c *Client) writeTimeout() time.Duration {
 
 func (c *Client) exchange(m *Msg, a string) (r *Msg, rtt time.Duration, err error) {
 	var co *Conn
-	if c.Net == "" {
-		co, err = DialTimeout("udp", a, c.dialTimeout())
-	} else {
-		co, err = DialTimeout(c.Net, a, c.dialTimeout())
+
+	network := "udp"
+	if c.Net != "" {
+		network = c.Net
 	}
+
+	if c.TLS {
+		co, err = DialTimeoutWithTLS(network, a, c.dialTimeout())
+	} else {
+		co, err = DialTimeout(network, a, c.dialTimeout())
+	}
+
 	if err != nil {
 		return nil, 0, err
 	}
@@ -378,6 +387,29 @@ func Dial(network, address string) (conn *Conn, err error) {
 func DialTimeout(network, address string, timeout time.Duration) (conn *Conn, err error) {
 	conn = new(Conn)
 	conn.Conn, err = net.DialTimeout(network, address, timeout)
+	if err != nil {
+		return nil, err
+	}
+	return conn, nil
+}
+
+// DialWithTLS connects to the address on the named network with TLS.
+func DialWithTLS(network, address string) (conn *Conn, err error) {
+	conn = new(Conn)
+	conn.Conn, err = tls.Dial(network, address, nil)
+	if err != nil {
+		return nil, err
+	}
+	return conn, nil
+}
+
+// DialTimeoutWithTLS acts like DialWithTLS but takes a timeout.
+func DialTimeoutWithTLS(network, address string, timeout time.Duration) (conn *Conn, err error) {
+	var dialer net.Dialer
+	dialer.Timeout = timeout
+
+	conn = new(Conn)
+	conn.Conn, err = tls.DialWithDialer(&dialer, network, address, nil)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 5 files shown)


miekg:
nice! Can you add docs to the README saying this is supported. Any maybe some tests?

client.go
```diff
@@ -26,6 +27,8 @@ type Conn struct {
 type Client struct {
 	Net            string            // if "tcp" a TCP query will be initiated, otherwise an UDP one (default is "" for UDP)
 	UDPSize        uint16            // minimum receive buffer for UDP messages
+	TLS            bool              // enables TLS connection (port 853)
```
miekg: Do we *need* the bool, or does a nil TLSConfig imply no TLS?

rafaeljusto: I thought about someone using TLS with all default parameters, in this case TLSConfig would be nil and the TLS flag true. But anyway, if we use only TLSConfig, the user could just do new(TLS.Config) if he wants TLS with all the default values.

rafaeljusto:
Sure! I will work on docs and tests.

client.go
andrewtj: With dnscrypt and TLS as transport security options (and maybe DTLS as another option again) some sort of flag set would make sense to me.

miekg: @andrewtj and actual flag.FlagSet? Or a ClientOpt{} struct that can be filled with options? Or a plain string that we overload, just like Network?

andrewtj: @miekg I had a bit field in a new type in mind, but now that you mention it `Net`'s tempting—could that be reused?

andrewtj:
The commit message in `1c62bb27` contains `Resolves #297` which will [close](https://help.github.com/articles/closing-issues-via-commit-messages/) #297. Since #297 is about dnscrypt that's not really appropriate.

client.go
miekg: [ Quoting <notifications@github.com> in "Re: [dns] DNS over TLS (#300)..." ]
>> @@ -26,6 +27,8 @@ type Conn struct {
>>  type Client struct {
>>  	Net            string            // if "tcp" a TCP query will be initiated, otherwise an UDP one (default is "" for UDP)
>>  	UDPSize        uint16            // minimum receive buffer for UDP messages
>> +	TLS            bool              // enables TLS connection (port 853)
>
>@miekg I had a bit field in a new type in mind, but now that you mention it `Net`'s tempting—could that be reused?

But net allows "tcp6" and "udp4", should iwe support that as well?
So:

tcp6-tls
tcp4-tls
tcp-tls

for enabling this?

Seems rather clean to put this in the Net string.

/Miek

--
Miek Gieben

miekg:
Are you also planning some server side knobs tot listen on port 853 with
TLS config?

rafaeljusto:
@miekg The idea of using the net is good for me. We could rewrite it in **Client.exchange** to avoid errors in stdlib (net.UnknownNetworkError). I was thinking on something like this:

```go
network := "udp"
tls := false

switch c.Net {
case "tcp-tls":
  network = "tcp"
  tls = true
case "tcp4-tls":
  network = "tcp4"
  tls = true
case "tcp6-tls":
  network = "tcp6"
  tls = true
default:
  if c.Net != "" {
    network = c.Net
  }
}
```

I will have a look in **server.go** to check the necessary changes to also enable DNS over TLS.

@andrewtj I tried a "git commit --amend" to remove the "Closes", but I don't think it worked. So we could reopen the task if this PR is accepted in the future (still need to work on it after @miekg comments).

rafaeljusto:
The downside of using net with our custom values is that will be not intuitive for the API user, that will need to read the library docs to understand how to enable TLS.

miekg:
[ Quoting <notifications@github.com> in "Re: [dns] DNS over TLS (#300)..." ]
>@miekg The idea of using the net is good for me. We could rewrite it in **Client.exchange** to avoid errors in stdlib (net.UnknownNetworkError). I was thinking on something like this:
>
>```go
>network := "udp"
>tls := false
>
>switch c.Net {
>case "tcp-tls":
>  network = "tcp"
>  tls = true
>case "tcp4-tls":
>  network = "tcp4"
>  tls = true
>case "tcp6-tls":
>  network = "tcp6"
>  tls = true
>default:
>  if c.Net != "" {
>    network = c.Net
>  }
>}
>```

Yes, something like that can work.

>I will have a look in **server.go** to check the necessary changes to also enable DNS over TLS.

Cool.

>@andrewtj I tried a "git commit --amend" to remove the "Closes", but I don't think it worked. So we could reopen the task if this PR is accepted in the future (still need to work on it after @miekg comments).

I will reopen it, no worries.

/Miek

--
Miek Gieben

rafaeljusto:
Done. One problem that remains is when calling `ListenAndServe("127.0.0.1:0", "tcp-tls", handler)` it will always fail, because a TLS server without the certificates don't make sense. So we could add another function, like:

```go
func ListenAndServeTLS(addr string, network string, config *tls.Config, handler Handler) error
```

@miekg  What do you think?

miekg:
[ Quoting <notifications@github.com> in "Re: [dns] DNS over TLS (#300)..." ]
>Done. One problem that remains is when calling `ListenAndServe("127.0.0.1:0", "tcp-tls", handler)` it will always fail, because a TLS server without the certificates don't make sense. So we could add another function, like:
>
>```go
>func ListenAndServeTLS(addr string, network string, config *tls.Config, handler Handler) error
>```
>
>@miekg  What do you think?

mimicing the net/http interface here makes sense. I like how the net/http is
structured and a lot of thing in go dns already emulate that.

So go ahead create a `ListenAndServeTLS` function.

/Miek

--
Miek Gieben

rafaeljusto:
Done! =)

server.go
```diff
@@ -92,13 +93,38 @@ func HandleFailed(w ResponseWriter, r *Msg) {
...
 func ListenAndServe(addr string, network string, handler Handler) error {
 	server := &Server{Addr: addr, Net: network, Handler: handler}
 	return server.ListenAndServe()
 }
 
+// ListenAndServeTLS acts identically to ListenAndServe, except that it
+// expects TLS connections. Additionally, files containing a certificate and
+// matching private key for the server must be provided. If the certificate
+// is signed by a certificate authority, the certFile should be the concatenation
+// of the server's certificate, any intermediates, and the CA's certificate.
```
miekg: can we just point to net/http's ListenAndServe for the extra docs?

server.go
```diff
@@ -92,13 +93,38 @@ func HandleFailed(w ResponseWriter, r *Msg) {
...
-// ListenAndServe Starts a server on addresss and network speficied. Invoke handler
+// ListenAndServe Starts a server on address and network specified Invoke handler
 // for incoming queries.
 func ListenAndServe(addr string, network string, handler Handler) error {
 	server := &Server{Addr: addr, Net: network, Handler: handler}
 	return server.ListenAndServe()
 }
 
+// ListenAndServeTLS acts identically to ListenAndServe, except that it
+// expects TLS connections. Additionally, files containing a certificate and
```
miekg: TLS connection (which are TCP only).  maybe?

README.md
```diff
@@ -62,6 +62,7 @@ Send pull request if you want to be listed here.
 * EDNS0, NSID;
 * AXFR/IXFR;
 * TSIG, SIG(0);
+* DNS over TLS;
```
miekg: Think we can make some more nice about this feature in the README :) It's a completely new transport...

server.go
rafaeljusto: You are thinking on something like bellow?

```go
// ListenAndServeTLS acts like http.ListenAndServeTLS, more information in
// http://golang.org/pkg/net/http/#ListenAndServeTLS
func ListenAndServeTLS(addr, certFile, keyFile string, handler Handler) error
```

README.md
rafaeljusto: Sure! Will work on that.

PR #302, opened 2016-01-12
Added new functions: TrimDomain(), AddDomain(), GenerateZoneFile()

TomOnTime:
* TrimDomain() Trims the domain off a FQDN.
  * AddDomain() Adds a domain to a shortname.
  * GenerateZoneFile() Create a pretty BIND-style Zonefile.

### DIFF
--- a/dnsutil/util.go
+++ b/dnsutil/util.go
@@ -1,0 +1,79 @@
+// Package dnsutil contains higher-level methods useful with the dns
+// package.  While package dns implements the DNS protocols itself,
+// these functions are related but not directly required for protocol
+// processing.  They are often useful in preparing input/output of the
+// functions in package dns.
+package dnsutil
+
+import (
+	"strings"
+
+	"github.com/miekg/dns"
+)
+
+// AddDomain adds origin to s if s is not already a FQDN.
+// Note that the result may not be a FQDN.  If origin does not end
+// with a ".", the result won't either.
+// This implements the zonefile convention (specified in RFC 1035,
+// Section "5.1. Format") that "@" represents the
+// apex (bare) domain. i.e. AddOrigin("@", "foo.com.") returns "foo.com.".
+func AddOrigin(s, origin string) string {
+	// ("foo.", "origin.") -> "foo." (already a FQDN)
+	// ("foo", "origin.") -> "foo.origin."
+	// ("foo"), "origin" -> "foo.origin"
+	// ("@", "origin.") -> "origin." (@ represents the apex (bare) domain)
+	// ("", "origin.") -> "origin." (not obvious)
+	// ("foo", "") -> "foo" (not obvious)
+
+	if dns.IsFqdn(s) {
+		return s // s is already a FQDN, no need to mess with it.
+	}
+	if len(origin) == 0 {
+		return s // Nothing to append.
+	}
+	if s == "@" || len(s) == 0 {
+		return origin // Expand apex.
+	}
+
+	if origin == "." {
+		return s + origin // AddOrigin(s, ".") is an expensive way to add a ".".
+	}
+
+	return s + "." + origin // The simple case.
+}
+
+// TrimDomainName trims origin from s if s is a subdomain.
+// This function will never return "", but returns "@" instead (@ represents the apex (bare) domain).
+func TrimDomainName(s, origin string) string {
+	// An apex (bare) domain is always returned as "@".
+	// If the return value ends in a ".", the domain was not the suffix.
+	// origin can end in "." or not. Either way the results should be the same.
+
+	if len(s) == 0 {
+		return "@" // Return the apex (@) rather than "".
+	}
+	// Someone is using TrimDomainName(s, ".") to remove a dot if it exists.
+	if origin == "." {
+		return strings.TrimSuffix(s, origin)
+	}
+
+	// Dude, you aren't even if the right subdomain!
+	if !dns.IsSubDomain(origin, s) {
+		return s
+	}
+
+	slabels := dns.Split(s)
+	olabels := dns.Split(origin)
+	m := dns.CompareDomainName(s, origin)
+	if len(olabels) == m {
+		if len(olabels) == len(slabels) {
+			return "@" // origin == s
+		}
+		if (s[0] == '.') && (len(slabels) == (len(olabels) + 1)) {
+			return "@" // TrimDomainName(".foo.", "foo.")
+		}
+	}
+
+	// Return the first (len-m) labels:
+	return s[:slabels[len(slabels)-m]-1]
+}
--- a/dnsutil/util_test.go
+++ b/dnsutil/util_test.go
@@ -1,0 +1,130 @@
+package dnsutil
+
+import "testing"
+
+func TestAddOrigin(t *testing.T) {
+	var tests = []struct{ e1, e2, expected string }{
+		{"@", "example.com", "example.com"},
+		{"foo", "example.com", "foo.example.com"},
+		{"foo.", "example.com", "foo."},
+		{"@", "example.com.", "example.com."},
+		{"foo", "example.com.", "foo.example.com."},
+		{"foo.", "example.com.", "foo."},
+		// Oddball tests:
+		// In general origin should not be "" or "." but at least
+		// these tests verify we don't crash and will keep results
+		// from changing unexpectedly.
+		{"*.", "", "*."},
+		{"@", "", "@"},
+		{"foobar", "", "foobar"},
+		{"foobar.", "", "foobar."},
+		{"*.", ".", "*."},
+		{"@", ".", "."},
+		{"foobar", ".", "foobar."},
+		{"foobar.", ".", "foobar."},
+	}
+	for _, test := range tests {
+		actual := AddOrigin(test.e1, test.e2)
+		if test.expected != actual {
+			t.Errorf("AddOrigin(%#v, %#v) expected %#v, go %#v\n", test.e1, test.e2, test.expected, actual)
+		}
+	}
+}
+
+func TestTrimDomainName(t *testing.T) {
+
+	// Basic tests.
+	// Try trimming "example.com" and "example.com." from typical use cases.
+	var tests_examplecom = []struct{ experiment, expected string }{
+		{"foo.example.com", "foo"},
+		{"foo.example.com.", "foo"},
+		{".foo.example.com", ".foo"},
+		{".foo.example.com.", ".foo"},
+		{"*.example.com", "*"},
+		{"example.com", "@"},
+		{"example.com.", "@"},
+		{"com.", "com."},
+		{"foo.", "foo."},
+		{"serverfault.com.", "serverfault.com."},
+		{"serverfault.com", "serverfault.com"},
+		{".foo.ronco.com", ".foo.ronco.com"},
+		{".foo.ronco.com.", ".foo.ronco.com."},
+	}
+	for _, dom := range []string{"example.com", "example.com."} {
+		for i, test := range tests_examplecom {
+			actual := TrimDomainName(test.experiment, dom)
+			if test.expected != actual {
+				t.Errorf("%d TrimDomainName(%#v, %#v): expected (%v) got (%v)\n", i, test.experiment, dom, test.expected, actual)
+			}
+		}
+	}
+
+	// Paranoid tests.
+	// These test shouldn't be needed but I was weary of off-by-one errors.
+	// In theory, these can't happen because there are no single-letter TLDs,
+	// but it is good to exercize the code this way.
+	var tests = []struct{ experiment, expected string }{
+		{"", "@"},
+		{".", "."},
+		{"a.b.c.d.e.f.", "a.b.c.d.e"},
+		{"b.c.d.e.f.", "b.c.d.e"},
+		{"c.d.e.f.", "c.d.e"},
+		{"d.e.f.", "d.e"},
+		{"e.f.", "e"},
+		{"f.", "@"},
+		{".a.b.c.d.e.f.", ".a.b.c.d.e"},
+		{".b.c.d.e.f.", ".b.c.d.e"},
+		{".c.d.e.f.", ".c.d.e"},
+		{".d.e.f.", ".d.e"},
+		{".e.f.", ".e"},
+		{".f.", "@"},
+		{"a.b.c.d.e.f", "a.b.c.d.e"},
+		{"a.b.c.d.e.", "a.b.c.d.e."},
+		{"a.b.c.d.e", "a.b.c.d.e"},
+		{"a.b.c.d.", "a.b.c.d."},
+		{"a.b.c.d", "a.b.c.d"},
+		{"a.b.c.", "a.b.c."},
+		{"a.b.c", "a.b.c"},
+		{"a.b.", "a.b."},
+		{"a.b", "a.b"},
+		{"a.", "a."},
+		{"a", "a"},
+		{".a.b.c.d.e.f", ".a.b.c.d.e"},
+		{".a.b.c.d.e.", ".a.b.c.d.e."},
+		{".a.b.c.d.e", ".a.b.c.d.e"},
+		{".a.b.c.d.", ".a.b.c.d."},
+		{".a.b.c.d", ".a.b.c.d"},
+		{".a.b.c.", ".a.b.c."},
+		{".a.b.c", ".a.b.c"},
+		{".a.b.", ".a.b."},
+		{".a.b", ".a.b"},
+		{".a.", ".a."},
+		{".a", ".a"},
+	}
+	for _, dom := range []string{"f", "f."} {
+		for i, test := range tests {
+			actual := TrimDomainName(test.experiment, dom)
+			if test.expected != actual {
+				t.Errorf("%d TrimDomainName(%#v, %#v): expected (%v) got (%v)\n", i, test.experiment, dom, test.expected, actual)
+			}
+		}
+	}
+
+	// Test cases for bugs found in the wild.
+	// These test cases provide both origin, s, and the expected result.
+	// If you find a bug in the while, this is probably the easiest place
+	// to add it as a test case.
+	var tests_wild = []struct{ e1, e2, expected string }{
+		{"mathoverflow.net.", ".", "mathoverflow.net"},
+		{"mathoverflow.net", ".", "mathoverflow.net"},
+		{"", ".", "@"},
+		{"@", ".", "@"},
+	}
+	for i, test := range tests_wild {
+		actual := TrimDomainName(test.e1, test.e2)
+		if test.expected != actual {
+			t.Errorf("%d TrimDomainName(%#v, %#v): expected (%v) got (%v)\n", i, test.e1, test.e2, test.expected, actual)
+		}
+	}
+
+}
--- a/labels.go
+++ b/labels.go
@@ -4,6 +4,7 @@ package dns
 
 // SplitDomainName splits a name string into it's labels.
 // www.miek.nl. returns []string{"www", "miek", "nl"}
+// .www.miek.nl. returns []string{"", "www", "miek", "nl"},
 // The root label (.) returns nil. Note that using
 // strings.Split(s) will work in most cases, but does not handle
 // escaped dots (\.) for instance.
@@ -102,6 +103,7 @@ func CountLabel(s string) (labels int) {
 
 // Split splits a name s into its label indexes.
 // www.miek.nl. returns []int{0, 4, 9}, www.miek.nl also returns []int{0, 4, 9}.
+// .www.miek.nl. returns []int{0, 1, 5, 10}
 // The root name (.) returns nil. Also see SplitDomainName.
 func Split(s string) []int {
 	if s == "." {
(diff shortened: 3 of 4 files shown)


zone.go
```diff
@@ -0,0 +1,244 @@
...
+	if origin[len(origin)-1] == '.' {
+		suffix = origin[:len(origin)-1]
+	}
+	if name == suffix {
+		return "@"
+	}
+	return strings.TrimSuffix(name, "."+suffix)
+}
+
+type Zone struct {
```
miekg: I once added a similar zone structure to ldns: http://git.nlnetlabs.nl/ldns/tree/ldns/zone.h?h=release-1.6.17#n42 This was also a simple list; this of course has scaling issues. Another thing it that most users of this lib have wildly different ideas on how to implement the concept of a Zone, that's one of the reason there isn't a zone structure in go dns.

zone.go
```diff
@@ -0,0 +1,244 @@
...
+	"bytes"
+	"fmt"
+	"io"
+	"log"
+	"sort"
+	"strings"
+)
+
+// AddDomain adds origin to s if s is not already a FQDN.
+func AddDomain(origin, s string) string {
```
miekg: I can see the benefit of these helpers. But take for instance `IsEdns0` in `default.go` - which I added at the time, because I thought: will be useful for some. People just code this out themselves. The a) don't expect such a function to be present or b) don't find it in the docs. In retrospect I think most functions in `default.go` should not have been added. I'm sorry, but I feel the same about these.

zone.go
```diff
@@ -0,0 +1,244 @@
...
+	// If you want the result to end in a ".", pass an origin that ends in a "."
+	// ("origin.", "foo") -> "foo.origin."
+	// ("origin", "foo") -> "foo.origin"
+	// ("origin.", "foo.") -> "foo." (already a FQDN)
+	// ("origin.", "@") -> "origin." (@ represents the apex domain)
+	// ("origin.", "") -> "origin." (not obvious)
+	if len(origin) == 0 || origin[0] == '.' {
+		log.Fatalf("Invalid origin: %#v\n", origin)
```
miekg: Always try to return an error instead of Fatalf-ing

zone.go
TomOnTime: Logically one can't know if utility functions like IsEdns0, AddDomain, TrimDomain are used by project outside of this module.  I agree that documenting them would help people find them.  I know that getting started using the module was difficult for me because the documentation was so brief.

Also, while a function such as IsEdns0() may seem obvious to some, the minutiae of DNS is staggeringly huge. Such utility functions prevent stupid bugs from haunting the internet.  For example without CountLabel, most people would just use strings.Count(s, "."), without IsSubDomain most people would just use strings.IsSubDomain(s, domain).

That said...

It sounds to me like this project only wants contributions in a very specific domain of functionality.  I'm not sure that that domain is, but I think it would make the project easier to contribute to if that domain was described in the README.md.  Feel free to contact me off-line if you'd like to know what I think it is, based on reverse engineering the comments posted on this PR.

zone.go
TomOnTime: It only needs a struct so that it can be sorted. I could create a simple `type sortableRR []RR` type just to create an interface that matches Sort.  I could re-factor the code to not need Origin and DefaultTtl by assuming the RRs have FQDNs and proper TTL set already.

zone.go
miekg: [ Quoting <notifications@github.com> in "Re: [dns] Added new functions: Trim..." ]
>Logically one can't know if utility functions like IsEdns0, AddDomain, TrimDomain are used by project outside of this module.  I agree that documenting them would help people find them.  I know that getting started using the module was difficult for me because the documentation was so brief.
>
>Also, while a function such as IsEdns0() may seem obvious to some, the minutiae of DNS is staggeringly huge. Such utility functions prevent stupid bugs from haunting the internet.  For example without CountLabel, most people would just use strings.Count(s, "."), without IsSubDomain most people would just use strings.IsSubDomain(s, domain).
>
>That said...
>
>It sounds to me like this project only wants contributions in a very specific domain of functionality.  I'm not sure that that domain is, but I think it would make the project easier to contribute to if that domain was described in the README.md.  Feel free to contact me off-line if you'd like to know what I think it is, based on reverse engineering the comments posted on this PR.

Fair enough. Having a more offline discussion makes sense. And I of course
welcome contributions, but I also don't want to bloat the API. This goes back
to the line (that is in the README :) ):

  If there is stuff you should know as a DNS programmer there isn't a convenience
  function for it.

But that is of course really open to all kinds of interpretation(s).

I don't want to alienate people by not accepting PRs and I don't want to make Go
DNS a "everything and the kitchen sink" kind of package. So it's a balance.

Happy to discuss this some more.

/Miek

--
Miek Gieben

asergeyev: Could it be similar to IDN when things live within this library but in separate import path?

miekg: [ Quoting <notifications@github.com> in "Re: [dns] Added new functions: Trim..." ]
>> @@ -0,0 +1,244 @@
>> +package dns
>> +
>> +import (
>> +	"bytes"
>> +	"fmt"
>> +	"io"
>> +	"log"
>> +	"sort"
>> +	"strings"
>> +)
>> +
>> +// AddDomain adds origin to s if s is not already a FQDN.
>> +func AddDomain(origin, s string) string {
>
>Could it be similar to IDN when things live within this library but in separate import path?

Hmmm, hadn't considered that. Nice idea.

/Miek

--
Miek Gieben

miekg:
Lets go with that. Add a `util` dir and put TrimDomain() and AddDomain() (still not sure about GenerateZoneFile) in there.

andrewtj:
> On 21 Jan 2016, at 8:21 AM, Miek Gieben <notifications@github.com> wrote:
> 
> Lets go with that. Add a util dir and put TrimDomain() and AddDomain() (still not sure about GenerateZoneFile) in there.
> 

`io` packages utility functions in `io/ioutil`, should we follow that convention? Might avoid some ambiguity.

miekg:
[ Quoting <notifications@github.com> in "Re: [dns] Added new functions: Trim..." ]
>
>> On 21 Jan 2016, at 8:21 AM, Miek Gieben <notifications@github.com> wrote:
>>
>> Lets go with that. Add a util dir and put TrimDomain() and AddDomain() (still not sure about GenerateZoneFile) in there.
>>
>
>`io` packages utility functions in `io/ioutil`, should we follow that convention? Might avoid some ambiguity.

yep, like that suggestion too, dns/dnsutil it is.

Thanks!

miekg:
@TomOnTime you agree to this as well? And if so, are you willing to rework this a little?

TomOnTime:
Oh good!  I was going to suggest dnsutil too!   I'll try to get to this over the weekend.

TomOnTime:
PTAL

dnsutil/util.go
```diff
@@ -0,0 +1,91 @@
+package dnsutil
+
+import "strings"
+
+// AddDomain adds origin to s if s is not already a FQDN.
+func AddDomain(origin, s string) string {
```
miekg: To be in sync with the names of the functions in `labels.go`, this is better named `AddDomainName` or maybe `AddOrigin`  and then reverse the order of the parameters.

dnsutil/util.go
```diff
@@ -0,0 +1,91 @@
...
+	// If you want the result to end in a ".", pass an origin that ends in a "."
```
miekg: Should we return an error on a non-qualified origin?

dnsutil/util.go
```diff
@@ -0,0 +1,91 @@
...
+	// ("origin.", "foo") -> "foo.origin."
+	// ("origin", "foo") -> "foo.origin"
+	// ("origin.", "foo.") -> "foo." (already a FQDN)
+	// ("origin.", "@") -> "origin." (@ represents the apex domain)
```
miekg: That `@` is specific should be moved to the function's doc.

dnsutil/util.go
```diff
@@ -0,0 +1,91 @@
...
+		return origin
+	}
+	if origin == "." {
+		if s[len(s)-1] == '.' {
+			return s
+		} else {
+			return s + "."
+		}
+	}
+	if s == "*" {
```
miekg: why is this special cased?

dnsutil/util.go
```diff
@@ -0,0 +1,91 @@
...
+		return "*." + origin
+	}
+	if s[len(s)-1] == '.' {
+		return s
+	}
+	return s + "." + origin
+}
+
+// TrimDomain trims the origin from a FQDN if it exists.
+func TrimDomain(origin, s string) string {
```
miekg: TrimDomainName (also believe it's more natural to flip the order of the arguments)

dnsutil/util.go
```diff
@@ -0,0 +1,91 @@
...
+	// The apex domain is always returned as "@".
```
miekg: Should be moved to the function's doc. Not sure if `apex` is directly understood by every consumer of this function

dnsutil/util.go
```diff
@@ -0,0 +1,91 @@
+package dnsutil
```
miekg: package doc?

dnsutil/util.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	// We're being asked to trim something that is empty, so we special-case:
+	if len(s) == 0 {
+		return "@"
+	}
+	if s == "." {
+		return "."
+	}
+
+	lod := origin[len(origin)-1] == '.' // Is the last char of origin a dot?
```
miekg: `IsFqn(origin)` - maybe we should copy to this package as well btw.

dnsutil/util.go
```diff
@@ -0,0 +1,91 @@
...
+	lsd := s[len(s)-1] == '.'           // Is the last char of s a dot?
+
+	if lod == lsd { // Either they both end in ".", or neither do.
+		if origin == s {
+			return "@"
+		}
+		if origin[0] == '.' {
+			return strings.TrimSuffix(s, origin)
+		} else {
+			d_origin := "." + origin
```
miekg: try not to use underscores in var names

dnsutil/util.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	if s == "." {
+		return "."
+	}
+
+	lod := origin[len(origin)-1] == '.' // Is the last char of origin a dot?
+	lsd := s[len(s)-1] == '.'           // Is the last char of s a dot?
+
+	if lod == lsd { // Either they both end in ".", or neither do.
+		if origin == s {
```
miekg: these if's might be better rewritten as a switch?

dnsutil/util_test.go
```diff
@@ -0,0 +1,110 @@
+package dnsutil
+
+import "testing"
+
+// func AddDomain
+
+func TestAddDomain_nodot(t *testing.T) {
```
miekg: TestAddDomainNameNoDot

dnsutil/util_test.go
```diff
@@ -0,0 +1,110 @@
...
+			t.Errorf("%v: expected (%v) got (%v)\n", test.experiment, test.expected, actual)
+		}
+	}
+}
+
+func TestAddDomain_dot(t *testing.T) {
+	var tests = []struct{ experiment, expected string }{
+		{"@", "example.com."},
+		{"foo", "foo.example.com."},
+		{"foo.", "foo."},
```
miekg: this repeats the same test as above...

dnsutil/util_test.go
```diff
@@ -0,0 +1,110 @@
...
+				t.Errorf("TrimDomain(%#v, %#v): expected (%v) got (%v)\n", dom, test.experiment, test.expected, actual)
+			}
+		}
+	}
+}
+
+func TestTrimDomain_shortdomain(t *testing.T) {
+	// This test shouldn't be needed but I was weary of off-by-one errors
+	// due to a very short TLD.
+	var tests = []struct{ experiment, expected string }{
```
miekg: move tests that are actually different into TestStripDomainNameNormal (and drop Normal)

dnsutil/zone.go
```diff
@@ -0,0 +1,209 @@
...
+type zoneGenData struct {
+	Origin     string
+	DefaultTtl uint32
+	Records    []dns.RR
+}
+
+func (z *zoneGenData) Len() int      { return len(z.Records) }
+func (z *zoneGenData) Swap(i, j int) { z.Records[i], z.Records[j] = z.Records[j], z.Records[i] }
+func (z *zoneGenData) Less(i, j int) bool {
+	a, b := z.Records[i], z.Records[j]
```
miekg: to properly sort names you will need to resort to the wireformat, see https://github.com/miekg/dns/blob/dfefa36553955889dbb9ef892d4bd116ec095ce2/dnssec.go#L587

If think that something that *efficently* (is so far possible with go dns) sorts a list of RRs is useful

dnsutil/zone.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+	return a.String() < b.String()
+}
+
+func WriteZoneFile(w io.Writer, records []dns.RR, origin string, defaultTtl uint32) error {
+	z := &zoneGenData{
+		Origin:     origin,
+		DefaultTtl: defaultTtl,
+	}
+	for _, rec := range records {
```
miekg: there shouldn't be a need to copy these per rr.

dnsutil/zone_test.go
```diff
@@ -0,0 +1,297 @@
...
+
+	// Compare:
+	if buf2.String() != expected {
+		t.Fatalf("Regenerated zonefile does not match: got=(\n%v\n)\nexpected=(\n%v\n)\n", buf2.String(), expected)
+	}
+}
+
+// func WriteZoneFile
+
+func TestWriteZoneFile_simple(t *testing.T) {
```
miekg: These might be better written as ExampleTests that check the output

dnsutil/zone.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+	if rrtypeA == dns.TypeMX {
+		ta2, tb2 := a.(*dns.MX), b.(*dns.MX)
+		pa, pb := ta2.Preference, tb2.Preference
+		return pa < pb
+	}
+	return a.String() < b.String()
+}
+
+func WriteZoneFile(w io.Writer, records []dns.RR, origin string, defaultTtl uint32) error {
```
miekg: If we are going to add something like this, we should focus on making something that allows users to override the way a RR is printed. I.e. stuff like: omit class or ttl, only print the rdata (or a subset of the rdata). I.e. something super flexible. I have no good ideas (yet) on how something like that should look.

dnsutil/util.go
TomOnTime: I'll leave that to another PR.

dnsutil/util.go
TomOnTime: Refactored out of existence.

dnsutil/util.go
TomOnTime: Done

dnsutil/util_test.go
TomOnTime: Consolidated.

dnsutil/util_test.go
TomOnTime: Agreed

dnsutil/util_test.go
TomOnTime: Done

dnsutil/zone.go
TomOnTime: Fixed

dnsutil/zone.go
TomOnTime: I prefer this sort order because it groups hostnames in a way that makes it easier for people to see patterns: Grouping all records in a subdomain next to each other.  I see the value of having it sorted in DNSSEC order, but it isn't as readable, but the code is structured so it would be easy to abstract out the ordering code and create a WriteZoneFileOrdered() function that lets you supply a custom sort function.

dnsutil/util.go
TomOnTime: Done

dnsutil/util.go
TomOnTime: Ok

dnsutil/util.go
TomOnTime: I'd rather leave it as is.  Often user-facing code displays FQDNs without the trailing dot.  Being able to include (or not include) the "." at the end of the origin makes it easy for this one function to work in either case.  That's why I didn't call this MakeFqdn() (and AppendIfNotFqdn() is a bit convoluted).

dnsutil/util.go
TomOnTime: Done

dnsutil/util.go
TomOnTime: Good catch.  It was needed for an earlier implementation but not any more. Removed. (actually... refactored out)

dnsutil/util.go
TomOnTime: Done.

dnsutil/util.go
TomOnTime: Done.

dnsutil/util.go
asergeyev: I think you need to remove empty line between doc and package. You can check doc running ```godoc github.com/miekg/dns/dnsutil``` or use http-server of godoc (assuming you edit package inside GODOC).  Common practice to use "Package pkgname" as start of first sentence but godoc would work without it too.

dnsutil/zone.go
asergeyev: Maybe there should be WriteZoneFile should accept either constant for type of label ordering or comparator. I've encountered needs to have zones written in odd order before.  

Also I don't really know if it's useful but another way of writing zonefiles is by grouping things under $ORIGIN subsections of zone.

TomOnTime: How about a "flag int" parameter that can be used in the future to supply flags for various sorting and formatting options?

dnsutil/util.go
asergeyev: Looks great now!

dnsutil/util.go
```diff
@@ -0,0 +1,78 @@
...
+// functions in package dns.
+package dnsutil
+
+import (
+	"strings"
+
+	"github.com/miekg/dns"
+)
+
+// AddDomain adds origin to s if s is not already a FQDN.
```
miekg: AddOrigin.

"The zonefile convention" is specified in RFC 1035, Section "5.1. Format", regarding the '@' sign.

dnsutil/util_test.go
```diff
@@ -0,0 +1,130 @@
...
+		for i, test := range tests {
+			actual := TrimDomainName(test.experiment, dom)
+			if test.expected != actual {
+				t.Errorf("%d TrimDomainName(%#v, %#v): expected (%v) got (%v)\n", i, test.experiment, dom, test.expected, actual)
+			}
+		}
+	}
+}
+
+func TestTrimDomainNameShortSuffix(t *testing.T) {
```
miekg: Could you collopse this into 2 two test function for both AddOrigin and TrimDomainName

dnsutil/zone.go
miekg: Can we give it a sort.Sort like parameter that sorts for you?

Don't really like magical flags for this...

dnsutil/zone.go
```diff
@@ -0,0 +1,212 @@
...
+	"fmt"
+	"io"
+	"log"
+	"sort"
+	"strings"
+
+	"github.com/miekg/dns"
+)
+
+type zoneGenData struct {
```
miekg: Can you put this change in a separate PR? Still not sure I want it or not and in what form. We can then merge the other functions

dnsutil/util_test.go
TomOnTime: Done

dnsutil/util.go
TomOnTime: Added

dnsutil/zone.go
TomOnTime: I've removed it.  I'll make a new PR soon.

labels_test.go
```diff
@@ -57,6 +58,8 @@ func TestSplit(t *testing.T) {
 			t.Logf("%s %v", s, Split(s))
 		}
 	}
+	f := Split(".www.miek.nl.")
+	fmt.Printf("TEST: %#v\n", f)
```
miekg: what's this actually?

Fine with extending the test below.

labels.go
```diff
@@ -102,6 +103,7 @@ func CountLabel(s string) (labels int) {
 
 // Split splits a name s into its label indexes.
 // www.miek.nl. returns []int{0, 4, 9}, www.miek.nl also returns []int{0, 4, 9}.
+// .www.miek.nl. returns []int{0, 1, 5, 10}
```
miekg: hmmm, this might be a bug, because the domain name is not valid.... good also call it garbage in, garbage out.

labels_test.go
TomOnTime: Oops. That was supposed to be deleted before I pushed.  Fixed.

miekg:
Now there are conflicts... could you fix as well, then I press the glorious merge button.

TomOnTime:
I squashed the commits.

Issue #312, opened 2016-02-22
Support for larger DNS responses


prisamuel:
I'm seeing the same error message as #214 when using this library for querying SRV records from Consul's DNS interface. It appears to be a similar issue.

When the message size is smaller than 512, my requests succeed. But when the response size increases to 538 B I get an error with "dns: buffer size too small".

A typical Consul SRV response looks like this:
`bash-4.3# dig example.service.consul SRV

; <<>> DiG 9.10.2 <<>> example.service.consul SRV
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 42745
;; flags: qr aa rd ra; QUERY: 1, ANSWER: 2, AUTHORITY: 0, ADDITIONAL: 2

;; QUESTION SECTION:
;example.service.consul. IN SRV

;; ANSWER SECTION:
example.service.consul. 0 IN SRV 1 1 32866 core-i-0a2b2rewr58f-172.17.5.34.node.dc1.consul.
example.service.consul. 0 IN SRV 1 1 32866 core-i-7fafffwerwwff-172.17.4.151.node.dc1.consul.
...
...

;; ADDITIONAL SECTION:
core-i-0a2b2rewr58f-172.17.5.34.node.dc1.consul. 0 IN A 172.17.5.34
core-i-7fafffwerwwff-172.17.4.151.node.dc1.consul. 0 IN A 172.17.4.151
...
...
;; Query time: 3 msec
;; SERVER: 10.0.42.1#53(10.0.42.1)
;; WHEN: Fri Feb 19 12:58:10 UTC 2016
;; MSG SIZE rcvd: 538`

On the RFC 2675 https://tools.ietf.org/html/rfc2675
A field that specifies the length in bytes of the UDP header and UDP data. The minimum length is 8 bytes because that is the length of the header. The field size sets a theoretical limit of 65,535 bytes (8 byte header + 65,527 bytes of data) for a UDP datagram. The practical limit for the data length which is imposed by the underlying IPv4 protocol is 65,507 bytes (65,535 − 8 byte UDP header − 20 byte IP header)

This is well below the specified limit. Will there be support for this on this library?

andrewtj:
There's a couple of things to cover here. I'll start with the protocol and then go to the practicals.

[...]

PR #315, opened 2016-02-23
Mirror Bufsize in OPT RR

miekg:
When the Bufsize is set we use a larger buffer to capture the
return packet. We didn't add a OPT RR to advertise this larger
size to the remote server leading to all kinds of mayhem.

When no OPT RR is set, create one and set the Bufsize.

Fixes: #312

### DIFF
--- a/client.go
+++ b/client.go
@@ -192,6 +192,8 @@ func (c *Client) exchange(m *Msg, a string) (r *Msg, rtt time.Duration, err erro
 	// Otherwise use the client's configured UDP size.
 	if opt == nil && c.UDPSize >= MinMsgSize {
 		co.UDPSize = c.UDPSize
+		// In this case also set the the EDNS0 bufsize to allow for larger responses.
+		m.SetEdns0(c.UDPSize, false /* dnssec */ )
 	}
 
 	co.TsigSecret = c.TsigSecret
--- a/client_test.go
+++ b/client_test.go
@@ -419,3 +419,42 @@ func TestTruncatedMsg(t *testing.T) {
 		t.Fail()
 	}
 }
+
+// Issue #312.
+func TestClientBufferSize(t *testing.T) {
+	c := new(Client)
+	type Test struct {
+		bufsize uint16
+		expected uint16
+	}
+	tests := []Test{
+		{0, 0},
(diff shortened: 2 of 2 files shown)


miekg:
2 things
1. Default to 4k in Exchange() by default?
2. Maybe also check the OPT rr (if it *is* there) and double check the bufsizes?

andrewtj:
I thought the behaviour of not adding an `OPT` was intentional. I wonder if changing this will catch anyone out.

miekg:
[ Quoting <notifications@github.com> in "Re: [dns] Mirror Bufsize in OPT RR ..." ]
>I thought the behaviour of not adding an `OPT` was intentional. I wonder if changing this will catch anyone out.

True .. and this might not be the *best* way forward. It also modifies the
message which might lead to further problems.

Then we still need something (boolean?) to enable this. Still don't know of a
good name for such a flag.

Leaves the problem of Exchange() - which maybe can't be tackled by not fiddling
with it at all.

/Miek

--
Miek Gieben

andrewtj:
Maybe a new abstraction is needed. Something that can be configured with a list of servers and optionally a UDP payload size, and that will then only take queries by Name and Type, returning answers or an error. Hiding away all the nitty-gritty of setting UDP payload sizes, retrying on timeout and retrying over TCP on truncation.

miekg:
[ Quoting <notifications@github.com> in "Re: [dns] Mirror Bufsize in OPT RR ..." ]
>Maybe a new abstraction is needed. Something that can be configured with a list
>of servers and optionally a UDP payload size, and that will then only take
>queries by Name and Type, returning answers or an error. Hiding away all the
>nitty-gritty of setting UDP payload sizes, retrying on timeout and retrying
>over TCP on truncation.

I never liked this "Resolver" object like api, it is really handy for a one-off
query, but I've found it to become unwieldy when doing more serious work (when
for instance using net:dns in Perl or Ldns)

/Miek

--
Miek Gieben

andrewtj:
Yeah fair enough. Otherwise I think the options available cover this task well for anyone familiar with the DNS. For those who aren't the API is probably intimidating, but I don't think that's something to fix within this project.

PR #319, opened 2016-02-27
Fix code around dynamic updates, allowing for multiple prereq/inserts/removals per update

middelink:
An update msg can have multiple, different prerequests and also multiple, different remove and inserts. The old code did a reset of the ANSWER and AUTHORITY section each time a rrset was added.

Made a slight optimization that the first time an rrset is added, we
take it length as the initial size. Helps for the one-time add cases.

Added test to prevent regressions.

### DIFF
--- a/update_test.go
+++ b/update_test.go
@@ -6,6 +6,7 @@ import (
 )
 
 func TestDynamicUpdateParsing(t *testing.T) {
+	// This does nothing with dynamic updates, move elsewhere?
 	prefix := "example.com. IN "
 	for _, typ := range TypeToString {
 		if typ == "OPT" || typ == "AXFR" || typ == "IXFR" || typ == "ANY" || typ == "TKEY" ||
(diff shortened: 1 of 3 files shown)


update_test.go
```diff
@@ -6,6 +6,7 @@ import (
 )
 
 func TestDynamicUpdateParsing(t *testing.T) {
+	// This does nothing with dynamic updates, move elsewhere?
```
miekg: It does actually, rr parsing with any rdata should work (for all types), i.e.
`miek.nl. IN A` is vaild.

update_test.go
```diff
@@ -83,3 +84,60 @@ func TestRemoveRRset(t *testing.T) {
...
+	m.Used([]RR{rr_rrset2})
+	m.RRsetNotUsed([]RR{rr_rrset3})
+
+	// and now the updates.
+	m.RemoveName([]RR{rr_remove1})
+	m.RemoveRRset([]RR{rr_remove2})
+	m.Remove([]RR{rr_remove3})
+	m.Insert([]RR{rr_insert})
+
+	expect := `;; opcode: UPDATE, status: NOERROR, id: 1234
```
miekg: If we want to do a check like this it's better to write an Example test.

update_test.go
```diff
@@ -83,3 +84,60 @@ func TestRemoveRRset(t *testing.T) {
 		t.Errorf("actual msg:\n%v", tmp)
 	}
 }
+
+func TestPreReqAndRemovals(t *testing.T) {
+	// Build a list of multiple prereqs and then somes removes followed by an insert.
+	// We should be able to add multiple prereqs and updates.
+	m := new(Msg)
+	m.SetUpdate("some_zone.")
```
miekg: use `example.org.` (and friends) for this like this.

update_test.go
middelink: Done

update_test.go
middelink: I agree this function should be tested, but in zscan_test.go as it is defined in zscan.go. All it does here is parse partial RRs, which is fine but not specific to updates.go. In fact I can parse partial RRs everywhere I want ^^

But never mind, it is not worth the discussion, al long as it is tested.

update_test.go
middelink: Done. unless you wanted me to change names in the NewRRs as well.

middelink:
Fuck this git shit. Rebased the branch to the new head of your branch and now it lists all the in between commits as mine. Killing this PR once more.

PR #320, opened 2016-02-27
Fix code around dynamic updates, allowing for multiple prereq/inserts/removals per update

middelink:
An update msg can have multiple, different prerequests and also multiple, different remove and inserts. The old code did a reset of the ANSWER and AUTHORITY section each time a rrset was added.

Made a slight optimization that the first time an rrset is added, we
take it length as the initial size. Helps for the one-time add cases.

Added test to prevent regressions.

### DIFF
--- a/update_test.go
+++ b/update_test.go
@@ -6,6 +6,7 @@ import (
 )
 
 func TestDynamicUpdateParsing(t *testing.T) {
+	// This does nothing with dynamic updates, move elsewhere?
 	prefix := "example.com. IN "
 	for _, typ := range TypeToString {
 		if typ == "OPT" || typ == "AXFR" || typ == "IXFR" || typ == "ANY" || typ == "TKEY" ||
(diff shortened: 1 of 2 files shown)


update_test.go
```diff
@@ -83,3 +83,60 @@ func TestRemoveRRset(t *testing.T) {
...
+	m.Used([]RR{rr_rrset2})
+	m.RRsetNotUsed([]RR{rr_rrset3})
+
+	// and now the updates.
+	m.RemoveName([]RR{rr_remove1})
+	m.RemoveRRset([]RR{rr_remove2})
+	m.Remove([]RR{rr_remove3})
+	m.Insert([]RR{rr_insert})
+
+	expect := `;; opcode: UPDATE, status: NOERROR, id: 1234
```
miekg: Could you rework this into an Example Test then you don't need the manual expect check it will be done for you

middelink: Ah, i see you did not get the inline responses of the close PR :)

I tried, doesn't work as the generated output contained trailing tabs in the partial RRs. The example output checker seems to trim the white space from its expected output. So they never match.

miekg: Ah, trailing tabs... Hmmm that's a bit ugly from our side. I'll add a TODO in there for future cleanup.

Issue #306, opened 2016-01-25
Split(".www.miek.nl.") -> []string{"", "www", "miek", "nl"},


miekg:
This might be a bug (or not). It's is basically an illegal domain name that you then split...

andrewtj:
Change the interface, add a new func or make a note of it for a someday-maybe `dns2`?

miekg:
We could return `[]{0}` for this... but yeah returning an error is better...

andrewtj:
Maybe a simple breaking change log could make things like this easier. A low effort way to do it would be to just put '#breaking` or similar in commit messages and then extract into the readme a list of commits that break things. This would be even more useful for changes that silently break things like #281 seems to have done to rrda (see fcambus/rrda#11).

miekg:
[ Quoting <notifications@github.com> in "Re: [dns] Split(".www.miek.nl.") ->..." ]
>Maybe a simple breaking change log could make things like this easier. A low effort way to do it would be to just put '#breaking` or similar in commit messages and then extract into the readme a list of commits that break things. This would be even more useful for changes that silently break things like #281 seems to have done to rrda (see fcambus/rrda#11).

I agree. Should still be careful, but bugs need to be fixed even though that could
break backwards compat.

I also like the '#breaking' idea.

/Miek

--
Miek Gieben

andrewtj:
[Sketched up](https://gist.github.com/andrewtj/fe8feae87427f3b4c592) a way it could work.

miekg:
Coming back to this specific issue and looking through other label function, such as CountLabel for instance, they don't return a value plus error. So either we break them *all*, *or* we assume and document that in the input should be a properly formatted dns name/label. I'm leaning towards the latter (garbage in, garbage out)

andrewtj:
That sounds okay. If there's demand in the future error returning variants could always go in `dnsutil`.

miekg:
[ Quoting <notifications@github.com> in "Re: [dns] Split(".www.miek.nl.") ->..." ]
>That sounds okay. If there's demand in the future error returning variants could always go in `dnsutil`.

Ok, let's agree to that. They are some prinln("deprecated") littered in the
source and at some point all of these functions/methods should be deprecated
(pref.  in one big bang release)

/Miek

--
Miek Gieben

PR #321, opened 2016-02-27
Improve docs on the label functions

miekg:
Explicitly mention that the domain s must be valid before
given to these functions.

Fixes: #306


andrewtj:
`s/syntacticly/syntactically` otherwise lgtm

PR #322, opened 2016-02-27
Udp buffer 2

miekg:
Take number 2

### DIFF
--- a/client.go
+++ b/client.go
@@ -26,14 +26,15 @@ type Conn struct {
 // A Client defines parameters for a DNS client.
 type Client struct {
 	Net            string            // if "tcp" or "tcp-tls" (DNS over TLS) a TCP query will be initiated, otherwise an UDP one (default is "" for UDP)
-	UDPSize        uint16            // minimum receive buffer for UDP messages
+	UDPSize        uint16            // minimum receive buffer for UDP messages, should be set to 2048 or 4096 in modern applications.
 	TLSConfig      *tls.Config       // TLS connection configuration
 	DialTimeout    time.Duration     // net.DialTimeout, defaults to 2 seconds
 	ReadTimeout    time.Duration     // net.Conn.SetReadTimeout value for connections, defaults to 2 seconds
 	WriteTimeout   time.Duration     // net.Conn.SetWriteTimeout value for connections, defaults to 2 seconds
 	TsigSecret     map[string]string // secret(s) for Tsig map[<zonename>]<base64 secret>, zonename must be fully qualified
 	SingleInflight bool              // if true suppress multiple outstanding queries for the same Qname, Qtype and Qclass
-	// When true SetUDPSize will use UDPSize and will *add* an EDNS0 OPT RR to the message when sending
+
+	// When true SetUDPSize will *add* an EDNS0 OPT RR with UDPSize to the message when sending
 	// it. If there already is an OPT RR it will use the UDP size specified in there.
 	SetUDPSize bool
 
--- a/client.go
+++ b/client.go
@@ -33,7 +33,11 @@ type Client struct {
 	WriteTimeout   time.Duration     // net.Conn.SetWriteTimeout value for connections, defaults to 2 seconds
 	TsigSecret     map[string]string // secret(s) for Tsig map[<zonename>]<base64 secret>, zonename must be fully qualified
 	SingleInflight bool              // if true suppress multiple outstanding queries for the same Qname, Qtype and Qclass
-	group          singleflight
+	// When true SetUDPSize will use UDPSize and will *add* an EDNS0 OPT RR to the message when sending
+	// it. If there already is an OPT RR it will use the UDP size specified in there.
+	SetUDPSize bool
+
+	group singleflight
 }
 
 // Exchange performs a synchronous UDP query. It sends the message m to the address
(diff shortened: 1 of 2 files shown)


miekg:
@andrewtj for you consideration

client.go
```diff
@@ -192,6 +199,10 @@ func (c *Client) exchange(m *Msg, a string) (r *Msg, rtt time.Duration, err erro
 	// Otherwise use the client's configured UDP size.
 	if opt == nil && c.UDPSize >= MinMsgSize {
 		co.UDPSize = c.UDPSize
+		// In this case also set the the EDNS0 bufsize to allow for larger responses.
+		if c.SetUDPSize {
+			m.SetEdns0(c.UDPSize, true /* dnssec */)
```
michaelharo: Setting dnssec to true seems like an unexpected side effect of enabling larger payloads.  It might result in even larger responses that don't fit into a UDP response.  If nothing else I think this behavior should be noted in the doc string for SetUDPSIze.

miekg: Yeah, I'm wondering about this as well. Maybe SetUDPSize should be an int:
* 0: off
* 1: enable, without DNSSEC
* 2: enable, with DNSSEC

?

michaelharo: I can't decide if using constants or two bools in the struct is better (SetUDPSize, EnableDNSSEC).  The nice part about using struct members is that the code is just
```
if c.SetUDPSize {
  m.SetEDNS0(c.UDPSize, c.EnableDNSSEC)
}
```

Perhaps SetUDPSize is the wrong bool name for the client though?  Is it really SetEDNS0?

At some point the code might need to handle different EDNS versions as well, so maybe an EDNS struct that has {UDPSize, DNSSEC, Version, ...future options...} is a better method to abstract the EDNS options?  Then m.SetEDNS (note not m.SetEDNS0) could take an options struct instead of an argument list.  If this route is selected then the Client struct should have an EDNSOptions member that is passed around instead.

```
type EDNSOptions struct {
  Version int
  UDPSize int
  DNSSEC bool
}

type Client struct {
  ... existing stuff ...
  EDNSOptions *EDNSOptions
  ... existing stuff ...
}

func (c *Client) exchange(m *Msg, a string) (r *Msg, rtt time.Duration, err error) {
  ... existing code ...
  if c.EDNSOptions != nil {
    m.SetEDNS(c.EDNSOptions)
  }
  ... existing code...
}

// not exactly sure how this should work, but for discussion...
func (dns *Msg) SetEDNS(opts *EDNSOptions) {
  if opts == nil {
    return
  }
  edns := dns.GetEDNS()
  if edns == nil {
    edns = dns.AddEDNS()  // adds an EDNS OPT RR to dns
  }
  edns.SetVersion(opts.Version)
  edns.SetDNSSec(opts.DNSSEC)
  edns.SetUDPSize(max(MinMsgSize, opts.UDPSize))  
}
```
I haven't placed much thought into what other EDNS options exist to know if this is a valid forward looking solution.

miekg: [ Quoting <notifications@github.com> in "Re: [dns] Udp buffer 2 (#322)..." ]
>> @@ -192,6 +199,10 @@ func (c *Client) exchange(m *Msg, a string) (r *Msg, rtt time.Duration, err erro
>>  	// Otherwise use the client's configured UDP size.
>>  	if opt == nil && c.UDPSize >= MinMsgSize {
>>  		co.UDPSize = c.UDPSize
>> +		// In this case also set the the EDNS0 bufsize to allow for larger responses.
>> +		if c.SetUDPSize {
>> +			m.SetEdns0(c.UDPSize, true /* dnssec */)
>
>I can't decide if using constants or two bools in the struct is better (SetUDPSize, EnableDNSSEC).  The nice part about using struct members is that the code is just
>```
>if c.SetUDPSize {
>  m.SetEDNS0(c.UDPSize, c.EnableDNSSEC)
>}
>```

That is true, but EnableDNSSEC *also* implies adding a EDNS0 OPT RR, regardless
of the UDPSize, which if the latter isn't set should also have a default of
4k...? But, maybe two bools is better.

>Perhaps SetUDPSize is the wrong bool name for the client though?  Is it really SetEDNS0?

Yep. I think SetEDNS0 is better, to be inline with the rest of the lib, it has
to be SetEdns0 though.

>At some point the code might need to handle different EDNS versions as well, so maybe an EDNS struct that has {UDPSize, DNSSEC, Version, ...future options...} is a better method to abstract the EDNS options?  Then m.SetEDNS (note not m.SetEDNS0) could take an options struct instead of an argument list.  If this route is selected then the Client struct should have an EDNSOptions member that is passed around instead.

*That* is a valid point, but I am

a) not convinced this is going to happen any time soon (could be mistaken
though)
b) warrents the complications it gives to solving the current problem (.i.e
people sending msgs with bufsizes that are too small).

>I haven't placed much thought into what other EDNS options exist to know if this is a valid forward looking solution.

I *do* like the Config/Option struct to pass options around though...

andrewtj:
I'm still not really sold on this. It doesn't seem like much of a saver when a `Msg` still needs to be constructed and `.SetQuestion()` called.

I'm also dubious that it'll solve the problem of folks sending messages with small buffer sizes. If they don't know about the existing options, why would new ones get on their radar? Doc pointers around the `ErrTruncated` constant and `.Exchange()` might be more effective.

That aside, I'd lean toward the `0 = Off, 1 = Enable, 2 = Enable with DNSSEC` approach. I'll mull over the struct-config option a bit more though.

miekg:
Good points, I rather improve the code than add more magical struct members (which are indeed hard to find). *Iff* you correctly create your `Msg` both Exchange and c.Exchange do the right thing, which is also something that will break with this PR.

So I'm with you in: better improve the docs than add more codes.

miekg:
I'm going to close this PR. I agree with @andrewtj sentiment, we should not try to fix this here, the *Msg is the interface to these function use that to convey the intent. We could improve the docs, will send another PR for that.

Now that we have dnsutil we *might* want to add some higher level function that provide some more hand holding.

Issue #316, opened 2016-02-25
Retrieving and setting RDLEN


matje:
There is no easy way to retrieve the RDLEN of an RR. Some statements after a twitter discussion are:

- `PackRR` does not set it (maybe it should).
- A Len(RR) and Len(RR_Header) might be nice.

andrewtj:
Can you provide some more detail?

miekg:
Twitter chat starts here: https://twitter.com/pletterpet/status/702118581291581441

miekg:
Note this is issue when parsing from strings, getting them from the wire will set the length.

andrewtj:
Making `PackRR` set `Rdlength` could have side effects.

miekg:
@andrewtj yep that could very well be (didn't look in detail). Making a simple Len is prolly sufficient for this usecase:
```
func Len(r RR) int {
   return r.len()
}
```
Or we could make `len` exported on all RRs. The former is more in line with what we do with `Copy` right now so that would be my preference.

matje:
#333

PR #326, opened 2016-02-29
Expose RR and RR Header Len

matje:
Related to issue #316

### DIFF
--- a/dns.go
+++ b/dns.go
@@ -32,8 +32,8 @@ type RR interface {
 	String() string
 	// copy returns a copy of the RR
 	copy() RR
-	// len returns the length (in octets) of the uncompressed RR in wire format.
-	len() int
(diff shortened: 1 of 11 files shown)


miekg:
I think it makes most sense to do this in the same way as we did `Copy`, https://github.com/miekg/dns/blob/85b661b2a6fc95a5a83e66d7730c4bc0b6e9c99e/msg.go#L1893

This would be a minimal I think Len() would then automatically work for `RR_Header` as that is an RR as well.

Also: currently the tests don't compile

andrewtj:
Although not relevant to implementing this as `func Len(RR) int`, for future reference `types_generate.go` builds `ztypes.go` via `go generate`.

matje:
@miekg It's still on my todo list

matje:
Easier to close this PR and create a new one with the easier fix, see #333

PR #327, opened 2016-03-02
Better docs on client buffer size

miekg:
Extend the documentation for Exchange to make it clear that it is
the callers responsibility to enable larger buffers.


miekg:
@andrewtj So tiny tweaks in the documentation.

andrewtj:
Looks good to me :+1:

PR #329, opened 2016-03-07
TSIG: lowercase algorithm before use

miekg:
These strings are domain names, so we should lowercase them before
using them.

Also add some tests for Tsig generation and verification.

(/ht ldns release).


andrewtj:
Good pickup. Is there a test covering multi-envelope sequences?

miekg:
probably no, we have some in `xfr_test.go` but not multi envelope afaict

PR #333, opened 2016-03-17
Len returns the rr len of the uncompressed format.

matje:
An easier fix than in #326


miekg:
Maybe would should also mention RR_Header in the docs, but fair enough.. merging.

PR #335, opened 2016-03-25
fix building on plan9

mischief:
note that i'm unsure if this code really works, but it allows github.com/xenolf/lego/acme to build on 9front.

### DIFF
--- a/udp_other.go
+++ b/udp_other.go
@@ -1,4 +1,4 @@
-// +build !linux
+// +build !linux,!plan9
 
 package dns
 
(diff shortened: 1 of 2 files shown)


miekg:
dunno either, it's also impossible for me to verify and probably hard to make a test for this...
See no reason not to merge this.

mischief:
can try to run tests later, i'll send more PRs if something goes wrong. thanks!

miekg:
would be appreciated! Thanks!

Issue #339, opened 2016-04-09
extended rcode don't work


miekg:
The extended rcode we write in the opt record do not work:

~~~
rr.Hdr.Ttl = rr.Hdr.Ttl&0x00FFFFFF | (uint32(v) << 24)
~~~
This write the exact rcode, for instance BADVERS, which is 16. *But* this hold the upper 8 bits:
> EXTENDED-RCODE  Forms upper 8 bits of extended 12-bit RCODE.  Note
> that EXTENDED-RCODE value "0" indicates that an
> unextended RCODE is in use (values "0" through "15").

This means if we want to store 16 (BADVERS) we should write 1here.

PR #340, opened 2016-04-09
Put the correct extended rcode in OPT

miekg:
We were off by a value of 15. This fixes it. Hard to come up with a test
as writing and reading it yourself will be consistent.


miekg:
This fix is working for me when running CoreDNS and querying with `kdig`

miekg:
Fixes #339

PR #345, opened 2016-04-14
Add Client.Timeout to allow limiting total exchange duration

wbond:
This change allows limiting the duration of the complete dial, write, read cycle, as opposed to controlling them individually.

### DIFF
--- a/client.go
+++ b/client.go
@@ -28,9 +28,10 @@ type Client struct {
 	Net            string            // if "tcp" or "tcp-tls" (DNS over TLS) a TCP query will be initiated, otherwise an UDP one (default is "" for UDP)
 	UDPSize        uint16            // minimum receive buffer for UDP messages
 	TLSConfig      *tls.Config       // TLS connection configuration
-	DialTimeout    time.Duration     // net.DialTimeout, defaults to 2 seconds
-	ReadTimeout    time.Duration     // net.Conn.SetReadTimeout value for connections, defaults to 2 seconds
-	WriteTimeout   time.Duration     // net.Conn.SetWriteTimeout value for connections, defaults to 2 seconds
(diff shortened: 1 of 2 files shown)


client_test.go
```diff
@@ -419,3 +419,51 @@ func TestTruncatedMsg(t *testing.T) {
...
+		}
+		done <- struct{}{}
+	}()
+
+	select {
+	case <-done:
+	case <-time.After(abortAfter):
+	}
+
+	length := time.Now().Sub(start)
```
miekg: you can just use start.Since iirc

client.go
```diff
@@ -192,12 +201,20 @@ func (c *Client) exchange(m *Msg, a string) (r *Msg, rtt time.Duration, err erro
 	}
 
 	co.TsigSecret = c.TsigSecret
-	co.SetWriteDeadline(time.Now().Add(c.writeTimeout()))
+	writeDeadline := deadline
```
michaelharo: Perhaps a helper function instead of these 4 additional lines of code multiple times?

Something like
```
func deadlineOrTimeout(deadline time.Time, timeout time.Duration) time.Time {
  if deadline.IsZero() {
    return time.Now().Add(timeout)
  }
  return deadline
}
```

Then in the main function use something like
```
co.SetWriteDeadline(deadlineOrTimeout(deadline, c.writeTimeout())
...
co.SetReadDeadline(deadlineOrTimeout(deadline, c.readTimeout())`
...
```

miekg:
Can you make the changes as sketched above? Then this is ready to be merged.

wbond:
Yes, I'll be updating the PR in the next day. Thanks for the feedback!

wbond:
I've just pushed up a revised commit that integrates the suggested changes.

Issue #341, opened 2016-04-10
EDNS ExtendedRcode return uint size


McStork:
Hi,
After the recent update of ExtendedRcode, the return type which was uint8 should be upgraded to at least uint16, as [it can now return](https://github.com/miekg/dns/commit/b5f8285e2ab8f825c92ba68ddbbfe55c534b1064) an integer of a size up to 12bits.

miekg:
yep, you're right and `uint8` was never good. Pondering just `int` or `uint16` as the fixed return code...

PR #347, opened 2016-04-18
set ExtendedRcode return type to int instead of uint8

McStork:
This PR sets the return type of method `ExtendedRcode` to `int`, to be consistent with the keys type of `map` [RcodeToString](https://github.com/miekg/dns/blob/89d9c5e74bc17cc9edb5d44d4f2f93a7ac77e365/msg.go#L127), and to close https://github.com/miekg/dns/issues/341.


miekg:
Indeed, a blatant bug, although this is not backwards compat, it is serious enough to just merge it. Also: hopefully no a very much used piece of go dns.

PR #350, opened 2016-05-03
small patch for EDNS0 cookie support

shuque:
If you're interested, a small patch to add EDNS0 cookie support.

### DIFF
--- a/edns.go
+++ b/edns.go
@@ -299,8 +299,8 @@ func (e *EDNS0_SUBNET) String() (s string) {
 //e.Cookie = "24a5ac.."
 //o.Option = append(o.Option, e)
 type EDNS0_COOKIE struct {
-	Code uint16                   // Always EDNS0COOKIE
-	Cookie string                 // Hex-encoded cookie data
+	Code   uint16 // Always EDNS0COOKIE
+	Cookie string // Hex-encoded cookie data
 }
 
 func (e *EDNS0_COOKIE) pack() ([]byte, error) {
@@ -313,7 +313,7 @@ func (e *EDNS0_COOKIE) pack() ([]byte, error) {
 
 func (e *EDNS0_COOKIE) Option() uint16        { return EDNS0COOKIE }
 func (e *EDNS0_COOKIE) unpack(b []byte) error { e.Cookie = hex.EncodeToString(b); return nil }
-func (e *EDNS0_COOKIE) String() string        { return string(e.Cookie) }
+func (e *EDNS0_COOKIE) String() string        { return e.Cookie }
 
 // The EDNS0_UL (Update Lease) (draft RFC) option is used to tell the server to set
 // an expiration on an update RR. This is helpful for clients that cannot clean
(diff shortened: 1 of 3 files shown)


edns.go
```diff
@@ -286,6 +289,32 @@ func (e *EDNS0_SUBNET) String() (s string) {
...
+//o := new(dns.OPT)
+//o.Hdr.Name = "."
+//o.Hdr.Rrtype = dns.TypeOPT
+//e := new(dns.EDNS0_COOKIE)
+//e.Code = dns.EDNS0COOKIE
+//e.Cookie = "24a5ac.."
+//o.Option = append(o.Option, e)
+type EDNS0_COOKIE struct {
+	Code uint16                   // Always EDNS0COOKIE
+	Cookie string                 // Hex-encoded cookie data
```
miekg: Has this code been gofmt-ed?

edns.go
```diff
@@ -286,6 +289,32 @@ func (e *EDNS0_SUBNET) String() (s string) {
...
+	h, err := hex.DecodeString(e.Cookie)
+	if err != nil {
+		return nil, err
+	}
+	return h, nil
+}
+
+func (e *EDNS0_COOKIE) Option() uint16        { return EDNS0COOKIE }
+func (e *EDNS0_COOKIE) unpack(b []byte) error { e.Cookie = hex.EncodeToString(b); return nil }
+func (e *EDNS0_COOKIE) String() string        { return string(e.Cookie) }
```
miekg: This does not need to string conversion, e.Cookie is already a string

miekg:
some nits and could update the README and add the RFC number (is that already known) to it?

edns.go
shuque: Yup, will fix.

shuque:
Thanks. I'll fix the nits and update README. 
RFC not yet published, but is close and the official IANA EDNS option code is already allocated (10) - instead I can cite the latest draft:
https://tools.ietf.org/html/draft-ietf-dnsop-cookies-10

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] small patch for EDN..." ]
>Thanks. I'll fix the nits and update README.
>RFC not yet published, but is close and the official IANA EDNS option code is already allocated (10) - instead I can cite the latest draft:
>https://tools.ietf.org/html/draft-ietf-dnsop-cookies-10

that will work as well, I just want all the drafts/RFCs implemented to be listed
there.

edns.go
```diff
@@ -286,6 +289,32 @@ func (e *EDNS0_SUBNET) String() (s string) {
 	return
 }
 
+// The Cookie EDNS0 option
+//
+//o := new(dns.OPT)
```
miekg: another nit: should be a space here as well

shuque: Ah yes, looks like you fixed the spaces already. Let me know if you wanted me to do anything else.

bortzmeyer:
Cookies are now  RFC 7873 so you may want to update README.md.

PR #352, opened 2016-05-04
Check length before packing OPT

### LINKED ISSUE #351: Panic on edns
Hey,

These days i received panic from dns servers. I checked debug logs and i found the problem.

The Log:
panic: runtime error: index out of range

The Line:
https://github.com/miekg/dns/blob/master/msg.go#L605

Thanks.

miekg:
fixes #351

### DIFF
--- a/msg.go
+++ b/msg.go
@@ -596,7 +596,7 @@ func packStructValue(val reflect.Value, msg []byte, off int, compression map[str
 				for j := 0; j < val.Field(i).Len(); j++ {
 					element := val.Field(i).Index(j).Interface()
 					b, e := element.(EDNS0).pack()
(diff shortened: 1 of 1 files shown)


andrewtj:
Is there an example that demonstrates this bug?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Check length before..." ]
>Is there an example that demonstrates this bug?

No, only a crash report, truncated packet? But not sure how you end up in this
part of the code.

/Miek

--
Miek Gieben

andrewtj:
Yeah neither. I tried to replicate it this morning but couldn't. The commit looks correct though.

PR #356, opened 2016-05-23
Generate pack

miekg:
Remove reflection from the most used types.

This also sets the stage for further development and port the rest of the RRs over.


miekg:
Ok, this starts to look good, @cbranch you want to double check (or bench again?) your bench figures looks slightly better for $reasons

miekg:
@andrewtj @asergeyev it is a bit of a monster PR, but could you look through a bit?

Mostly cut and paste, msg_generate.go generates zmsg.go which packs/unpacks RR using function msg_helper.go. Also the packet's header is packed and unpacked without reflection: 30% speedup when using exdns/reflect and queryperf (50Kqps to 62Kqps on my laptop)

asergeyev:
Ack. Looks interesting.  I will look and run your new bench code tonight.

miekg:
Thank you!

andrewtj:
Is setting `Rdlength` likely to upset anyone?

Is it worth generating DNSSEC and DNS Update equality helpers while you're at it?

asergeyev:
Looks clean IMO... I only see minor stuff.

What is reason for changing TTL from 3600 to 16 in couple tests?

Any reasoning for not having separate pack/unpack for 48bit vs using boolean arg?

unpackString has few constructs like ```s = append(s, `\t`...)``` I don't see reason for replacing it to ```s = append(s, '\t')```

Piece ```txtTmp := make([]byte, 256*4+1)``` seem not clear where size comes from. Also not clear if it's easier to make static array with ```var txtTmp [1025]byte``` (either way works).

From nitpicking category... reverseInt16 and so can use length argument in make() 

Commented import in msg_generate can be removed. There is commented code too... In case if you want to clean it up.

I'd suggest adding AAAA benchmarks and compare copy vs consecutive offsetting and assigning.

asergeyev:
Maybe it could be useful to run ```perf record <server command>``` and ```perf report``` with your queryperf setup to see hot places in code if you want to look at things that are actually important for critical path, not for code benchmarking...

andrewtj:
> On 1 Jun 2016, at 11:17 AM, Alex Sergeyev <notifications@github.com> wrote:
> 
> unpackString has few constructs like s = append(s, `\t`...) I don't see reason for replacing it to s = append(s, '\t')
> 
> Piece txtTmp := make([]byte, 256*4+1) seem not clear where size comes from. Also not clear if it's easier to make static array with var txtTmp [1025]byte (either way works).
> 

I'm to blame for the TXT string stuff. The 4 comes from the `\DDD` byte presentation. Not sure why I went for 256 (rather than 255) and the `+ 1` though.

miekg:
>Looks clean and great... I only see minor stuff.
>
>What is reason for changing TTL from 3600 to 16 in couple tests?

So that I can spot the TTL in the binary message output better :-) Though I had some off-by-one issues and needed to debug this. No other reason

>Any reasoning for not having separate pack/unpack for 48bit vs using boolean arg?

Done.

>unpackString has few constructs like s = append(s, `\t`...) I don't see reason for not replacing it to s = >append(s, '\t')

Done.

>Piece txtTmp := make([]byte, 256*4+1) seem not clear where size comes from. Also not clear if it's >easier to make static array with var txtTmp [1025]byte (either way works fine though).

Andrew commented on that.

>From nitpicking category... reverseInt16 and so can use length argument in make()

Done.

>Commented import in msg_generate can be removed. There is commented code too... In case if you >want to clean it up.

Done. The commented out Slice stuff in msg_generated is needed for nsec and a few other record. Should be fixed in upcoming and smaller PRs.

> From performance paranoia I'd suggest adding AAAA benchmarks and compare copy vs consecutive > offsetting and assigning (such a subtle difference if any though...)

Benchmark added, but I've not bothered with actually checking this.

Perf the whole server a good idaea.

miekg:
> Is it worth generating DNSSEC and DNS Update equality helpers while you're at it?

Which ones to you mean?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Generate pack (#356..." ]
>Is setting `Rdlength` likely to upset anyone?

Should be fine, I thought we did the same thing when packing already (too lazy
to dig up the code though)

>
>Is it worth generating DNSSEC and DNS Update equality helpers while you're at it?
>
>---
>You are receiving this because you authored the thread.
>https://github.com/miekg/dns/pull/356#issuecomment-222851020
/Miek

--
Miek Gieben

miekg:
```
# Overhead  Command  Shared Object        Symbol                                                     
# ........  .......  ...................  ...........................................................
#
     8.67%  reflect  reflect              [.] runtime.pcvalue                                        
     7.00%  reflect  reflect              [.] runtime.readvarint                                     
     6.65%  reflect  reflect              [.] runtime.adjustpointers                                 
     4.57%  reflect  reflect              [.] runtime.step                                           
     3.88%  reflect  reflect              [.] runtime.mallocgc                                       
     2.37%  reflect  reflect              [.] runtime.gentraceback                                   
     2.02%  reflect  reflect              [.] runtime.adjustframe                                    
     1.47%  reflect  reflect              [.] runtime.findfunc                                       
     1.44%  reflect  reflect              [.] github.com/miekg/dns.packDomainName                    
     1.32%  reflect  reflect              [.] runtime.duffcopy                                       
     1.10%  reflect  reflect              [.] runtime.heapBitsSetType                                
     1.01%  reflect  reflect              [.] runtime.makeslice                                      
     1.01%  reflect  reflect              [.] runtime.memmove                                        
     0.81%  reflect  reflect              [.] github.com/miekg/dns.packStructValue                   
     0.80%  reflect  reflect              [.] runtime/internal/atomic.Cas                            
     0.77%  reflect  reflect              [.] runtime.(*mspan).sweep.func1                           
     0.67%  reflect  reflect              [.] main.handleReflect                                     
     0.65%  reflect  reflect              [.] github.com/miekg/dns.UnpackDomainName                  
     0.61%  reflect  reflect              [.] runtime.memclr                                         
     0.60%  reflect  reflect              [.] github.com/miekg/dns.unpackMsgHdr                      
     0.60%  reflect  reflect              [.] runtime.systemstack                                    
     0.59%  reflect  reflect              [.] reflect.(*structType).Field                            
     0.56%  reflect  reflect              [.] runtime/internal/atomic.Xchg                           
     0.53%  reflect  reflect              [.] runtime.funcdata                                       
     0.53%  reflect  [kernel.kallsyms]    [k] 0xffffffff813f69b5                                     
     0.53%  reflect  reflect              [.] runtime.funcspdelta                                    
     0.51%  reflect  reflect              [.] github.com/miekg/dns.(*Server).serve 
```
from perf...not sure what it says... Use less memory, see what we can tweak in (Un)PackDomainName?

asergeyev:
I don't have anything new. Looks like most overhead in your perf is from stack walking and it could be either from deep recursion or from too many go-routines running. Which is likely latter in your case. Surely performance problems are not in packing/unpacking code (and pack|unpack name is a hard problem to deal with... could be whole separate ticket to see if people have ideas around those funcs)

miekg:
OK. Thanks. Will be merging this soon

miekg:
manually merged. Thanks all!

Issue #311, opened 2016-02-15
Missing lock in ServeMux.HandleRemove causes concurrent r/w map access


eswierk:
Running a test program compiled with the latest go 1.6 release candidate, I'm seeing occasional "fatal error: concurrent map read and map write" panics. The panic is a result of a new runtime check (https://github.com/golang/go/commit/50c50420).

The stack trace points to ServeMux.match where it reads the map mux.z. Some other goroutine must be writing the same map concurrently. The only place I can find an access unprotected by a lock is ServeMux.HandleRemove.

I can make the crash go away by wrapping the delete(mux.z, ...) with calls to mux.m.Lock/Unlock (and removing the "don't need a mutex here" comment ;-).

eswierk:
Just realized I was debugging an old fork. The problem was fixed by https://github.com/miekg/dns/commit/a6742d53 .

PR #357, opened 2016-05-29
Create better msg.Len() benchmark

miekg:
See #311 for the discussion.


miekg:
too lazy to fix merge conflicts, trivial do to later (if we care)

PR #359, opened 2016-06-04
Seed math/rand with crypto/rand

rolandshoemaker:
The performance arguments given in #48 and #102 are perfectly valid reasons to not use `crypto/rand` for every message ID but that doesn't mean you have to use the default seed ([which is literally just `1`](https://golang.org/src/math/rand/rand.go#L178)) when using `math/rand`.

This patch simply reads out a random `int64` using `crypto/rand` on package initialization and uses it to seed the default `math/rand` source. While there are obviously other more viable attacks this should make message ID spoofing/prediction much harder when using the default `Id`.

This also changes `id()` to just use a rand `int32` and bound it to the max size of a `uint16` instead of using `time.Now` since it's a relatively easy to guess variable so it shouldn't actually provide that many useful bits (since someone analyzing the messages will generally know the period the messages were generated in).

### DIFF
--- a/msg.go
+++ b/msg.go
@@ -11,14 +11,27 @@ package dns
 //go:generate go run msg_generate.go
 
 import (
+	crand "crypto/rand"
+	"encoding/binary"
 	"encoding/hex"
+	"math"
 	"math/big"
 	"math/rand"
 	"net"
 	"reflect"
 	"strconv"
-	"time"
 )
+
+func init() {
+	// Initialize default math/rand source using crypto/rand to provide better
+	// security without the performance trade-off.
+	buf := make([]byte, 8)
+	_, err := crand.Read(buf)
+	if err != nil {
+		panic(err)
+	}
+	rand.Seed(int64(binary.BigEndian.Uint64(buf) % math.MaxInt64))
+}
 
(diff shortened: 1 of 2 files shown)


miekg:
Nice!

Don't like the panic though as that is a backwards incomp. change. Can't we just default to the current setup then? I.e. seed with 1?

rolandshoemaker:
Good point, I've also cleaned up the various integer bounding by leaving it to the compiler (it's much cleverer than me anyway).

msg.go
```diff
@@ -11,15 +11,32 @@ package dns
...
 
+func init() {
+	// Initialize default math/rand source using crypto/rand to provide better
+	// security without the performance trade-off.
+	buf := make([]byte, 8)
+	_, err := crand.Read(buf)
+	if err != nil {
+		// Failed to read from cryptographic source, warn user and
+		// fallback to default initial seed (1) by returning early
+		log.Printf("Failed to seed math/rand with crypto/rand.Read for message ID generation: %s\n", err)
```
miekg: prolly nobody ever will read this and the random is no worse then now, so I would just drop the log - leave the comment though!

PR #363, opened 2016-06-08
Simplify packing and unpacking

michaelharo:


msg_helpers.go
```diff
@@ -53,9 +44,7 @@ func unpackDataAAAA(msg []byte, off int) (net.IP, int, error) {
 	if off+net.IPv6len > len(msg) {
 		return nil, len(msg), &Error{err: "overflow unpacking aaaa"}
 	}
-	aaaa := net.IP{msg[off], msg[off+1], msg[off+2], msg[off+3], msg[off+4],
-		msg[off+5], msg[off+6], msg[off+7], msg[off+8], msg[off+9], msg[off+10],
-		msg[off+11], msg[off+12], msg[off+13], msg[off+14], msg[off+15]}
+	aaaa := net.IP(msg[off : off+net.IPv6len])
```
michaelharo: I'm unsure if this needed to be a byte copy or if referencing the original buffer was acceptable.

andrewtj:
Making the `net.IP`'s references will probably subtly break code.

I think I avoided the `fmt.Sprintf` machinery for performance reasons. It's probably not important though.

msg_helpers.go
asergeyev: I think it might be risky for situation when someone reuses same buffer to read messages into. If you go to append(make([]byte, 0, net.IPv6len), msg[..]) I'd recommend making a silly benchmark to see if that's more efficient than constructing array directly.

Similarly above with replacing net.IPv4 to net.IP typecasting. Might be better to use similar net.IP{....} construct.

michaelharo:
I debated allocating a new buffer and copying into it for the unpack functions.  It wasn't clear to me how that would be necessary.  Would you like me to create new buffers to copy bytes into from the msg buffer?

I was also debating changing some of the uint functions to use the functions in the encoding/binary package.  Do you have thoughts on that?

msg_helpers.go
```diff
@@ -314,23 +288,11 @@ func unpackString(msg []byte, off int) (string, int, error) {
...
-				var buf [3]byte
-				bufs := strconv.AppendInt(buf[:0], int64(b), 10)
-				s = append(s, '\\')
-				for i := 0; i < 3-len(bufs); i++ {
-					s = append(s, '0')
-				}
-				for _, r := range bufs {
-					s = append(s, r)
-				}
+				s = append(s, fmt.Sprintf("\\%03d", b)...)
```
asergeyev: Also silly benchmark could be great. On other hand

```
   if b < 9 {
       s = append (s, '\\', '0', '0', '0'+b)
   } else if b < 32 {
       s = append (s, '\\', '0', '0'+(b/10), '0'+(b%10))
   } else if b >= 200 {
       s = append (s, '\\', '2', '0'+((b/10)%10), '0'+(b%10))
   } else if b > 127 {
       s = append (s, '\\', '1', '0'+((b/10)%10), '0'+(b%10))
   }
```

Might be worth micro-benchmarking too.

michaelharo: How often is unpackString used?  I'm wondering if it's better to optimize for readability or performance.

asergeyev: Good point, I did not see it's not a core func. Up to @miekg of course but I don't have more questions here.

msg_helpers.go
michaelharo: Even after adding the append(....) you suggested I'm seeing better performance vs calling net.IPv4(...).

asergeyev: cool, thanks for checking.

msg_helpers.go
miekg: benchmarking this makes sense, unpackString is used to txt/spf etc. Rather not re-introduce reflection for fmt.Sprintf here...

miekg:
@andrewtj what do you mean with "Making the net.IP's references will probably subtly break code."?

miekg:
Thanks for the other comments. This all looks good, I am only wondering about using `fmt.Sprintf`

andrewtj:
> On 8 Jun 2016, at 4:57 PM, Miek Gieben <notifications@github.com> wrote:
> 
> @andrewtj what do you mean with "Making the net.IP's references will probably subtly break code."?

If the net.IP is backed by a slice from msg, mutating msg mutates the IP. So the net.IP is only sane as long as msg isn't reused.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Simplify packing an..." ]
>
>> On 8 Jun 2016, at 4:57 PM, Miek Gieben <notifications@github.com> wrote:
>>
>> @andrewtj what do you mean with "Making the net.IP's references will probably subtly break code."?
>
>If the net.IP is backed by a slice from msg, mutating msg mutates the IP. So the net.IP is only sane as long as msg isn't reused.

Ack, we don't want that.

Still I am a bit confused (blame it on the lack of coffee). This is (one of the) 
lines you're talking about:

aaaa := append(make(net.IP, 0, net.IPv6len), msg[off:off+net.IPv6len]...)


/Miek

--
Miek Gieben

andrewtj:
Ahh, I see where you're coming from - my comment was way earlier in the day on a now out of date commit that had `a := net.IP(msg[off : off+net.IPv4len])` and the like, which was addressed in a27ea45.

The context to the comment wouldn't have been lost if I'd made it on the commit rather than on the PR, so that's what I'll try to remember to do next time.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Simplify packing an..." ]
>Ahh, I see where you're coming from - my comment was way earlier in the day on a now out of date commit that had `a := net.IP(msg[off : off+net.IPv4len])` and the like, which was addressed in a27ea45.
>
>The context to the comment wouldn't have been lost if I'd made it on the commit rather than on the PR, so that's what I'll try to remember to do next time.

Ah. Thanks.

The PR/comment mechanism used by @github isn't the best in the world. Wish they 
would use something else/better.

/Miek

--
Miek Gieben

michaelharo:
I removed the Sprintf change so that the rest of this change may be pulled.

miekg:
thanks. cleanup like this make all the difference in the long run.

PR #364, opened 2016-06-08
Use encoding/binary's conversion functions when possible.

michaelharo:


miekg:
Can you use PutUint64 for the uint48 functions as well? Or does that become too hacky?

Assuming this has no impact on performance? It might actually be faster? (Sorry to keep going on about performance - it is just the theme of the month :) )

michaelharo:
I couldn't decide if it was worth it.  I think we'd have to allocate a
buffer and then copy into it starting at index 2 and then pass that to the
PutUint64.  It's extra work so I assume it would be slower, but I haven't
benchmarked.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Use encoding/binary..." ]
>I couldn't decide if it was worth it.  I think we'd have to allocate a
>buffer and then copy into it starting at index 2 and then pass that to the
>PutUint64.  It's extra work so I assume it would be slower, but I haven't
>benchmarked.

Ok, let it be :) I'll merged this as-is.

Thanks!

/Miek

--
Miek Gieben

PR #366, opened 2016-06-08
Remove temporary buffer used by packTxtString.

michaelharo:
This buffer was used to copy a string into it before copying that copy
into msg.  The intermediate buffer was never modified so there is no
need to make a copy.

### DIFF
--- a/msg.go
+++ b/msg.go
@@ -403,20 +403,17 @@ Loop:
 	return string(s), off1, nil
 }
 
-func packTxt(txt []string, msg []byte, offset int, tmp []byte) (int, error) {
+func packTxt(txt []string, msg []byte, offset int) (int, error) {
 	var err error
+	if offset >= len(msg) {
+		return offset, ErrBuf
+	}
 	if len(txt) == 0 {
-		if offset >= len(msg) {
-			return offset, ErrBuf
-		}
 		msg[offset] = 0
 		return offset, nil
 	}
 	for i := range txt {
-		if len(txt[i]) > len(tmp) {
-			return offset, ErrBuf
-		}
-		offset, err = packTxtString(txt[i], msg, offset, tmp)
+		offset, err = packTxtString(txt[i], msg, offset)
 		if err != nil {
 			return offset, err
 		}
@@ -424,14 +421,13 @@ func packTxt(txt []string, msg []byte, offset int, tmp []byte) (int, error) {
 	return offset, err
 }
 
-func packTxtString(s string, msg []byte, offset int, tmp []byte) (int, error) {
-	lenByteOffset := offset
-	if offset >= len(msg) || len(s) > len(tmp) {
+func packTxtString(s string, msg []byte, offset int) (int, error) {
+	if offset >= len(msg) {
 		return offset, ErrBuf
 	}
+	lenByteOffset := offset
 	offset++
-	bs := tmp[:len(s)]
-	copy(bs, s)
+	bs := []byte(s)
 	for i := 0; i < len(bs); i++ {
 		if len(msg) <= offset {
 			return offset, ErrBuf
@@ -551,7 +547,6 @@ func unpackTxtString(msg []byte, offset int) (string, int, error) {
 // Pack a reflect.StructValue into msg.  Struct members can only be uint8, uint16, uint32, string,
 // slices and other (often anonymous) structs.
 func packStructValue(val reflect.Value, msg []byte, off int, compression map[string]int, compress bool) (off1 int, err error) {
-	var txtTmp []byte
 	lenmsg := len(msg)
 	numfield := val.NumField()
 	for i := 0; i < numfield; i++ {
@@ -588,10 +583,7 @@ func packStructValue(val reflect.Value, msg []byte, off int, compression map[str
 					}
 				}
 			case `dns:"txt"`:
-				if txtTmp == nil {
-					txtTmp = make([]byte, 256*4+1)
-				}
-				off, err = packTxt(fv.Interface().([]string), msg, off, txtTmp)
+				off, err = packTxt(fv.Interface().([]string), msg, off)
 				if err != nil {
 					return lenmsg, err
 				}
@@ -830,10 +822,7 @@ func packStructValue(val reflect.Value, msg []byte, off int, compression map[str
 			case `dns:"txt"`:
 				fallthrough
 			case "":
-				if txtTmp == nil {
-					txtTmp = make([]byte, 256*4+1)
-				}
-				off, err = packTxtString(fv.String(), msg, off, txtTmp)
+				off, err = packTxtString(fv.String(), msg, off)
 				if err != nil {
 					return lenmsg, err
 				}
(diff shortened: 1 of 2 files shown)


miekg:
Yes, but I believe the original idea was to re-use the buffer, hence it was an input parameter.

msg.go
```diff
@@ -403,35 +403,31 @@ Loop:
...
 	var err error
+	if offset >= len(msg) {
+		return offset, ErrBuf
+	}
 	if len(txt) == 0 {
-		if offset >= len(msg) {
-			return offset, ErrBuf
-		}
 		msg[offset] = 0
 		return offset, nil
```
michaelharo: Is this the correct behavior when len(txt) == 0?

Is this any different from
```
if len(txt) == 0 {
    return offset, nil
}
```
without doing the 0 assignment to the returned offset?

michaelharo:
It looks like currently a new 1k+1b buffer is allocated per RR that has a string type in it.
I see the string buffers used by the following RR types: (type: count)
GPOS: 3
HINFO: 2
NAPTR: 3
TKEY: 2
TXT: potentially > 1

Count listed above minus 1 is the savings from reusing the buffer.  Is this worth it?  If yes, then I think the code should be changed so that a nil buffer could be passed to indicate create it locally as many uses of this function allocate a buffer immediately before calling the function.

I wrote a quick benchmark to try to test this.  Please let me know how it's flawed.  This test assumes there are either 1 or 3 strings to be converted per RR and that the buffer isn't shared across RRs.

```
func BenchmarkWithoutBuffer3(b *testing.B) {
  s1 := "My string is this long"
  s2 := "My other string is long too"
  s3 := "My last string is not short."
  b.ReportAllocs()
  b.ResetTimer()
  for i := 0; i < b.N; i++ {
    _ = []byte(s1)
    _ = []byte(s2)
    _ = []byte(s3)
  }
}
func BenchmarkWithBuffer3(b *testing.B) {
  s1 := "My string is this long"
  s2 := "My other string is long too"
  s3 := "My last string is not short."
  b.ReportAllocs()
  b.ResetTimer()
  for i := 0; i < b.N; i++ {
    buf := make([]byte, 1025)
    copy(buf, s1)
    copy(buf, s2)
    copy(buf, s3)
  }
}

func BenchmarkWithoutBuffer1(b *testing.B) {
  s1 := "My string is this long"
  b.ReportAllocs()
  b.ResetTimer()
  for i := 0; i < b.N; i++ {
    _ = []byte(s1)
  }
}
func BenchmarkWithBuffer1(b *testing.B) {
  s1 := "My string is this long"
  b.ReportAllocs()
  b.ResetTimer()
  for i := 0; i < b.N; i++ {
    buf := make([]byte, 1025)
    copy(buf, s1)
  }
}
```

The results from running this:
```
BenchmarkWithoutBuffer3-6       50000000                37.6 ns/op             0 B/op          0 allocs/op
BenchmarkWithBuffer3-6          20000000                57.0 ns/op             0 B/op          0 allocs/op
BenchmarkWithoutBuffer1-6       100000000               13.3 ns/op             0 B/op          0 allocs/op
BenchmarkWithBuffer1-6          30000000                47.2 ns/op             0 B/op          0 allocs/op

msg.go
miekg: [ Quoting <notifications@github.com> in "Re: [miekg/dns] Remove temporary bu..." ]
>>  	if len(txt) == 0 {
>> -		if offset >= len(msg) {
>> -			return offset, ErrBuf
>> -		}
>>  		msg[offset] = 0
>>  		return offset, nil
>
>Is this the correct behavior when len(txt) == 0?
>
>Is this any different from
>```
>if len(txt) == 0 {
>    return offset, nil
>}
>```
>without doing the 0 assignment to the returned offset?

yes, you need to see the length of the txt in pkt to 0 to signal; no text

michaelharo: should offset be incremented or is that happening somewhere else?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Remove temporary bu..." ]
>It looks like currently a new 1k+1b buffer is allocated per RR that has a string type in it.

The main use of re-using that buffer was if there are multiple txt segments you 
don't need to allocate a new buffer for each segment. This is, I think, not true
(or should be fixed) in the world where we generate zmsg.go from msg_generate.go

>I see the string buffers used by the following RR types: (type: count)
>GPOS: 3
>HINFO: 2
>NAPTR: 3
>TKEY: 2
>TXT: potentially > 1
>
>Count listed above minus 1 is the savings from reusing the buffer.  Is this worth it?  If yes, then I think the code should be changed so that a nil buffer could be passed to indicate create it locally as many uses of this function allocate a buffer immediately before calling the function.
>
>I wrote a quick benchmark to try to test this.  Please let me know how it's flawed.  This test assumes there are either 1 or 3 strings to be converted per RR and that the buffer isn't shared across RRs.
>
>```
>func BenchmarkWithoutBuffer3(b *testing.B) {
>  s1 := "My string is this long"
>  s2 := "My other string is long too"
>  s3 := "My last string is not short."
>  b.ReportAllocs()
>  b.ResetTimer()
>  for i := 0; i < b.N; i++ {
>    _ = []byte(s1)
>    _ = []byte(s2)
>    _ = []byte(s3)
>  }
>}
>func BenchmarkWithBuffer3(b *testing.B) {
>  s1 := "My string is this long"
>  s2 := "My other string is long too"
>  s3 := "My last string is not short."
>  b.ReportAllocs()
>  b.ResetTimer()
>  for i := 0; i < b.N; i++ {
>    buf := make([]byte, 1025)
>    copy(buf, s1)
>    copy(buf, s2)
>    copy(buf, s3)
>  }
>}
>
>func BenchmarkWithoutBuffer1(b *testing.B) {
>  s1 := "My string is this long"
>  b.ReportAllocs()
>  b.ResetTimer()
>  for i := 0; i < b.N; i++ {
>    _ = []byte(s1)
>  }
>}
>func BenchmarkWithBuffer1(b *testing.B) {
>  s1 := "My string is this long"
>  b.ReportAllocs()
>  b.ResetTimer()
>  for i := 0; i < b.N; i++ {
>    buf := make([]byte, 1025)
>    copy(buf, s1)
>  }
>}
>```
>
>The results from running this:
>```
>BenchmarkWithoutBuffer3-6       50000000                37.6 ns/op             0 B/op          0 allocs/op
>BenchmarkWithBuffer3-6          20000000                57.0 ns/op             0 B/op          0 allocs/op
>BenchmarkWithoutBuffer1-6       100000000               13.3 ns/op             0 B/op          0 allocs/op
>BenchmarkWithBuffer1-6          30000000                47.2 ns/op             0 B/op          0 allocs/op
>
>---
>You are receiving this because you commented.
>https://github.com/miekg/dns/pull/366#issuecomment-224695107
/Miek

--
Miek Gieben

msg.go
miekg: [ Quoting <notifications@github.com> in "Re: [miekg/dns] Remove temporary bu..." ]
>>  	if len(txt) == 0 {
>> -		if offset >= len(msg) {
>> -			return offset, ErrBuf
>> -		}
>>  		msg[offset] = 0
>>  		return offset, nil
>
>should offset be incremented or is that happening somewhere else?

hmm, good question

/Miek

--
Miek Gieben

miekg: We should check and add a test for this.

miekg:
coming back to this I think the original function is correct; it correctly creates the binary that is seen on the wire.

PR #369, opened 2016-06-11
Skip reflection for most types

miekg:
Make the reflection types a black list:

CAA, HIP, IPSECKEY, NSEC3, NSEC, SIG, TSIG, URI,

All other types don't use reflection anymore.

Return a pointer to the header when there is no rdata, this restores old
behavior. The rest of the conversion mostly hangs on getting size-hex
right, but then packStruct and packStructValue and the unpack variant
can be killed.

Generate pack and unpack for all embedded types as well.

### DIFF
--- a/msg.go
+++ b/msg.go
@@ -587,35 +587,6 @@ func packStructValue(val reflect.Value, msg []byte, off int, compression map[str
 						return lenmsg, err
 					}
 				}
-			case `dns:"txt"`:
-				if txtTmp == nil {
-					txtTmp = make([]byte, 256*4+1)
-				}
-				off, err = packTxt(fv.Interface().([]string), msg, off, txtTmp)
-				if err != nil {
-					return lenmsg, err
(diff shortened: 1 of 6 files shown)


miekg:
@asergeyev this works, except for:
~~~
% go t                                                                                                                   ~/g/src/github.com/miekg/dns cleanup-noreflect +
--- FAIL: TestPrivateByteSlice (0.00s)
	privaterr_test.go:84: offset is not matching to length of Private RR: 17!=40
	privaterr_test.go:89: got error unpacking ISBN: dns: buffer size too small
FAIL
exit status 1
~~~
I'm afraid I've changed some (hidden) semantics on the Len() stuff. Could you check wth is going on here? Thanks!

miekg:
@asergeyev nvm, the `pack` for the privaterr type didn't pack the header. Working.  Now failing elsewhere, but not as severe.

PR #384, opened 2016-07-12
(WIP) Added support for more than one handler

mbhinder:
This commit changes the match function for handlers to return more than
one handler in the mux.

The ServeDNS function iterates over all the handlers and buffers any
messages written onto the wire by them. The message from the final
handler is combined with the other messages and finally written out onto
the wire.

### DIFF
--- a/server.go
+++ b/server.go
@@ -132,10 +132,10 @@ func ActivateAndServe(l net.Listener, p net.PacketConn, handler Handler) error {
 	return server.ActivateAndServe()
 }
 
-func (mux *ServeMux) match(q string, t uint16) Handler {
+func (mux *ServeMux) match(q string, t uint16) []Handler {
 	mux.m.RLock()
 	defer mux.m.RUnlock()
-	var handler Handler
+	var handler []Handler
 	b := make([]byte, len(q)) // worst case, one label of length q
 	off := 0
 	end := false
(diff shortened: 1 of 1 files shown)


mbhinder:
I needed this change for something that I'm working on. Any interest in getting this merged into master? Please treat it as a WIP PR for initial feedback. If there is any interest, I'll add unit test cases. However, I've tested it manually and everything works.

server.go
```diff
@@ -148,20 +148,18 @@ func (mux *ServeMux) match(q string, t uint16) Handler {
 			}
 		}
 		if h, ok := mux.z[string(b[:l])]; ok { // 'causes garbage, might want to change the map key
-			if t != TypeDS {
```
mbhinder: Should TypeDS only return one handler - the parent one?

miekg:
Not feeling really inclined to merge this. As you mention yourself you have trouble right away with the TypeDS handler and it also adds more allocations to a critical query path.

May I ask what you are attempting to do and why you want this?

mbhinder:
We could keep the existing behavior of the TypeDS handler as is but I completely understand.

In our code we use handlers for two different patterns like `foo.com` and `bar.foo.com` The actual DNS result is returned from the `foo.com` handler and the `bar.foo.com` handler has some other logic in it for  firewall logic. In the current code, the `foo.com` handler is not invoked.

mbhinder:
Also, regarding more allocations, I'd just like add that in most cases which have only one callback, the code path is mostly unchanged.

miekg:
I see. I think another way to do is to have 1 handler for foo.com (or .) and re-implement the specific handling you'll need. 

Also I seem to remember that I've implement the surrounding bits that you can define your own `mux` and do whatever you like.

mbhinder:
Thanks for looking at it, will look at the other ways of implementing this in our code. 👍

miekg:
np, yeah, the default hander is `dns.DefaultServeMux`, but you should be able to add your own. If not then that def. is a bug.

PR #385, opened 2016-07-12
Added support for more than one handler

mbhinder:
This commit changes the match function for handlers to return more than
one handler in the mux.

The ServeDNS function iterates over all the handlers and buffers any
messages written onto the wire by them. The message from the final
handler is combined with the other messages and finally written out onto
the wire.


mbhinder:
Created a duplicate PR by mistake.

PR #386, opened 2016-07-12
Some golint fixes

mbhinder:
Noticed some golint errors, cleaned them up. There are more, can try to go through them as well.

### DIFF
--- a/msg.go
+++ b/msg.go
@@ -58,7 +58,7 @@ var (
 	ErrTruncated     error = &Error{err: "failed to unpack truncated message"} // ErrTruncated indicates that we failed to unpack a truncated message. We unpacked as much as we had so Msg can still be used, if desired.
 )
 
-// Id, by default, returns a 16 bits random number to be used as a
+// Id by default, returns a 16 bits random number to be used as a
 // message id. The random provided should be good enough. This being a
 // variable the function can be reassigned to a custom function.
 // For instance, to make it return a static value:
--- a/server.go
+++ b/server.go
@@ -147,7 +147,7 @@ func (mux *ServeMux) match(q string, t uint16) Handler {
 				b[i] |= ('a' - 'A')
 			}
 		}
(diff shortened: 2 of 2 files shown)


defaults.go
```diff
@@ -82,7 +82,7 @@ func (dns *Msg) SetIxfr(z string, serial uint32, ns, mbox string) *Msg {
 	dns.Question = make([]Question, 1)
 	dns.Ns = make([]RR, 1)
 	s := new(SOA)
-	s.Hdr = RR_Header{z, TypeSOA, ClassINET, defaultTtl, 0}
+	s.Hdr = RR_Header{z, TypeSOA, ClassINET, defaultTTL, 0}
```
miekg: The Ttl -> TTL ship has sailed. Go DNS uses Ttl (although TTL would have been better in hindsight)

ztypes.go
```diff
@@ -282,16 +282,16 @@ func (rr *DNAME) len() int {
 func (rr *DNSKEY) len() int {
 	l := rr.Hdr.len()
 	l += 2 // Flags
-	l += 1 // Protocol
-	l += 1 // Algorithm
+	l++    // Protocol
```
miekg: I like the old style because it's inline with `l +=2` directly above it

msg.go
```diff
@@ -58,7 +58,7 @@ var (
 	ErrTruncated     error = &Error{err: "failed to unpack truncated message"} // ErrTruncated indicates that we failed to unpack a truncated message. We unpacked as much as we had so Msg can still be used, if desired.
 )
 
-// Id, by default, returns a 16 bits random number to be used as a
+// Id by default, returns a 16 bits random number to be used as a
```
miekg: The doc changes are still welcome :) Although I nacked the rest of your changes. Are you willing to make that change?

mbhinder: Done, thanks for reviewing

miekg:
Thanks. Doc fixes are always useful

PR #388, opened 2016-07-18
readme - add dockness to user list

bamarni:
Which is a resolver for docker machines.

And thanks for this great lib :+1:


miekg:
Pull another change, before this one. Care to rebase?

bamarni:
Sure, I've put it just before the last one in case it happens again.

Issue #389, opened 2016-07-18
panics raised in compressionLenHelper() and packDomainName()


johnmah:
i believe that there is a race condition that is causing panics around the code that modifies dns.Compress flag when determining the size of a compressed packet. unfortunately i do not have the original domain(s) that cause the panic at this time.

note the stack trace for PackDomainName() and packDomainName, in theory i would think that these parameters should be almost identical.

i am going to try and fix this by adding a function that determines the size of the compressed package without modifying the internal dns.Compress state.

```
panic: assignment to entry in nil map

goroutine 19713 [running]:
panic(0x86e0e0, 0xc82019b060)
        /usr/local/go/src/runtime/panic.go:464 +0x3e6
github.com/miekg/dns.compressionLenHelper(0x0, 0xc8204749a0, 0x11)
        /jenkins/workspace/dnsproxy-tcpproxy-build/dnsproxy/src/github.com/miekg/dns/msg.go:969 +0x150
github.com/miekg/dns.(*Msg).Len(0xc820495e60, 0x0)
        /jenkins/workspace/dnsproxy-tcpproxy-build/dnsproxy/src/github.com/miekg/dns/msg.go:915 +0xa51
github.com/miekg/dns.(*Msg).PackBuffer(0xc820495e60, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
        /jenkins/workspace/dnsproxy-tcpproxy-build/dnsproxy/src/github.com/miekg/dns/msg.go:737 +0x3ab
github.com/miekg/dns.(*Msg).Pack(0xc820495e60, 0x0, 0x0, 0x0, 0x0, 0x0)
        /jenkins/workspace/dnsproxy-tcpproxy-build/dnsproxy/src/github.com/miekg/dns/msg.go:665 +0x53
github.com/miekg/dns.(*response).WriteMsg(0xc82011bb80, 0xc820495e60, 0x0, 0x0)
        /jenkins/workspace/dnsproxy-tcpproxy-build/dnsproxy/src/github.com/miekg/dns/server.go:671 +0x275
main.(*DNSServer).handleForward(0xc820192140, 0x7f56d6448e18, 0xc82011bb80, 0xc820147050)
        /jenkins/workspace/dnsproxy-tcpproxy-build/dnsproxy/dnserver.go:533 +0x96d
main.(*DNSServer).(main.handleForward)-fm(0x7f56d6448e18, 0xc82011bb80, 0xc820147050)
[...]

PR #390, opened 2016-07-19
don't modify Compress flag, use special UncompressedLen function to c…

johnmah:
…alculate uncompressed size for msg.

the basis for this pull request to demonstrate the cause of the panics mentioned in https://github.com/miekg/dns/issues/389. it admittedly duplicates the base functionality of the Msg.Len() function but in production solves the issues around panics. you might want to solve this using another approach but i don't see one that doesn't involve changing the existing semantics of Msg.Len() (i.e.: passing in a boolean into Len() that denotes whether or not you want the compressed or uncompressed length).

a test scenario would be to start with a Msg with Compress enabled and create two goroutines that repeatedly call PackBuffer. eventually it will cause the panic.

### DIFF
--- a/msg.go
+++ b/msg.go
@@ -732,12 +732,9 @@ func (dns *Msg) PackBuffer(buf []byte) (msg []byte, err error) {
 
 	// We need the uncompressed length here, because we first pack it and then compress it.
 	msg = buf
-	compress := dns.Compress
-	dns.Compress = false
-	if packLen := dns.Len() + 1; len(msg) < packLen {
+	if packLen := dns.UncompressedLen() + 1; len(msg) < packLen {
 		msg = make([]byte, packLen)
 	}
-	dns.Compress = compress
 
 	// Pack it in: header and then the pieces.
 	off := 0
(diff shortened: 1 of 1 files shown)


msg.go
```diff
@@ -732,12 +732,9 @@ func (dns *Msg) PackBuffer(buf []byte) (msg []byte, err error) {
 
 	// We need the uncompressed length here, because we first pack it and then compress it.
 	msg = buf
-	compress := dns.Compress
```
miekg: Hmm, yes, this (admittely) was always a hack

miekg: But this also means you accessing the msg from two different go routines...?

johnmah: it's possible. let's say you implement a caching DNS server that caches a DNS response, and you serve up two requests for the same domain (which has been cached) at roughly the same time.

miekg: [ Quoting <notifications@github.com> in "Re: [miekg/dns] don't modify Compre..." ]
>> @@ -732,12 +732,9 @@ func (dns *Msg) PackBuffer(buf []byte) (msg []byte, err error) {
>>
>>  	// We need the uncompressed length here, because we first pack it and then compress it.
>>  	msg = buf
>> -	compress := dns.Compress
>
>it's possible. let's say you implement a caching DNS server that caches a DNS response, and you serve up two requests for the same domain (which has been cached) at roughly the same time.

But then you would have a problem on the query ID as well? Asked diff.: how do
fix the query id there?

/Miek

--
Miek Gieben

johnmah: we copy the *Msg, then SetReply() to "rewrite" the response. i suppose a deep-copy would work as well (and be safer). it's a different, but related problem.

at the end of the day mutating the Compress flag and using Len() will give racy results.

miekg:
I'm not convinced. Is a test possible that shows you're right?

PR #393, opened 2016-08-07
Enlist zaccone/goPubIP as library user.

zaccone:
Update README.md file and add zaccone/goPubIP application as a library user.


miekg:
Merged this other stuff, care to rebase to resolve the conflict?

PR #400, opened 2016-09-01
adds tests for shutdown bind port

favadi:
# Intro

After calling `Shutdown` of DNS server, the bound port still open. It only happens if the DNS server is running on Linux and use UDP.

### DIFF
--- a/server_test.go
+++ b/server_test.go
@@ -695,6 +695,7 @@ func testShutdownBindPort(t *testing.T, protocol string, port string) {
 	if err := s.Shutdown(); err != nil {
 		t.Fatal(err)
 	}
+	time.Sleep(500 * time.Millisecond)
 	go func() {
 		if err := s.ListenAndServe(); err != nil {
 			t.Fatal(err)
--- a/server_test.go
+++ b/server_test.go
@@ -691,17 +691,17 @@ func testShutdownBindPort(t *testing.T, protocol string, port string) {
 			t.Log(err)
 		}
 	}()
-	time.Sleep(500 * time.Millisecond)
+	time.Sleep(100 * time.Millisecond)
 	if err := s.Shutdown(); err != nil {
 		t.Fatal(err)
 	}
-	time.Sleep(500 * time.Millisecond)
+	time.Sleep(100 * time.Millisecond)
 	go func() {
 		if err := s.ListenAndServe(); err != nil {
 			t.Fatal(err)
 		}
 	}()
-	time.Sleep(500 * time.Millisecond)
+	time.Sleep(100 * time.Millisecond)
 }
 
 func TestShutdownBindPortUDP(t *testing.T) {
--- a/server_test.go
+++ b/server_test.go
@@ -681,17 +681,22 @@ kFsxKCqxAnBVGEWAvVZAiiTOxleQFjz5RnL0BQp9Lg2cQe+dvuUmIAA=
 func testShutdownBindPort(t *testing.T, protocol string, port string) {
 	handler := NewServeMux()
 	handler.HandleFunc(".", func(w ResponseWriter, r *Msg) {})
+	startedCh := make(chan struct{})
 	s := &Server{
 		Addr:    net.JoinHostPort("127.0.0.1", port),
 		Net:     protocol,
 		Handler: handler,
+		NotifyStartedFunc: func() {
+			startedCh <- struct{}{}
+		},
 	}
 	go func() {
 		if err := s.ListenAndServe(); err != nil {
 			t.Log(err)
 		}
 	}()
-	time.Sleep(100 * time.Millisecond)
+	<-startedCh
+	t.Logf("DNS server is started on: %s", s.Addr)
 	if err := s.Shutdown(); err != nil {
 		t.Fatal(err)
 	}
@@ -701,7 +706,8 @@ func testShutdownBindPort(t *testing.T, protocol string, port string) {
 			t.Fatal(err)
 		}
 	}()
-	time.Sleep(100 * time.Millisecond)
+	<-startedCh
+	t.Logf("DNS server is started on: %s", s.Addr)
 }
 
 func TestShutdownBindPortUDP(t *testing.T) {
--- a/server_test.go
+++ b/server_test.go
@@ -677,3 +677,36 @@ zDCJkckCgYEAndqM5KXGk5xYo+MAA1paZcbTUXwaWwjLU+XSRSSoyBEi5xMtfvUb
 kFsxKCqxAnBVGEWAvVZAiiTOxleQFjz5RnL0BQp9Lg2cQe+dvuUmIAA=
 -----END RSA PRIVATE KEY-----`)
 )
+
+func testShutdownBindPort(t *testing.T, protocol string, port string) {
+	handler := NewServeMux()
+	handler.HandleFunc(".", func(w ResponseWriter, r *Msg) {})
+	s := &Server{
+		Addr:    net.JoinHostPort("127.0.0.1", port),
+		Net:     protocol,
+		Handler: handler,
+	}
+	go func() {
+		if err := s.ListenAndServe(); err != nil {
+			t.Log(err)
+		}
+	}()
+	time.Sleep(500 * time.Millisecond)
+	if err := s.Shutdown(); err != nil {
+		t.Fatal(err)
+	}
+	go func() {
+		if err := s.ListenAndServe(); err != nil {
+			t.Fatal(err)
+		}
+	}()
+	time.Sleep(500 * time.Millisecond)
+}
+
+func TestShutdownBindPortUDP(t *testing.T) {
+	testShutdownBindPort(t, "udp", "1153")
+}
+
+func TestShutdownBindPortTCP(t *testing.T) {
+	testShutdownBindPort(t, "tcp", "1154")
+}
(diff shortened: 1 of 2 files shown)


favadi:
This PR is a WIP, I added the test cases and expect the CI test to fail the test `TestShutdownBindPortUDP` with error: 

```
listen udp 127.0.0.1:1153: bind: address already in use
```

favadi:
As I expected the build https://travis-ci.org/miekg/dns/jobs/156725743 fails:
```
--- FAIL: TestShutdownBindPortUDP (1.00s)
	server_test.go:700: listen udp 127.0.0.1:1153: bind: address already in use
```

I'm working on a fix.

secmask:
maybe there's bug in `udp_linux.go`, something like this
```
func setUDPSocketOptions4(conn *net.UDPConn) error {
	file, err := conn.File()
	if err != nil {
		return err
	}
........................
```
```
// File sets the underlying os.File to blocking mode and returns a copy.
// It is the caller's responsibility to close f when finished.
// Closing c does not affect f, and closing f does not affect c.
//
// The returned os.File's file descriptor is different from the connection's.
// Attempting to change properties of the original using this duplicate
// may or may not have the desired effect.
func (c *conn) File() (f *os.File, err error) {
	f, err = c.fd.dup()
	if err != nil {
		err = &OpError{Op: "file", Net: c.fd.net, Source: c.fd.laddr, Addr: c.fd.raddr, Err: err}
	}
	return
}
```
so, we must `Close` this `file`  descriptor in those functions.

favadi:
@secmask I've tried to add `defer file.Close()` to all function in udp_linux.go, but the test still fails.

secmask:
it passed on mine, you may try to comment call to `setUDPSocketOptions` in `server.go`.

favadi:
> it passed on mine, you may try to comment call to setUDPSocketOptions in server.go.

You are right, commit 403a8ccfa829d479469f3fab8ed58b93c5d753ee fixed the problem. I'm not sure why the test failed yesterday in my local VM.

I think the problem is fixed.

favadi:
> I think the problem is fixed.

I was wrong, the CI test passes now, but the test still fail in my local VM. Notice that the TCP port fails too! I'm not sure what's going on.

```
vagrant@vagrant-ubuntu-trusty-64:~$ go test -run '^TestShutdownBindPort' github.com/miekg/dns/...
--- FAIL: TestShutdownBindPortUDP (1.00s)
       	server_test.go:700: listen udp 127.0.0.1:1153: bind: address already in use
--- FAIL: TestShutdownBindPortTCP (1.00s)
       	server_test.go:700: listen tcp 127.0.0.1:1154: bind: address already in use
       	server_test.go:691: accept tcp 127.0.0.1:1154: use of closed network connection
FAIL
FAIL   	github.com/miekg/dns   	2.007s
ok     	github.com/miekg/dns/dnsutil   	0.005s
ok     	github.com/miekg/dns/idn       	0.004s
```

secmask:
> accept tcp 127.0.0.1:1154: use of closed network connection
this because the server connection is `Close()` 2 times, in serv loop (defer statement) and in `Shutdown(...)`
> TCP port fails too
can you check the state of TCP connection at this point , `netstat -ntlp|grep 1154` , TCP listerner may require a delay before we can rebind to the same port if we did not use `REUSE_ADDR`

favadi:
@secmask I'm on mobile right now so I can't check, but I believe go sets REUSE_ADDR for all connections unconditionally.

server_test.go
```diff
@@ -677,3 +677,36 @@ zDCJkckCgYEAndqM5KXGk5xYo+MAA1paZcbTUXwaWwjLU+XSRSSoyBEi5xMtfvUb
...
+	handler := NewServeMux()
+	handler.HandleFunc(".", func(w ResponseWriter, r *Msg) {})
+	s := &Server{
+		Addr:    net.JoinHostPort("127.0.0.1", port),
+		Net:     protocol,
+		Handler: handler,
+	}
+	go func() {
+		if err := s.ListenAndServe(); err != nil {
+			t.Log(err)
```
Gnouc: CI passed due to this line, you  must change `t.Log(err)` to `t.Fatal(err)`

favadi: The reason we use `t.Log` here because it will returns error when we call `s.ShutDown` below, and it is expected.

Gnouc: @favadi Hmm, in case you can't start a server, so how do you expect calling `s.ShutDown` successfully?

Gnouc: @favadi Ah right, got it, thanks.

Gnouc:
> but I believe go sets REUSE_ADDR for all connections unconditionally.

Not for all, only for `syscall.SOCK_STREAM` and `syscall.SOCK_SEQPACKET` socket type.

For udp listener, which is `syscall.SOCK_DGRAM`, you must set `syscall.SO_REUSEADDR` before binding the socket, but it's currently impossible with standard library.

miekg:
What's the state of this PR?

favadi:
> What's the state of this PR?

403a8ccfa829d479469f3fab8ed58b93c5d753ee fixes the problem in CI test, but I still get failures in my local machine. I suspect that something is wrong in the test setup, but haven't have the time to look more carefully. Maybe I will do it this week.

miekg:
ping?

favadi:
I'm looking at this problem now.

server_test.go
```diff
@@ -677,3 +677,36 @@ zDCJkckCgYEAndqM5KXGk5xYo+MAA1paZcbTUXwaWwjLU+XSRSSoyBEi5xMtfvUb
...
+		}
+	}()
+	time.Sleep(500 * time.Millisecond)
+	if err := s.Shutdown(); err != nil {
+		t.Fatal(err)
+	}
```
favadi: @miekg if I add a small time.Sleep duration here, the test will never fail. But if I don't sometimes it fails. I guess the linux network stack still needs a small amount time after closing the connection before  it can be used again. I might depends on some kernel parameters but I can't figure out exactly.

So what can we do about it?

- adds a small sleep to Shutdown function if os is linux
- accepts it and documents somewhere .

favadi: I think user should adds the sleep if needs themselves. This MR is ready to merge.

server_test.go
```diff
@@ -677,3 +677,37 @@ zDCJkckCgYEAndqM5KXGk5xYo+MAA1paZcbTUXwaWwjLU+XSRSSoyBEi5xMtfvUb
...
+		Addr:    net.JoinHostPort("127.0.0.1", port),
+		Net:     protocol,
+		Handler: handler,
+	}
+	go func() {
+		if err := s.ListenAndServe(); err != nil {
+			t.Log(err)
+		}
+	}()
+	time.Sleep(500 * time.Millisecond)
```
miekg: Can we get rid of these random sleeps? Why is this 500 ms btw? Can it be 100? 1000?

udp_linux.go
```diff
@@ -21,6 +21,7 @@ func setUDPSocketOptions4(conn *net.UDPConn) error {
 	if err != nil {
 		return err
 	}
+	defer file.Close()
```
favadi: I removed those defer usages.

server_test.go
favadi: As I said above, I don't know how much time it needs. So I tested and the test passes every time with 100ms. If I removed those sleep completely sometimes test will fails. If you concern about those sleep statements, I think it is OK about comment out the tests and accept it is a known behavior.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] adds tests for shut..." ]
>favadi commented on this pull request.
>
>> @@ -677,3 +677,37 @@ zDCJkckCgYEAndqM5KXGk5xYo+MAA1paZcbTUXwaWwjLU+XSRSSoyBEi5xMtfvUb
> kFsxKCqxAnBVGEWAvVZAiiTOxleQFjz5RnL0BQp9Lg2cQe+dvuUmIAA=
> -----END RSA PRIVATE KEY-----`)
> )
>+
>+func testShutdownBindPort(t *testing.T, protocol string, port string) {
>+	handler := NewServeMux()
>+	handler.HandleFunc(".", func(w ResponseWriter, r *Msg) {})
>+	s := &Server{
>+		Addr:    net.JoinHostPort("127.0.0.1", port),
>+		Net:     protocol,
>+		Handler: handler,
>+	}
>+	go func() {
>+		if err := s.ListenAndServe(); err != nil {
>+			t.Log(err)
>+		}
>+	}()
>+	time.Sleep(500 * time.Millisecond)
>
>As I said above, I don't know how much time it needs. So I tested and the test passes every time with 100ms. If I removed those sleep completely sometimes test will fails. If you concern about those sleep statements, I think it is OK about comment out the tests and accept it is a known behavior.

Yes, I see. It would be better to use the startup notification that is available
in server to check if it started. That way we don't need sleeps at all.

Yes, I've used sleeps as well for getting tests to pass :) But we should start
to do the right thing, and wait for it to have started. This is prolly faster
and less flaky.

secmask:
@favadi relate question http://unix.stackexchange.com/questions/17218/how-long-is-a-tcp-local-socket-address-that-has-been-bound-unavailable-after-clo

miekg:
To clarify: let's see how easy it is to remove the sleeps otherwise I'll just merge. Is good to have this test.

favadi:
@miekg I replaced the sleep after start with notify channel.

miekg:
Awesome. Thanks!

PR #403, opened 2016-09-12
Always set tokenUpper when setting token

rolandshoemaker:
This fixes parsing bugs for certain records, mainly NSEC3. Previously a presentation version of the type bit map `A NS CAA` would be parsed as `A NS NS` as `tokenUpper` was not being properly set when setting `token` which mean't the previous value was used instead.


miekg:
yes thanks!

Issue #406, opened 2016-10-02
edns flag extensions


miekg:
We should probably add: opt.Z() and opt.SetZ() and a generic .Flags() that returns the flags value as an int (also SetFlags() for setting that section of the OPT record)

PR #407, opened 2016-10-04
Simple function to clear the DO bit from an OPT resource record.

shane-kerr:


miekg:
Also see #406. Probably an oversight (from me, not to give SetDo a parameter).

I think btw we can get away with this, but redefining SetDo() as
```
func SetDo(do bool...) {}
```
checking for the number of args, and if it's 1, use that value, otherwise to the current thing.

shane-kerr:
Okay I'll update the branch accordingly - I am a Go n00b so didn't really consider the hand-crated optional parameter rout. :wink:

shane-kerr:
Okay, this is now changed to support an optional argument.

Unfortunately there does not appear to be an easy way in Go to require that there be only one argument. You can return an error, but anyone who accidentally passes 2+ parameters probably is not checking errors carefully. I guess it could call `panic()` in that case?

edns_test.go
```diff
@@ -15,6 +15,18 @@ func TestOPTTtl(t *testing.T) {
...
 	}
+	e.SetDo(true)
+	if !e.Do() {
+		t.Errorf("DO bit should still be non-zero")
+	}
+	e.SetDo(false)
+	if e.Do() {
+		t.Errorf("DO bit should be zero")
+	}
+	e.SetDo(false)
```
miekg: does e.SetDo() also still work?

edns.go
```diff
@@ -128,8 +128,17 @@ func (rr *OPT) Do() bool {
 }
 
 // SetDo sets the DO (DNSSEC OK) bit.
-func (rr *OPT) SetDo() {
-	rr.Hdr.Ttl |= _DO
+func (rr *OPT) SetDo(do ...bool) {
+	// if we pass an argument, set the DO bit to that value
```
miekg: Put this comment in the exported doc string above the function. Also tell we will silently ignore more than 1 argument.

miekg:
No, panic() in a lib is usually a bad idea.

edns_test.go
shane-kerr: Actually this was tested, but there was a missing test or two. I made a comment explaining comprehensive tests for this, and annotated each call to explain which case is covered. Hopefully that makes it clear (and hopefully I didn't mess it up).

edns.go
shane-kerr: Okay, done.

shane-kerr:
The last commit addresses the review comments - thanks!

miekg:
Nice. Thanks.

PR #413, opened 2016-10-17
Respect range of RFC6895

marcusgrando:


miekg:
Geh :-)

asergeyev:
ouch :(  looked at the RCODE section probably when was writing those tests...

PR #418, opened 2016-11-02
Enables checking of tranports on a per-Msg basis.

fbettag:
In order to not be a participant of DNS amplification attacks,
one needs to limit record/response sizes for UDP connections.
One solution would be to curry the function for the Handler and pass it the listening transport,
but this would mean i need different muxes for different listeners and keep the up2date.

To avoid this uglyness, i added w.Tcp() and w.Udp() to determine
wether the writer's transport is either or, thus being able to limit
response sizes easily and conditionally.

### DIFF
--- a/server.go
+++ b/server.go
@@ -40,6 +40,8 @@ type ResponseWriter interface {
 	// Hijack lets the caller take over the connection.
 	// After a call to Hijack(), the DNS package will not do anything with the connection.
 	Hijack()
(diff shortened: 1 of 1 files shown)


miekg:
you don't need that, you can just use: https://github.com/miekg/exdns/blob/master/reflect/reflect.go#L69

fbettag:
at least we have it documented now ;)

PR #419, opened 2016-11-21
Add nil check for interface value in ActivateAndServe

sanimej:
This will avoid the panic if the library user inadvertently initializes Server.PacketConn with a nil pointer

https://github.com/docker/docker/issues/28112

Signed-off-by: Santhosh Manohar <santhosh@docker.com>

### DIFF
--- a/server.go
+++ b/server.go
@@ -389,7 +389,9 @@ func (srv *Server) ActivateAndServe() error {
 		if srv.UDPSize == 0 {
 			srv.UDPSize = MinMsgSize
 		}
-		if t, ok := pConn.(*net.UDPConn); ok {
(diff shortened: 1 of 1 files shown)


miekg:
Why should the dns package do this?

sanimej:
Because its a basic validation of the inputs to an exported function.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Add nil check for i..." ]
>Because its a basic validation of the inputs to an exported function.

Yes, I was afraid we silently discard stuff and then "do nothing", but that's
not the case.

/Miek

--
Miek Gieben

Issue #420, opened 2016-11-23
TXT data is parsed incorrectly


gibson042:
https://github.com/miekg/dns/blob/46df8c9462fa287d54225d6bdebbcfd41da434f7/parse_test.go#L429 requires that `t.example.com. 3600 IN TXT aaa aaa` is considered equivalent to `t.example.com.	3600	IN	TXT	"aaa aaa"`, but that does not match real-world nameserver behavior, which expects `t.example.com. IN TXT "aaa" "aaa"` per RFC 1035:

[Section 3.3.14 TXT RDATA format](https://tools.ietf.org/html/rfc1035#section-3.3.14):
> TXT-DATA        One or more \<character-string>s.

[Section 5.1 Master File Format](https://tools.ietf.org/html/rfc1035#section-5.1) (emphasis mine):
> \<character-string> is expressed in one or two ways: as **a contiguous set of characters without interior spaces**, or as a string beginning with a " and ending with a ".  Inside a " delimited string any character can occur, except for a " itself, which must be quoted using \ (back slash).

asergeyev:
IMO when people manually write TXT records without quotes, they don't expect anything but keeping spaces internal to the whole text and not a "shellwords" quoting.   With that said I doubt many people who don't use quotes will be affected when their canonical representation uses either "... ..." or "..." "..." form.

edmonds:
> vIMO when people manually write TXT records without quotes, they don't expect anything but keeping spaces internal to the whole text and not a "shell" "words" quoting.

Sometimes I **forget** to escape a TXT record properly, but I **expect** a DNS parser to parse according to the standard. Going against the standard and guessing what the user may have meant instead would be very unexpected behavior.

> With that said I doubt many people who don't use quotes will be affected when their canonical representation uses either "... ..." or "..." "..." form.

They will be pretty affected when the "..." "..." form is concatenated without spaces and the contents end up malformed. E.g. https://tools.ietf.org/html/rfc7208#section-3.3:

    3.3.  Multiple Strings in a Single DNS Record

       As defined in [RFC1035], Sections 3.3 and 3.3.14, a single text DNS
       record can be composed of more than one string.  If a published
       record contains multiple character-strings, then the record MUST be
       treated as if those strings are concatenated together without adding
       spaces.  For example:

          IN TXT "v=spf1 .... first" "second string..."

       is equivalent to:

          IN TXT "v=spf1 .... firstsecond string..."

       TXT records containing multiple strings are useful in constructing
[...]

PR #421, opened 2016-11-28
Fix TXT RDATA parsing

gibson042:
Fixes gh-420

This PR updates parsing  of valid TXT RDATA for RFC compliance, which includes recognizing unquoted whitespace as separating \<character-string>s (in the first two commits, 975337868b3168fe5993f6b5ed5b6257914edaf0 and 04a88b794228d0e7bd5366628af462cbf8bde199) and removing special treatment of `\t`/`\n`/`\r` sequences (in the third, 730b0c7ceab6a3351743ba5d2603853255732c21). To keep tests passing—and fixing a previously unreported bug—it also applies the latter change to owner names (in the fourth, 774c47fdee15765ca75a088ae2646c2293a74b4a).


update_test.go
```diff
@@ -10,7 +10,8 @@ func TestDynamicUpdateParsing(t *testing.T) {
 	for _, typ := range TypeToString {
 		if typ == "OPT" || typ == "AXFR" || typ == "IXFR" || typ == "ANY" || typ == "TKEY" ||
 			typ == "TSIG" || typ == "ISDN" || typ == "UNSPEC" || typ == "NULL" || typ == "ATMA" ||
-			typ == "Reserved" || typ == "None" || typ == "NXT" || typ == "MAILB" || typ == "MAILA" {
+			typ == "Reserved" || typ == "None" || typ == "NXT" || typ == "MAILB" || typ == "MAILA" ||
+			typ == "UINFO" {
```
miekg: What happened here?

Rest looks good.

gibson042: Tests were failing, because `setUINFO` uses `endingToTxtSlice` and assumes that the resulting slice is non-empty. I figured this was the most straightforward fix, since [UINFO](http://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#dns-parameters-4) is an "IANA-Reserved" nonstandard type anyway. But if you prefer, I could instead update `setUINFO`.

PR #424, opened 2016-12-02
For consistency with other types, allow empty UINFO RDATA

gibson042:
Ref https://github.com/miekg/dns/pull/421#discussion_r90610949


miekg:
thanks for the cleanup!

PR #427, opened 2016-12-09
Fix tcp6-tls support in (*Server).ListenAndServe().

tmthrgd:
In the switch statement srv.Net is matched for tcp6-tls but then compared against tcp6 within the case statement. This causes tcp6-tls to be equivalent to tcp-tls and not specific to IPv6. The `network = "tcp6"` line was previously unreachable.

This change corrects this and ensures tcp6-tls listens on IPv6 only.


miekg:
Thanks!

Issue #429, opened 2016-12-14
Support generic RRs


Arachnid:
I'm writing a DNS server that gets its data from a database that stores (type, class, ttl, data) tuples; data is the uninterpreted RDATA for the DNS response.

There doesn't seem to be any good way to construct an RR from this; all the methods assume either text format or a known-at-compile-time RR type. RFC3597 provides an option for a kludge, but it isn't great.

It'd be really handy if the library either:

 - Provided an `UnpackRData(header, data)` method
 - Provided a GenericRR that contains a `data []byte`
 - Exported all the methods implemented by `dns.RR` so one can write their own RRs.

miekg:
[ Quoting <notifications@github.com> in "[miekg/dns] Support generic RRs (#4..." ]
>I'm writing a DNS server that gets its data from a database that stores (type, class, ttl, data) tuples; data is the uninterpreted RDATA for the DNS response.

No domain name is stored here?

>There doesn't seem to be any good way to construct an RR from this; all the methods assume either text format or a known-at-compile-time RR type. RFC3597 provides an option for a kludge, but it isn't great.
>
>It'd be really handy if the library either:
>
> - Provided an `UnpackRData(header, data)` method
> - Provided a GenericRR that contains a `data []byte`
> - Exported all the methods implemented by `dns.RR` so one can write their own RRs.

You can actually do private RRs https://miek.nl/2014/September/21/idn-and-private-rr-in-go-dns/
But that might even be farther from what you need.

If you store the data in wireformat and you also create the wireformat for the
header you can just use UnpackRR to get the native go dns RR.

/Miek

--
Miek Gieben

Arachnid:
> No domain name is stored here?

Not in plaintext; I plan to use the name from the query to provide this field.

> You can actually do private RRs https://miek.nl/2014/September/21/idn-and-private-rr-in-go-dns/
But that might even be farther from what you need.

I don't think that's what we need - we need to be able to create arbitrary existing public DNS RRs from data that's provided at runtime.

> If you store the data in wireformat and you also create the wireformat for the
header you can just use UnpackRR to get the native go dns RR.

Since name isn't stored explicitly at all, and TTL is stored in a separate record, we can't do this, unfortunately.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Support generic RRs..." ]
>> If you store the data in wireformat and you also create the wireformat for the
>header you can just use UnpackRR to get the native go dns RR.
>
[...]

PR #431, opened 2016-12-19
Add support for packing and unpacking Rdata

Arachnid:
This adds support for packing and unpacking Rdata separately in #429.

Packing is currently broken, and I'm afraid I'm rather stumped as to why this is the case - if you can look at the modifications I made to msg.go and the code generation in the second commit, I'd be most grateful.


andrewtj:
@Arachnid are you still interested in this?

PR #432, opened 2016-12-21
Update RSA Exponent Limit

everclear:
publicKeyRSA() should allow an exponent of F5.


miekg:
does the go crypto library support this?

everclear:
The exponent is an `int` in the crypto library and this size exponent works just fine on a 64bit OS.
I've altered the check to also ensure that expo fits within the max capacity of the int type on a 32bit OS.

miekg:
Has this issue been fixed: https://github.com/golang/go/issues/3161

miekg:
Stale PR and not responding to questions.

PR #434, opened 2017-01-03
Push test coverage profile to Coveralls

corny:
- Push test coverage to Coveralls
- Add coverage badge to README
- Test Go 1.7 as well
- Don't use miekg/dns when testing forks

To use the coverage badge @miekg has to sign in at https://coveralls.io/ and enable it for this repository. Example of the result: https://coveralls.io/github/corny/dns?branch=improve-tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,5 +3,17 @@ sudo: false
 go:
   - 1.5
   - 1.6
+  - 1.7
+
+before_install:
+  - go get github.com/mattn/goveralls
+
+  # don't use the miekg/dns when testing forks
+  - mkdir -p $GOPATH/src/github.com/miekg
+  - ln -s $TRAVIS_BUILD_DIR $GOPATH/src/github.com/miekg/ || true
+
 script:
-  - go test -race -v -bench=.
+  # run tests with coverage
+  - goveralls -v
+  # run benchmarks
+  - go test -v -bench=. -run=^$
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -3,5 +3,17 @@ sudo: false
 go:
   - 1.5
   - 1.6
+  - 1.7
+
+before_install:
+  - go get github.com/mattn/goveralls
```
miekg: Thanks for the PR, but this pulls in a 3rd party dependency for go dns. I don't think I want that (even for testing). What does this do?

corny:
Currently the build fails because Coveralls is not set up yet:
```
Bad response status from coveralls: 422 - {"message":"Couldn't find a repository matching this job.","error":true}
```

.travis.yml
corny: It's required to push the coverage profile from travis to coveralls. It's the easiest way to track the test coverage.

corny: I updated the description of my pull request.

corny: It does not affect the tests when you run them locally. It's only used by travis.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Improve test suite ..." ]
>corny commented on this pull request.
>
>
>
>> @@ -3,5 +3,17 @@ sudo: false
> go:
>   - 1.5
>   - 1.6
>+  - 1.7
>+
>+before_install:
>+  - go get github.com/mattn/goveralls
>
>It's required to push the coverage profile from travis to coveralls. It's the easiest way to track the test coverage.

I'm sorry but I'll have to NACK this PR. I don't want to introduce a 3rd party
dependency - even when testing Go DNS.

corny:
Ok, I renamed this PR and created another one with my changes except Coveralls: #435 

Do you know another way to track and display the test coverage? Since the coverage is only around 63 % it should be improved a lot.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Push test coverage ..." ]
>Do you know another way to track and display the test coverage? Since the coverage is only around 63 % it should be improved a lot.

Yes, this should be improved. For CoreDNS we use codedov and a somewhat hacky
script.

Of course *just* having a badge won't increase the test coverage...

corny:
Ok, I didn't know codecov before. I would love to see this in `miekg/dns` as well.

PR #442, opened 2017-01-27
Unpack header bits if message lenght equals offset

matje:
`Unpack` should obviously have checks whether the packet has the expected length. But you know that it contains the DNS header if `UnpackMsgHdr` does not return an error. At that point it is safe to unpack the DNS header wire into the `Msg` struct. 

If unpacked, we can test if our name server has the right bits set when encountering bad DNS queries.

### DIFF
--- a/msg.go
+++ b/msg.go
@@ -759,9 +759,6 @@ func (dns *Msg) Unpack(msg []byte) (err error) {
 	if dh, off, err = unpackMsgHdr(msg, off); err != nil {
 		return err
 	}
-	if off == len(msg) {
-		return ErrTruncated
-	}
 
 	dns.Id = dh.Id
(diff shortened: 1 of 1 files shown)


miekg:
Can you update your description to tell why you are making the change? Thanks.

matje:
Done!

miekg:
> 	return errTruncated after setting header bits

why? What fixes this?

matje:
It fixes that you can easily examine what was in the (too) short DNS packet. For example:
1. I send a query with QDCOUNT = 2
2. I get a FORMERR back
3. I want to test that a FORMERR is returned, but cannot because the `msg.Rcode` was not set because `off == len(msg)` and the values are not set in `dns.Msg`. 

Obviously I could check it by looking into the wire data, but that's not what you use a DNS library for ;)

PR #443, opened 2017-02-01
compress more records

gpaul:
The SRV and PX records' fields should be compressed. The individual commit messages quote the relevant sections from RFCs.

### DIFF
--- a/types.go
+++ b/types.go
@@ -535,7 +535,7 @@ type SRV struct {
 	Priority uint16
 	Weight   uint16
 	Port     uint16
-	Target   string `dns:"domain-name"`
+	Target   string `dns:"cdomain-name"`
 }
 
 func (rr *SRV) String() string {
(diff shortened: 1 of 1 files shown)


miekg:
2052 is obsoleted by 2728 which is updated by 6335 https://www.rfc-editor.org/info/rfc6335

The same chain can be made for https://www.rfc-editor.org/info/rfc2163

gpaul:
Thanks for the quick response - I don't find where the statements regarding compression of SRV and PX are overturned in the newer RFCs. The word 'compress' doesn't seem to occur in 6335 at all. 

RFC2163 also says 
"All normal DNS conventions, like
default values, wildcards, abbreviations and message compression,
apply also for all the components of the PX RR. In particular <name>,
MAP822 and MAPX400, as <domain-name> elements, must have the final
 "." (root) when they are fully qualified."

gpaul:
My mistake - I see you meant 2782 instead of 2728. You are correct about SRV not using compression.

gpaul:
I followed the obsoletion chain for rfc2163 and you are correct. Closing this PR :+1:

PR #444, opened 2017-02-01
Improve performance by addressing some low hanging fruit.

tmthrgd:
This pull request introduces some very minor performance improvements and reduces the garbage produced when calling several functions.

The first commit removes an unused bytes.Buffer from dns/idn.encode. The second commit reduces the amount of garbage produced when calling CertificateToDANE, HashName, (*SIG).Sign and TsigGenerate.

No functionality or APIs are changed.

### DIFF
--- a/dane.go
+++ b/dane.go
@@ -6,7 +6,6 @@ import (
 	"crypto/x509"
 	"encoding/hex"
 	"errors"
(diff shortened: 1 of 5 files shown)


miekg:
Nice. Thanks!

PR #447, opened 2017-02-04
Work around golang/go#11833 predictable random issue in Id.

tmthrgd:
In certain circumstances crypto/rand.Reader will return non-random bytes, see golang/go#11833. The most likely case is near boot, and as init is run when the go program is started, it's possible that a non-random seed could be used. While this is very unlikely to ever be an issue, it is a very easy fix and it is preferable to be resilient.

Instead of seeding the global math/rand rng during init, a separate math/rand.Rand is seeded upon the first call to Id. This also avoids polluting the global math/rand rng which might be seeded elsewhere.

If crypto/rand.Reader fails, math/rand.Int63 will be called to provide a seed. This is better than the current fallback to a seed of 1.

This change introduces no noticeable performance overhead as the global math/rand rng already uses a sync.Mutex internally.

The current implementation was introduced in 3f53d75 from #359.

### DIFF
--- a/msg.go
+++ b/msg.go
@@ -80,8 +80,14 @@ func id() uint16 {
 		idRand = rand.New(rand.NewSource(seed))
 	}
 
-	// The call to Uint32 must be within the mutex lock
-	// because *rand.Rand is not safe for concurrent use.
+	// The call to idRand.Uint32 must be within the
+	// mutex lock because *rand.Rand is not safe for
+	// concurrent use.
+	//
+	// There is no added performance overhead to calling
+	// idRand.Uint32 inside a mutex lock over just
+	// calling rand.Uint32 as the global math/rand rng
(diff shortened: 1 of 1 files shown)


msg.go
```diff
@@ -66,11 +53,39 @@ var (
...
+		} else {
+			seed = rand.Int63()
+		}
+
+		idRand = rand.New(rand.NewSource(seed))
+	}
+
+	// The call to Uint32 must be within the mutex lock
+	// because *rand.Rand is not safe for concurrent use.
+	id := uint16(idRand.Uint32())
```
miekg: Can you add 'This change introduces no noticeable performance overhead as the global math/rand rng already uses a sync.Mutex internally.' to this comment as well. Thanks!

tmthrgd: Done! I made the comment slightly clearer.

PR #449, opened 2017-02-08
Implement attempts in DNS Client

jvgutierrez:
This PR implements attempts in DNS client resolution.

If Client.attempts is not set defaults to 1, mimicking the behaviour the library had before adding the attempts implementation

### DIFF
--- a/client.go
+++ b/client.go
@@ -35,6 +35,7 @@ type Client struct {
 	WriteTimeout   time.Duration     // net.Conn.SetWriteTimeout value for connections, defaults to 2 seconds - overridden by Timeout when that value is non-zero
 	TsigSecret     map[string]string // secret(s) for Tsig map[<zonename>]<base64 secret>, zonename must be fully qualified
 	SingleInflight bool              // if true suppress multiple outstanding queries for the same Qname, Qtype and Qclass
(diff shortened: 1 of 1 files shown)


miekg:
I rather leave this out of the library and let user implement their own re-try logic. 
Should retry only be done for UDP? Does it makes sense for TCP? Retry on time out as well? etc. etc.

jvgutierrez:
@miekg IMHO this should be the way to support attempts setting from resolv.conf

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Implement attempts ..." ]
>@miekg IMHO this should be the way to support attempts setting from resolv.conf

hmmm, not so sure. Another question that arises; should a client try another
nameserver while this one is being re-tried? How do you convey that information?
What if you have 1000 diff queries inflight and they all retry. I mean that
retry behavior should be smarter then just an for-loop and might also contain
backoff and jitter.

I think it is best to keep this code super simple and allow (force?) user to
define their own retry stuff.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Implement attempts ..." ]
>Closed #449.

In the future is better to open a bug first to discuss various trade off, before
heading off and implementing it (granted this was a small change)

PR #451, opened 2017-02-13
Refactored udp*.go to remove redundancy, and make them compile on appengine

bramp:
App Engine even though is Linux, does not allow referencing the syscall package. Thus I reorganised
the udp*.go files a little to make this work. As best I can the functionality is unchanged, and all
this continues to compiles on darwin, linux, windows, plan9 and appengine.

Namely:
* Moved all Linux specific code from udp.go into udp_linux.go (e.g setUDPSocketOptions)
* Deleted udp_plan9.go which was almost a copy of udp.go.
* Added build tags to stop appengine using the linux build, and instead using the udp_other.go.


miekg:
lgtm. thanks!

PR #455, opened 2017-02-14
Update GoDoc reference to EDNS0_SUBNET.SourceNetmask field.

aggrolite:
Hi,

I was following the example of setting edns-client-subnet option via GoDoc and noticed the struct field was misnamed:

`e.SourceNetMask undefined (type *dns.EDNS0_SUBNET has no field or method SourceNetMask, but does have SourceNetmask)`


miekg:
yep. Thanks.

PR #458, opened 2017-02-16
Add support for Ed25519 DNSSEC signing from RFC 8080

tmthrgd:
This pull request adds support for the Ed25519 algorithm in DNSSEC from [RFC 8080](https://tools.ietf.org/html/rfc8080). RFC 8080 also specifies the use of Ed448 but there is no standard library implementation of Ed448 so it is not included, this should not be an issue as Ed25519 will always see more use than Ed448.

I'm not sure if this pull request meets the 'Depends only on the standard library.' feature listed in the README. Ed25519 support comes from a Golang sub-repositories, in particular [golang.org/x/crypto/ed25519](https://godoc.org/golang.org/x/crypto/ed25519). The sub-repositories are described with:
> These packages are part of the Go Project but outside the main Go tree. They are developed under looser compatibility requirements than the Go core. Install them with "go get".

Note: the examples included in the RFC are invalid and were corrected in [RFC Errata Report (4935)](https://www.rfc-editor.org/errata_search.php?eid=4935).

### DIFF
--- a/dnssec.go
+++ b/dnssec.go
@@ -310,6 +310,12 @@ func (rr *RRSIG) Sign(k crypto.Signer, rrset []RR) error {
 
 	switch rr.Algorithm {
 	case ED25519:
+		// ed25519 signs the raw message and performs hashing internally.
+		// All other supported signature schemes operate over the pre-hashed
+		// message, and thus ed25519 must be handled separately here.
+		//
+		// The raw message is passed directly into sign and crypto.Hash(0) is
+		// used to signal to the crypto.Signer that the data has not been hashed.
 		signature, err := sign(k, append(signdata, wire...), crypto.Hash(0), rr.Algorithm)
 		if err != nil {
 			return err
--- a/dnssec.go
+++ b/dnssec.go
@@ -19,6 +19,8 @@ import (
 	"sort"
 	"strings"
 	"time"
+
+	"golang.org/x/crypto/ed25519"
 )
 
 // DNSSEC encryption algorithm codes.
@@ -38,6 +40,8 @@ const (
 	ECCGOST
 	ECDSAP256SHA256
 	ECDSAP384SHA384
+	ED25519
+	ED448
 	INDIRECT   uint8 = 252
 	PRIVATEDNS uint8 = 253 // Private (experimental keys)
 	PRIVATEOID uint8 = 254
@@ -56,6 +60,8 @@ var AlgorithmToString = map[uint8]string{
 	ECCGOST:          "ECC-GOST",
 	ECDSAP256SHA256:  "ECDSAP256SHA256",
 	ECDSAP384SHA384:  "ECDSAP384SHA384",
+	ED25519:          "ED25519",
+	ED448:            "ED448",
 	INDIRECT:         "INDIRECT",
 	PRIVATEDNS:       "PRIVATEDNS",
 	PRIVATEOID:       "PRIVATEOID",
@@ -73,6 +79,7 @@ var AlgorithmToHash = map[uint8]crypto.Hash{
 	ECDSAP256SHA256:  crypto.SHA256,
 	ECDSAP384SHA384:  crypto.SHA384,
 	RSASHA512:        crypto.SHA512,
+	ED25519:          crypto.Hash(0),
 }
 
 // DNSSEC hashing algorithm codes.
@@ -301,16 +308,26 @@ func (rr *RRSIG) Sign(k crypto.Signer, rrset []RR) error {
 		return ErrAlg
 	}
 
-	h := hash.New()
-	h.Write(signdata)
-	h.Write(wire)
+	switch rr.Algorithm {
+	case ED25519:
+		signature, err := sign(k, append(signdata, wire...), crypto.Hash(0), rr.Algorithm)
+		if err != nil {
+			return err
+		}
 
-	signature, err := sign(k, h.Sum(nil), hash, rr.Algorithm)
-	if err != nil {
-		return err
+		rr.Signature = toBase64(signature)
+	default:
+		h := hash.New()
+		h.Write(signdata)
+		h.Write(wire)
+
+		signature, err := sign(k, h.Sum(nil), hash, rr.Algorithm)
+		if err != nil {
+			return err
+		}
+
+		rr.Signature = toBase64(signature)
 	}
-
-	rr.Signature = toBase64(signature)
 
 	return nil
 }
@@ -352,6 +369,9 @@ func sign(k crypto.Signer, hashed []byte, hash crypto.Hash, alg uint8) ([]byte, 
 		// 	signature = append(signature, intToBytes(r1, 20)...)
 		// 	signature = append(signature, intToBytes(s1, 20)...)
 		// 	rr.Signature = signature
+
+	case ED25519:
+		return signature, nil
 	}
 
 	return nil, ErrAlg
@@ -452,6 +472,17 @@ func (rr *RRSIG) Verify(k *DNSKEY, rrset []RR) error {
 		h.Write(signeddata)
 		h.Write(wire)
 		if ecdsa.Verify(pubkey, h.Sum(nil), r, s) {
+			return nil
+		}
+		return ErrSig
+
+	case ED25519:
+		pubkey := k.publicKeyED25519()
+		if pubkey == nil {
+			return ErrKey
+		}
+
+		if ed25519.Verify(pubkey, append(signeddata, wire...), sigbuf) {
 			return nil
 		}
 		return ErrSig
@@ -576,6 +607,17 @@ func (k *DNSKEY) publicKeyDSA() *dsa.PublicKey {
 	pubkey.Parameters.G = big.NewInt(0).SetBytes(g)
 	pubkey.Y = big.NewInt(0).SetBytes(y)
 	return pubkey
+}
+
+func (k *DNSKEY) publicKeyED25519() ed25519.PublicKey {
+	keybuf, err := fromBase64([]byte(k.PublicKey))
+	if err != nil {
+		return nil
+	}
+	if len(keybuf) != ed25519.PublicKeySize {
+		return nil
+	}
+	return keybuf
 }
 
 type wireSlice [][]byte
--- a/dnssec_keyscan.go
+++ b/dnssec_keyscan.go
@@ -184,6 +184,15 @@ func readPrivateKeyED25519(m map[string]string) (ed25519.PrivateKey, error) {
 			if len(p1) != 32 {
 				return nil, ErrPrivKey
 			}
+			// RFC 8080 and Golang's x/crypto/ed25519 differ as to how the
+			// private keys are represented. RFC 8080 specifies that private
+			// keys be stored solely as the seed value (p1 above) while the
+			// ed25519 package represents them as the seed value concatenated
+			// to the public key, which is derived from the seed value.
+			//
+			// ed25519.GenerateKey reads exactly 32 bytes from the passed in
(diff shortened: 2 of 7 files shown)


miekg:
Thank you. This is a useful addition. I'm however torn about the golang.org/x dep that it adds; yes we could vendor, use the new `dep`, but it has been something we managed to avoid. I think for that reason alone merging this would be impossible. Is there any chance this code will move into the std lib?

Code looks good btw.

tmthrgd:
I don't have an easy solution for the golang.org/x dep. Some things to note though:
* ed25519 was added to golang.org/x/crypto in https://go-review.googlesource.com/22030,
* golang.org/x/crypto/ed25519 hasn't changed since it was included (see https://go.googlesource.com/crypto/+log/master/ed25519),

tmthrgd:
I don't have an easy solution for the golang.org/x dep. Some things to note though:
* ed25519 was added to golang.org/x/crypto in https://go-review.googlesource.com/22030,
* golang.org/x/crypto/ed25519 hasn't changed since it was included (see https://go.googlesource.com/crypto/+log/master/ed25519),
* the distinct but related package, [golang.org/x/crypto/curve25519](https://godoc.org/golang.org/x/crypto/curve25519), was vendored into the standard library in https://go-review.googlesource.com/30822,

tmthrgd:
Unfortunately I don't have an easy solution for the golang.org/x dep. Some things to note though:
* ed25519 was added to golang.org/x/crypto in https://go-review.googlesource.com/22030 on 25th/26th April 2016,
* golang.org/x/crypto/ed25519 hasn't had a single commit or change since it was included (see https://go.googlesource.com/crypto/+log/master/ed25519),
* the distinct but related package, [golang.org/x/crypto/curve25519](https://godoc.org/golang.org/x/crypto/curve25519), was vendored into the standard library in https://go-review.googlesource.com/30822,
* golang.org/x/crypto/curve25519 will be used internally by crypto/tls from go1.8 (see https://go-review.googlesource.com/30824),
* [golang.org/x/crypto/chacha20poly1305](https://godoc.org/golang.org/x/crypto/chacha20poly1305) has also been vendored into the standard library to be used internally by crypto/tls (see https://go-review.googlesource.com/30957), and
* golang.org/x/crypto/ed25519 is used by [golang.org/x/crypto/ssh](https://godoc.org/golang.org/x/crypto/ssh), which is also a sub-repository, (see https://godoc.org/golang.org/x/crypto/ed25519?importers).

It's quite unfortunate that ed25519 isn't yet part of the standard library. I'm happy for this not to merged and just left open for the indefinite future until either ed25519 becomes crypto/ed25519 or some other solution is arrived at. I've really got no idea when ed25519 might become part of the standard library.

I guess other possible solutions are to simply import golang.org/x/crypto/ed25519 and consider the sub-repositories to be loosely part of the standard library or to vendor it. Neither is ideal of course.

miekg:
I have no problem leaving this PR open until enough stuff sits in the standard lib.

andrewtj:
The example RRSIG records have unbalanced parenthesis. I guess the parsing machinery doesn't explicitly check that it has found the end of an entry?

Thanks for doing this. I'm using it for testing another (non-Go) implementation 👍

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Add support for Ed2..." ]
>The example RRSIG records have unbalanced parenthesis. I guess the parsing machinery doesn't explicitly check that it has found the end of an entry?

Geh, interesting bug. It does know when the RR ends, so by virtue of that we
don't check the parenthesis.

/Miek

--
Miek Gieben

tmthrgd:
@andrewtj You're right, and it is an interesting bug. I didn't even notice it, principally because there was no parsing error. It's a mistake in [RFC 8080](https://tools.ietf.org/html/rfc8080) though, and I missed it when I filed the [errata](https://www.rfc-editor.org/errata_search.php?rfc=8080). I'm not familiar enough with the DNS record format to know where it should go. You're really more than welcome! I was very much hoping that getting an implementation out there would help quicken adoption of Ed25519!

@miekg Could that be handled by simply keeping a counter that is incremented with opening parenthesis and decremented with closing parenthesis, and then checking it's equal to zero at the end of the record?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Add support for Ed2..." ]
>@andrewtj You're right, and it is an interesting bug. I didn't even notice it, principally because there was no parsing error. It's a mistake in [RFC 8080](https://tools.ietf.org/html/rfc8080) though, and I missed it when I filed the [errata](https://www.rfc-editor.org/errata_search.php?rfc=8080). I'm not familiar enough with the DNS record format to know where it should go. You're really more than welcome! I was very much hoping that getting an implementation out there would help quicken adoption of Ed25519!

oh, nice. Should file an errata ! :)

>@miekg Could that be handled by simply keeping a counter that is incremented with opening parenthesis and decremented with closing parenthesis, and then checking it's equal to zero at the end of the record?

(haven't look at the code). Yes, but I believe I'm already doing that.

miekg:
The code check closing braces and errors on that, not superfluous opening braces

miekg:
named-checkzones barf: dns_rdata_fromtext: example.org:8: unbalanced parentheses.
https://github.com/miekg/dns/issues/488

miekg:
The popularity of the algo's and the fact that we don't support them becomes problematic; we should revisit this and probably start vendoring things.

andrewtj:
My only concern with vendoring is the possibility of missing upstream updates. A Travis job should be able to warn of that happening though.

tmthrgd:
@miekg I just resolved the merge conflicts associate with this pull request.

I still think you can get away without vendoring in this case as the [golang.org/x/crypto/ed25519](https://godoc.org/golang.org/x/crypto/ed25519) package has proved [very stable](https://github.com/golang/crypto/commits/master/ed25519). So far there have been no changes to the API in it's history. Even if it must be vendored, I think that's a very approachable problem.

tmthrgd:
The changes have been rebased onto the master branch and the test failure noted in https://github.com/miekg/dns/pull/458#issuecomment-302004827 has been fixed. I'm not sure if this matches the current coding style as I haven't otherwise changed the code since the 16th of February.

miekg:
> My only concern with vendoring is the possibility of missing upstream updates. A Travis job should be able to warn of that happening though.
Some kind of signal we be good, yes.

I'll review this, get this merged; then setup `godep` to vendor this.

cc @fastest963 we might then also start vendoring the x/net/ stuff and pull out our internal "fork" of it

dnssec.go
```diff
@@ -301,16 +308,26 @@ func (rr *RRSIG) Sign(k crypto.Signer, rrset []RR) error {
 		return ErrAlg
 	}
 
-	h := hash.New()
-	h.Write(signdata)
-	h.Write(wire)
+	switch rr.Algorithm {
+	case ED25519:
+		signature, err := sign(k, append(signdata, wire...), crypto.Hash(0), rr.Algorithm)
```
miekg: I haven't read the rfc/draft. Why does this need to be done like this? (some comments maybe?).
What's crypto.Hash(0)? And why can't we do the double `h.Write`?

dnssec_keyscan.go
```diff
@@ -166,6 +171,30 @@ func readPrivateKeyECDSA(m map[string]string) (*ecdsa.PrivateKey, error) {
...
+		switch k {
+		case "privatekey":
+			p1, err := fromBase64([]byte(v))
+			if err != nil {
+				return nil, err
+			}
+			if len(p1) != 32 {
+				return nil, ErrPrivKey
+			}
+			_, p, err = ed25519.GenerateKey(bytes.NewReader(p1))
```
miekg: The interface in the ed25519 pkgs uses GenerateKey to reconstruct a key? There is no e.g. setBytes equiv?

tmthrgd: @miekg Unfortunately the representation of ed25519 private keys in Golang and RFC 8080 differ. RFC 8080 uses only the seed (or the random private half), while Golang uses `seed || public-key` – in ed25519 the public key is derived from the seed. `ed25519.GenerateKey` uses the first 32-bytes read from `rand` (`bytes.NewReader(p1)`) as the seed/private-half and then derives the public half from this. I'm using `ed25519.GenerateKey` in the same fashion as [`ED25519_keypair_from_seed` from BoringSSL](https://github.com/google/boringssl/blob/f21650709a6f76e829ddcc77fe221c9d6a5c12de/include/openssl/curve25519.h#L94-L102).

I'll completely agree that this isn't particularly obvious and I can add documentation to this. I'm not sure exactly what that documentation should be though.

miekg: Something like your comment, maybe somewhat condensed?

dnssec.go
tmthrgd: @miekg That's not directly related to the draft but to how ed25519 operates. Most signature schemes operate on hashed data while ed25519 performs the hashing internally and thus requires the original message. While there is a variant of ed25519 that uses pre-hashing, RFC 8080 does not use it – very few things do. Basically ed25519 needs to be passed the raw message without any hashing so it's special cased here.

The crypto.Hash(0) is basically used as a placeholder to mean 'not hashed', it's checked in [`(ed25519.PrivateKey).Sign`](https://github.com/golang/crypto/blob/bd6f299fb381e4c3393d1c4b1f0b94f5e77650c8/ed25519/ed25519.go#L55-L57).

It might be clearer if I switch this to calling [ed25519.Sign](https://github.com/golang/crypto/blob/bd6f299fb381e4c3393d1c4b1f0b94f5e77650c8/ed25519/ed25519.go#L96) directly.

tmthrgd: @miekg Scratch calling `ed25519.Sign` directly, that would prevent the use of opaque `crypto.Signers`. I'll improve the documentation here instead.

dnssec_keyscan.go
tmthrgd: @miekg I've got something worked out now, but I'm not sure whether to draw the parallel to ED25519_keypair_from_seed.

```
// ...
//
// This is used in the same manner as |ED25519_keypair_from_seed|
// from BoringSSL.
```

Thoughts?

miekg: OK. Not sure about referencing a C impl. in the comments makes sense for a Go library though? Did you push your changes?

tmthrgd: @miekg That was my thinking too. I've pushed them now, I was just waiting to hear what you thought first.

dnssec_keyscan.go
```diff
@@ -166,6 +171,39 @@ func readPrivateKeyECDSA(m map[string]string) (*ecdsa.PrivateKey, error) {
 	return p, nil
 }
 
+func readPrivateKeyED25519(m map[string]string) (ed25519.PrivateKey, error) {
+	var p ed25519.PrivateKey
+	// TODO: validate that the required flags are present
```
miekg: Should we do this TODO here now as well? Or just garbage in garbage out?

tmthrgd: @miekg I think there is a compelling argument to be strict from the get go, but at the same time this is consistent with all the other key types so I'd be inclined to leave it. It could be worth making them all stricter in a separate pull request though.

tmthrgd:
Does anyone know of any other implementations that could be used for interop testing? I would hate for this pull request to land and later find out that it entrenches some mistaken or invalid behaviour.

andrewtj:
It's in the BIND 9.12 RC and I think Knot and PowerDNS have some support too.

miekg:
planning to merge, add the dep and using `dep` today-ish

miekg:
Merged as #591.

Closing this one. Thanks everybody!

PR #459, opened 2017-02-17
Fix data race in error handling.

devnev:
The response message must copied regardless of whether there was an
error or not, otherwise two concurrent queries may modify the response
as they write it out.

Thanks @Bplotka @kdima for their work on debugging this race.


Bplotka:
+1

I think it is worth mentioning that in this case the fixed "error case" is actual a heavily used case to detect if the message is Truncated (in https://github.com/skynetservices/skydns/blob/master/server/stub.go#L106) and without this fix it very often causes a panic in `(*Msg) Len()` method.

Any reason why we don't always Copy the msg?

miekg:
Nice one! Thanks for finding and fixing this

PR #463, opened 2017-02-21
limiting domain names to 255/63 octets/labels

ogud:
title says it all

### DIFF
--- a/msg.go
+++ b/msg.go
@@ -327,6 +327,7 @@ End:
 // UnpackDomainName unpacks a domain name into a string.
 func UnpackDomainName(msg []byte, off int) (string, int, error) {
 	s := make([]byte, 0, 64)
+	labels := 0
 	off1 := 0
 	lenmsg := len(msg)
 	ptr := 0 // number of pointers followed
@@ -369,6 +370,15 @@ Loop:
 					}
 				}
 			}
+			// never exceed the allowed label count lenght (63) 
+			if labels >= 63 { 
+				return "", lenmsg, &Error{err: "name exceeds 63 labels"}
+			}
+			labels += 1
+			// never exceed the allowed doman name length (255 octets) 
+			if len(s) >= 255 {
+				return "", lenmsg, &Error{err: "name exceeded allowed 255 octets"}
+			}
 			s = append(s, '.')
 			off += c
 		case 0xC0:
@@ -388,6 +398,9 @@ Loop:
 			if ptr++; ptr > 10 {
 				return "", lenmsg, &Error{err: "too many compression pointers"}
 			}
+			// pointer should guarantee that it advances and points forwards at least
+			// but the condition on previous three lines guarantees that it's
+			// at least loop-free
 			off = (c^0xC0)<<8 | int(c1)
 		default:
 			// 0x80 and 0x40 are reserved


msg.go
```diff
@@ -369,6 +370,15 @@ Loop:
 					}
 				}
 			}
+			// never exceed the allowed label count lenght (63) 
+			if labels >= 63 { 
+				return "", lenmsg, &Error{err: "name exceeds 63 labels"}
```
gibson042: There's no such limit in DNS; [RFC 1035](https://tools.ietf.org/html/rfc1035#section-2.3.4) requires the wire-format length of _each_ label to be 63 octets or less, which is already guaranteed by the `c & 0xC0` check.

msg.go
```diff
@@ -369,6 +370,15 @@ Loop:
...
+			}
+			labels += 1
+			// never exceed the allowed doman name length (255 octets) 
+			if len(s) >= 255 {
```
gibson042: This is erroneously checking presentation-format length, but the [RFC 1035 limit](https://tools.ietf.org/html/rfc1035#section-2.3.4) applies to _wire_ format. As a result, it rejects the following perfectly valid (if unusual) domain name:
```
\.\(\)\;\.\(\)\;\.\(\)\;.32-octets-including-length--001.32-octets-including-length--002.32-octets-including-length--003.32-octets-including-length--004.32-octets-including-length--005.32-octets-including-length--006.32-octets-including-length--007.13-octet-TLD.
```

Also, note that 255 octets is an acceptable length; it's 256 that is invalid.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] limiting domain nam..." ]
>gibson042 commented on this pull request.

Thanks for double checking this. I'll take a look.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] limiting domain nam..." ]
>gibson042 commented on this pull request.
>
>The checks in this merge are incorrect; it should be reverted.
>
>> @@ -369,6 +370,15 @@ Loop:
> 					}
> 				}
> 			}
>+			// never exceed the allowed label count lenght (63)
>+			if labels >= 63 {
>+				return "", lenmsg, &Error{err: "name exceeds 63 labels"}
>
>There's no such limit in DNS; [RFC 1035](https://tools.ietf.org/html/rfc1035#section-2.3.4) requires the wire-format length of _each_ label to be 63 octets or less, which is already guaranteed by the `c & 0xC0` check.
>
>> @@ -369,6 +370,15 @@ Loop:
> 					}
> 				}
> 			}
>+			// never exceed the allowed label count lenght (63)
>+			if labels >= 63 {
>+				return "", lenmsg, &Error{err: "name exceeds 63 labels"}
>+			}
>+			labels += 1
>+			// never exceed the allowed doman name length (255 octets)
>+			if len(s) >= 255 {
>
>This is erroneously checking presentation-format length, but the [RFC 1035 limit](https://tools.ietf.org/html/rfc1035#section-2.3.4) applies to _wire_ format. As a result, it rejects the following perfectly valid (if unusual) domain name:
>```
>\.\(\)\;\.\(\)\;\.\(\)\;.32-octets-including-length--001.32-octets-including-length--002.32-octets-including-length--003.32-octets-including-length--004.32-octets-including-length--005.32-octets-including-length--006.32-octets-including-length--007.13-octet-TLD.
>```
>
>Also, note that 255 octets is an acceptable length; it's 256 that is invalid.
>

Yes. :-( Care to send revert PR? We still have an issue here though.

miekg:
@ogud any comments before stuff gets yanked out?

ogud:
I'm working on a fix

Give me 24 hours, suffering from jet lag

Ólafur

On March 7, 2017 1:29:08 PM GMT+00:00, Miek Gieben <notifications@github.com> wrote:
>@ogud any comments before stuff gets yanked out?
>
>-- 
>You are receiving this because you were mentioned.
>https://github.com/miekg/dns/pull/463#issuecomment-284720995

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] limiting domain nam..." ]
>I'm working on a fix
>
>Give me 24 hours, suffering from jet lag

Ack. Thanks

PR #469, opened 2017-03-08
Limit length fix to account for escaped bytes 

ogud:
Here is a "fix" for the length check to account for "non-printable" characters in presentation format

### DIFF
--- a/msg.go
+++ b/msg.go
@@ -327,6 +327,7 @@ End:
 // UnpackDomainName unpacks a domain name into a string.
 func UnpackDomainName(msg []byte, off int) (string, int, error) {
 	s := make([]byte, 0, 64)
+	labels := 0
 	off1 := 0
 	lenmsg := len(msg)
 	ptr := 0 // number of pointers followed
@@ -369,6 +370,15 @@ Loop:
 					}
 				}
 			}
+			// never exceed the allowed label count lenght (63) 
+			if labels >= 63 { 
+				return "", lenmsg, &Error{err: "name exceeds 63 labels"}
+			}
+			labels += 1
+			// never exceed the allowed doman name length (255 octets) 
+			if len(s) >= 255 {
+				return "", lenmsg, &Error{err: "name exceeded allowed 255 octets"}
+			}
 			s = append(s, '.')
 			off += c
 		case 0xC0:
@@ -388,6 +398,9 @@ Loop:
 			if ptr++; ptr > 10 {
 				return "", lenmsg, &Error{err: "too many compression pointers"}
 			}
+			// pointer should guarantee that it advances and points forwards at least
+			// but the condition on previous three lines guarantees that it's
+			// at least loop-free
 			off = (c^0xC0)<<8 | int(c1)
 		default:
 			// 0x80 and 0x40 are reserved
--- a/msg.go
+++ b/msg.go
@@ -331,6 +331,7 @@ func UnpackDomainName(msg []byte, off int) (string, int, error) {
 	off1 := 0
 	lenmsg := len(msg)
 	ptr := 0 // number of pointers followed
+	maxLen := 255 // max length of domain name in wire format 
 Loop:
 	for {
 		if off >= lenmsg {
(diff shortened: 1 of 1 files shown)


msg.go
```diff
@@ -327,9 +327,11 @@ End:
 // UnpackDomainName unpacks a domain name into a string.
 func UnpackDomainName(msg []byte, off int) (string, int, error) {
 	s := make([]byte, 0, 64)
+	labels := 0
 	off1 := 0
 	lenmsg := len(msg)
 	ptr := 0 // number of pointers followed
+	maxLen := 255 // max length of domain name in wire format
```
ogud: sounds good MaxDomainNameLen

msg.go
```diff
@@ -364,11 +367,21 @@ Loop:
...
 							s = append(s, r)
 						}
+						maxLen += 3 // unprintable char is 3 extra bytes
 					} else {
 						s = append(s, b)
 					}
 				}
 			}
+			// never exceed the allowed label count lenght (63) 
+			if labels >= 63 {
```
ogud: Fine

msg.go
```diff
@@ -364,11 +367,21 @@ Loop:
...
 						s = append(s, b)
 					}
 				}
 			}
+			// never exceed the allowed label count lenght (63) 
+			if labels >= 63 { 
+				return "", lenmsg, &Error{err: "name exceeds 63 labels"}
+			}
+			labels += 1
+			// never exceed the allowed doman name length (255 octets)
```
gibson042: doman → domain

msg.go
```diff
@@ -364,11 +367,21 @@ Loop:
...
 				}
 			}
+			// never exceed the allowed label count lenght (63) 
+			if labels >= 63 { 
+				return "", lenmsg, &Error{err: "name exceeds 63 labels"}
+			}
+			labels += 1
+			// never exceed the allowed doman name length (255 octets) 
+			if len(s) >= maxLen {
+				return "", lenmsg, &Error{err: "name exceeded allowed 255 octets"}
```
gibson042: You might also consider special treatment for names ending with the "`.local`" TLD, since [RFC 6762](https://tools.ietf.org/html/rfc6762#appendix-C) defines those as link-local and allows them to occupy 256 octets in wire format for mDNS.

msg.go
```diff
@@ -388,6 +401,9 @@ Loop:
 			if ptr++; ptr > 10 {
 				return "", lenmsg, &Error{err: "too many compression pointers"}
 			}
+			// pointer should guarantee that it advances and points forwards at least
+			// but the condition on previous three lines guarantees that it's
+			// at least loop-free
 			off = (c^0xC0)<<8 | int(c1)
```
gibson042: Probably out of scope for this PR, but [RFC 1035](https://tools.ietf.org/html/rfc1035#section-4.1.4) requires compression pointers to reference _prior_ occurrences, so the arbitrary 10-hop limit could be removed by tracking the starting offset of the most recent pointer and adding a second check at the top of the loop like
```go
if off >= ptrOffset {
	return "", lenmsg, &Error{err: "non-prior compression pointer"}
}
```

msg.go
andrewtj: I don't think the additional complexity would be worth it as I don't think 256 octet names exist in the wild.

gibson042: > Compliant Multicast DNS implementations MUST support names up to 255 bytes plus a terminating zero, i.e., 256 bytes total.

Of course, I don't know if this project aims for mDNS compliance.

andrewtj: Pointing to the RFC doesn't help substantiate that 256 octet names exist in the wild. I don't think mDNSResponder supports 256 octet names (I haven't been able to get `DNSServiceRegisterRecord` to register such a record) and I believe that covers Android, iOS and macOS. Even if mDNSResponder did, such long names aren't going to be produced unless a developer drops below the high-level DNS-SD APIs offered on those platforms. So to the best of my knowledge 256 octet names exist only in the spec.

miekg: I agree with @andrewtj, this sounds a bit too corner case-y. Think the current check is fine.

miekg:
Think we should just revert 0b729df06cb29cf7a3ff59b876e9a2ecb7fac2b1 and redo this slightly.

msg.go
andrewtj: BIND and Knot don't seem to accept forward pointers (if I've read their code correctly), maybe  that implementation has died out?

miekg:
Ack. But let's start a bug to discuss fix that.

miekg:
I've reverted the original commit in 25ac7f1

PR #470, opened 2017-03-10
Replace Atoi with ParseUint where appropriate

moogar0880:
I came across an interesting problem where uint field parsing can exhibit some odd behavior when parsing out of bound input. This updates the various functions in `scan_rr` that parse unsigned integers to use `strconv.ParseUint` instead of `strconv.Atoi`, which returns more appropriate errors.

Below is an example highlighting some of the parsing oddities
```go
package main

import (
	"fmt"

	"github.com/miekg/dns"
)

func main() {
	input := []string{
		"example.com. 3600 IN MX -1 mail.example.com.",
		"example.com. 3600 IN MX 65536 mail.example.com.",
	}

	for _, inp := range input {
		rr, _ := dns.NewRR(inp)
		fmt.Println(rr.String())
	}
}
```

Which is accepted and yields the following output, instead of an error:
```
example.com.	3600	IN	MX	65535 mail.example.com.
example.com.	3600	IN	MX	0 mail.example.com.
```

### DIFF
--- a/dnssec_keyscan.go
+++ b/dnssec_keyscan.go
@@ -36,7 +36,7 @@ func (k *DNSKEY) ReadPrivateKey(q io.Reader, file string) (crypto.PrivateKey, er
 		return nil, ErrPrivKey
 	}
 	// TODO(mg): check if the pubkey matches the private key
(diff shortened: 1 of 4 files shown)


miekg:
The things you trip over :) Merged.

PR #471, opened 2017-03-14
Client can have an explicit local address

matje:


client.go
```diff
@@ -446,6 +450,18 @@ func DialWithTLS(network, address string, tlsConfig *tls.Config) (conn *Conn, er
 	return conn, nil
 }
 
+// Client.DialTimeoutEx acts like DialTimeout, but uses the client's settings.
```
miekg: Why the Ex suffix?

miekg:
Might be hard, but can you add a test of some sorts?

client.go
matje: I guess not to mix it with `Client.dialTimeout`

matje:
> Might be hard, but can you add a test of some sorts?

The existing tests use `exchange` and therefor use the new calls. Could you be more specific what kind of tests you are looking for?

client.go
miekg: Can't we just look at the value of LocalAddr and use it if is not nil? Think a new method is quite a big hammer for this feature.

matje: Where do you mean? In `DialTimeout` there is no access to `Client` or we have to add it as a parameter.

matje: Alternatively we could make the new functions unexported, as they are only relevant to this file.

miekg: Why can't use just Dial for your self and wrap that inside a `dns.Conn`?

matje: Not entirely sure what you are suggesting, but we are calling `client.Exchange` to perform a synchronous query. But we want to be in control of the local address. The changes in this PR address that. Not sure how to achieve that with `Dial` without having to copy a lot of code.

insomniacslk:
Hello @matje, any plan to proceed on this PR? This is a very interesting feature :-)

matje:
@insomniacslk I do want this feature to make it into master, but it seems to be stuck on tests. I asked @miekg what kind of tests he was looking for but never got an answer. 

I see your PR #502 adds a test. If that is what we are looking for to resolve this PR then I am happy to add it. Or maybe easier to get your PR merged (it's basically the same code).

insomniacslk:
Either of those work for me - on the implementation, I tried to stay as close as possible to your code. On the test, took a bit more to make it in a way that could make sense, but it worked :-)

Issue #426, opened 2016-12-08
NSEC3 HashLength is set incorrectly


andrewtj:
Looks like padding is required to parse an NSEC3 record and padding is generated when unpacking.  From [RFC 5155](https://tools.ietf.org/html/rfc5155#section-1.3):
```
   Base32:  the "Base 32 Encoding with Extended Hex Alphabet" as
      specified in [RFC4648].  Note that trailing padding characters
      ("=") are not used in the NSEC3 specification.
```

Additionally when parsing `NSEC3`'s `HashLength` is set incorrectly which results in them packing okay but failing to unpack.

Test case:

```go
package main

import (
	"log"

	"github.com/miekg/dns"
)

func main() {
	rr, err := dns.NewRR(". 0 IN NSEC3 0 0 0 0 C0 TYPE0")
	if err != nil {
		log.Fatalf("new rr failed: %s (unexpected)", err)
	}
	m := new(dns.Msg)
	m.Answer = []dns.RR{rr}
	_, err = m.Pack()

	log.Printf("bug 1: parsing NextDomain requires padding")
	if err == nil {
		log.Fatal("expected pack to fail")
	} else {
		log.Printf("pack failed: %s (expected)", err)
	}
	rr, err = dns.NewRR(". 0 IN NSEC3 0 0 0 0 CO====== TYPE0")
	if err != nil {
		panic(err)
	}
	m.Answer = []dns.RR{rr}
	buf, err := m.Pack()
	if err != nil {
		log.Fatalf("pack failed: %s (unexpected)", err)
	}

	log.Printf("bug 2: HashLength isn't set correctly (set to: %d)",
		rr.(*dns.NSEC3).HashLength)
	err = m.Unpack(buf)
	if err == nil {
		log.Fatalf("expected unpack to fail")
	} else {
		log.Printf("unpack failed: %s (expected)", err)
	}

	rr.(*dns.NSEC3).HashLength = 1
	m.Answer = []dns.RR{rr}
	buf, err = m.Pack()
	if err != nil {
		log.Fatalf("pack failed: %s (unexpected)", err)
	}
	err = m.Unpack(buf)
	if err != nil {
		log.Fatalf("unpack failed: %s (unexpected)", err)
	}
	nd := m.Answer[0].(*dns.NSEC3).NextDomain
	switch nd {
	case "CO======":
		log.Printf("bug3: next domain is padded on unpack")
	case "CO":
		log.Fatalf("next domain is unpadded (unexpectedly)")
[...]

PR #474, opened 2017-03-21
Only pack NSEC3/NSEC3PARAM salt when it is not empty

### LINKED ISSUE #473: Salt is not packed for NSEC3 records
#392 was meant to stop packing the salt from NSEC3/NSEC3PARAM records if the salt was empty (i.e. had the `-` value) but instead just completely stopped packing salts. This causes signature verification failure for signatures of NSEC3/NSEC3PARAM records.

rolandshoemaker:
Instead of just not doing it at all.

Fixes #473.

### DIFF
--- a/msg_generate.go
+++ b/msg_generate.go
@@ -139,8 +139,10 @@ return off, err
 			case st.Tag(i) == `dns:"base64"`:
 				o("off, err = packStringBase64(rr.%s, msg, off)\n")
 
-			case strings.HasPrefix(st.Tag(i), `dns:"size-hex:SaltLength`): // Hack to fix empty salt length for NSEC3
-				o("if rr.%s == \"-\" { /* do nothing, empty salt */ }\n")
+			case strings.HasPrefix(st.Tag(i), `dns:"size-hex:SaltLength`):
+				// directly write instead of using o() so we get the error check in the correct place
+				field := st.Field(i).Name()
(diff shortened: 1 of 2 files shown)


andrewtj:
This change won't survive the next run of `go generate` as `zmsg.go` is produced by `msg_generate.go`. I've spiked a fix in #426.

rolandshoemaker:
Realized I was editing a generated file and have pushed the relevant `msg_generate.go` fixes and the generated code.

Just saw your comments on #426, happy to pull this and defer to your changeset if that works better for you guys. Would be very good to get this fixed ASAP though as this is breaking in the wild DNSSEC deployments which is always a super fun thing to debug.

msg_generate.go
```diff
@@ -139,8 +139,10 @@ return off, err
 			case st.Tag(i) == `dns:"base64"`:
 				o("off, err = packStringBase64(rr.%s, msg, off)\n")
 
-			case strings.HasPrefix(st.Tag(i), `dns:"size-hex:SaltLength`): // Hack to fix empty salt length for NSEC3
-				o("if rr.%s == \"-\" { /* do nothing, empty salt */ }\n")
+			case strings.HasPrefix(st.Tag(i), `dns:"size-hex:SaltLength`):
+				// directly write instead of using o() so we get the error check in the correct place
+				field := st.Field(i).Name()
+				fmt.Fprintf(b, "// Only pack salt if value is not \"-\", i.e. empty\nif rr.%s != \"-\" {\noff, err = packStringHex(rr.%s, msg, off)\nif err != nil {\nreturn off, err\n}\n}\n", field, field)
```
miekg: Please use a string literal here to make it more readable.

miekg:
This can go in before #426

PR #478, opened 2017-03-28
Domain name limits

gibson042:
Domain names must not exceed 255 octets in wire format.

Ref gh-463
Ref gh-469


msg.go
```diff
@@ -400,6 +412,12 @@ Loop:
 	if len(s) == 0 {
 		s = []byte(".")
 	}
+	// error if the name is too long, but don't throw it away
+	if len(s) >= maxLen {
```
shane-kerr: Could be `} else if len(s) >= maxlen {`. ;-)

gibson042: Sure. :)

miekg:
lgtm.

PR #479, opened 2017-03-29
NSEC/3 Cover+Match cleanup

### LINKED ISSUE #313: nsecx.go
Stuff in that file is half done and has been for more than 2 years. Either fix or delete.

rolandshoemaker:
Removes NSEC `Cover` + `Match` functions and the denialer interface and cleans up the same NSEC3 methods to fix all the TODOs. This is a breaking change but given that the NSEC methods were already broken (i.e. only ever returned `false`) if anyone was using them seriously their software was already broken and they should stop. Also adds a table test for `Cover` which checks each of the possible outcomes and a few more tests for `Match`.

Fixes #313.

### DIFF
--- a/nsecx.go
+++ b/nsecx.go
@@ -52,15 +52,16 @@ func HashName(label string, ha uint8, iter uint16, salt string) string {
 func (rr *NSEC3) Cover(name string) bool {
 	nameHash := HashName(name, rr.Hash, rr.Iterations, rr.Salt)
 	owner := strings.ToUpper(rr.Hdr.Name)
-	labels := Split(owner)
-	if len(labels) < 2 {
+	parts := strings.SplitN(owner, ".", 2)
+	if len(parts) < 2 {
 		return false
 	}
-	if !strings.HasSuffix(strings.ToUpper(name), owner[labels[1]:]) { // name is outside zone
+	ownerHash := parts[0]
+	ownerZone := parts[1]
+	if !IsSubDomain(ownerZone, strings.ToUpper(name)) { // name is outside owner zone
 		return false
 	}
 
-	ownerHash := strings.TrimRight(owner[labels[0]:labels[1]], ".")
 	nextHash := rr.NextDomain
 	if ownerHash == nextHash { // empty interval
 		return false
@@ -81,15 +82,15 @@ func (rr *NSEC3) Cover(name string) bool {
 func (rr *NSEC3) Match(name string) bool {
 	nameHash := HashName(name, rr.Hash, rr.Iterations, rr.Salt)
 	owner := strings.ToUpper(rr.Hdr.Name)
-	labels := Split(owner)
-	if len(labels) < 2 {
+	parts := strings.SplitN(owner, ".", 2)
+	if len(parts) < 2 {
 		return false
 	}
-	if !strings.HasSuffix(strings.ToUpper(name), owner[labels[1]:]) {
-		// name is outside zone
+	ownerHash := parts[0]
+	ownerZone := parts[1]
+	if !IsSubDomain(ownerZone, strings.ToUpper(name)) { // name is outside owner zone
 		return false
 	}
-	ownerHash := strings.TrimRight(owner[labels[0]:labels[1]], ".")
 	if ownerHash == nameHash {
 		return true
 	}
--- a/nsecx.go
+++ b/nsecx.go
@@ -2,6 +2,7 @@ package dns
 
 import (
 	"crypto/sha1"
+	"fmt"
 	"hash"
 	"strings"
 )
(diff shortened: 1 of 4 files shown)


nsecx.go
```diff
@@ -48,62 +48,49 @@ func HashName(label string, ha uint8, iter uint16, salt string) string {
...
 	}
-	hash := strings.ToUpper(rr.Hdr.Name[labels[0] : labels[1]-1]) // -1 to remove the dot
-	if hash == rr.NextDomain {
-		return false // empty interval
-	}
-	if hash > rr.NextDomain { // last name, points to apex
-		// hname > hash
-		// hname > rr.NextDomain
-		// TODO(miek)
+	if !strings.HasSuffix(strings.ToUpper(name), owner[labels[1]:]) { // name is outside zone
```
miekg: Thanks for doing this cleanup!

My nsec3 is getting rusty :) What are you checking here? Also the ToUpper might be expensive. Can dns.IsSubdomain be used here?

nsecx.go
```diff
@@ -48,62 +48,49 @@ func HashName(label string, ha uint8, iter uint16, salt string) string {
...
 
-// Match implements the Denialer interface.
+// Match returns true if a name matches the NSEC3 record
 func (rr *NSEC3) Match(name string) bool {
-	// FIXME(miek): Check if we are in the same zone
-	hname := HashName(name, rr.Hash, rr.Iterations, rr.Salt)
-	labels := Split(rr.Hdr.Name)
+	nameHash := HashName(name, rr.Hash, rr.Iterations, rr.Salt)
+	owner := strings.ToUpper(rr.Hdr.Name)
+	labels := Split(owner)
```
miekg: CountLabel does not allocate

nsecx.go
```diff
@@ -48,62 +48,49 @@ func HashName(label string, ha uint8, iter uint16, salt string) string {
...
 }
 
-// Match implements the Denialer interface.
+// Match returns true if a name matches the NSEC3 record
 func (rr *NSEC3) Match(name string) bool {
-	// FIXME(miek): Check if we are in the same zone
-	hname := HashName(name, rr.Hash, rr.Iterations, rr.Salt)
-	labels := Split(rr.Hdr.Name)
+	nameHash := HashName(name, rr.Hash, rr.Iterations, rr.Salt)
+	owner := strings.ToUpper(rr.Hdr.Name)
```
miekg: why the upper?

nsecx.go
rolandshoemaker: This checks that the provided name is a subdomain of the parent of the owner name of the record (i.e. is in the owners zone) which avoids weird cases where the hashed name may appear to be covered but isn't actually in the relevant zone. `dns.IsSubdomain` would work perfectly here except it seems to be case insensitive so it will fail to detect that `asd.test.com` is a subdomain of `TEST.com` which could introduce some hard to debug issues.

I'd be happy to make `IsSubdomain` insensitive if that works for you, it should just require two `strings.ToLower` operations, which are pretty cheap in the overall picture.

nsecx.go
rolandshoemaker: The `NextDomain` field and the return from `HashName` are both uppercase so this is required to do proper lexographic checking, we can either upcase only the owner or downcase the other two. Only doing one should be cheapest.

nsecx.go
rolandshoemaker: (or I guess I could just `strings.ToUpper` the name as I have done in the most recent commit... I think it's _relatively_ performant)

nsecx.go
```diff
@@ -48,62 +48,50 @@ func HashName(label string, ha uint8, iter uint16, salt string) string {
...
+// Cover returns true if a name is covered by the NSEC3 record
 func (rr *NSEC3) Cover(name string) bool {
-	// FIXME(miek): check if the zones match
-	// FIXME(miek): check if we're not dealing with parent nsec3
-	hname := HashName(name, rr.Hash, rr.Iterations, rr.Salt)
-	labels := Split(rr.Hdr.Name)
-	if len(labels) < 2 {
+	nameHash := HashName(name, rr.Hash, rr.Iterations, rr.Salt)
+	owner := strings.ToUpper(rr.Hdr.Name)
+	parts := strings.SplitN(owner, ".", 2)
```
miekg: Please use CountLabel, this break with '\.' in names.

nsecx.go
```diff
@@ -48,62 +48,50 @@ func HashName(label string, ha uint8, iter uint16, salt string) string {
...
-// Match implements the Denialer interface.
+// Match returns true if a name matches the NSEC3 record
 func (rr *NSEC3) Match(name string) bool {
-	// FIXME(miek): Check if we are in the same zone
-	hname := HashName(name, rr.Hash, rr.Iterations, rr.Salt)
-	labels := Split(rr.Hdr.Name)
-	if len(labels) < 2 {
+	nameHash := HashName(name, rr.Hash, rr.Iterations, rr.Salt)
+	owner := strings.ToUpper(rr.Hdr.Name)
+	parts := strings.SplitN(owner, ".", 2)
```
miekg: Here, as well. You can't (sadly) use `strings.Split`

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] NSEC/3 Cover+Match ..." ]
>rolandshoemaker commented on this pull request.
>
>>  	if len(labels) < 2 {
> 		return false
> 	}
>-	hash := strings.ToUpper(rr.Hdr.Name[labels[0] : labels[1]-1]) // -1 to remove the dot
>-	if hash == rr.NextDomain {
>-		return false // empty interval
>-	}
>-	if hash > rr.NextDomain { // last name, points to apex
>-		// hname > hash
>-		// hname > rr.NextDomain
>-		// TODO(miek)
>+	if !strings.HasSuffix(strings.ToUpper(name), owner[labels[1]:]) { // name is outside zone
>
>This checks that the provided name is a subdomain of the parent of the owner name of the record (i.e. is in the owners zone) which avoids weird cases where the hashed name may appear to be covered but isn't actually in the relevant zone. `dns.IsSubdomain` would work perfectly here except it seems to be case insensitive so it will fail to detect that `asd.test.com` is a subdomain of `TEST.com` which could introduce some hard to debug issues.
>
>I'd be happy to make `IsSubdomain` insensitive if that works for you, it should just require two `strings.ToLower` operations, which are pretty cheap in the overall picture.

IsSubdomain with lowercasing ourselves sgtm.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] NSEC/3 Cover+Match ..." ]
>rolandshoemaker commented on this pull request.
>>  func (rr *NSEC3) Match(name string) bool {
>-	// FIXME(miek): Check if we are in the same zone
>-	hname := HashName(name, rr.Hash, rr.Iterations, rr.Salt)
>-	labels := Split(rr.Hdr.Name)
>+	nameHash := HashName(name, rr.Hash, rr.Iterations, rr.Salt)
>+	owner := strings.ToUpper(rr.Hdr.Name)
>
>The `NextDomain` field and the return from `HashName` are both uppercase so this is required to do proper lexographic checking, we can either upcase only the owner or downcase the other two. Only doing one should be cheapest.

yep you're right. Waaay too long that I took a proper look at NSEC3

nsecx.go
rolandshoemaker: Would a name containing `\.` be valid in any record? It is outside of the common LDH label rules.

Also how would I use `CountLabel` to split the hash from the zone name, from my reading of the docs it seems to just return how many labels the name contains, not the label indices?

gibson042: [RFC 7719](https://tools.ietf.org/html/rfc7719#section-2): "Any label in a domain name can contain any octet value." The LDH "preferred name syntax" restricts host names, but not every domain name is a host name (a trivial example being `_sip._tcp.example.com`).

As for your question, `Split` provides the  label indices, using `NextLabel` to find each in turn.

rolandshoemaker: Fair enough, DNS is... weird.

rolandshoemaker:
Ping

miekg:
Yeah. Sorry I was traveling a lot. Saw some commits flying by. I'll take a
closer look tomorrow but should be good to go.

On 8 Apr 2017 9:03 pm, "Roland Bracewell Shoemaker" <
notifications@github.com> wrote:

Ping

—
You are receiving this because you commented.

<https://github.com/miekg/dns/pull/479#issuecomment-292742198>, or mute the
thread

PR #489, opened 2017-05-19
parsing: error on unbalanced braces

### LINKED ISSUE #488: unbalanced braces not flagged as syntax error
~~~
	IN      RRSIG MX 15 2 3600 (
              1440021600 1438207200 3613 example.com. (
              oL9krJun7xfBOIWcGHi7mag5/hdZrKWw15jPGrHpjQeRAvTdszaPD+QLs3f
              x8A4M3e23mRZ9VrbpMngwcrqNAg== )
~~~
Is not seen as an error.

named-checkzone think it is an error:
~~~
dns_rdata_fromtext: example.org:8: unbalanced parentheses
zone example.org/IN: loading from master file example.org fail
~~~

miekg:
When done parsing, check if we have balanced braces, if not error out.

Fixes #488

### DIFF
--- a/parse_test.go
+++ b/parse_test.go
@@ -1527,3 +1527,14 @@ func TestParseAVC(t *testing.T) {
 		}
 	}
 }
+
+func TestUnbalancedParens(t *testing.T) {
+	sig := `example.com. 3600 IN RRSIG MX 15 2 3600 (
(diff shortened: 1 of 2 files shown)


miekg:
cc @tmthrgd @andrewtj

PR #496, opened 2017-06-04
xfr: return sane error when !RcodeSuccess

### LINKED ISSUE #467: Misleading error for failed/refused XFR
It seems that the library does not evaluate the return code of the messages received during XFR transfers. When the server returns a REFUSED or SERVAIL code for an XFR query, instead of passing through the error code, an ominous "No SOA" error is returned.

dig

```
; <<>> DiG 9.8.3-P1 <<>> @yadifa.sidnlabs.nl types.wb.sidnlabs.nl axfr +all
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: SERVFAIL, id: 5274
;; flags: qr; QUERY: 1, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 0

;; QUESTION SECTION:
;types.wb.sidnlabs.nl.		IN	AXFR

; Transfer failed.
```

miekg/dns

> dns: no SOA

miekg:
When the server returns a non succesful rcode, return that to the caller
in stead of the "bad soa" of before. "dns: bad xfr rcode: <RCODE>" is
now returned.

Fixes #467

### DIFF
--- a/xfr.go
+++ b/xfr.go
@@ -1,6 +1,7 @@
 package dns
 
 import (
+	"fmt"
 	"time"
 )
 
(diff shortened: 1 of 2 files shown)


SuperQ:
\o/

PR #498, opened 2017-07-06
Fix ignored err variables.

alrs:
I found some places where err variables were being ignored.


client.go
```diff
@@ -316,6 +316,9 @@ func (co *Conn) ReadMsgHeader(hdr *Header) ([]byte, error) {
 		}
 		p = make([]byte, l)
 		n, err = tcpRead(r, p)
+		if err != nil {
```
miekg: err is not ignored here, it is checked below on line 333

alrs:
I've made the requested change, @miekg.

PR #499, opened 2017-07-20
document RCodes from the IANA registry

pmoroney:
I pulled these from https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml
I kept having to look up which one was NXDomain ;)
This is quite verbose so I can remove some of the columns or change the formatting if you want.

### DIFF
--- a/types.go
+++ b/types.go
@@ -116,26 +116,26 @@ const (
 	ClassANY    = 255
 
 	// Message Response Codes.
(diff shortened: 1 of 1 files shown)


miekg:
lgtm, but can you put the URL after "// Message Response Codes, see ... " ?

pmoroney:
Added :)

PR #500, opened 2017-08-01
Only check last 2 records for EDNS

miekg:
Last records of additional section hold the OPT record. We need to check
2 because TSIG can be there as well.

### DIFF
--- a/defaults.go
+++ b/defaults.go
@@ -146,12 +146,15 @@ func (dns *Msg) IsTsig() *TSIG {
 // found or nil.
 func (dns *Msg) IsEdns0() *OPT {
 	// EDNS0 is at the end of the additional section, start there.
-	// We might want to change this to *only* look at the last two
-	// records. So we see TSIG and/or OPT - this a slightly bigger
-	// change though.
+	j := 0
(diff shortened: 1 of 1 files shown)


edmonds:
If I understand correctly, @raybellis's XPF proposal (https://tools.ietf.org/html/draft-bellis-dnsop-xpf-02) would append one or more meta RR's to the additional section.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Only check last 2 r..." ]
>If I understand correctly, @raybellis's XPF proposal (https://tools.ietf.org/html/draft-bellis-dnsop-xpf-02) would append one or more meta RR's to the additional section.

*sigh* thanks for the heads up.

andrewtj:
[RFC 6891](https://tools.ietf.org/html/rfc6891#section-6) says "The OPT RR MAY be placed anywhere within the additional data section". I could see an implemention placing an OPT RR first to make packing the most RR Sets in the additional section simpler (though I'm not aware of one).

miekg:
Ok. I'll close this.

Issue #501, opened 2017-08-07
Add ability to specify local address and port


insomniacslk:
For certain applications it is necessary to be able to specify a source port to send the packets from. For example, to force traversing a specific ECMP path.

Go's `net` package offers this functionality via `DialTCP`[1] and `DialUDP`[2], but not through `Dial` or `DialTimeout`. I've experimented by modifying just `dns.Client.DialTimeout` with positive results, see https://github.com/insomniacslk/dns/commit/5b4e81257e0f5bc060ad078baf312d55d3efd24e and a PoC implementation using it, https://github.com/insomniacslk/nsidenumerator/blob/master/experimental/nsidenumerator.go .

However this is far from being complete or clean, it breaks the current API, and in general it's just a proof of concept. Additionally, `DialTCP` and `DialUDP` do not resolve hostnames, while `Dial` and `DialTimeout` do, so this should be also addressed. 

Would this request fit `dns`'s roadmap?

[1] https://golang.org/pkg/net/#DialTCP
[2] https://golang.org/pkg/net/#DialUDP

miekg:
https://github.com/miekg/dns/pull/471 and assorted discussion.

insomniacslk:
Another option is wrapping around a `net.Dialer` with a custom `LocalAddr` (and timeout where needeD) and call its `Dial` method

insomniacslk:
the thread on PR #471  seems to be stuck, especially on the interface to export. If that  is not going to move forward I can make a new PR + tests

miekg:
Do we need code in dns to make this work, is the first question I would ask? If the answer is yes we can bikeshed on the API. Following the std lib in these discussion is a good idea; I like that API :-)

insomniacslk:
I like that API too -  adding a `Client`-specific method that honours the instance variables is the cleanest in my opinion. I just noticed that the PR is stuck on naming and testing, and offered to help to get this done, unless @matje or you do that.

On the "do we need this in `dns`", my opinion is "yes, despite being borderline" (it touches UDP, not DNS), but of course the same goal can be achieved differently.

But I would answer differently to a similar question. My next step after this is to control the underlying IP layer (in particular I need to set the TTL). Should this be implemented in `dns`? I don't think so, because it is far beyond the scope of this library.

miekg:
[...]

PR #502, opened 2017-08-07
Implemented custom dialling with local address and timeout

insomniacslk:
Implemented custom dialling with local address and timeout as Client attributes.

This is a slightly different implementation of https://github.com/miekg/dns/pull/471 , and enables the user to set a custom local address. E.g.:

```
msg := &dns.Msg{/* something here */}
c := new(dns.Client)
c.LocalAddr = &net.UDPAddr{IP: net.ParseIP("127.0.0.1"), Port: 12345, Zone: ""}
resp, ttl, err := client.Exchange(msg, "server.example.org:53")
```

See also the related tests for usage.

### DIFF
--- a/client.go
+++ b/client.go
@@ -37,6 +37,7 @@ type Client struct {
 	TsigSecret     map[string]string // secret(s) for Tsig map[<zonename>]<base64 secret>, zonename must be fully qualified
 	SingleInflight bool              // if true suppress multiple outstanding queries for the same Qname, Qtype and Qclass
 	group          singleflight
+	LocalAddr      net.Addr // local address to use for outgoing connections. If nil, a local address is automatically chosen
 }
 
 // Exchange performs a synchronous UDP query. It sends the message m to the address
@@ -131,6 +132,69 @@ func ExchangeConn(c net.Conn, m *Msg) (r *Msg, err error) {
 	return r, err
 }
 
+// Dial connects to the address on the named network.
+func Dial(network, address string) (conn *Conn, err error) {
+	conn = new(Conn)
+	conn.Conn, err = net.Dial(network, address)
+	if err != nil {
+		return nil, err
+	}
+	return conn, nil
+}
+
+// DialTimeout acts like Dial but takes a timeout.
+func DialTimeout(network, address string, timeout time.Duration) (conn *Conn, err error) {
+	conn = new(Conn)
+	conn.Conn, err = net.DialTimeout(network, address, timeout)
+	if err != nil {
+		return nil, err
+	}
+	return conn, nil
+}
+
+// DialWithTLS connects to the address on the named network with TLS.
+func DialWithTLS(network, address string, tlsConfig *tls.Config) (conn *Conn, err error) {
+	conn = new(Conn)
+	conn.Conn, err = tls.Dial(network, address, tlsConfig)
+	if err != nil {
+		return nil, err
+	}
+	return conn, nil
+}
+
+// DialTimeoutWithTLS acts like DialWithTLS but takes a timeout.
+func DialTimeoutWithTLS(network, address string, tlsConfig *tls.Config, timeout time.Duration) (conn *Conn, err error) {
+	var dialer net.Dialer
+	dialer.Timeout = timeout
+
+	conn = new(Conn)
+	conn.Conn, err = tls.DialWithDialer(&dialer, network, address, tlsConfig)
+	if err != nil {
+		return nil, err
+	}
+	return conn, nil
+}
+
+// deadlineOrTimeout chooses between the provided deadline and timeout
+// by always preferring the deadline so long as it's non-zero (regardless
+// of which is bigger), and returns the equivalent deadline value.
+func deadlineOrTimeout(deadline time.Time, timeout time.Duration) time.Time {
+	if deadline.IsZero() {
+		return time.Now().Add(timeout)
+	}
+	return deadline
+}
+
+// deadlineOrTimeoutOrCtx returns the earliest of: a context deadline, or the
+// output of deadlineOrtimeout.
+func deadlineOrTimeoutOrCtx(ctx context.Context, deadline time.Time, timeout time.Duration) time.Time {
+	result := deadlineOrTimeout(deadline, timeout)
+	if ctxDeadline, ok := ctx.Deadline(); ok && ctxDeadline.Before(result) {
+		result = ctxDeadline
+	}
+	return result
+}
+
 // Exchange performs a synchronous query. It sends the message m to the address
 // contained in a and waits for a reply. Basic use pattern with a *dns.Client:
 //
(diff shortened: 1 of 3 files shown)


insomniacslk:
See https://github.com/miekg/dns/issues/501 and https://github.com/miekg/dns/pull/471 for context

client.go
```diff
@@ -178,6 +179,28 @@ func (c *Client) ExchangeContext(ctx context.Context, m *Msg, a string) (
...
+func (c *Client) DialTimeoutWithTLS(network, address string) (conn *Conn, err error) {
+	dialer := net.Dialer{Timeout: c.dialTimeout(), LocalAddr: c.LocalAddr}
+
+	conn = new(Conn)
+	if conn.Conn, err = tls.DialWithDialer(&dialer, network, address, c.TLSConfig); err != nil {
+		return nil, err
+	}
+	return conn, nil
+}
+
```
matje: It probably makes sense to move these methods below the functions they are referring to, so under `DialTimeout` and `DialTimeoutWithTLS`.

insomniacslk: I generally agree with you, but looking at this file's layout, I see, in order:
1) type definitions
2) various `Exchange*` functions
3) all the struct methods of `Client`
4) the various `Dial*` functions (non-struct methods)

If I move those two new `Client` methods below the referenced functions (i.e. after `4`) we would fragment the `Client` methods (`3`) along the file, making it less natural to follow the file content.

What about instead moving `Dial` and `DialTimeout` above, i.e. before any `Client` method is defined? CC @miekg as this would touch other parts of the file

matje: I see your argument now for putting the methods here. Either way is fine with me.

insomniacslk: I've move those two functions earlier in the file, let's see what @miekg thinks of this :-)

insomniacslk:
@miekg , friendly ping - can you please review this?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Implemented custom ..." ]
>@miekg , friendly ping - can you please review this?

maaaaaybe tonight - busy with work on coredns atm.

client.go
```diff
@@ -178,6 +242,28 @@ func (c *Client) ExchangeContext(ctx context.Context, m *Msg, a string) (
...
+// Client.DialTimeoutWithTLS acts like DialTimeoutWithTLS, but uses the client's configuration
+func (c *Client) DialTimeoutWithTLS(network, address string) (conn *Conn, err error) {
+	dialer := net.Dialer{Timeout: c.dialTimeout(), LocalAddr: c.LocalAddr}
+
+	conn = new(Conn)
+	if conn.Conn, err = tls.DialWithDialer(&dialer, network, address, c.TLSConfig); err != nil {
```
miekg: Is the `WithDialer` a thing we've fully explored?

With every new transport (granted they are not added daily), we get a new  `DialWithTLS` and `DialWithTimeoutXXX`... would it make sense to use a net.Dailer and plumb that through as "the new way"?

Code-wise this all looks fine.

insomniacslk: If I understand this correctly, this is similar to what I suggested at https://github.com/miekg/dns/issues/501#issuecomment-320787105 (point 2). I am in favour of a new API that encompasses a `net.Dialer`, while keeping the old API alive but deprecated (until it gets removed).

For this specific change, I'd keep it scoped, so the refactoring could be done just after.

insomniacslk: super-friendly ping :)

Once (if) this is merged I will attempt some rework of the API and submit for review

miekg:
circling back here. Today I finally have time to properly review and give feedback. (well later today :) )

miekg:
Merging this will make it part of the API and then we need to deprecate it somehow again...

As mentioned. I think we should drop this and go for a `WithDialer` option immediately.

miekg:
If the WithDialer code looks as clean as this one I have no problem merging that.

insomniacslk:
Sorry for the delay. I had some time today to think about an API redesign. What I did so far:
- created a generic method `Client.dialWithDialer` that encompasses timeout, local address, network and whether to use TLS or not (return values are unchanged);
- modified `Client.exchange` (the unexported one) to take as input a context, a Msg and a resolver address (return values are unchanged);
- all the rest of the API is unchanged, but some functions/methods are deprecated and moved to a `client_compat.go` file to make it easy to distinguish the old and new API.

I will also add a crapload of tests to cover as many cases as possible.

I have a working prototype that does what said above (except the test coverage). What do you think of this plan? To be clear, the existing API is unchanged and in some parts deprecated. Such existing API is reimplemented on top of the new API.

One thing of note: Exchange and ExchangeContext behave differently despite what documented (the latter honours `SingleInflight`, the former and all the rest of the code don't). Shall I backport this feature to Exchange (and similarly named functions/methods) or not?

insomniacslk:
that actually makes sense, as `Exchange` is not attached to any client, but `SingleInflight` is client-specific. I'll keep the current behaviour

miekg:
Yes this sounds good!

insomniacslk:
Note to self: also update `doc.go` :)

insomniacslk:
Sorry, holidays and a small mountain accident kept me away from this.

@miekg I have one doubt on the design where I'd like your input.

Background: there are several ways to set a timeout, and using a `net.Dialer` adds one more. At the moment there are:
* dial timeout
* read timeout
* write timeout
* client timeout (overrides all the previous ones)
* context timeout (overrides client timeout)

With the dialer, we are adding a new timeout (`net.Dialer.Timeout`) which, just like the context timeout, should override any other timeout of the ones above (except context timeout, but just because context and dialer are never used together).

I am also for deprecating `ExchangeContext` in favour of `Client.ExchangeWithDialer` and `Client.DialWithDialer` (and the existing methods are just wrappers around it), since it is only used for timeouts, and I agree with Dave Cheney that context should not be used for cancellation either, https://dave.cheney.net/2017/08/20/context-isnt-for-cancellation .

My proposed solution on this specific topic is to:
* deprecate (but not remove) `ExchangeContext` and other `*Context` methods, and wrap them around the `*WithDialer`. Context is only used for timeout and I don't like the idea of carrying around both a Context and a Dialer
* context timeout becomes dialer timeout (i.e. the wrapper takes the context timeout but then uses a dialer under the hood)
* dialer timeout overrides all the other timeouts, if present, just like the context timeout does now (this helps maintain the current behaviour in the existing API)

Does this plan make sense or do you have a preference to keep `*Context` methods around?

insomniacslk:
And another question for @miekg : what do you think of using GoMock (https://github.com/golang/mock) for better testing? This would allow us to mock tls.DialWithDialer, net.Dialer.Dial and other library functions so I can do proper unit testing in addition to integration tests

miekg:
(also on holiday)

I don't like those mocking frameworks. Also adds a non stdlib dependency.
Of course not hanging those, actually testing those will require more code.

Re: your other question: assume net.Dialer timeouts take precedence over
all others.

On Aug 29, 2017 20:49, "insomniac" <notifications@github.com> wrote:

And another question for @miekg <https://github.com/miekg> : what do you
think of using GoMock (https://github.com/golang/mock) for better testing?
This would allow us to mock tls.DialWithDialer, net.Dialer.Dial and other
library functions so I can do proper unit testing in addition to
integration tests

—
You are receiving this because you were mentioned.

<https://github.com/miekg/dns/pull/502#issuecomment-325759403>, or mute the
thread

insomniacslk:
I have created https://github.com/miekg/dns/pull/511 since it's a larger change than the one here (that I will close)

miekg:
closing favor of #511

PR #503, opened 2017-08-08
variable shadowing of token

timesselens:
when zone files parsed with ParseZone(...) contain $INCLUDE statements the file to be included is never parsed because l := <- c shadows the next occurrence of l.token to open the file.

added testcase as proof, the testcase was created from working backwards from the error. Originally the reported l.token used in os.Open() was "\n".

proposed fix: use switch with a ; to localize the variable


miekg:
Nice! thank you!

PR #504, opened 2017-08-10
Fix TSIG bug releated to ID substitution

matje:
TSIG accounts for ID substitution. This means if the ID in the DNS
message is changed by for example a forwarder, TSIG calculation should
use the original message ID (from the TSIG RR).

I have a test for this as well, but it seems tsig_test.go has been
removed, so not sure where to put it now.


miekg:
Nice one. Just create tsig_test.go (can' remember why it was removed)

matje:
I looked at the wrong place (an old version). Added the test to `tsig_test.go`

PR #507, opened 2017-08-16
make IXFR standard compliant

kahnert:
make IXFR work with QIP

### DIFF
--- a/xfr.go
+++ b/xfr.go
@@ -51,18 +51,18 @@ func (t *Transfer) In(q *Msg, a string) (env chan *Envelope, err error) {
 	env = make(chan *Envelope)
 	go func() {
 		if q.Question[0].Qtype == TypeAXFR {
-			go t.inAxfr(q.Id, env)
+			go t.inAxfr(q, env)
 			return
 		}
 		if q.Question[0].Qtype == TypeIXFR {
-			go t.inIxfr(q.Id, env)
+			go t.inIxfr(q, env)
 			return
 		}
 	}()
 	return env, nil
 }
 
-func (t *Transfer) inAxfr(id uint16, c chan *Envelope) {
+func (t *Transfer) inAxfr(q *Msg, c chan *Envelope) {
 	first := true
 	defer t.Close()
 	defer close(c)
@@ -77,7 +77,7 @@ func (t *Transfer) inAxfr(id uint16, c chan *Envelope) {
 			c <- &Envelope{nil, err}
 			return
 		}
-		if id != in.Id {
+		if q.Id != in.Id {
 			c <- &Envelope{in.Answer, ErrId}
 			return
 		}
@@ -110,9 +110,11 @@ func (t *Transfer) inAxfr(id uint16, c chan *Envelope) {
 	}
 }
 
-func (t *Transfer) inIxfr(id uint16, c chan *Envelope) {
+func (t *Transfer) inIxfr(q *Msg, c chan *Envelope) {
 	serial := uint32(0) // The first serial seen is the current server serial
-	first := true
+	axfr := true
+	n := 0
+	qser := q.Ns[0].(*SOA).Serial
 	defer t.Close()
 	defer close(c)
 	timeout := dnsTimeout
@@ -126,21 +128,15 @@ func (t *Transfer) inIxfr(id uint16, c chan *Envelope) {
 			c <- &Envelope{nil, err}
 			return
 		}
-		if id != in.Id {
+		if q.Id != in.Id {
 			c <- &Envelope{in.Answer, ErrId}
 			return
 		}
-		if first {
-			if in.Rcode != RcodeSuccess {
-				c <- &Envelope{in.Answer, &Error{err: fmt.Sprintf(errXFR, in.Rcode)}}
-				return
-			}
-			// A single SOA RR signals "no changes"
-			if len(in.Answer) == 1 && isSOAFirst(in) {
-				c <- &Envelope{in.Answer, nil}
-				return
-			}
-
+		if in.Rcode != RcodeSuccess {
+			c <- &Envelope{in.Answer, &Error{err: fmt.Sprintf(errXFR, in.Rcode)}}
+			return
+		}
+		if n == 0 {
 			// Check if the returned answer is ok
 			if !isSOAFirst(in) {
 				c <- &Envelope{in.Answer, ErrSoa}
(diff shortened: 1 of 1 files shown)


andrewtj:
Can you explain what the problem is and why your patch is a good solution? (It's not obvious to me from what you've provided; may be because I'm unfamiliar with IXFR and this part of the code base.)

kahnert:
I'd setup a bind and tested my patch against it. It run into a timeout, so my patch will break bind, don't use it!

But now I see the different behavior of QIP. QIP always answers with many small messages of "len(in.Answer) == 1". And the current code can't handle this (see line 139, it always matches and returns right after the first line).

I'll provide a working patch for both (bind and QIP). Than you can review it.

kahnert:
On Thu, Aug 17, 2017 at 07:08:00AM +0000, Miek Gieben wrote:
> From the patch, this is strange `rs := q.Ns[0].(*SOA).Serial`. The SOA is put in the auth section for an ixfr?

I need the requested serial number of the query to determine, if the
single line answer is all, because it's also the current serial number.
I renamed this variable to qser (query serial).

And "last" is necessary to determine, if the last SOA block started.
If every answer message only contains a single line, you can't end as
soon as you see the current serial number.

This patch is working for me with QIP and also with bind. Not sure about
other nameserver.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] make IXFR work with..." ]
>I'd setup a bind and tested my patch against it. It run into a timeout, so my patch will break bind, don't use it!
>
>But now I see the different behavior of QIP. QIP always answers with many small messages of "len(in.Answer) == 1". And the current code can't handle this (see line 139, it always matches and returns right after the first line).
>
>I'll provide a working patch for both (bind and QIP). Than you can review it.

Really? `len(in.Answer) == 1` for ixfr. This happens for axfr as well? Are there 
any config options you can tweak, because this sounds really inefficient...

(still there is a bug in the lib here as well). If the 1 msg per response is 
true we could create a testcase.

kahnert:
On Thu, Aug 17, 2017 at 05:37:18PM +0000, Miek Gieben wrote:
> Really? `len(in.Answer) == 1` for ixfr.

Sure, it is.


> This happens for axfr as well?

Well, yes, but doesn't matter because that's covered by the current
code.


> Are there any config options you can tweak, because this sounds really
> inefficient...

I neither programmed QIP nor am I a DNS admin; I simply want the data
and my patch allows me to get it, which the current lib code doesn't.

And I'm not sure if it's inefficient as long as the server is optimized
for small packages (the general use case for a DNS / DHCP server). I've
done no performance tests and won't do it either.

xfr.go
```diff
@@ -55,7 +55,7 @@ func (t *Transfer) In(q *Msg, a string) (env chan *Envelope, err error) {
 			return
 		}
 		if q.Question[0].Qtype == TypeIXFR {
-			go t.inIxfr(q.Id, env)
```
miekg: please change `inAxfr` as well to take a *dns.Msg for symmetry

xfr.go
```diff
@@ -110,9 +110,12 @@ func (t *Transfer) inAxfr(id uint16, c chan *Envelope) {
...
 }
 
-func (t *Transfer) inIxfr(id uint16, c chan *Envelope) {
+func (t *Transfer) inIxfr(q *Msg, c chan *Envelope) {
+	id := q.Id
 	serial := uint32(0) // The first serial seen is the current server serial
-	first := true
+	axfr := true
+	n := 0
+	qser := q.Ns[0].(*SOA).Serial
```
miekg: That we have a Ns section at all, let alone a SOA record is not guaranteed; you need to check for this.

kahnert:
> We also need a test for this.

I won't be able to setup a test domain for this, I'm not a DNS admin here. If you provide a mocking framework for IXFR requests, this test should be a smooth one.

> please change `inAxfr` as well to take a *dns.Msg for symmetry

It's no public API, it's inside an anonymous function. But it's your lib, if it's makes you happy, consider it as done. ;)

> That we have a Ns section at all, let alone a SOA record is not guaranteed; you need to check for this.

You have to shoot yourself in the foot, to make `qser := q.Ns[0].(*SOA).Serial` fail.
Something like this:
```
m := new(dns.Msg)
m.SetIxfr("example.com.", 7, ".", ".")
m.Ns[0] = new(dns.NS)
```
But why would you do so? Even so, if you mess with data sets, a panic is an adequate response. And do you want me to cover every messing around with the Msg struct, like this:
```
m := new(dns.Msg)
m.SetIxfr("example.com.", 7, ".", ".")
m.Question[0].Qtype = dns.TypeNS
```
Which will result into an open channel without receiving data nor a close event? What kind of error you would like me to return? Maybe `ErrMess error = &Error{err: "Stop messing around!"}`? /teasing

Seriously, do you really think this can fail under any reasonable scenario? Than let me know which error you like to see in this case.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] make IXFR work with..." ]
>> We also need a test for this.
>
>I won't be able to setup a test domain for this, I'm not a DNS admin here. If you provide a mocking framework for IXFR requests, this test should be a smooth one.
>
>> please change `inAxfr` as well to take a *dns.Msg for symmetry
>
>It's no public API, it's inside an anonymous function. But it's your lib, if it's makes you happy, consider it as done. ;)
>
>> That we have a Ns section at all, let alone a SOA record is not guaranteed; you need to check for this.
>
>You have to shoot yourself in the foot, to make `qser := q.Ns[0].(*SOA).Serial` fail.
>Something like this:
>```
>m := new(dns.Msg)
>m.SetIxfr("example.com.", 7, ".", ".")
>m.Ns[0] = new(dns.NS)
>```
>But why would you do so? Even so, if you mess with data sets, a panic is an adequate response. And do you want me to cover every messing around with the Msg struct, like this:
>```
>m := new(dns.Msg)
>m.SetIxfr("example.com.", 7, ".", ".")
>m.Question[0].Qtype = dns.TypeNS
>```
>Which will result into an open channel without receiving data nor a close event? What kind of error you would like me to return? Maybe `ErrMess error = &Error{err: "Stop messing around!"}`? /teasing
>
>Seriously, do you really think this can fail under any reasonable scenario? Than let me know which error you like to see in this case.

Ah wait, this is what *you* put in the API, *not* what we get from network. Yes; 
disregard my comment then.

miekg:
this is now actually: "make IXFR RFC compliant"?

miekg:
can you squash your commits and update the final  commit message?

Issue #509, opened 2017-08-21
SetIxfr with empty ns and mbox -> Malformed Packet


kahnert:
With "dig" the ns and mbox field is empty (inside the packet a byte with value 0 for both strings).
But SetIxfr will create malformed packets which causes "dns: bad xfr rcode: 1".

BTW: Parsing the error string to get the code number is kinda cumbersome. Wouldn't it be better to have an own error type which holds the rcode number? Something like `type XfrError struct{ Rcode int }`?

miekg:
Yes `SetIxfr` will happily do what it is told to do.

Agree with the typed errors. That would have been a better idea, sadly that horse has left the barn.

kahnert:
> Yes `SetIxfr` will happily do what it is told to do.

And how do you tell `SetIxfr` to create a correct packet with empty ns
and mbox? `SetIxfr("example.com.", 7, "", "")` won't do the trick.


> Agree with the typed errors. That would have been a better idea, sadly that horse has left the barn.

You only export the error interface, so this should work without
breaking anything:

`type XfrError struct { Rcode int }

func (e *XfrError) Error() string {
    if e == nil {
        return "dns: <nil>"
    }
    return "dns: bad xfr rcode: " + strconv.Itoa(e.Rcode)
}`

And use it in xfr.go `c <- &Envelope{in.Answer, &XfrError{in.Rcode}}`.

But I understand if you don't take the chance to break stuff like
`err.(*dns.Error)`. Maybe something for a dns.v2.

Would also be nice to return the error string via `RcodeToString[e.Rcode]`
in dns.v2.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] SetIxfr with emtpy ..." ]
>And how do you tell `SetIxfr` to create a correct packet with empty ns
>and mbox? `SetIxfr("example.com.", 7, "", "")` won't do the trick.

SetIxfr("example.com.", 7, ".", ".")

should work.

>> Agree with the typed errors. That would have been a better idea, sadly that horse has left the barn.
>
>You only export the error interface, so this should work without
>breaking anything:
>
>`type XfrError struct { Rcode int }
>
>func (e *XfrError) Error() string {
>    if e == nil {
>        return "dns: <nil>"
>    }
>    return "dns: bad xfr rcode: " + strconv.Itoa(e.Rcode)
>}`
>
>And use it in xfr.go `c <- &Envelope{in.Answer, &XfrError{in.Rcode}}`.
[...]

PR #510, opened 2017-08-22
added XfrError type for better error handling

kahnert:
The XfrError type shouldn't break old code, as long as nobody made a type assertion to dns.Error on AXFR or IXFR errors.

See discussion in issue https://github.com/miekg/dns/issues/509

### DIFF
--- a/xfr.go
+++ b/xfr.go
@@ -65,7 +65,7 @@ func (t *Transfer) In(q *Msg, a string) (env chan *Envelope, err error) {
 			return
 		}
 		if q.Question[0].Qtype == TypeIXFR {
-			go t.inIxfr(q, env)
+			go t.inIxfr(q.Id, env)
 			return
 		}
 	}()
@@ -120,12 +120,9 @@ func (t *Transfer) inAxfr(id uint16, c chan *Envelope) {
 	}
 }
 
-func (t *Transfer) inIxfr(q *Msg, c chan *Envelope) {
-	id := q.Id
+func (t *Transfer) inIxfr(id uint16, c chan *Envelope) {
 	serial := uint32(0) // The first serial seen is the current server serial
 	first := true
-	last := false
-	qser := q.Ns[0].(*SOA).Serial
 	defer t.Close()
 	defer close(c)
 	timeout := dnsTimeout
@@ -143,11 +140,17 @@ func (t *Transfer) inIxfr(q *Msg, c chan *Envelope) {
 			c <- &Envelope{in.Answer, ErrId}
 			return
 		}
-		if in.Rcode != RcodeSuccess {
-			c <- &Envelope{in.Answer, &XfrError{in.Rcode}}
-			return
-		}
 		if first {
+			if in.Rcode != RcodeSuccess {
+				c <- &Envelope{in.Answer, &XfrError{in.Rcode}}
+				return
+			}
+			// A single SOA RR signals "no changes"
+			if len(in.Answer) == 1 && isSOAFirst(in) {
+				c <- &Envelope{in.Answer, nil}
+				return
+			}
+
 			// Check if the returned answer is ok
 			if !isSOAFirst(in) {
 				c <- &Envelope{in.Answer, ErrSoa}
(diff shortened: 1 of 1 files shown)


miekg:
@andrewtj think this looks good, slightly backwards incompat, but I wouldn't know why you would want to do type assertion on the old error.

andrewtj:
I agree that adding `XfrError` is unlikely to catch anyone out but I'm not sure the type as-is is sufficient. Are there any other cases it should cover? (I don't have the time to answer this for myself, sorry.)

miekg:
you mean making *more* errors types or extending this on to cover more cases?

andrewtj:
I wanted to make sure `XfrError` is a suitable catch-all for AXFR/IXFR errors. I still haven't read the protocol or code so I don't really know if an RCODE alone is good enough.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] added XfrError type..." ]
>I wanted to make sure `XfrError` is a suitable catch-all for AXFR/IXFR errors. I still haven't read the protocol or code so I don't really know if an RCODE alone is good enough.

Ah, ok. From the top of my head:

connection errors.
no SOA records
TSIG failures
non-signed envelopes

... and maybe more.

When I have some time, I'll try to dig in some more.

andrewtj:
@miekg It sounds like we could potentially end up with a bunch of types (for *XFR alone) if we go down this path.

@kahnert what were you doing with the RCODE once it was parsed? Grabbing a string from `dns.RcodeToString` or something more elaborate?

I wonder if some sort of type that contains a static string describing what was being attempted, the request, the response and an `error` interface would work more generally. That ought to capture enough information to debug a problem or implement a use-case specific work around. Does that sound like it would have been helpful in your case @kahnert?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] added XfrError type..." ]
>@miekg It sounds like we could potentially end up with a bunch of types (for *XFR alone) if we go down this path.

agreed.

>I wonder if some sort of type that contains a static string describing what was being attempted, the request, the response and an `error` interface would work more generally. That ought to capture enough information to debug a problem or implement a use-case specific work around. Does that sound like it would have been helpful in your case @kahnert?

+1 Think this a better approach.

kahnert:
> It sounds like we could potentially end up with a bunch of types

Which doesn't have to be bad by default. Still better than dynamically generated error strings which have to be parsed to extract the actual error code.

> what were you doing with the RCODE once it was parsed?

Decide how to deal with this error, e.g.: `RcodeNotAuth` -> wrong nameserver / typo (if domain was just added) or obsolete domain (if previously working), `RcodeRefused` -> missing permissions (ask DNS admins), ...

And also:

> Grabbing a string from `dns.RcodeToString`

for the error log, yes.

> I wonder if some sort of type that contains a static string describing what was being attempted

Would work for me, but it sounds a little bit tedious on the other side (inside the lib):

```
switch in.Rcode {
case RcodeSuccess:
    return nil
case RcodeFormatError:
    return ErrForm
case RcodeServerFailure:
    return ErrSrvFail
[skipped long list: you know how it would look like]
}
```

instead of the proposed `&XfrError{in.Rcode}`.

And we could even go with `type XfrError int` because XfrError will never be `nil`, so `dns: <nil>` won't ever be used anyway.

andrewtj:
@kahnert I'm not sure why you think there'd be a switch statement. The static string field would only describe what was being attempted, eg: `"xfer: initial SOA query"` and the error field would describe the problem only, eg: `ErrRcode`. If you needed more detail like the exact RCODE, you'd retrieve that from the message in the response field. The intention being to cover any error in the transfer process in detail with a minimal amount of new API.

kahnert:
@andrewtj I'm not sure if I understand your idea. Do you want to introduce a `RR` type called `ErrRcode`? If that's the case, than I can't see the advantage over an extra error type for xfer. Anyway, it's not my lib, I'm merely a user of it. Everything I say is, that it's cumbersome to parse the actual Rcode out of a string.

andrewtj:
@kahnert [`ErrRcode`](https://github.com/miekg/dns/blob/e4205768578dc90c2669e75a2f8a8bf77e3083a4/msg.go#L43) is an existing error. My suggestion is a new type that contains a string, a request, a response and an error interface. Something like:

```go
type XfrError struct {
	State             string // this should be static; a new type might be better
	Request, Response *Msg
	Error             error
}
```

kahnert:
@andrewtj I've no idea how to populate this struct. The only thing I would like to have is an easy way to get the Rcode number without parsing a string. I can't see any benefit of your suggested type. But anyway, go for it, if you think this would be the better approach. There is no immanent need to accept this PR.

miekg:
I'm going to close this as we need a (re)think on potentially a more generic error type that captures more (all?) errors.

PR #511, opened 2017-08-29
Redesigned Client API around net.Dialer

insomniacslk:
This change is aimed at redesigning the `dns.Client` API to wrap around a
`net.Dialer` struct. A Dialer exposes several interesting features, in particular
a `Timeout` (that does not depend on a `Context`) and `LocalAddr` (that allows to
specify local IP and port to use for the requests).

The central elements of the API now become `Client.DialWithDialer` and
`Client.ExchangeWithDialer`. The existing functions wrap around it. The existing
API however is unchanged, for backward compatibility, but some parts are now
deprecated.

A few hints to simplify the review:

* client.go has been split into client.go and client_compat.go . The latter
  contains the deprecated API, so it is immediate to identify what should be
  used and what should be moved away from. In particular, I have deprecated
  `ExchangeContext`, `ExchangeConn`, `DialTimeout`, `DialWithTLS`,
  `DialTimeoutWithTLS` and `ExchangeContext`. All of these functions now wrap
  around `Client.DialWithDialer` or `Client.ExchangeWithDialer`
* the tests are mostly unchanged. I have added some more and made a few small
  improvements, but the idea is that the original tests have to pass
* I have added a function `msgHash`, to be used with `singleflight` functions.
  This function returns the sha256 of a unique string representing a DNS
  question. Technically the string generated before the sha256 should be already
  unique, so passing through sha256 is not strictly necessary. I have not strong
  opinions on this, we just need the hash to return a unique string to identify
  a request
* a new function `Client.DialWithDialer` has been introduced. This function
  uses a `net.Dialer` to establish connections, and provides at least a
  per-request Timeout and a LocalAddr attribute that can be used to specify a
  source host and port from which to connect from
* a new function `Client.ExchangeWithDialer` has been introduced, too. This
  function behaves like the original `Client.Exchange`, but introduces an
  optional `net.Dialer` as input. This function is a wrapper around
  `Client.DialWithDialer`
* `Client.ExchangeContext` has been deprecated. Context here is only used for
  timeouts, and `net.Dialer` is a better alternative to to the same. Therefore
  `Client.ExchangeContext` is a wrapper around `Client.ExchangeWithDialer`


insomniacslk:
More context at https://github.com/miekg/dns/pull/502

miekg:
I haven't made any time to look at this. Will commit to review tomorrow.

client.go
```diff
@@ -30,104 +31,36 @@ type Client struct {
...
+	Timeout        time.Duration     // a cumulative timeout for dial, write and read, defaults to 0 (disabled) - overrides DialTimeout, ReadTimeout, WriteTimeout when non-zero. Can be overridden with net.Dialer.Timeout (see Client.ExchangeWithDialer and Client.DialWithDialer) or context.Context.Deadline (see the deprecated ExchangeContext)
+	DialTimeout    time.Duration     // net.DialTimeout, defaults to 2 seconds, or net.Dialer.Timeout if expiring earlier - overridden by Timeout when that value is non-zero
 	ReadTimeout    time.Duration     // net.Conn.SetReadTimeout value for connections, defaults to 2 seconds - overridden by Timeout when that value is non-zero
 	WriteTimeout   time.Duration     // net.Conn.SetWriteTimeout value for connections, defaults to 2 seconds - overridden by Timeout when that value is non-zero
 	TsigSecret     map[string]string // secret(s) for Tsig map[<zonename>]<base64 secret>, zonename must be fully qualified
 	SingleInflight bool              // if true suppress multiple outstanding queries for the same Qname, Qtype and Qclass
 	group          singleflight
 }
 
+// msgHash is a helper function to get a unique key given the triple
```
miekg: Please do this in a different PR.

client.go
```diff
@@ -30,104 +31,36 @@ type Client struct {
 	Net            string            // if "tcp" or "tcp-tls" (DNS over TLS) a TCP query will be initiated, otherwise an UDP one (default is "" for UDP)
 	UDPSize        uint16            // minimum receive buffer for UDP messages
 	TLSConfig      *tls.Config       // TLS connection configuration
-	Timeout        time.Duration     // a cumulative timeout for dial, write and read, defaults to 0 (disabled) - overrides DialTimeout, ReadTimeout and WriteTimeout when non-zero
-	DialTimeout    time.Duration     // net.DialTimeout, defaults to 2 seconds - overridden by Timeout when that value is non-zero
+	Timeout        time.Duration     // a cumulative timeout for dial, write and read, defaults to 0 (disabled) - overrides DialTimeout, ReadTimeout, WriteTimeout when non-zero. Can be overridden with net.Dialer.Timeout (see Client.ExchangeWithDialer and Client.DialWithDialer) or context.Context.Deadline (see the deprecated ExchangeContext)
```
miekg: Wouldn't it be better to just put a Dialer in here? Why not?

client.go
```diff
@@ -202,40 +104,65 @@ func (c *Client) writeTimeout() time.Duration {
...
+	if dialer == nil {
+		realDialer = net.Dialer{}
+	} else {
+		realDialer = net.Dialer(*dialer)
+	}
+	realDialer.Timeout = c.getTimeoutForRequest(dialer, c.writeTimeout())
+
 	network := "udp"
-	tls := false
+	use_tls := false
```
miekg: don't use underscores in Go names, also this does *not* have to be changed for this PR, so please bump to a new one.

client.go
```diff
@@ -253,12 +180,13 @@ func (c *Client) exchange(ctx context.Context, m *Msg, a string) (r *Msg, rtt ti
 	}
 
 	co.TsigSecret = c.TsigSecret
-	co.SetWriteDeadline(deadlineOrTimeoutOrCtx(ctx, deadline, c.writeTimeout()))
+	// write with the appropriate write timeout
+	co.SetWriteDeadline(time.Now().Add(c.getTimeoutForRequest(dialer, c.writeTimeout())))
```
miekg: why are you removing/rewriting deadlineOrTimeotOrCtx? Because we don't have ctx anymore?

client_compat.go
```diff
@@ -0,0 +1,101 @@
...
+	conn, err = client.DialWithDialer(&net.Dialer{Timeout: timeout}, address)
+	if err != nil {
+		return nil, err
+	}
+	return conn, nil
+}
+
+// DialWithTLS connects to the address on the named network with TLS.
+func DialWithTLS(network, address string, tlsConfig *tls.Config) (conn *Conn, err error) {
+	println("dns: DialWithTLS: this function is deprecated")
```
miekg: don't think we can (or should) put these warnings here though...

client.go
insomniacslk: The only reason was to avoid breaking code that uses this (public) interface. Anyone creating a `Client` with a Timeout would see their code failing the build.
What to do here depends on how inclined you are to push a breaking change

client.go
insomniacslk: OK

client.go
insomniacslk: OK for renaming it.

I am not sure I get it right when you say this does not have to be changed for this PR. Do you mean "leave the variable name unchanged" or something else? In the former case, I changed it because the variable `tls` was clashing with the package name `tls`, and I found that `use_tls` explained better what it is used for

miekg:
Yes just leave tls unless absolutely necessary to change it for this pr

client.go
insomniacslk: Yes but not only. Background: this part of the diff went through a couple of iterations, where I first removed the function, and eventually reinstated it. When I reinstated it, the content was slightly different, and I was looking for a name that was more descriptive of its behaviour.

client_compat.go
insomniacslk: I am not aware of a blessed way to generate warnings in Go, so I just did what already done in `ExchangeConn`. According to the documentation, `println` is unstable API and may disappear.

On the "whether we should", I think it's useful to notify the user of the risk of their code breaking with future versions of the library. While it may be annoying to read on stdout/logs, it is a clear indicator of a bigger risk.

I don't have strong opinions on this, I can remove them if you think this is wrong.

insomniacslk:
unfortunately the rename of `tls` is necessary because it clashes with the package name from Go's standard library. I can rename the import, but I would rather rename an internal variable than a stl package import

insomniacslk:
Will wait for your feedback before submitting an updated diff

client.go
miekg: I don't meaning removing anything, just *adding* a net.Dailer in here...

client.go
miekg: Can we *not* change it now, so we can just focus on getting a net.Dialer in ?

client_compat.go
miekg: Yes, remove these println's; nothing is deprecated here, we adding new stuff.

client.go
insomniacslk: ...I have no idea why I didn't consider this since the beginning. OK, will add a Dialer here. This means that I can remove `Client.DialWithDialer` and `Client.ExchangeWithDialer` and just leave `Client.Dial` and `Client.Exchange` that will implicitly use the `Client.Dialer` if it's not nil. Does this sound right?

client.go
insomniacslk: ok, will revert this part

client_compat.go
insomniacslk: will do

client.go
insomniacslk: Went through the code again, and refreshed my memory. I remembered other reasons to rewrite that function:

* all the parameters are now timeouts, not deadlines (since context is gone, and Dialer has no deadline either). This makes them easily comparable, which is not the case when mixing deadlines and timeouts
* context is gone, deadline is gone, dialer is coming, so basically this function would have a different name and a different content too. Not sure if it still makes sense to leave it as it is, as most of its content has to change anyway

As per the above I still believe that this function should be rewritten within this PR. Since we are moving a Dialer under Client, it could be done like this:

* remove deadlineOrTimeoutOrCtx anyway (it's not part of the API, it's unexported, and as said above it's not relevant anymore in the original form)
* leave this new method (whatever name we want to give it) attached to Client, and simplify it by just passing a time.Duration. The dialer timeout can be accessed inside as `c.Dialer.Timeout`. In other words `func (c *Client) getTimeoutForRequest(timeout time.Duration) time.Duration`.

What do you think of the above?

client.go
insomniacslk: done, diff incoming

client.go
insomniacslk: done, diff incoming

client.go
insomniacslk: done, diff incoming

client.go
insomniacslk: done as I suggested above so you can review the new diff, then decide how to proceed. Diff incoming

miekg:
> unfortunately the rename of tls is necessary because it clashes with the package name from Go's standard library.

Ack, rename the var then.

client.go
```diff
@@ -44,91 +44,74 @@ type Client struct {
...
 
-	defer co.Conn.Close()
-
-	opt := m.IsEdns0()
-	// If EDNS0 is used use that for size.
-	if opt != nil && opt.UDPSize() >= MinMsgSize {
-		co.UDPSize = opt.UDPSize()
+func (c *Client) Dial(address string) (conn *Conn, err error) {
+	// create a new dialer with the appropriate timeout
+	var realDialer net.Dialer
```
miekg: why not: `var d *net.Dialer`?

client.go
```diff
@@ -142,100 +125,35 @@ func ExchangeConn(c net.Conn, m *Msg) (r *Msg, err error) {
...
-func (c *Client) writeTimeout() time.Duration {
-	if c.WriteTimeout != 0 {
-		return c.WriteTimeout
-	}
-	return dnsTimeout
+	return r, rtt, err
 }
 
-func (c *Client) exchange(ctx context.Context, m *Msg, a string) (r *Msg, rtt time.Duration, err error) {
+func (c *Client) exchange(m *Msg, a string) (r *Msg, rtt time.Duration, err error) {
```
miekg: Why wouldn't we want a context here?

client.go
insomniacslk: That's an old name from when there was a second dialer passed as argument. Will rename to `d`

client.go
insomniacslk: If we have a dialer we just don't need a context. The only property of the context used here is the deadline, but there is already a timeout implemented in the dialer in the stdlib.
From the design perspective, using a context would also makes this code more complex, by introducing several different timeouts (context deadline, dialer timeout, default timeout, read/write/dial timeout). A dialer is in my opinion a better place for timeouts, and it is already implemented and handled in the standard library. On the other side, a context carries a deadline, not a timeout (there's quite a difference): a deadline may be already expired when called, while a timeout starts with the request, and can be reused, e.g. for writing and then for reading. More in general Context it's under a lot of discussions on whether it should be used again for what it has been created, i.e. for context-specific data, not for deadlines or other things.
That's my opinion of course, always happy to discuss!

insomniacslk:
I've reverted `client_compat.go` but of course the diff is not much more readable :(

I've also addressed the rest of your feedback, with the exception of the context, as per my previous comment

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Redesigned Client A..." ]
>I've reverted `client_compat.go` but of course the diff is not much more readable :(
>
>I've also addressed the rest of your feedback, with the exception of the context, as per my previous comment

Try to look again this evening (and other open PRs on this repo)

client.go
```diff
@@ -30,8 +31,9 @@ type Client struct {
 	Net            string            // if "tcp" or "tcp-tls" (DNS over TLS) a TCP query will be initiated, otherwise an UDP one (default is "" for UDP)
 	UDPSize        uint16            // minimum receive buffer for UDP messages
 	TLSConfig      *tls.Config       // TLS connection configuration
-	Timeout        time.Duration     // a cumulative timeout for dial, write and read, defaults to 0 (disabled) - overrides DialTimeout, ReadTimeout and WriteTimeout when non-zero
-	DialTimeout    time.Duration     // net.DialTimeout, defaults to 2 seconds - overridden by Timeout when that value is non-zero
+	Dialer         *net.Dialer       // a net.Dialer used to set local address, timeouts and more
+	Timeout        time.Duration     // a cumulative timeout for dial, write and read, defaults to 0 (disabled) - overrides DialTimeout, ReadTimeout, WriteTimeout when non-zero. Can be overridden with net.Dialer.Timeout (see Client.ExchangeWithDialer and Client.Dialer) or context.Context.Deadline (see the deprecated ExchangeContext)
```
miekg: Please put some line breaks in this line - I know there is no limit in Go, but this is pushing it.

client.go
```diff
@@ -30,8 +31,9 @@ type Client struct {
 	Net            string            // if "tcp" or "tcp-tls" (DNS over TLS) a TCP query will be initiated, otherwise an UDP one (default is "" for UDP)
 	UDPSize        uint16            // minimum receive buffer for UDP messages
 	TLSConfig      *tls.Config       // TLS connection configuration
-	Timeout        time.Duration     // a cumulative timeout for dial, write and read, defaults to 0 (disabled) - overrides DialTimeout, ReadTimeout and WriteTimeout when non-zero
-	DialTimeout    time.Duration     // net.DialTimeout, defaults to 2 seconds - overridden by Timeout when that value is non-zero
+	Dialer         *net.Dialer       // a net.Dialer used to set local address, timeouts and more
```
miekg: Why is `net.Dailer` a pointer here, but in `Dial` it is used as-is `var d net.Dialer`

client.go
```diff
@@ -467,6 +388,24 @@ func (co *Conn) Write(p []byte) (n int, err error) {
...
 
+// Return the appropriate timeout for a specific request
+func (c *Client) getTimeoutForRequest(timeout time.Duration) time.Duration {
+	var requestTimeout time.Duration
+	if c.Timeout != 0 {
+		requestTimeout = c.Timeout
+	} else {
+		requestTimeout = timeout
+	}
+	// net.Dialer.Timeout has priority if smaller than the timeouts computed so
```
miekg: Is this documented anywhere else?

client.go
insomniacslk: it's a pointer so it can be `nil`, which means "I don't want to set any Dialer, use your defaults".

In `Dial` I do this:
```
    var d net.Dialer 
    if c.Dialer == nil {
        d = net.Dialer{}
    } else { 
        d = net.Dialer(*c.Dialer) 
    } 
```
which eventually uses another dialer instance - the one from `Client`, if any, or a default one if not

client.go
insomniacslk: Will do, not a big fan of long lines either

client.go
insomniacslk: Yes, at line 35 (the one after the long one that I will break in the next commit):

```
DialTimeout    time.Duration     // net.DialTimeout, defaults to 2 seconds, or net.Dialer.Timeout if
expiring earlier - overridden by Timeout when that value is non-zero
```

While I also updated `docs.go`, I didn't call this out explicitly to stay true to the style of the rest of the document (i.e. no information overload). The current text there is:

```
More advanced options are availabe using a net.Dialer and the corresponding API.
For example it is possible to set a timeout, or to specify a source IP address
and port to use for the connection:

        c := new(dns.Client)
        laddr := net.UDPAddr{
                IP: net.ParseIP("[::1]"),
                Port: 12345,
                Zone: "",
        }
        d := net.Dialer{
                Timeout: 200 * time.Millisecond,
                LocalAddr: &laddr,
        }
        in, rtt, err := c.ExchangeWithDialer(&d, m1, "8.8.8.8:53")
```

client.go
insomniacslk: `gofmt` keeps squeezing spaces and breaks the alignment of this and the other (original ) lines. Do you want me to keep the alignment?

insomniacslk: I've respected `gofmt` for now, which squeeze away the alignment. Have a look and let me know if you prefer the old format, I'll have to disable the `gofmt` plugin to get it aligned :D

client.go
```diff
@@ -27,11 +28,15 @@ type Conn struct {
...
 type Client struct {
-	Net            string            // if "tcp" or "tcp-tls" (DNS over TLS) a TCP query will be initiated, otherwise an UDP one (default is "" for UDP)
-	UDPSize        uint16            // minimum receive buffer for UDP messages
-	TLSConfig      *tls.Config       // TLS connection configuration
-	Timeout        time.Duration     // a cumulative timeout for dial, write and read, defaults to 0 (disabled) - overrides DialTimeout, ReadTimeout and WriteTimeout when non-zero
-	DialTimeout    time.Duration     // net.DialTimeout, defaults to 2 seconds - overridden by Timeout when that value is non-zero
+	Net       string      // if "tcp" or "tcp-tls" (DNS over TLS) a TCP query will be initiated, otherwise an UDP one (default is "" for UDP)
+	UDPSize   uint16      // minimum receive buffer for UDP messages
+	TLSConfig *tls.Config // TLS connection configuration
+	Dialer    *net.Dialer // a net.Dialer used to set local address, timeouts and more
```
miekg: I don't see the change? This should be `Dialer net.Dialer`?

insomniacslk: there is intentionally no change, I explained above why I used a pointer. Reporting here for ease:

```
it's a pointer so it can be nil, which means "I don't want to set any Dialer, use your defaults".

In Dial I do this:

    var d net.Dialer 
    if c.Dialer == nil {
        d = net.Dialer{}
    } else { 
        d = net.Dialer(*c.Dialer) 
    } 
which eventually uses another dialer instance - the one from Client, if any, or a default one if not
```

insomniacslk: In other words I want `Dialer` to be `nil` by default, in which case the defaults are used. If set, the defaults will be overridden

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Redesigned Client A..." ]
>In Dial I do this:
>
>    var d net.Dialer
>    if c.Dialer == nil {
>        d = net.Dialer{}
>    } else {
>        d = net.Dialer(*c.Dialer)
>    }


Yes, I see, OK

/Miek

--
Miek Gieben

insomniacslk:
Thanks!

client.go
tmthrgd: @insomniacslk Using a pointer isn't needed to do that. If it's `nil`, you're just explicitly using the zero value anyway. The following would achieve the exact same operation and is less code; I don't see why a pointer is preferable:

```
type Client struct {
	...
	Dialer net.Dialer
	...
}

func (c *Client) Dial(address string) (conn *Conn, err error) {
	d := c.Dialier // this creates a copy of Dialer
	...
}
```

PR #514, opened 2017-09-11
Document when the AuthenticatedData flag should be trusted

rolandshoemaker:
Hopefully this will help others not have to dive down the DNSSEC RFC rabbit hole.

### DIFF
--- a/msg.go
+++ b/msg.go
@@ -77,9 +77,12 @@ type MsgHdr struct {
 	RecursionDesired   bool
 	RecursionAvailable bool
 	Zero               bool
-	AuthenticatedData  bool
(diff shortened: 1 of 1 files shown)


miekg:
NACK in this form. Should probably not say anything about usage but just reference to the relevant RFC:
~~~
AuthenticatedData bool // RFC 6840 Section 5.7
~~~
And then also for all the other header bits

rolandshoemaker:
Makes sense, I'll update this PR do document the most recent relevant doc for each header bit.

miekg:
Closing PRs to clear my head. Please re-open when fixed/done.

Issue #516, opened 2017-09-18
Incorrect implementation of EDNS Extended RCODE


m13253:
According to [RFC6891](https://tools.ietf.org/html/rfc6891#section-6.1.3):
```
   EXTENDED-RCODE
      Forms the upper 8 bits of extended 12-bit RCODE (together with the
      4 bits defined in [RFC1035].
```

But the [current implementation](https://github.com/miekg/dns/blob/5a15a35f5f671cf6ad8b316b50607c76e5e87fb6/edns.go#L104) is:
```go
func (rr *OPT) ExtendedRcode() int {
	return int((rr.Hdr.Ttl&0xFF000000)>>24) + 15
}
func (rr *OPT) SetExtendedRcode(v uint8) {
	if v < RcodeBadVers { // Smaller than 16.. Use the 4 bits you have!
		return
	}
	rr.Hdr.Ttl = rr.Hdr.Ttl&0x00FFFFFF | (uint32(v-15) << 24)
}
```

RFC requires extended 12-bit RCODE being put to two places:
- Upper 8 bits in `OPT RR`
- Lower 4 bits in original `RCODE`

While the implementation subtracts 15 from `v`, then puts it into `OPT RR`.

Hopefully we only have `BADVERS` in EDNS specification, and by coincidence `BADVERS` encodes to the same value by both the correct and wrong algorithm. It is necessary to fix it before any other RCODE is standardized.

PR #517, opened 2017-09-18
Fix EDNS Extended RCODE

m13253:
This pull request fixes issue #516.

### DIFF
--- a/edns.go
+++ b/edns.go
@@ -102,15 +102,13 @@ func (rr *OPT) SetVersion(v uint8) {
 }
 
 // ExtendedRcode returns the EDNS extended RCODE field (the upper 8 bits of the TTL).
-// Make sure you "OR" with the 4-bit RCODE field of the header to form the full result.
-func (rr *OPT) ExtendedRcode() uint16 {
-	return uint16((rr.Hdr.Ttl&0xFF000000)>>20)
+func (rr *OPT) ExtendedRcode() int {
+	return int((rr.Hdr.Ttl&0xFF000000)>>24)
 }
 
 // SetExtendedRcode sets the EDNS extended RCODE field.
-// Lower 4 bits are discarded, as they should be set in the 4-bit RCODE field of the header.
-func (rr *OPT) SetExtendedRcode(v uint16) {
-	rr.Hdr.Ttl = rr.Hdr.Ttl&0x00FFFFFF | (uint32(v&0xFF0) << 20)
+func (rr *OPT) SetExtendedRcode(v uint8) {
+	rr.Hdr.Ttl = rr.Hdr.Ttl&0x00FFFFFF | (uint32(v) << 24)
 }
 
 // UDPSize returns the UDP buffer size.
--- a/edns.go
+++ b/edns.go
@@ -102,16 +102,15 @@ func (rr *OPT) SetVersion(v uint8) {
 }
 
 // ExtendedRcode returns the EDNS extended RCODE field (the upper 8 bits of the TTL).
-func (rr *OPT) ExtendedRcode() int {
-	return int((rr.Hdr.Ttl&0xFF000000)>>24) + 15
+// Make sure you "OR" with the 4-bit RCODE field of the header to form the full result.
+func (rr *OPT) ExtendedRcode() uint16 {
+	return uint16((rr.Hdr.Ttl&0xFF000000)>>20)
 }
 
 // SetExtendedRcode sets the EDNS extended RCODE field.
-func (rr *OPT) SetExtendedRcode(v uint8) {
-	if v < RcodeBadVers { // Smaller than 16.. Use the 4 bits you have!
-		return
-	}
-	rr.Hdr.Ttl = rr.Hdr.Ttl&0x00FFFFFF | (uint32(v-15) << 24)
+// Lower 4 bits are discarded, as they should be set in the 4-bit RCODE field of the header.
+func (rr *OPT) SetExtendedRcode(v uint16) {
+	rr.Hdr.Ttl = rr.Hdr.Ttl&0x00FFFFFF | (uint32(v&0xFF0) << 20)
 }
 
 // UDPSize returns the UDP buffer size.
--- a/edns_test.go
+++ b/edns_test.go
@@ -62,7 +62,7 @@ func TestOPTTtl(t *testing.T) {
 	}
 
 	e.SetExtendedRcode(42)
-	if e.ExtendedRcode() != 42 {
-		t.Errorf("set 42, expected %d, got %d", 42-15, e.ExtendedRcode())
+	if e.ExtendedRcode() != (42 & 0xFF0) {
+		t.Errorf("set 42, expected %d, got %d", 42 & 0xFF0, e.ExtendedRcode())
 	}
 }
--- a/edns_test.go
+++ b/edns_test.go
@@ -62,7 +62,7 @@ func TestOPTTtl(t *testing.T) {
 	}
 
 	e.SetExtendedRcode(42)
-	if e.ExtendedRcode() != (42 & 0xFF0) {
-		t.Errorf("set 42, expected %d, got %d", 42 & 0xFF0, e.ExtendedRcode())
+	if e.ExtendedRcode() != 42 {
+		t.Errorf("set 42, expected %d, got %d", 42, e.ExtendedRcode())
 	}
 }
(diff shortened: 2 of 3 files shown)


edns_test.go
```diff
@@ -62,7 +62,7 @@ func TestOPTTtl(t *testing.T) {
 	}
 
 	e.SetExtendedRcode(42)
-	if e.ExtendedRcode() != 42 {
-		t.Errorf("set 42, expected %d, got %d", 42-15, e.ExtendedRcode())
+	if e.ExtendedRcode() != (42 & 0xFF0) {
```
miekg: Why doesn't ExtendedRcode return 42?

m13253: The original code stores `v - 15` into OPT RR.
The correct behavior is to store `v >> 4` into OPT RR.

The currently only available extended RCODE (`RcodeBadVers = 16`), produces the coincidence that `16 - 15 == 16 >> 4`, therefore the existing code is not causing problem with **real-life** DNS packets at present.

By adapting the correct behavior, I need to discard the lower 4 bits from `v` to make sure the function still behaves correctly with `e.SetExtendedRcode(RcodeBadVers)`.
Therefore the discarded lower 4 bits is intended to cause an input of 42 to produce an output of 32.

Another method to fix this problem is to ask the caller to discard the lower 4 bits themselves before calling `SetExtendedRcode`, making the output still 42. But this might break any existing third-party code.

---

By the way, `msg.go` is already using `SetExtendedRcode` [the wrong way](https://github.com/miekg/dns/blob/e4205768578dc90c2669e75a2f8a8bf77e3083a4/msg.go#L708). It discarded the lower 4 bits **before** calling, resulting `SetExtendedRcode` always falling into the empty `return` statement. I have also modified it to use the consistent behavior.

I have also updated the docstring, and intentionally changed the function signature to bring in incompatibility. I think this change will not introduce problem to third-party code without being noticed.

miekg:
I'm hoping we can get away with pushing this even though it is backward incompatible, we changing the function signature anyway from the wrong `uint8`. So if we're going to do that; we should probably make it sane to use as well

edns_test.go
miekg: What I meant; if we properly fix this (by dropping `uint8`) this should just return 42?

m13253: > What I meant; if we properly fix this (by dropping `uint8`) this should just return 42?

No I don't think so.

Extended Rcode only records the upper 8 bits. The lower 4 bits is still in the old place.
You should either drop the last 4 bits yourself, or ask the caller to drop the 4 bits for you.
For the former, 42 should not just return 42. For the latter, existing users might be confused.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Fix EDNS Extended R..." ]
>m13253 commented on this pull request.
>
>
>
>> @@ -62,7 +62,7 @@ func TestOPTTtl(t *testing.T) {
> 	}
>
> 	e.SetExtendedRcode(42)
>-	if e.ExtendedRcode() != 42 {
>-		t.Errorf("set 42, expected %d, got %d", 42-15, e.ExtendedRcode())
>+	if e.ExtendedRcode() != (42 & 0xFF0) {
>
>> What I meant; if we properly fix this (by dropping `uint8`) this should just return 42?
>
>No I don't think so.
>
>Extended Rcode only records the upper 8 bits. The lower 4 bits is still in the old place.
>You should either drop the last 4 bits yourself, or ask the caller to drop the 4 bits for you.
>For the former, 42 should not just return 42. For the latter, existing users might be confused.

I don't understand: why should I care about the implemenation as a user of this package? 42 in, 42 out?

m13253:
> I don't understand: why should I care about the implemenation as a user of this package? 42 in, 42 out?

As a simple explanation: 42 is not a valid input for this function. It has lower 4 bits set.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Fix EDNS Extended R..." ]
>> I don't understand: why should I care about the implemenation as a user of this package? 42 in, 42 out?
>
>As a simple explanation: 42 is not a valid input for this function. It has lower 4 bits set.

ah :-) I stand corrected. I'm sorry, but I don't have time to reread all the 
extended rcode: But it is *not* a continues range from 15 to ... N ? This might 
might warrant some usuability comments around the public exported code.

m13253:
> But it is *not* a continues range from 15 to ... N ?

According to [RFC6891](https://tools.ietf.org/html/rfc6891#section-6.1.3):
```
   EXTENDED-RCODE
      Forms the upper 8 bits of extended 12-bit RCODE (together with the
      4 bits defined in [RFC1035].  Note that EXTENDED-RCODE value 0
      indicates that an unextended RCODE is in use (values 0 through
      15).
```

The OPT record only stores the upper 8 bits, representing 16, 32, 48, 64 ... 4080
The old header stores the lower 4 bits, representing 0 to 15.
To represent the code 42, you need to put 32 into OPT record (stored as 2), then put 10 into the old header. You **can not** use only the OPT record to represent 42.

Since the biggest code number today is only 16.
So to represent the code 16, you need to put 16 into OPT record (stored as 1), then put 0 into the old header.
The previous code also stores 1 into OPT record by accident (calculated from 16-15). That's why this problem is not discovered for years.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Fix EDNS Extended R..." ]
>> But it is *not* a continues range from 15 to ... N ?
>
>According to [RFC6891](https://tools.ietf.org/html/rfc6891#section-6.1.3):
>```
>   EXTENDED-RCODE
>      Forms the upper 8 bits of extended 12-bit RCODE (together with the
>      4 bits defined in [RFC1035].  Note that EXTENDED-RCODE value 0
>      indicates that an unextended RCODE is in use (values 0 through
>      15).
>```
>
>The OPT record only stores the upper 8 bits, representing 16, 32, 48, 64 ... 4080
>The old header stores the lower 4 bits, representing 0 to 15.
>To represent the code 42, you need to put 32 into OPT record (stored as 2), then put 10 into the old header. You **can not** use only the OPT record to represent 42.
>
>Since the biggest code number today is only 16.
>So to represent the code 16, you need to put 16 into OPT record (stored as 1), then put 0 into the old header.
>The previous code also stores 1 into OPT record by accident (calculated from 16-15). That's why this problem is not discovered for years.

my my... Should we then fiddle with the signature of SetExtendedRcode even more 
and return on error on wrong values?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Fix EDNS Extended R..." ]
>> But it is *not* a continues range from 15 to ... N ?
>
>According to [RFC6891](https://tools.ietf.org/html/rfc6891#section-6.1.3):
>```
>   EXTENDED-RCODE
>      Forms the upper 8 bits of extended 12-bit RCODE (together with the
>      4 bits defined in [RFC1035].  Note that EXTENDED-RCODE value 0
>      indicates that an unextended RCODE is in use (values 0 through
>      15).
>```
>
>The OPT record only stores the upper 8 bits, representing 16, 32, 48, 64 ... 4080
>The old header stores the lower 4 bits, representing 0 to 15.
>To represent the code 42, you need to put 32 into OPT record (stored as 2), then put 10 into the old header. You **can not** use only the OPT record to represent 42.

this is not true. You get 12 bits and you can use then 0..2^12-1

Question is how should SetExtendedRcode behave, looking through it (again), the 
uint8 makes sense, but from an API standpoint is not obvious, Setting 42 and 
then getting that value back is impossibly. OK

This this needs a least a documentation update on what SetExtendedRcode actually 
does. The smaller than 16 test is clearly bullshit, and I don't trust the rest 
anymore.

Can you rework you PR to the minimal possible with the current API and added 
docs?

m13253:
> Can you rework you PR to the minimal possible with the current API and added
> docs?

I didn't quite understand this sentence. Sorry.
Do you want the "discard 4 bits" behavior, or the "42 in, 42 out" behavior?

So I submitted another method to fix. This time "42 in, 42 out".
I changed every piece of existing code calling `SetExtendedRcode` to match this new behavior.

Please review this new commit (diff it with `HEAD~~`).
Then let me know which method do you want. The old one, or the new one?
I will then follow your decision, and improve the docs to match the change.

edns.go
```diff
@@ -103,15 +103,12 @@ func (rr *OPT) SetVersion(v uint8) {
 
 // ExtendedRcode returns the EDNS extended RCODE field (the upper 8 bits of the TTL).
```
miekg: Thanks.. Should we update the function doc as well? To make it clearer what we're doing here?

m13253: > Thanks.. Should we update the function doc as well? To make it clearer what we're doing here?

I agree. We should improve the docs.

And sorry for the delay.
I will submit another PR to improve the docs several days later.

PR #519, opened 2017-09-19
Simplify compressed length code

rolandshoemaker:
Removes a bunch of duplication in the `msg.Len` path and exposes the method used to calculate the compressed length of a RR given a provided compression map.

### DIFF
--- a/msg.go
+++ b/msg.go
@@ -915,66 +915,29 @@ func compressedLen(dns *Msg, compress bool) int {
 	l := 12 // Message header is always 12 bytes
 	compression := map[string]int{}
 
+	lenFunc := Len
+	if compress {
+		lenFunc = func(r RR) int {
+			return CompressedLen(compression, r)
+		}
+	}
+
 	for i := 0; i < len(dns.Question); i++ {
 		l += dns.Question[i].len()
 		if compress {
 			compressionLenHelper(compression, dns.Question[i].Name)
 		}
 	}
-	for i := 0; i < len(dns.Answer); i++ {
-		if dns.Answer[i] == nil {
-			continue
-		}
-		l += dns.Answer[i].len()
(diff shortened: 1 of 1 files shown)


rolandshoemaker:
```
master HEAD
BenchmarkMsgLength-4                	 1000000	      1576 ns/op
BenchmarkMsgLengthNoCompression-4   	20000000	       108 ns/op

compress-len HEAD
BenchmarkMsgLength-4                	 1000000	      1552 ns/op
BenchmarkMsgLengthNoCompression-4   	10000000	       123 ns/op
```

Appears to have minimal impact.

Reason for exporting CompressLen was to help work on a external implementation of a fit message in X bytes method although I now notice there is an issue open for a similar method here (#223). I'd be happy to leave this unexported and instead just upstream the rest of my work on that here (I feel like there could be utility for leaving it exported but I can't really come up with another good reason off the top of my head...).

miekg:
From 108ns/op to 123 ns/op that

rolandshoemaker:
So I think the most of the latency increase here comes from using a nested loop that the compiler can't optimize as much as it'd like. If I pull that out we get down to ~9ns difference, although this is unlikely to be a statistically significant change (`0.063 > 0.05`)

```
name                      old time/op  new time/op  delta
MsgLength-4               1.79µs ± 4%  1.92µs ±15%   ~     (p=0.310 n=5+5)
MsgLengthNoCompression-4   126ns ± 8%   135ns ± 7%   ~     (p=0.063 n=5+5)
```

but ¯\_(ツ)_/¯, if you're not happy with any increase here I'm happy to just close this out.

miekg:
No, we should not merge anything that increases time spend here. If you can get a lower I'll merge :)

Closing for now.

PR #524, opened 2017-09-20
Correctly set the Source IP to the received Destination IP with ipv4/ipv6

fastest963:
Previously, the oob data was just stored and sent to WriteMsgUDP but it ignores the Src field when writing. Instead, now it is setting the Src to the original Dst and handling IPv4 IPs over IPv6 correctly.

Unfortunately, we still need the Windows-specific file since `ReadMsgUDP` is not implemented in Windows:
```
// BUG(mikio): On NaCl, Plan 9 and Windows, the ReadMsgUDP and
// WriteMsgUDP methods of UDPConn are not implemented.
```
Additionally, `SetControlMessage` is not supported on Windows either.

I've tested nothing broke with Windows, and Linux ipv4/ipv6 works as expected.

### DIFF
--- a/udp.go
+++ b/udp.go
@@ -27,8 +27,19 @@ func ReadFromSessionUDP(conn *net.UDPConn, b []byte) (int, *SessionUDP, error) {
 	return n, &SessionUDP{raddr, oob[:oobn]}, err
 }
 
-// WriteToSessionUDP acts just like net.UDPConn.WritetTo(), but uses a *SessionUDP instead of a net.Addr.
+// WriteToSessionUDP acts just like net.UDPConn.WriteTo(), but uses a *SessionUDP instead of a net.Addr.
 func WriteToSessionUDP(conn *net.UDPConn, b []byte, session *SessionUDP) (int, error) {
-	n, _, err := conn.WriteMsgUDP(b, session.context, session.raddr)
+	oob := correctSource(session.context)
+	n, _, err := conn.WriteMsgUDP(b, oob, session.raddr)
 	return n, err
+}
+
+// correctSource takes oob data and returns new oob data with the Src equal to the Dst
+func correctSource(oob []byte) []byte {
+	dst, err := parseUDPSocketDst(oob)
+	// If the destination could not be determined, ignore.
+	if err != nil || dst == nil {
+		return nil
+	}
+	return marshalUDPSocketSrc(dst)
 }
--- a/udp_windows.go
+++ b/udp_windows.go
@@ -4,10 +4,12 @@ package dns
 
 import "net"
 
+// SessionUDP holds the remote address
 type SessionUDP struct {
 	raddr *net.UDPAddr
 }
 
+// RemoteAddr returns the remote network address.
 func (s *SessionUDP) RemoteAddr() net.Addr { return s.raddr }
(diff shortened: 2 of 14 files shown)


server_test.go
```diff
@@ -59,7 +59,9 @@ func RunLocalUDPServerWithFinChan(laddr string) (*Server, string, chan struct{},
 	fin := make(chan struct{}, 0)
 
 	go func() {
-		server.ActivateAndServe()
+		if err := server.ActivateAndServe(); err != nil {
+			panic(err)
```
miekg: hmmm, `panic` in a lib
This was probably to reason to, uhm, ignore that error

udp.go
```diff
@@ -4,6 +4,11 @@ package dns
 
 import (
 	"net"
+	"os"
+	"syscall"
+
+	"golang.org/x/net/ipv4"
```
miekg: We only use the std lib, so this is a *big* change and will mean we can't merge it

server_test.go
fastest963: Forgot to remove this. It was returning an error and stalling all of the tests so I needed to figure out what the error was.

udp.go
fastest963: Didn't realize that restriction. I'll work on removing that.

fastest963:
@miekg Updated to remove any external package dependencies. I also added tests to verify the Linux implementation. I also realized last time I contributed I forgot to update CONTRIBUTORS so I did that as well.

udp_windows.go
```diff
@@ -21,9 +23,8 @@ func ReadFromSessionUDP(conn *net.UDPConn, b []byte) (int, *SessionUDP, error) {
 	return n, session, err
 }
 
-// WriteToSessionUDP acts just like net.UDPConn.WritetTo(), but uses a *SessionUDP instead of a net.Addr.
+// WriteToSessionUDP acts just like net.UDPConn.WriteTo(), but uses a *SessionUDP instead of a net.Addr.
 func WriteToSessionUDP(conn *net.UDPConn, b []byte, session *SessionUDP) (int, error) {
 	n, err := conn.WriteTo(b, session.raddr)
 	return n, err
 }
-
```
fastest963: There were two empty lines at the end of the file.

socket/controlmessage.go
```diff
@@ -0,0 +1,118 @@
+package socket
```
miekg: this is all copied verbatim from .... ? Might need a comments in these file. Do we need keep in sync with these files? (we basically vendoring by copying the code?)

fastest963:
@miekg Added a comment to the package that it was copied from `github.com/x/net/internal/socket`. We don't need to necessarily keep in sync with them unless the Linux kernel definitions change. I also copied the test and added it.

fastest963:
@miekg good call. Didn't think about that. Moved into an internal directory.

miekg:
This breaks a test in CoreDNS: https://travis-ci.org/coredns/coredns/jobs/281471373

fastest963:
The problem was actually with #511. I fixed it in #526

fastest963:
@miekg The problem was actually with #511. I fixed it in #526.

PR #526, opened 2017-09-30
Fix DialTimeout always using udp

fastest963:


miekg:
LOL. Thanks for finding and fixing!

miekg:
cc @insomniacslk fyi

insomniacslk:
Damn, didn't catch it, sorry! Will look into what I overlooked in tests tonight after the conference

insomniacslk:
wanted to write an unit test for that but with the latest commit it fails to build:

```
$ go build
udp_linux.go:18:2: use of internal package not allowed
```

Started at commit https://github.com/miekg/dns/commit/032fbabc82455d0649e332bdf83767f134fc9431 . Not sure why the Travis-CI build is not failing though

fastest963:
@insomniacslk Are you running `go build` from `$GOPATH/src/github.com/miekg/dns`? In order for the internal package import to be allowed it needs to be from a package that's within `github.com/miekg/dns`. Try moving your code to that path or symlinking it if moving it is annoying.

insomniacslk:
I already noticed that it did work from the gopath, just didn't know that it's treated differently. Thanks for the clarification!

On September 30, 2017 7:52:01 PM PDT, James Hartig <notifications@github.com> wrote:
>@insomniacslk Are you running `go build` from
>`$GOPATH/src/github.com/miekg/dns`? In order for the internal package
>import to be allowed it needs to be from a package that's within
>`github.com/miekg/dns`. Try moving your code to that path or symlinking
>it if moving it is annoying.
>
>-- 
>You are receiving this because you were mentioned.
>https://github.com/miekg/dns/pull/526#issuecomment-333349636

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

insomniacslk:
There is no simple way to test that function. The timeout is on the connection establishment, which is not easy to simulate in a Travis-CI environment.
E.g. I cannot use iptables safely because the tests run concurrently and it could affect other test results.
I cannot use a nonexisting non-routable IP either because it may give false negatives in the future.
And using an eBPF program to just drop the SYNs from this specific connection would be quite complex (but possibly this is the only reliable way).

cc @miekg what do you think?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Fix DialTimeout alw..." ]
>There is no simple way to test that function. The timeout is on the connection establishment, which is not easy to simulate in a Travis-CI environment.
>E.g. I cannot use iptables safely because the tests run concurrently and it could affect other test results.
>I cannot use a nonexisting non-routable IP either because it may give false negatives in the future.
>And using an eBPF program to just drop the SYNs from this specific connection would be quite complex (but possibly this is the only reliable way).
>
>cc @miekg what do you think?

Some things are just too difficult to test in such an environment. We have a ton 
of unit/integration tests, but as some point you can/need to give up.

We do have tests that reach out on the internet (axfr tests for instance), if 
you allow for that, would it be doable?

insomniacslk:
It's possible to make a network request towards a host:port that is configured to drop packets. I did it on insomniac.slackware.it port 5053, and created PR https://github.com/miekg/dns/pull/531 . We can use `miek.nl` if you prefer. I used my host to get this tested quickly.

fastest963:
@insomniacslk in this particular case just testing the success case would've caught this bug, spinning up a local udp and tcp server and verifying that DialTimeout connects and doesn't error would've caught this bug. To test the timeout case, can you make a `net.Listener` on local and then never call `Accept`?

insomniacslk:
@fastest963 you mean something like

```
l, err := net.Listen("tcp", "[::1]:0")
// no call to l.Accept()
if err != nil {
    ...
}
...
conn, err := DialTimeout("tcp", l.Addr().String()")
if err ...
```

This wouldn't work for this test, as the socket is listening already, so the three-way handshake would complete just fine (also double-checked with `tcpdump` to be sure). This test case is about the `connect` timeout unfortunately :(

PR #527, opened 2017-10-02
Simplify compressed length code

tmthrgd:
This is #519 by @rolandshoemaker with different code.

There is no difference in the speed of `msg.Len` with compression, but there is a noticeable speed up without compression.

I have no idea if this is worth merging or not, but I could see some obvious inefficiencies in that PR and I decided to address them.

```
name                      old time/op  new time/op  delta
MsgLength-8               1.65µs ± 6%  1.65µs ± 6%     ~     (p=0.968 n=10+9)
MsgLengthNoCompression-8   132ns ± 6%    15ns ± 3%  -88.61%  (p=0.000 n=10+10)
```

Original pull request (#519): "Removes a bunch of duplication in the msg.Len path ~and exposes the method used to calculate the compressed length of a RR given a provided compression map~."

### DIFF
--- a/msg.go
+++ b/msg.go
@@ -919,9 +919,9 @@ func compressedLen(dns *Msg, compress bool) int {
 			l += r.len()
 			compressionLenHelper(compression, r.Name)
 		}
-		l += comperssionLenSlice(compression, dns.Answer)
-		l += comperssionLenSlice(compression, dns.Ns)
-		l += comperssionLenSlice(compression, dns.Extra)
+		l += compressionLenSlice(compression, dns.Answer)
+		l += compressionLenSlice(compression, dns.Ns)
+		l += compressionLenSlice(compression, dns.Extra)
 	} else {
 		for _, r := range dns.Question {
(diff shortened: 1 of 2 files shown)


msg.go
```diff
@@ -913,67 +913,55 @@ func (dns *Msg) Len() int { return compressedLen(dns, dns.Compress) }
...
 			}
-			compressionLenHelperType(compression, dns.Ns[i])
 		}
 	}
-	for i := 0; i < len(dns.Extra); i++ {
-		if dns.Extra[i] == nil {
+	return l
+}
+
+func comperssionLenSlice(c map[string]int, rs []RR) int {
```
miekg: typo

miekg:
This is welcome.

msg.go
tmthrgd: @miekg Fixed.

rolandshoemaker:
Awesome! 👍 this is much better than my original attempt.

PR #528, opened 2017-10-03
Keep compatibility with go1.6

juniorz:
Although f282f80e243cc2bf8f6410c30d821b93b794e168 removes 1.6 from travis builds, it's not clear if the intent is to completely drop support to 1.6.

Compatibility with go1.6 can be kept by simply using `golang.org/x/net/context` rather than `context`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: go
 sudo: false
 go:
+  - 1.6.x
   - 1.7.x
   - 1.8.x
   - tip
--- a/client.go
+++ b/client.go
@@ -6,7 +6,6 @@ import (
 	"bytes"
 	"crypto/tls"
 	"encoding/binary"
-	"golang.org/x/net/context"
 	"io"
 	"net"
 	"strings"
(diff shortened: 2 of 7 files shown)


fastest963:
@juniorz The problem with changing it back to `golang.org/x/net/context` is that it would break anyone running 1.7 or 1.8 since `context` would is not equivalent to `golang.org/x/net/context`. In go 1.9 that is fixed with the introduction of type aliases. If you upgrade to go 1.9 you won't need to make your code use the new `context` since they're now considered equivalent.

juniorz:
I'm confused about how it would break since the tests worked. Are you saying the API is expected to use a `context.Context` in everything but 1.6 (where the API is expected to use `golang.org/x/net/context.Context`)?

What about using build tags to achieve this?

miekg:
We want to use the (now) std lib context package. If you want to use go 1.6 you'll have to maintain your own fork. This has come up before: AppEngine seems stuck in the past.

So we can't merge this.

Issue #529, opened 2017-10-03
Failed to unpack EDNS0_SUBNET with prefix-length=0


m13253:
A prefix of length 0 is well defined in the standard, and is necessary to indicate "a resolver that the client's address information must **not** be used when resolving this query".

Test server code:
https://gist.github.com/m13253/f354f601bb741d1c0ef903ee2104e936

Test client command line:
```
$ dig -p5533 @localhost google.com +subnet=0.0.0.0/0
```

Result:
```
; <<>> DiG 9.10.4-P5 <<>> -p5533 @localhost google.com +subnet=0.0.0.0/0
; (2 servers found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: FORMERR, id: 25579
;; flags: qr; QUERY: 0, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 0

;; WARNING: EDNS query returned status FORMERR - retry with '+noedns'

;; Query time: 0 msec
;; SERVER: ::1#5533(::1)
;; WHEN: Tue Oct 03 18:28:32 UTC 2017
;; MSG SIZE  rcvd: 12
```

m13253:
I found it actually a bug of `dig`. Not our fault.

Nonetheless, we will have to be compatible with `dig` -- That's the _de facto_ standard.

Pull request submitted in #530.

miekg:
Is there an upstream bug for this? We might want to reference that in the PR you submmited...

miekg:
Sort of dupliced in #581

PR #530, opened 2017-10-03
Fix EDNS0_SUBNET compatibility with dig

m13253:
Do not complain about AddressFamily=0, if SubnetNetmask is also 0.

Fixes issue #529


miekg:
is this still needed?

m13253:
> is this still needed?

Sorry for the delay.
I think it is okay to merge.

I will report the problem to dig next week,
because I don't have enough time these days.

PR #531, opened 2017-10-03
Added test for DialTimeout

insomniacslk:
Added test for `DialTimeout`, to address the concerns at https://github.com/miekg/dns/pull/526 .

The test currently makes a network call to a TCP port configured to drop packets, as it is not easy to simulate in an isolated manner a connection timeout on Travis-CI.

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -34,6 +34,21 @@ func TestDialUDP(t *testing.T) {
 	}
 }
 
+func TestDialTimeoutTCP(t *testing.T) {
+	m := new(Msg)
+	m.SetQuestion("miek.nl.", TypeSOA)
+
+	// insomniac.slackware.it is configured to drop connections to 5053 for this
+	// test to succeed
+	conn, err := DialTimeout("tcp", "insomniac.slackware.it:5053", time.Millisecond*100)
+	if err == nil {
+		t.Fatalf("expected timeout error, but no error returned")
+	}
+	if conn != nil {
+		t.Fatalf("expected nil value for conn, got %v instead", conn)
+	}
+}
+
 func TestClientSync(t *testing.T) {
 	HandleFunc("miek.nl.", HelloServer)
 	defer HandleRemove("miek.nl.")
--- a/client_test.go
+++ b/client_test.go
@@ -34,18 +34,23 @@ func TestDialUDP(t *testing.T) {
 	}
 }
 
-func TestDialTimeoutTCP(t *testing.T) {
+func TestDialTimeoutTCPNoTimeout(t *testing.T) {
+	// this tests the non-timeout case of a call to DialTimeout
+	s, addrstr, err := RunLocalTCPServer("[::1]:0")
+	if err != nil {
+		t.Fatalf("unable to run test server: %v", err)
+	}
+	defer s.Shutdown()
+
 	m := new(Msg)
 	m.SetQuestion("miek.nl.", TypeSOA)
 
-	// insomniac.slackware.it is configured to drop connections to 5053 for this
-	// test to succeed
-	conn, err := DialTimeout("tcp", "insomniac.slackware.it:5053", time.Millisecond*100)
-	if err == nil {
-		t.Fatalf("expected timeout error, but no error returned")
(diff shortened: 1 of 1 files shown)


client_test.go
```diff
@@ -34,6 +34,21 @@ func TestDialUDP(t *testing.T) {
 	}
 }
 
+func TestDialTimeoutTCP(t *testing.T) {
+	m := new(Msg)
+	m.SetQuestion("miek.nl.", TypeSOA)
+
+	// insomniac.slackware.it is configured to drop connections to 5053 for this
+	// test to succeed
+	conn, err := DialTimeout("tcp", "insomniac.slackware.it:5053", time.Millisecond*100)
```
miekg: How to do get to the time out?

Not to sure this test adds anything

insomniacslk: it tries to connect to that host:port that is configured with a DROP rule on netfilter, and this will trigger the connect timeout. I can clarify the comment if necessary

insomniacslk: Additionally I would configure the same DROP rule miek.nl rather than on my own host since you already use miek.nl's server for some tests. This would help keeping the tests more consistent, and leave you in control of where it connects to

miekg: yeah.. I know :) Also not perfect. We might want push this to something more official, but $$$. We should be conservative in adding remote endpoints that require fiddling and make (more) tests depend on that.

insomniacslk: I totally agree with you. On the short term: are you OK to move this test to point to a port on miek.nl configured to drop connections?

On the long term: what other provider would you see fit? And how much money would it cost? This is a very useful and used library, I am pretty sure it wouldn't be hard to raise some money or sponsorship to have a better infra (and I could contribute as well)

miekg: Doing it on miek.nl is just as bad - the only reason that I originally did it was because I saw no other option and I was less likely to leave the project.

We should drop this PR and open an issue to discuss what we need from an external test server, how we pay for it and how we provision it.

insomniacslk: Agreed. I am thinking of other options for this specific test case, and thought that if we can assign a new local IP (e.g. `127.0.0.3` or similar) and set a DROP rule on that IP:port this test can be still valid. I can update the PR and see how does Travis-CI behave

insomniacslk:
...aaand the GCE-based testing environment does not support IPv6, cannot test with sudo (to run iptables) and use `::1` to listen for tests

insomniacslk:
In other words, I cannot use both `sudo` and test IPv6 on Travis-CI :(

insomniacslk:
@miekg given that I can't easily test the timeout case, I'd rather at least test the non-timeout case as suggested by @fastest963 in https://github.com/miekg/dns/pull/526 . I will update this PR, then I'll open an issue for the testing infra as you suggested

insomniacslk:
worth merging or dropping?

client_test.go
```diff
@@ -34,6 +34,26 @@ func TestDialUDP(t *testing.T) {
...
+		t.Fatalf("unable to run test server: %v", err)
+	}
+	defer s.Shutdown()
+
+	m := new(Msg)
+	m.SetQuestion("miek.nl.", TypeSOA)
+
+	conn, err := DialTimeout("tcp", addrstr, time.Millisecond*100)
+	if err != nil {
+		t.Fatalf("failed to dial with timeout: %v", err)
```
miekg: Nit; these errors should start with a capital letter

insomniacslk: Will fix it once at home, no personal laptop with me today :(

insomniacslk:
Re. what we are testing here: only the successful queries, so basically *not* testing the timeout expiry. Unfortunately as we discussed above the current testing infra doesn't let us easily/cleanly test the negative case. But at least this positive case would have caught the bug in #526 so I feel there's still some value added

miekg:
Ack. But as such I don't think this adds anything.

insomniacslk:
So better dropping it? This would still increase coverage and test a case that we failed to catch earlier, but I'm not going to fight for it :)

PR #532, opened 2017-10-10
Allow parsing resolv.conf from io.Reader

sarneaud:
This allows projects that use this parser to write unit tests without
writing temporary files to the filesystem.

### DIFF
--- a/clientconfig_test.go
+++ b/clientconfig_test.go
@@ -1,10 +1,10 @@
 package dns
 
 import (
-	"bytes"
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 )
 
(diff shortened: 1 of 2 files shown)


clientconfig_test.go
```diff
@@ -20,14 +21,34 @@ nameserver 10.28.10.2
 nameserver 11.28.10.1` // <- NOTE: NO newline.
 
 func testConfig(t *testing.T, data string) {
+	cc, err := ClientConfigFromReader(bytes.NewBufferString(data))
```
tmthrgd: Why not use `strings.NewReader` here? It's not important but it does avoid an allocation that `bytes.NewBufferString` has to make.

sarneaud: Yep, that makes sense.

miekg:
looks sane.

PR #533, opened 2017-10-12
Go version bump in CI

insomniacslk:
Bumped Go version to 1.8 and 1.9 to keep in line with the last two releases


insomniacslk:
Abandoning this as it pulled in the changes from another branch

PR #535, opened 2017-10-13
Optimize CompareDomainName

miekg:
Optimize CompareDomainName:
old: BenchmarkCompareDomainName-2   	 1000000	      1869 ns/op	      64 B/op	       2 allocs/op
new: BenchmarkCompareDomainName-2   	 2000000	       854 ns/op	      64 B/op	       2 allocs/op

This removes the strings.ToLower and fixes the documentation. It also
does not Fqdn's the names anymore (the documentation said we didn't, now
the documentation is right again).
Unlike what the documentation said we are comparing in a ignore-case
manor, add helper function equal that does this without calling
strings.ToLower.

### DIFF
--- a/labels.go
+++ b/labels.go
@@ -1,7 +1,5 @@
 package dns
 
-import "strings"
-
 // Holds a bunch of helper functions for dealing with labels.
 
 // SplitDomainName splits a name string into it's labels.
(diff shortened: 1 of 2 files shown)


johnbelamaric:
Nice, improvement. LGTM.

andrewtj:
Is it worth documenting that escapes aren't decoded?

miekg:
How is that different from the ToLower we did?

miekg:
In the end we should probably make something like equal public and 100% right (and fast)

PR #545, opened 2017-11-03
cleaups: gofmt -w -s

miekg:
Various cleanups. Remove superfluous nil check and fmt.Errorf where it
makes sense.


miekg:
I don't understand why travis is acting this way.

Locally tests pass.

PR #546, opened 2017-11-03
server: drop graceful handling

miekg:
Drop all graceful handling. There is just too much locking in
waitgrouping going on for very little gain; deal with it.

This actually breaks a test - which I'm looking into.

### DIFF
--- a/server.go
+++ b/server.go
@@ -296,22 +296,10 @@ type Server struct {
 	DecorateReader DecorateReader
 	// DecorateWriter is optional, allows customization of the process that writes raw DNS messages.
 	DecorateWriter DecorateWriter
-
-	// Graceful shutdown handling
-
-	inFlight sync.WaitGroup
-
-	lock    sync.RWMutex
-	started bool
 }
 
 // ListenAndServe starts a nameserver on the configured address in *Server.
 func (srv *Server) ListenAndServe() error {
-	srv.lock.Lock()
-	defer srv.lock.Unlock()
-	if srv.started {
-		return &Error{err: "server already started"}
-	}
 	addr := srv.Addr
 	if addr == "" {
 		addr = ":domain"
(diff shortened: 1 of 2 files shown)


miekg:
cc @oif

miekg:
go t -v -run=TestTimeout 
=== RUN   TestTimeout
--- PASS: TestTimeout (0.00s)
PASS
ok  	github.com/miekg/dns	1.016s

miekg:
% go t 
--- FAIL: TestTimeout (0.05s)
	client_test.go:527: exchange took longer (50.521853ms) than specified Timeout (11ms)

uhm?

server.go
```diff
@@ -520,24 +458,15 @@ func (srv *Server) serveUDP(l *net.UDPConn) error {
...
 		m, s, err := reader.ReadUDP(l, rtimeout)
-		srv.lock.RLock()
-		if !srv.started {
-			srv.lock.RUnlock()
-			return nil
-		}
-		srv.lock.RUnlock()
 		if err != nil {
-			continue
+			return err
```
miekg: Is a closed listener the only way that this will err? cc @andrewtj @asergeyev

miekg:
This is big change - backward compatible but rather big. cc @andrewtj @asergeyev

server.go
andrewtj: Perhaps some resource running out might bubble up?

miekg: I wondered about that too. The TCP listener has  this:
~~~
	for {
		rw, err := l.Accept()
		if err != nil {
			if neterr, ok := err.(net.Error); ok && neterr.Temporary() {
				continue
			}
			return err
		}
~~~

miekg:
fyi: for anyone reading along: I'm not against graceful stopping a server; but it should be very light weight *and* it is probably only useful for TCP conns.

miekg:
Going in.

miekg:
This exposes an old(er) race: https://travis-ci.org/coredns/coredns/jobs/299780481
wondering to revert this one, or forward fix this branch. I.e. reverting also has a race (or maybe it is that old race? Don't know yet)

PR #547, opened 2017-11-03
Test: remove all Logf/Log

miekg:
Move some of them to Errorf and friends, but most of them are just
gone: This make go test -v actually readable.


miekg:
And whole bunch of tests removed.

miekg:
cc @andrewtj @asergeyev 
(should probably measure code coverage at some point)

miekg:
Pushing this, because I want to push some golint and go vet cleanups as well (and passing tests would be nice)

PR #556, opened 2017-11-08
Tests updates

miekg:
Use :0 for loopback testing. This is more portable between testing environments.
Add testRR that calls NewRR and throws error away - apply it everywhere where needed.


miekg:
1.8 breaks, 1.9 works - restarting a 1.8 job - if this is a 1.8 thing I'm inclined to drop that on the spot.

Issue #555, opened 2017-11-08
dnsutil.TrimDomainName tests fail


miekg:
These are not covered by travis, and surprise: they fail.

miekg:
~~~
--- FAIL: TestTrimDomainName (0.00s)
	util_test.go:56: 1 TrimDomainName("foo.example.com.", "example.com"): expected foo, got foo.example.com.
	util_test.go:56: 3 TrimDomainName(".foo.example.com.", "example.com"): expected .foo, got .foo.example.com.
	util_test.go:56: 6 TrimDomainName("example.com.", "example.com"): expected @, got example.com.
	util_test.go:56: 0 TrimDomainName("foo.example.com", "example.com."): expected foo, got foo.example.com
	util_test.go:56: 2 TrimDomainName(".foo.example.com", "example.com."): expected .foo, got .foo.example.com
	util_test.go:56: 4 TrimDomainName("*.example.com", "example.com."): expected *, got *.example.com
	util_test.go:56: 5 TrimDomainName("example.com", "example.com."): expected @, got example.com
	util_test.go:107: 2 TrimDomainName("a.b.c.d.e.f.", "f"): expected a.b.c.d.e, got a.b.c.d.e.f.
	util_test.go:107: 3 TrimDomainName("b.c.d.e.f.", "f"): expected b.c.d.e, got b.c.d.e.f.
	util_test.go:107: 4 TrimDomainName("c.d.e.f.", "f"): expected c.d.e, got c.d.e.f.
	util_test.go:107: 5 TrimDomainName("d.e.f.", "f"): expected d.e, got d.e.f.
	util_test.go:107: 6 TrimDomainName("e.f.", "f"): expected e, got e.f.
	util_test.go:107: 7 TrimDomainName("f.", "f"): expected @, got f.
	util_test.go:107: 8 TrimDomainName(".a.b.c.d.e.f.", "f"): expected .a.b.c.d.e, got .a.b.c.d.e.f.
	util_test.go:107: 9 TrimDomainName(".b.c.d.e.f.", "f"): expected .b.c.d.e, got .b.c.d.e.f.
	util_test.go:107: 10 TrimDomainName(".c.d.e.f.", "f"): expected .c.d.e, got .c.d.e.f.
	util_test.go:107: 11 TrimDomainName(".d.e.f.", "f"): expected .d.e, got .d.e.f.
	util_test.go:107: 12 TrimDomainName(".e.f.", "f"): expected .e, got .e.f.
[...]

PR #559, opened 2017-11-08
util.TrimDomainName() fails when origin doesn't end in dot

tlimoncelli:
Fixes https://github.com/miekg/dns/issues/555

### DIFF
--- a/dnsutil/util.go
+++ b/dnsutil/util.go
@@ -57,8 +57,16 @@ func TrimDomainName(s, origin string) string {
 		return strings.TrimSuffix(s, origin)
 	}
 
+	//original_origin := origin
+	if !strings.HasSuffix(origin, ".") {
+		origin = origin + "."
+	}
+	original_s := s
+	if !strings.HasSuffix(s, ".") {
+		s = s + "."
+	}
 	if !dns.IsSubDomain(origin, s) {
-		return s
+		return original_s
 	}
 
 	slabels := dns.Split(s)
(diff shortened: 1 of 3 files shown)


miekg:
Thanks!

tlimoncelli:
I made a couple more small changes.  PTAL

dnsutil/util.go
```diff
@@ -36,7 +38,7 @@ func AddOrigin(s, origin string) string {
 	}
 
 	if origin == "." {
-		return s + origin // AddOrigin(s, ".") is an expensive way to add a ".".
+		return s + origin // AddOrigin(s, ".") is an easy way to add a "." without creating "..".
```
miekg: Just drop this comment. Doesn't add anything imo

dnsutil/util.go
```diff
@@ -57,8 +59,12 @@ func TrimDomainName(s, origin string) string {
 		return strings.TrimSuffix(s, origin)
 	}
 
+	original_s := s
+	s = AddOrigin(s, ".")
```
miekg: you prolly can't upcall to dns.Fqdn() which is basically what this does?

dnsutil/util.go
```diff
@@ -57,8 +59,12 @@ func TrimDomainName(s, origin string) string {
 		return strings.TrimSuffix(s, origin)
 	}
 
+	original_s := s
```
miekg: Don't use underscores in Go names.

miekg:
Can you copy the .travis.yml changes from https://github.com/miekg/dns/pull/558/files ?

dnsutil/util.go
tlimoncelli: Done

dnsutil/util.go
tlimoncelli: Done

PR #561, opened 2017-11-10
Server: drop inflight waitgroup

miekg:
This drops the waitgroup in Server, the suspicion is this can make the server
fail to stop; doing this make graceful shutdown not work.

Add test that tries to find a race between starting on stopping race;
there was a data race on srv.Inflight.

The coredns' TestReadme doesn't race anymore with this as it did with
the more evasive PR #546.


miekg:
Upping test coverage in #562

PR #563, opened 2017-11-10
Net tests

miekg:


miekg:
@andrewtj we've never travis tested the xfr_test.go tests's and now they fail (fail locally as well). Can you take a look? The sidn frame is still up it seems.

andrewtj:
I think they may have removed bind10.sidnlabs.nl since it can't be resolved. PowerDNS and Yadifa don't seem to be AXFR enabled anymore. I have no IPv6 connectivity at the minute and spotty IPv4 connectivity to their servers which has made it a little frustrating to work out what's up. I don't think Travis has IPv6 support so I guess that'll need to be worked around too.

miekg:
Thanks. Yes, there is no ipv6 in travis :-( Question becomes if we depend so much on this external infra, should we continue with all these tests?

andrewtj:
There's some value in testing interoperability but the balance probably isn't right here. The only implementation specific interoperability issue I can recall is the Knot compressed root issue, and these tests wouldn't have caught that.

Probably culling the tests down to a single BIND install (because it supports everything and is common as muck) is the sweet spot. In my Rust crate I have a test helper that checks if BIND and its tools are available and if so sets up an instance in a temporary directory with a signed zone, TSIG/SIG0, etc. Something like that is probably a better fit.

miekg:
I'm inclined to remove (all lot of) test code in favor of a solution like #612; however I don't want to merge #612 into miekg/dns proper, but have this a seperate integration (ci) repo.

PR #565, opened 2017-11-14
Use workers instead spawning goroutines for each incoming DNS request

UladzimirTrehubenka:
There are two major issues:
unlimited using resources - performance test shows that on high load CoreDNS silently crashed;
performance drop due much time is spent on management goroutines instead serve DNS requests.

### DIFF
--- a/server.go
+++ b/server.go
@@ -262,6 +262,14 @@ type DecorateReader func(Reader) Reader
 // Implementations should never return a nil Writer.
 type DecorateWriter func(Writer) Writer
 
+// A message defines struct that is passed to worker
+type message struct {
+	m []byte
+	u *net.UDPConn
+	s *SessionUDP
+	t net.Conn
+}
+
 // A Server defines parameters for running an DNS server.
 type Server struct {
 	// Address to listen on, ":dns" if empty.
@@ -296,10 +304,24 @@ type Server struct {
 	DecorateReader DecorateReader
 	// DecorateWriter is optional, allows customization of the process that writes raw DNS messages.
 	DecorateWriter DecorateWriter
-
+	// Messages queue
+	queue chan *message
+	// Number of workers, if set to zero - use spawn goroutines instead
+	Workers int
 	// Shutdown handling
 	lock    sync.RWMutex
 	started bool
+}
+
+func (srv *Server) startWorkers() {
+	srv.queue = make(chan *message)
+	for i := 0; i < srv.Workers; i++ {
+		go func() {
+			for msg := range srv.queue {
+				srv.serve(msg)
+			}
+		}()
+	}
 }
 
 // ListenAndServe starts a nameserver on the configured address in *Server.
@@ -309,6 +331,15 @@ func (srv *Server) ListenAndServe() error {
 	if srv.started {
 		return &Error{err: "server already started"}
 	}
+
+	if srv.Handler == nil {
+		srv.Handler = DefaultServeMux
+	}
+
+	if srv.Workers > 0 {
+		srv.startWorkers()
+	}
+
 	addr := srv.Addr
 	if addr == "" {
 		addr = ":domain"
@@ -380,6 +411,15 @@ func (srv *Server) ActivateAndServe() error {
 	if srv.started {
 		return &Error{err: "server already started"}
 	}
+
+	if srv.Handler == nil {
+		srv.Handler = DefaultServeMux
+	}
+
+	if srv.Workers > 0 {
+		srv.startWorkers()
+	}
+
 	pConn := srv.PacketConn
 	l := srv.Listener
 	if pConn != nil {
@@ -420,6 +460,10 @@ func (srv *Server) Shutdown() error {
 	srv.started = false
 	srv.lock.Unlock()
 
+	if srv.Workers > 0 {
+		close(srv.queue)
+	}
+
 	if srv.PacketConn != nil {
 		srv.PacketConn.Close()
 	}
@@ -439,7 +483,6 @@ func (srv *Server) getReadTimeout() time.Duration {
 }
 
 // serveTCP starts a TCP listener for the server.
-// Each request is handled in a separate goroutine.
 func (srv *Server) serveTCP(l net.Listener) error {
 	defer l.Close()
 
@@ -452,10 +495,6 @@ func (srv *Server) serveTCP(l net.Listener) error {
 		reader = srv.DecorateReader(reader)
 	}
 
-	handler := srv.Handler
-	if handler == nil {
-		handler = DefaultServeMux
-	}
 	rtimeout := srv.getReadTimeout()
 	// deadline is not used here
 	for {
@@ -476,12 +515,16 @@ func (srv *Server) serveTCP(l net.Listener) error {
 		if err != nil {
 			continue
 		}
-		go srv.serve(rw.RemoteAddr(), handler, m, nil, nil, rw)
+		msg := &message{m, nil, nil, rw}
+		if srv.Workers > 0 {
+			srv.queue <- msg
+		} else {
+			go srv.serve(msg)
+		}
 	}
 }
 
 // serveUDP starts a UDP listener for the server.
-// Each request is handled in a separate goroutine.
 func (srv *Server) serveUDP(l *net.UDPConn) error {
 	defer l.Close()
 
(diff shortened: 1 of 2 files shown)


miekg:
Why?

server.go
```diff
@@ -12,8 +12,14 @@ import (
...
 
-// Maximum number of TCP queries before we close the socket.
-const maxTCPQueries = 128
+const (
+	// Maximum number of TCP queries before we close the socket.
+	maxTCPQueries = 128
+	// Maximum number of incoming DNS messages in queue.
+	maxQueueSize = 1000000
+	// Maximum number of workers.
+	maxWorkers = 100
```
miekg: Why a 100?

server.go
```diff
@@ -12,8 +12,14 @@ import (
 	"time"
 )
 
-// Maximum number of TCP queries before we close the socket.
-const maxTCPQueries = 128
+const (
+	// Maximum number of TCP queries before we close the socket.
+	maxTCPQueries = 128
+	// Maximum number of incoming DNS messages in queue.
+	maxQueueSize = 1000000
```
miekg: Why 1000000?

server.go
UladzimirTrehubenka: Observed that 100 is optimal value, <100 and >100 drops performance.

UladzimirTrehubenka: But probably we need more tests.

server.go
UladzimirTrehubenka: CoreDNS cannot handle 1M requests - so this queue size is not affected anything (prevents bottleneck).

miekg:
Couple a things:
* If you're proposing large changes, it's best to open a bug first
* Nor the comment, nor the PR has a decent description of what and why you are doing
* If this is performance it should add a benchmark test or show some improvements
* A buffered channel with a random number of elements in it (the other place this happens: https://github.com/miekg/dns/blob/master/scan.go#L171 should be removed) does not work
* A single benchmark to fix the numbers to N does not say anything. How about ARM, s390, i386?
* What happens when you hit the 100 goroutines?
* What happens when the buffer fills up?
* How much faster is this anyway?

miekg:
What do other think of this?

The speed is nice. Is the contant of '100' a problem?

johnbelamaric:
I asked @UladzimirTrehubenka to put some more details (like those in the email) onto this PR or onto an issue, I think others will need to see that to weigh in.

I think to know if the constant is a problem probably requires more empirical tests on other platforms. But if this feature is disabled with workers == 0, then the cost is low (well, if you consider maintaining two different paths low cost).

Silent crashes are bad...I like that it fixes that (it does, right?).

miekg:
> Silent crashes are bad...I like that it fixes that (it does, right?).

Good point, somehow I tunnel visioned on the constant.

UladzimirTrehubenka:
Actually this PR breaks nothing. By default Workers and QueueSize set to zero. These params are set during server object initialization (e.g. on CoreDNS side). Finally on AWS cluster with using handler that returns random A record for any request I got following numbers (dnsperf against test binary over network):
workers 0, size 0 - 92K;
workers 100, size 0 - 95K;
workers 100, size 1000000 - 100K.
BTW the project shows (for 100 workers) 38K for size=0 vs 47K for size=1000000.
For erratic handler on my local machine I got (for local testing doesn't matter what is the queue size):
workers 0, size 0 - 208K;
workers 50, size 0 - 209K;
workers 100, size 0 - 250K;
workers 200, size 0 - 246K;
workers 1000, size 0 - 234K.

server.go
```diff
@@ -296,19 +305,53 @@ type Server struct {
...
+		}()
+	}
 }
 
 // ListenAndServe starts a nameserver on the configured address in *Server.
 func (srv *Server) ListenAndServe() error {
-	srv.lock.Lock()
-	defer srv.lock.Unlock()
-	if srv.started {
+	if srv.isRunning() {
```
miekg: Is this need in this PR? (Not opposing the change - but it seems to do the same thing as earlier code, or is there a bug fixed?)

[it does clear out of awful locking we had in server]

Can you make this a seperate PR?

server.go
```diff
@@ -296,19 +305,53 @@ type Server struct {
...
 		return &Error{err: "server already started"}
 	}
+
+	if srv.Handler == nil {
+		srv.Handler = DefaultServeMux
+	}
```
miekg: Why is this on needed?

*scrolls down*

Ah you're pulling it out from the serveX functions; sensible change; can you make that also a new PR?

server.go
```diff
@@ -439,8 +473,8 @@ func (srv *Server) getReadTimeout() time.Duration {
 }
 
 // serveTCP starts a TCP listener for the server.
-// Each request is handled in a separate goroutine.
 func (srv *Server) serveTCP(l net.Listener) error {
+	srv.start()
```
miekg: why is `start()` need again here?

server.go
```diff
@@ -494,94 +524,87 @@ func (srv *Server) serveUDP(l *net.UDPConn) error {
...
-	w := &response{tsigSecret: srv.TsigSecret, udp: u, tcp: t, remoteAddr: a, udpSession: s}
+func (srv *Server) serve(in *message) {
+	var a net.Addr
+	var reader Reader
+	var idleTimeout time.Duration
+
+	if in.s != nil {
+		a = in.s.RemoteAddr()
+	} else if in.t != nil {
+		a = in.t.RemoteAddr()
```
miekg:
```
if in.s != nil {
    a = in.s.RemoteAddr()
}
if in.t != nil {
    a = in.t.RemoteAddr()
}
```
All the reader stuff should apply for both cases and can be outdented

server.go
```diff
@@ -494,94 +524,87 @@ func (srv *Server) serveUDP(l *net.UDPConn) error {
...
-	w.tsigStatus = nil
-	if w.tsigSecret != nil {
-		if t := req.IsTsig(); t != nil {
-			secret := t.Hdr.Name
-			if _, ok := w.tsigSecret[secret]; !ok {
-				w.tsigStatus = ErrKeyAlg
+	for q := 0; q < maxTCPQueries; q++ { // TODO(miek): make this number configurable?
+		req := new(Msg)
+		err := req.Unpack(in.m)
+		if err != nil { // Send a FormatError back
```
miekg: I'm not convinced the if-elseif-else is better than the goto we had.

server.go
```diff
@@ -296,19 +305,53 @@ type Server struct {
...
 	// DecorateWriter is optional, allows customization of the process that writes raw DNS messages.
 	DecorateWriter DecorateWriter
+	// Graceful shutdown handling
+	running int32
+	// Messages queue
+	queue chan *message
+	// Number of workers, if set to zero - use spawn goroutines instead
+	Workers int
+	// Size of DNS requests queue
+	QueueSize int
```
miekg: the #workers is one thing, but the QueueSize is quite another, if this doesn't perform better with QueueSize == 0 it's not worth adding. A buffered channel leads to weird "sometimes it is slow - or blocking" errors in prod when the thing finally fills up.

server.go
UladzimirTrehubenka: serveUDP() and serveTCP() have:
```
	handler := srv.Handler
	if handler == nil {
		handler = DefaultServeMux
	}
        ...
        go srv.serve(s.RemoteAddr(), handler, ...)
```
Why do we need to set handler each time on serveUDP() or serveTCP() call and then pass handler into serve() if we can set srv.Handler only once and serve() can just use srv.Handler?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Use workers instead..." ]
>UladzimirTrehubenka commented on this pull request.
>
>
>
>>  		return &Error{err: "server already started"}
> 	}
>+
>+	if srv.Handler == nil {
>+		srv.Handler = DefaultServeMux
>+	}
>
>serveUDP() and serveTCP() have:
>```
>	handler := srv.Handler
>	if handler == nil {
>		handler = DefaultServeMux
>	}
>        ...
>        go srv.serve(s.RemoteAddr(), handler, ...)
>```
>Why do we need to set handler each time on serveUDP() or serveTCP() call and then pass handler into serve() if we can set srv.Handler only once and serve() can just use srv.Handler?

Good question, it does look a bit odd doing this in the serving path.. And no 
comments (yeah!) on why this makes sense.

Can't think of a good reason right now.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Use workers instead..." ]
>Why do we need to set handler each time on serveUDP() or serveTCP() call and then pass handler into serve() if we can set srv.Handler only once and serve() can just use srv.Handler?

A quick hack to remove this crashes with 'go test'

UladzimirTrehubenka:
This is not enough just remove this - need to change serve() to use srv.Handler instead passed handler and set srv.Handler to DefaultServeMux (if handler is empty) in serveUDP() and serveTCP() as in the PR.

UladzimirTrehubenka:
BTW PR passed all UT and don't change default behavior.

PR #567, opened 2017-11-14
Add support for TKEY RRs

spsholleman:
- make sure Key and Data fields are variable length hex fields
- checkin output from 'go generate'
- add a TKEY specific test to ensure this stays working

### DIFF
--- a/dns_test.go
+++ b/dns_test.go
@@ -410,28 +410,29 @@ func TestMsgPackBuffer(t *testing.T) {
 
 // Make sure we can decode a TKEY packet from the string, modify the RR, and then pack it again.
 func TestTKEY(t *testing.T) {
-	// An example tkey RR captured
+	// An example TKEY RR captured.  There is no known accepted standard text format for a TKEY
+	// record so we do this from a hex string instead of from a text readable string.
 	tkeyStr := "0737362d6d732d370932322d3332633233332463303439663961662d633065612d313165372d363839362d6463333937396666656666640000f900ff0000000000d2086773732d747369670059fd01f359fe53730003000000b8a181b53081b2a0030a0100a10b06092a864882f712010202a2819d04819a60819706092a864886f71201020202006f8187308184a003020105a10302010fa2783076a003020112a26f046db29b1b1d2625da3b20b49dafef930dd1e9aad335e1c5f45dcd95e0005d67a1100f3e573d70506659dbed064553f1ab890f68f65ae10def0dad5b423b39f240ebe666f2886c5fe03819692d29182bbed87b83e1f9d16b7334ec16a3c4fc5ad4a990088e0be43f0c6957916f5fe60000"
 	tkeyBytes, err := hex.DecodeString(tkeyStr)
 	if err != nil {
-		t.Fatal("Unable to decode TKEY string ", err)
+		t.Fatal("unable to decode TKEY string ", err)
 	}
 	// Decode the RR
 	rr, tkeyLen, unPackErr := UnpackRR(tkeyBytes, 0)
 	if unPackErr != nil {
-		t.Fatal("Unable to decode TKEY RR", unPackErr)
+		t.Fatal("unable to decode TKEY RR", unPackErr)
 	}
 	// make space for it with some fudge room
 	msg := make([]byte, tkeyLen+1000)
 	offset, packErr := PackRR(rr, msg, 0, nil, false)
 	if packErr != nil {
-		t.Fatal("Unable to pack TKEY RR", packErr)
+		t.Fatal("unable to pack TKEY RR", packErr)
 	}
 	if offset != len(tkeyBytes) {
-		t.Fatalf("Mismatched TKEY RR size %d != %d", len(tkeyBytes), offset)
+		t.Fatalf("mismatched TKEY RR size %d != %d", len(tkeyBytes), offset)
 	}
 	if bytes.Compare(tkeyBytes, msg[0:offset]) != 0 {
-		t.Fatal("Mismatched TKEY data after rewriting bytes")
+		t.Fatal("mismatched TKEY data after rewriting bytes")
 	}
 	// Now add some bytes to this and make sure we can encode OtherData properly
 	tkey := rr.(*TKEY)
@@ -439,9 +440,9 @@ func TestTKEY(t *testing.T) {
 	tkey.OtherLen = 2
 	offset, packErr = PackRR(tkey, msg, 0, nil, false)
 	if packErr != nil {
-		t.Fatal("Unable to pack TKEY RR after modification", packErr)
+		t.Fatal("unable to pack TKEY RR after modification", packErr)
 	}
 	if offset != (len(tkeyBytes) + 2) {
-		t.Fatalf("Mismatched TKEY RR size %d != %d", offset, len(tkeyBytes)+2)
+		t.Fatalf("mismatched TKEY RR size %d != %d", offset, len(tkeyBytes)+2)
 	}
 }
--- a/dns_test.go
+++ b/dns_test.go
@@ -1,10 +1,10 @@
 package dns
 
 import (
+	"bytes"
 	"encoding/hex"
 	"net"
 	"testing"
-	"bytes"
 )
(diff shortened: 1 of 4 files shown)


dns_test.go
```diff
@@ -405,3 +407,41 @@ func TestMsgPackBuffer(t *testing.T) {
 		}
 	}
 }
+
+// Make sure we can decode a TKEY packet from the string, modify the RR, and then pack it again.
+func TestTKEY (t *testing.T) {
```
miekg: Thanks.

This needs gofmt

dns_test.go
```diff
@@ -405,3 +407,41 @@ func TestMsgPackBuffer(t *testing.T) {
...
+	// An example tkey RR captured
+	tkeyStr := "0737362d6d732d370932322d3332633233332463303439663961662d633065612d313165372d363839362d6463333937396666656666640000f900ff0000000000d2086773732d747369670059fd01f359fe53730003000000b8a181b53081b2a0030a0100a10b06092a864882f712010202a2819d04819a60819706092a864886f71201020202006f8187308184a003020105a10302010fa2783076a003020112a26f046db29b1b1d2625da3b20b49dafef930dd1e9aad335e1c5f45dcd95e0005d67a1100f3e573d70506659dbed064553f1ab890f68f65ae10def0dad5b423b39f240ebe666f2886c5fe03819692d29182bbed87b83e1f9d16b7334ec16a3c4fc5ad4a990088e0be43f0c6957916f5fe60000"
```
miekg: Can't recall is there is some kind of text presentation format for TKEY? Might be worth adding a comment.

dns_test.go
```diff
@@ -405,3 +407,41 @@ func TestMsgPackBuffer(t *testing.T) {
...
+	tkeyBytes, err := hex.DecodeString(tkeyStr)
+	if err != nil {
+		t.Fatal("Unable to decode TKEY string ", err)
```
miekg: All these errors need to start with lowercase.

types.go
```diff
@@ -1012,14 +1012,14 @@ type TKEY struct {
...
+	Key        string `dns:"size-hex:KeySize"`
 	OtherLen   uint16
-	OtherData  string
+	OtherData  string `dns:"size-hex:OtherLen"`
 }
 
 func (rr *TKEY) String() string {
 	// It has no presentation format
-	return ""
+	return fmt.Sprintf("%s Algo: %s KeyLen: %d OtherLen: %d", rr.Hdr.String(), rr.Algorithm, rr.KeySize, rr.OtherLen)
```
miekg: If we going to output this, we should
a) make sure it is a standard
b) make sure we can parse it (from text)

andrewtj:
What'd be involved in setting up an ephemeral test environment for TKEY transactions? Would it be possible to generate BIND configs etc in a temp dir and run the appropriate daemons on non-privileged ports?

miekg:
https://github.com/coredns/coredns/blob/master/.travis.yml
sets up an etcd (although we're moving that out), but seems doable.

dns_test.go
spsholleman: whoops, my bad - new project was setup right in my IDE.  will fix.

dns_test.go
spsholleman: Okay, can fix that.  curious though and want to understand.   I get it for errors returned back from the package but why does it matter here in the unit test?

miekg: style

(a grep shows there are style a few offenders)

dns_test.go
spsholleman: I can't find a standard accepted format for this, it is a fairly ephemeral RR.  More on this below.

dns_test.go
spsholleman: okay, updated

types.go
spsholleman: I added it for just being able to dump and debug it.  I can't find a well-known or standard format, if someone knows of one please point me to it.  I think that leaves us with three choices

1. Have no text representation (not a fan of that one because String() is useful for debug)
2. Come up with our own format and add the parser

Let me know what's preferred - I could try to add the parser.  Note that the current TSIG implementation is similar to this one.  There is a string representation but no parser.  The existing parser chokes on the TSIG string output when trying to parse it as 3597.

types.go
```diff
@@ -1012,14 +1012,17 @@ type TKEY struct {
...
 	KeySize    uint16
-	Key        string
+	Key        string `dns:"size-hex:KeySize"`
 	OtherLen   uint16
-	OtherData  string
+	OtherData  string `dns:"size-hex:OtherLen"`
 }
 
 func (rr *TKEY) String() string {
 	// It has no presentation format
```
miekg: This comment is now not true anymore; we just made one up. What does dig show for TKEY?
As said if we make a format we should also be able to parse it back in *and* it needs to be interoperable with other implementations.

I don't think we should output anything here...

andrewtj:
What's actually needed for a TKEY environment though? BIND and some sort of Kerberos server? I might have a go at adding a test environment as a part of replacing the SIDN tests (see #563) if someone can point me to a tutorial or write a summary.

On the text presentation of TKEY, for TSIG and OPT we emit a comment so we should do the same here imo.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Add support for TKE..." ]
>What's actually needed for a TKEY environment though? BIND and some sort of Kerberos server? I might have a go at adding a test environment as a part of replacing the SIDN tests (see #563) if someone can point me to a tutorial or write a summary.

Not sure, never used TKEY in an operational setting...

>On the text presentation of TKEY, for TSIG and OPT we emit a comment so we should do the same here imo.

Ack. Do we know what dig outputs for TKEY? I like to follow that lead (can look 
at the source)

bodgit:
> What's actually needed for a TKEY environment though? BIND and some sort of Kerberos server? I might have a go at adding a test environment as a part of replacing the SIDN tests (see #563) if someone can point me to a tutorial or write a summary.

Yes, pretty much. Either that or a Windows AD environment. Also a client that talks to both Kerberos and DNS if necessary. Setting up BIND seems to be pretty much covered here: http://jpmens.net/2012/06/29/dynamic-dns-updates-using-gss-tsig-and-kerberos/

For doing the actual updating, I don't actually ever see the TKEY records really, it's all under the hood when using `nsupdate`. How I've manually got it working is something like the following:
```
# ktutil
ktutil:  addent -password -p user@EXAMPLE.COM -k 1 -e aes256-cts-hmac-sha1-96
Password for user@EXAMPLE.COM:
ktutil:  wkt nsupdate.keytab
ktutil:  quit
# export KRB5CCNAME=/tmp/nsupdate.cc
# kinit -F -k -t nsupdate.keytab user@EXAMPLE.COM
# nsupdate -g
> server example.com
> realm EXAMPLE.COM
> zone example.com
> update add test.example.com 30 cname something.example.com.
> send
> answer
Answer:
;; ->>HEADER<<- opcode: UPDATE, status: NOERROR, id:   9850
;; flags: qr; ZONE: 1, PREREQ: 0, UPDATE: 1, ADDITIONAL: 1
;; ZONE SECTION:
;example.com.             IN      SOA

;; UPDATE SECTION:
test.example.com. 30      IN      CNAME   something.example.com.

;; TSIG PSEUDOSECTION:
1233807667.sig-hostname.example.com. 0 ANY TSIG gss-tsig. 1509978324 36000 28 FEEDDEADBEEF== 9850 NOERROR 0
```
If you use tcpdump and sniff TCP & UDP port 53 traffic you can then see the TKEY RR's in flight.

types.go
miekg: I can live with what do for TSIG's String():
~~~
	s := "\n;; TSIG PSEUDOSECTION:\n"
	s += rr.Hdr.String() +
		" " + rr.Algorithm +
~~~
And make it TKEY PSEUDOSECTION or something

Now need to parse from text, don't do that for the (wireonly) TSIG either

spsholleman: updated to be similar to TSIG - the parsing from text still somewhat worked so I left it in the newly added test as is right now.

PR #569, opened 2017-11-16
Fix go-routine leak

### LINKED ISSUE #586: Ndots = 1
Hi within my docker swarm cluster the resolv.conf is configured using options ndots0. It's required for service discovery. Within your lib you set ndots always to 1. As such I can't use the library for name resolution, it's also used by prometheus and as such it's also not usable in the cluster. :(
Images in docker swarm mode needs this setting to discover the services. When set to > 0 then the resolution didn't work.

miekg:
When and higher level (grammar or syntax) error was encountered when
parsing records from strings the routine would return, but the openend
channel would be left open and no other error would be triggered; this
means the goroutine starting z.lexer would be left hanging.

The fix is a bit of a hack; as we signal the upper error back down via
s.e and the trigger an "error" in the lexer, this at least ensure the
z.lexer function return and cleans up the goroutine.

Fixes #586
Fixes https://github.com/coredns/coredns/issues/1233

### DIFF
--- a/scan.go
+++ b/scan.go
@@ -174,15 +174,19 @@ func parseZoneHelper(r io.Reader, origin string, defttl *ttlState, file string, 
 }
 
 func parseZone(r io.Reader, origin string, defttl *ttlState, f string, t chan *Token, include int) {
+	s := scanInit(r)
+	c := make(chan lex)
+
 	defer func() {
 		if include == 0 {
 			close(t)
 		}
+		s.setErr(true)
 	}()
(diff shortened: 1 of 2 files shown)


miekg:
I'm struggling to come up with a meaningful test.

miekg:
testing the binary with the PR. This does not seem to have fixed it...

miekg:
which makes sense because it is even higher where I need to signal this.

PR #574, opened 2017-11-17
TSIG name must be presented in canonical form

matje:
Update the documentation to make clear that the zonename in the TsigSecret map must be in canonical form.

Personally I think the library has to do the canonicalization from string to wire format itself but having the documentation to be clear seems like a good idea in the first place.

### DIFF
--- a/client.go
+++ b/client.go
@@ -20,7 +20,7 @@ const tcpIdleTimeout time.Duration = 8 * time.Second
 type Conn struct {
 	net.Conn                         // a net.Conn holding the connection
 	UDPSize        uint16            // minimum receive buffer for UDP messages
-	TsigSecret     map[string]string // secret(s) for Tsig map[<zonename>]<base64 secret>, zonename must be in canonical form (lowercase, fqdn)
+	TsigSecret     map[string]string // secret(s) for Tsig map[<zonename>]<base64 secret>, zonename must be in canonical form (lowercase, fqdn, see RFC 4034 Section 6.2)
 	rtt            time.Duration
 	t              time.Time
 	tsigRequestMAC string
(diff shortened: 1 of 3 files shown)


client.go
```diff
@@ -20,7 +20,7 @@ const tcpIdleTimeout time.Duration = 8 * time.Second
 type Conn struct {
 	net.Conn                         // a net.Conn holding the connection
 	UDPSize        uint16            // minimum receive buffer for UDP messages
-	TsigSecret     map[string]string // secret(s) for Tsig map[<zonename>]<base64 secret>, zonename must be fully qualified
+	TsigSecret     map[string]string // secret(s) for Tsig map[<zonename>]<base64 secret>, zonename must be in canonical form (lowercase, fqdn)
```
miekg: I would reference the RFC defining it in all these.

miekg:
We should take the oppertunity to update client.go and make ErrSecret return the thing it was looking for as well.

matje:
Will do after lunch

matje:
My initial thought is to change the return from:
```
    return m, ErrSecret
```
to
```
    return m, fmt.Errorf("tsig name %s: %s", t.Hdr.Name, ErrSecret)
```
but that makes it no longer possible to allow the caller check for `err == dns.ErrSecret`.

Is there any better way of doing this?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] TSIG name must be p..." ]
>My initial thought is to change the return from:
>```
>    return m, ErrSecret
>```
>to
>```
>    return m, fmt.Errorf("tsig name %s: %s", t.Hdr.Name, ErrSecret)
>```
>but that makes it no longer possible to allow the caller check for `err == dns.ErrSecret`.
>
>Is there any better way of doing this?

Hmmm, yeah, we did export that error, so it's probably used like that, then 
let's *not* do this :(

client.go
matje: Done

Issue #575, opened 2017-11-17
travis/build fails: test ran too long


miekg:
from: https://travis-ci.org/miekg/dns/jobs/303528710

obvious suspect: #570 

~~~
 go get -t -v ./...
github.com/miekg/dns/internal/socket
github.com/miekg/dns
github.com/miekg/dns/dnsutil
github.com/miekg/dns/idn
615.49s$ go test $TESTS
=== RUN   TestDialUDP
--- PASS: TestDialUDP (0.00s)
=== RUN   TestClientSync
--- PASS: TestClientSync (0.00s)
=== RUN   TestClientLocalAddress
--- PASS: TestClientLocalAddress (0.00s)
=== RUN   TestClientTLSSyncV4
--- PASS: TestClientTLSSyncV4 (0.04s)
=== RUN   TestClientSyncBadID
--- PASS: TestClientSyncBadID (0.00s)
=== RUN   TestClientEDNS0
--- PASS: TestClientEDNS0 (0.00s)
=== RUN   TestClientEDNS0Local
--- PASS: TestClientEDNS0Local (0.00s)
=== RUN   TestClientConn
--- PASS: TestClientConn (0.00s)
=== RUN   TestTruncatedMsg
--- PASS: TestTruncatedMsg (0.00s)
=== RUN   TestTimeout
--- PASS: TestTimeout (0.00s)
=== RUN   TestConcurrentExchanges
SIGQUIT: quit
PC=0x488631 m=0 sigcode=0
goroutine 0 [idle]:
runtime.futex(0xb5bf40, 0x0, 0x0, 0x0, 0x0, 0x0, 0x45fe42, 0x0, 0x0, 0x7ffdf5151f00, ...)
	/home/travis/.gimme/versions/go/src/runtime/sys_linux_amd64.s:526 +0x21
runtime.futexsleep(0xb5bf40, 0x0, 0xffffffffffffffff)
	/home/travis/.gimme/versions/go/src/runtime/os_linux.go:45 +0x4b
runtime.notesleep(0xb5bf40)
	/home/travis/.gimme/versions/go/src/runtime/lock_futex.go:151 +0xc7
runtime.stopm()
	/home/travis/.gimme/versions/go/src/runtime/proc.go:1951 +0xe5
runtime.findrunnable(0xc420032500, 0x0)
	/home/travis/.gimme/versions/go/src/runtime/proc.go:2407 +0x2bc
runtime.schedule()
[...]

PR #576, opened 2017-11-17
Test: rework concurrentExchange

miekg:
See issue #575. Try to fix (and maybe trigger it)


miekg:
defer and errorf seem like a sensible change anyhow.

PR #580, opened 2017-11-22
Add fuzzing framework

miekg:
Add easy way to fuzz this dns library, put fuzz related code in fuzz.go
and have a small Makefile.fuzz to be used:

$ make -f Makefile.fuzz build
$ make -f Makefile.fuzz fuzz

Will build and fuzz the library. Both pack/unpack and NewRR are fuzz
targets, but we could open this up.


andrewtj:
This looks interesting. Why is `FuzzNewRR` only expected to get junk?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Add fuzzing framewo..." ]
>This looks interesting. Why is `FuzzNewRR` only expected to get junk?

As opposed to all other function that potentially can get data? This was just
one other thing that I added, more is welcome.

Both Pack and Unpack and NewRR have been fuzzed for a week right now.

andrewtj:
No, I was wondering if `FuzzNewRR` should expect to be given valid input by the fuzzer. If it gets a comment, or a valid RR that can be packed and unpacked, shouldn't it return 0? Then again I suppose if the fuzzer's run for a week and that case hasn't come up it's probably not likely to.

miekg:
Note it did found these: https://github.com/miekg/dns/pull/579

PR #588, opened 2017-11-24
Completely ignore EDNS0_SUBNET errors if invalid subnet, resolves #581

### LINKED ISSUE #581: Handle invalid network families for EDNS0 subnet option
Hi, while playing for a bit with DNS-over-TLS I found an issue using [miekg/dns](https://github.com/miekg/dns) in conjunction with [getdnsapi/stubby](https://github.com/getdnsapi/stubby) that uses [getdnsapi/getdns](https://github.com/getdnsapi/getdns) behind the scenes.

getdns has the following configuration option:

- `edns_client_subnet_private`: Use EDNS0 Client Subnet privacy so the client subnet is not sent to authoritative servers

when that option is disabled, all works just as expected, but when that option is enabled, the new TLS DNS request has some extra EDNS0 flags, including one "empty" subnet that is a "Family 0" address ([code](https://github.com/getdnsapi/getdns/blob/develop/src/stub.c#L152)).

Because of the "Family 0" address, miekg/dns just throws an `dns: bad address family` error while unpacking the EDNS0 message in [edns.go#L266](https://github.com/miekg/dns/blob/master/edns.go#L266) and the request is not resolved.

I tried stubby with other TLS resolvers and they work just fine with this "Family 0" address, so I was confused about why miekg/dns has this problem.

Reading the [RFC  7871](https://tools.ietf.org/html/rfc7871#section-7.1.2) specs, I found the following:

```
   If there is no ADDRESS set, i.e.,
   SOURCE PREFIX-LENGTH is set to 0, then FAMILY SHOULD be set to the
   transport over which the query is sent.  This is for
   interoperability; at least one major authoritative server will ignore
   the option if FAMILY is not 1 or 2, even though it is irrelevant if
   there are no ADDRESS bits.
```

So, from the RFC snippet:

- getdns is not respecting the RFC because he is not including the family of the address (I will write an issue there too)
- miekg/dns can be improved by ignoring the EDNS0 subnet option if the Family is not 1 or 2

I hope this will help anyone facing similar issues.

daareiza:
Per [RFC  7871](https://tools.ietf.org/html/rfc7871#section-7.1.2) spec, if an invalid EDNS0_SUBNET option comes in the requests, it can be ignored.


msg_helpers.go
```diff
@@ -408,14 +408,13 @@ Option:
 		off += int(optlen)
 	case EDNS0SUBNET, EDNS0SUBNETDRAFT:
 		e := new(EDNS0_SUBNET)
-		if err := e.unpack(msg[off : off+int(optlen)]); err != nil {
-			return nil, len(msg), err
+		if err := e.unpack(msg[off : off+int(optlen)]); err == nil {
```
miekg: `err == nil` ? Where are w detecting the subnet is wrong?

msg_helpers.go
```diff
@@ -408,14 +408,13 @@ Option:
...
+			edns = append(edns, e)
+			if code == EDNS0SUBNETDRAFT {
```
miekg: We should kill EDNS0SUBNETDRAFT. I'll do that in another PR.

msg_helpers.go
daareiza: Hi @miekg, sorry for the late response, if the [unpack](https://github.com/miekg/dns/blob/master/edns.go#L266) function throws an error, we just ignore it

miekg: Hmm, I'm a bit confused now. If you have a edns client subnet section and the length != 0; this should in theory hold something that is sane, no?

miekg: I think we should solve #581 in the unpack function, instead of ignoring all errors here?

daareiza: I thought about that too, but in that case we need some flag that tell us to append or not the subnet to the `edns` var, or maybe I am missing something?

msg_helpers.go
```diff
@@ -408,14 +408,15 @@ Option:
 		off += int(optlen)
 	case EDNS0SUBNET, EDNS0SUBNETDRAFT:
 		e := new(EDNS0_SUBNET)
-		if err := e.unpack(msg[off : off+int(optlen)]); err != nil {
+		if err := e.unpack(msg[off : off+int(optlen)]); err == nil {
+			edns = append(edns, e)
+			if code == EDNS0SUBNETDRAFT {
```
fastest963: This is gone in #589.

msg_helpers.go
daareiza: @miekg sorry, pinging you to see if you have any comment/suggestion about my [last comment](https://github.com/miekg/dns/pull/588#discussion_r153811138)

miekg:
See https://github.com/miekg/dns/issues/581#issuecomment-347910673

PR #590, opened 2017-11-27
Modified clientconfig to match ndots0

eBeyond:
My modifications expect that the order of NameList matches the order of requests to dns
miekg. miekg.test. 
will be queried in exact this order.

miekg should be queried before miekg.test. so the dns has got the capability to give an answer to this (as docker in swarm mode does)

### DIFF
--- a/clientconfig_test.go
+++ b/clientconfig_test.go
@@ -48,6 +48,14 @@ nameserver 11.28.10.1
 options ndots:16
 `
 
+const normalWithNdotsBelow0 string = `
+# Comment
+domain somedomain.com
+nameserver 10.28.10.2
+nameserver 11.28.10.1
+options ndots:-1
+`
+
 func testConfig(t *testing.T, data string, expectedNdots int) {
 	cc, err := ClientConfigFromReader(strings.NewReader(data))
 	if err != nil {
@@ -74,6 +82,7 @@ func TestWithOutNdots(t *testing.T)        { testConfig(t, normalWithOutNdots, 1
 func TestNdots0(t *testing.T)              { testConfig(t, normalWithNdots0, 0) }
 func TestNdots15(t *testing.T)             { testConfig(t, normalWithNdots15, 15) }
 func TestNdots16(t *testing.T)             { testConfig(t, normalWithNdots16, 15) }
+func TestNdotsBelow0(t *testing.T)         { testConfig(t, normalWithNdotsBelow0, 0) }
 
 func TestReadFromFile(t *testing.T) {
 	tempDir, err := ioutil.TempDir("", "")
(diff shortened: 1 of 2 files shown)


clientconfig_test.go
```diff
@@ -35,10 +71,18 @@ func testConfig(t *testing.T, data string) {
...
+	if cc.Ndots != expectedNdots {
+		t.Errorf("Ndots not properly parsed: (Expected: %d / Was: %d)", expectedNdots, cc.Ndots)
+	}
 }
 
-func TestNameserver(t *testing.T)          { testConfig(t, normal) }
-func TestMissingFinalNewLine(t *testing.T) { testConfig(t, missingNewline) }
+func TestNameserver(t *testing.T)          { testConfig(t, normal, 1) }
+func TestMissingFinalNewLine(t *testing.T) { testConfig(t, missingNewline, 1) }
+func TestWithOutNdots(t *testing.T)        { testConfig(t, normalWithOutNdots, 1) }
```
miekg: Instead of adding all these test function I would just add a single `TestNdots` tests, put all the new consts in a `map[string]int` and test the resulting ndots by just ranging over that map. Think even a resolv.conf with just `options ndots:N` is already valid, so this would be somewhat smaller.

clientconfig_test.go
```diff
@@ -91,10 +116,18 @@ func TestNameList(t *testing.T) {
 	} else if names[1] != "miek.nl.test." {
 		t.Errorf("NameList didn't return search last: %v", names[1])
 	}
+}
+func TestNameList_Ndots2(t *testing.T) {
```
miekg: Is this syntax allow by Go? go lint seems fine with it...

clientconfig_test.go
```diff
@@ -40,6 +40,31 @@ func testConfig(t *testing.T, data string) {
...
+	var ndotsVariants map[string]int = make(map[string]int)
+	ndotsVariants["options ndots:0"] = 0
+	ndotsVariants["options ndots:1"] = 1
+	ndotsVariants["options ndots:15"] = 15
+	ndotsVariants["options ndots:16"] = 15
+	ndotsVariants["options ndots:-1"] = 0
+	ndotsVariants[""] = 1
+
+	for key := range ndotsVariants {
+		testNdotsConfig(t, key, ndotsVariants[key])
```
miekg: just inline that function here, keeps everything localized in a spot.

clientconfig_test.go
```diff
@@ -40,6 +40,31 @@ func testConfig(t *testing.T, data string) {
...
+	if err != nil {
+		t.Errorf("error parsing resolv.conf: %v", err)
+	}
+	if cc.Ndots != expectedNdots {
+		t.Errorf("Ndots not properly parsed: (Expected: %d / Was: %d)", expectedNdots, cc.Ndots)
+	}
+}
+
+func TestNdots(t *testing.T) {
+	var ndotsVariants map[string]int = make(map[string]int)
```
miekg: you can do this in one go:
```
ndotsVariants := map[string]int {
    "options ndots:0": 0,
 ///etc.
}
```

PR #593, opened 2017-11-28
Provide configured string in callback function

spsholleman:
When using this package to run a DNS server, users of the package provide callback routines to be called when certain FQDN strings/patterns are matched during a DNS query.  For users adding intelligence to these callbacks it can be very useful to also be given in the callback what the configured pattern was, particularly if they are managing any context in other data structures outside of this library.  For example, a lookup of www.foobar.com could have been enabled by a configured pattern of www.foobar.com, foobar.com, or com.  It doesn't make sense to duplicate the logic of finding the longest matching pattern when it already exists in this package.

A private fork achieved this in a non-backwards compatible manner and this is a first attempt to get the same functionality in the upstream library in a backwards compatible manner.  I don't love the implementation but it does maintain backwards compatibility.  In addition, it adds this functionality by returning a struct and should assist similar changes in the future because it will be possible to add a single field to the struct.

Open to another method to achieve this if anyone has ideas, tried a variadic solution but because the typing is so strong it still requires at least one new argument.

### DIFF
--- a/server.go
+++ b/server.go
@@ -85,9 +85,9 @@ func (f HandlerFunc) ServeDNS(w ResponseWriter, r *Msg) {
 }
 
 type CallbackContext struct {
-	w ResponseWriter
-	r *Msg
-	pattern string
+	Writer  ResponseWriter
+	Req     *Msg
+	Pattern string
 }
 
 type HandlerContext func(*CallbackContext)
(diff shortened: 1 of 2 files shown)


miekg:
This is quite invasive *and* you don't need this as you can write your own Mux for the handling the callback: https://github.com/coredns/coredns/blob/master/core/dnsserver/server.go#L96

spsholleman:
I understand that you can wrap the callback into a different callback but inherently we want some of the output computed in the match() function - mainly what the configured pattern was that triggered this callback.  I don't think there is any way to wrap that in another Mux without computing it a second time and taking a performance hit.

I do kind of understand "invasive" but from what perspective?  It does maintain backwards compatibility, has a very low performance impact, and the diff is small (much of the diff is in the unit test when it directly used private functions).

Any suggestion on how this could be done in a less invasive way?

miekg:
I don't see why this needs to be in the libl; you can write your own mux and do what you want.

spsholleman:
By the way, still stewing on this feedback - need to experiment some but I will come back and comment or close this.

spsholleman:
alright, I get how to do it and what you mean now, closing this.  sorry for the noise.

Issue #571, opened 2017-11-17
Support ReadMsgUDP on Windows


fastest963:
Looks like with the addition of https://github.com/golang/go/issues/9252 in Go 1.10 we should be able to support #524 on Windows. We need to wait for https://go-review.googlesource.com/c/go/+/76391 to be merged into x/net first, but I thought I would open this issue for tracking. This would be a good point to switch to using `godep` as well.

miekg:
@fastest963 We're now (merged this few minutes ago) using dep to manage dependencies. Probably makes sense to drop our `internal` dir and properly vendor the net/tools ?

miekg:
Should this still be open?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Support ReadMsgUDP ..." ]
>Given that https://go-review.googlesource.com/c/go/+/76393 is merged into go 1.10, it looks like we should be able to support this. Are we comfortable with requiring go 1.10 (on Windows) or should I add a build flag to the file so it only opportunistically (Windows+1.10) does this?

Should probably wait until 1.11 then...

tmthrgd:
This was fixed in #738.

PR #594, opened 2017-11-29
Use and vendor golang.org/x/net/ipv4 and golang.org/x/net/ipv6

fastest963:
Removed the `internal/socket` and used `golang.org/x/net/ipv4` and `golang.org/x/net/ipv6` adding them with dep to the vendor folder.

Additionally, I removed the `udp_linux.go` in favor of using ipv4/ipv6 for cross-platform support. We don't quite have Windows support since that will be coming with go 1.10 (and then we can use build flags to exclude `udp_windows.go` on < 1.10) and is tracked in #571.

I tried to add tests to test as much of the OOB functionality as possible. I will be testing in production with a newer build of CoreDNS to verify everything on Linux.


fastest963:
Apparently travis-ci doesn't have any IPv6 support so the test for `[::1]:0` will log and not be run. Running that locally for me works but I'm not sure how to get that tested via travis-ci.

fastest963:
We don't have a way to test for IPv6 or IPv4 in `parseDstFromOOB` so I left it defaulting to IPv6 and then falling back to IPv4. If you have an IPv6 kernel and don't specifically listen on IPv4, then the out-of-band data will be in IPv6 format (with :ffff: in front of IPv4) no matter if the packet was actually from IPv4 or IPv6 so I assume that's probably the more common case. We could look at the first few bytes of the data and see if there's a level of `0` or `41` to determine the format but that wouldn't be cross platform. I don't think the overhead of doing both is worth the optimization to only 1 or the other at this point.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Use and vendor gola..." ]
>Apparently travis-ci doesn't have any IPv6 support so the test for `[::1]:0` will log and not be run. Running that locally for me works but I'm not sure how to get that tested via travis-ci.

No, you can't with travis :-(

miekg:
Is everything looking Ok from your side? If so; I will merge this.

miekg:
I'm going ahead with this. CoreDNS 1.0.0 will be announced today and 1.0.1 dev has started, meaning plenty of time for me to see if this works or not.

fastest963:
@miekg yes we've been running this in production with CoreDNS since the 30th and it seems to be working fine. Sorry, somehow I missed your comment.

johnbelamaric:
@fastest963 if you are running CoreDNS in production would you mind adding yourself to https://github.com/coredns/coredns/blob/master/ADOPTERS.md ? It will help as we move toward incubation status in the CNCF.

PR #595, opened 2017-12-04
Add semver

### LINKED ISSUE #592: semver?
I'm a bit ambivalent about doing semver, because it implies doing releases, which are just toil. OTOH it might make using this lib easier; because you can version check if you're on the right version (can see a need for this in coredns/coredns).

We could just version without actually releasing?
Can we automatically bump a minor an each PR merged? Start with 1.0.0 --> 1.0.1 etc?

miekg:
Add a version.go that has the semver version of this libary; now at
1.0.0. Use a struct so external code can easily check the for the
version without resulting to string parsing. Add String() function if
you want to access the version string.

Fixes #592

### DIFF
--- a/Makefile.release
+++ b/Makefile.release
@@ -1,0 +1,51 @@
+# Makefile for releasing.
+#
+# The release is controlled from version.go. The version found there is
+# used to tag the git repo, we're not building any artifects so there is nothing
+# to upload to github.
+#
+# * Up the version in version.go
+# * Run: make -f Makefile.release release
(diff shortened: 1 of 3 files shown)


fastest963:
Once this is merged, is it going to be tagged and marked as a release in github?

miekg:
hmm, good point. Seeing what gh-release can do here (already using it for coredns)

miekg:
ok. PTAL

PR #596, opened 2017-12-04
Lint: use ignore-this on generated files

miekg:
Make golint not complain about generated file - it's a bit of cheating,
but I think that is OK.


miekg:
Seems tkey has a change here as well?

miekg:
Merging this; probably means TKEY is now broken-ish

PR #597, opened 2017-12-04
Run go gen for TKEY

miekg:
TKEY's length changed, but we've missed the 'go generate'.

### DIFF
--- a/ztypes.go
+++ b/ztypes.go
@@ -592,9 +592,9 @@ func (rr *TKEY) len() int {
 	l += 2 // Mode
 	l += 2 // Error
 	l += 2 // KeySize
-	l += len(rr.Key) + 1
+	l += len(rr.Key)/2 + 1
 	l += 2 // OtherLen
-	l += len(rr.OtherData) + 1
+	l += len(rr.OtherData)/2 + 1
 	return l
 }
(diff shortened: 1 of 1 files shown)


spsholleman:
sorry, that was probably my fault in https://github.com/miekg/dns/pull/567 - not sure how that got missed because I definitely ran it to get the zmsg.go changes.

spsholleman:
however the +1 doesn't need to be there, isn't the /2 alone sufficient to represent the length?  did I use the wrong type for these fields in TKEY?

I can also modify TestTKEY to check the length against the string/bytes passed in and ensure we get back the same length.  that also highlights that the +1 isn't correct because we don't back to the same same len() value after UnpackRR().

miekg:
This is the patch for TKEY:
```
-       Key        string
+       Key        string `dns:"size-hex:KeySize"`
        OtherLen   uint16
-       OtherData  string
+       OtherData  string `dns:"size-hex:OtherLen"`
 }
~~~
So this go gen is correct *iff* that patch is correct. Depends how this is encoded on the on the wire; is that hex on wire or just octets?

miekg:
the wiredata for TKEY is not hex; so this is wrong regardless. Closing here. See #599 for the correct wiredata.

PR #598, opened 2017-12-06
Allow $INCLUDE to reference relative file

### LINKED ISSUE #537: $INCLUDE is not relative to file's location
It seems like the working/excution directory from coredns is used and not something set in the root mw.

Corefile:
```
  .:5053 {
    root /zones/
    auto {
      directory . (.*)
      transfer *
    }
    errors stdout
    log stdout
  }
```

Main zone (example.com):
```
@                   3600 IN SOA     ns.example.com. domains.example.com. (
                                      2017091317   ; serial
                                      15m          ; refresh
                                      5m           ; retry
                                      1w           ; expire
                                      12h    )     ; minimum

@                  86400 IN NS     ns.example.com

test IN A 127.0.0.1

; works
$INCLUDE /zones/test.example.com test

; doesn't work
$INCLUDE test.example.com test
```

miekg:
When using a relative file in an $INCLUDE the file is referenced from
the cwd from the calling processes; this changes it to be down from the
view point where the file exists.

Code from https://github.com/miekg/dns/issues/537#issuecomment-342932962

Fixes #537

### DIFF
--- a/scan.go
+++ b/scan.go
@@ -1,8 +1,10 @@
 package dns
 
 import (
+	"fmt"
 	"io"
 	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 )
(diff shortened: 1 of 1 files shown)


miekg:
merging on the basis the testing in https://github.com/miekg/dns/issues/537#issuecomment-346861654 was good enough.

Pretty hard to do a unit test 'cause I need to fake the CWD of the calling process

Issue #549, opened 2017-11-06
GSS-TSIG support


bodgit:
Would it be possible to add support for GSS-TSIG (RFC 3645)? This would make it possible to perform secure DNS updates to a Windows Active Directory environment, which AFAICT doesn't support normal TSIG updates. I figured maybe https://github.com/jcmturner/gokrb5 could be useful to do the Kerberos side of things.

captncraig:
Wow, I was actually just looking for this as well. https://technet.microsoft.com/en-us/library/cc961412.aspx?f=255&MSPPError=-2147217396 describes their security model a bit more. It kinda looks like greek to me though.

bodgit:
Yep, I looked at that page too :smiley:

I've managed to work out the steps needed to do this manually using a combination of `ktutil`, `kinit`, and `nsupdate -g` from a Linux host such that records will get added/updated/deleted in DNS, but I'm at a loss where to start with any code changes to achieve the same thing.

captncraig:
https://tools.ietf.org/html/rfc3645 has the dns queries for the handshake spelled out, but I am kinda stuck on the GSS_Init_sec_context part of generating the initial token or whatever. There's some go bindings to a c api that I can't figure out how to install, and there is that native go package that I can't quite figure out the gssapi interface for.

bodgit:
I'm guessing the bit section under "Generic Kerberos Client" on https://github.com/jcmturner/gokrb5 might be relevant. Especially given the final note:

> Now send the AP_REQ to the serivce. How this is done will be specific to the application use case.

Maybe that's the bit that gets sent as the TSIG in some form or another? (Complete guess)

miekg:
Apart from some TSIG and TKEY constants that aren't defined, what else would be needed from this package?

captncraig:
A good point. I wonder if this wouldn't be a better fit for the kerberos library to implement on that end. Maybe a `dns` package with a simple `ExchangeWithGssTsig(msg)` type function.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] GSS-TSIG support (#..." ]
>A good point. I wonder if this wouldn't be a better fit for the kerberos library to implement on that end. Maybe a `dns` package with a simple `ExchangeWithGssTsig(msg)` type function.

That's not even needed, you need to construct a dns.Msg with such a TSIG at the 
end and the unpack/pack will take care of it.

spsholleman:
[...]

PR #599, opened 2017-12-06
Add size-hex: modifier for len()

miekg:
This was missing and generated the wrong code for TKEY; it adds a +1 to
the amount. This should happen (technically).

I think the fallout is not super bad (of the +1) as we allocate a byte
more for when pack a message.

### DIFF
--- a/dns_test.go
+++ b/dns_test.go
@@ -422,6 +422,14 @@ func TestTKEY(t *testing.T) {
 	if unPackErr != nil {
 		t.Fatal("unable to decode TKEY RR", unPackErr)
 	}
+	// Make sure it's a TKEY record
+	if rr.Header().Rrtype != TypeTKEY {
+		t.Fatal("Unable to decode TKEY")
+	}
+	// Make sure we get back the same length
+	if rr.len() != len(tkeyBytes) {
+		t.Fatalf("Lengths don't match %d != %d", rr.len(), len(tkeyBytes))
+	}
 	// make space for it with some fudge room
 	msg := make([]byte, tkeyLen+1000)
 	offset, packErr := PackRR(rr, msg, 0, nil, false)
--- a/types_generate.go
+++ b/types_generate.go
@@ -181,6 +181,8 @@ func main() {
 				fallthrough
 			case st.Tag(i) == `dns:"base64"`:
 				o("l += base64.StdEncoding.DecodedLen(len(rr.%s))\n")
+			case strings.HasPrefix(st.Tag(i), `dns:"size-hex:`): // this has an extra field where the length is stored
+				o("l += len(rr.%s)/2\n")
 			case strings.HasPrefix(st.Tag(i), `dns:"size-hex`):
 				fallthrough
 			case st.Tag(i) == `dns:"hex"`:
--- a/ztypes.go
+++ b/ztypes.go
@@ -347,7 +347,7 @@ func (rr *HIP) len() int {
 	l++    // HitLength
 	l++    // PublicKeyAlgorithm
 	l += 2 // PublicKeyLength
-	l += len(rr.Hit)/2 + 1
+	l += len(rr.Hit) / 2
 	l += base64.StdEncoding.DecodedLen(len(rr.PublicKey))
 	for _, x := range rr.RendezvousServers {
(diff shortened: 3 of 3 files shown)


miekg:
interesting that this fails now...

miekg:
have plain octets probably also means we can read it from text, because you can't put all TKEY's in presentation format, i.e. null bytes?

bodgit:
This is how I've currently been creating my TKEY RR's with #567 merged, `b` is the Kerberos token marshaled into `[]byte`:
```golang
        now := time.Now().Unix()
        m.Extra[0] = &dns.TKEY{
                Hdr: dns.RR_Header{
                        Name:   keyname,
                        Rrtype: dns.TypeTKEY,
                        Class:  dns.ClassANY,
                        Ttl:    0,
                },
                Algorithm:  dns.Fqdn("gss-tsig"),
                Mode:       3,
                Inception:  uint32(now),
                Expiration: uint32(now),
                KeySize:    uint16(binary.Size(b)),
                Key:        hex.EncodeToString(b),
        }
```
If that helps.

miekg:
that looks correct. But the code in miekg/dns is generated from the struct, so it must match the wireformat - it essentially does the same.

spsholleman:
Wow, I'm kind of confused now.  It's working well for me and  I am glad to hear @bodgit has had some luck as well.  I'd like to get it straight.

I agree I messed up the 'go generate' portion and I'm still not entirely clear on how I did that.  I think that I just forgot to generate or checkin the part for len().  But note that to pack/unpack the RR you don't need to call len() - that's why it wasn't caught as being not correct.   The pack()/unpack() generation also clearly was done in https://github.com/miekg/dns/pull/567 because we see the change to zmsg.go and otherwise you would have had no diffs in zmsg.go here.  So I don't think it's accidentally working for people as mentioned in https://github.com/miekg/dns/issues/549 

What I'm most confused about is why you think it should be a 'string' just because the RFC said octet-stream.  What we then want is just a stream of bytes.   I thought we have some flexibility about how to represent a string field in this go package and then translate it to the right wire format when calling pack()/unpack() and that a hex encoded string would best represent the usage models.  And that's exactly what the 'dns:hex' attribute was buying us was to treat this string field as a hexEncoded string representation before translating it to the wire format.

I'm also still confused at the len() code generated for the size-hex field, I don't get why there is a +1 added to it.  I think it should just be len(KeyData)/2.   But as I came to learn, len() is not that important for getting stuff on / off the wire properly.   What is the +1 for?  Is the +1 trying to guard against an odd number length?

miekg:
Thanks for comment, prompted me to look some more.

I think you're right about is can just be `size-hex:KeySize` and the `len+1` is wrong. This may be an oversight in generating the `len` functions with using a size-<something> function.
Where the +1 came from: when encoding some of these the first byte would be used for the length, ala TXT RDATA.

Checking that len bit right now.

miekg:
yeah :( No special handling here: https://github.com/miekg/dns/blob/master/types_generate.go#L184

spsholleman:
Yeah, but I think hex is wrong as well if I understand it.  It too should not have the +1.    The could have some protection to round up to an even number if len is odd.

The only difference between the two is we can use another field to ensure the length is right for size-hex and for hex and you have to just go on string length.

miekg:
I think hex can work - because that only impacts the presentation format, len(hex)/2 *should* be the same length as just using string.

The +1 is just wrong in any case.

spsholleman:
Okay good - we're getting on the same page.  Let me know if I can help or you want me to clean this up.

miekg:
Thanks. I'll rework this PR, hopefully the fallout will be small

spsholleman:
Here is an enhancement to the TKEY test to ensure len() is good moving forward once fixed if you'd like to cherry pick it with these changes.

https://github.com/skyportsystems/dns/commit/8bb992ef67308387247dda3a5145d047514d8735

miekg:
In the end the +1 might be harmless, probably added this *years* ago for some reason. Anyway most minimal PR I could come up with; doesn't win any prizes for readability; but works. PTAL

spsholleman:
LGTM - thanks for looking at it.

PR #603, opened 2017-12-12
Two changes to assist with TSIG operations

spsholleman:
Below are two changes to assist with various TSIG operations.  The first is to introduce a flag that controls whether the TSIG verification is done immediately when data is read.  In order to be backwards compatible, the logic is inverted that when you want to skip you must explicitly set it.

Also, if a dns.Msg is created by Unpack()'ing from the wire format, save the wire format for others to be able to interpret and use it later if necessary.

Noted, these operations may not be useful to all but I know at least the ability to skip TSIG verification had been mentioned.

### DIFF
--- a/client.go
+++ b/client.go
@@ -348,7 +348,7 @@ func (co *Conn) Read(p []byte) (n int, err error) {
 // signature is calculated.
 func (co *Conn) WriteMsg(m *Msg) (err error) {
 	var out []byte
-	if t := m.IsTsig(); t != nil {
+	if t := m.IsTsig(); t != nil && co.TsigSecret != nil {
 		mac := ""
 		if _, ok := co.TsigSecret[t.Hdr.Name]; !ok {
 			return ErrSecret
--- a/client.go
+++ b/client.go
@@ -21,7 +21,6 @@ type Conn struct {
 	net.Conn                         // a net.Conn holding the connection
 	UDPSize        uint16            // minimum receive buffer for UDP messages
 	TsigSecret     map[string]string // secret(s) for Tsig map[<zonename>]<base64 secret>, zonename must be in canonical form (lowercase, fqdn, see RFC 4034 Section 6.2)
-	TsigVerifySkip bool              // Skip verifying a TSIG RR immediately at read time
 	rtt            time.Duration
 	t              time.Time
 	tsigRequestMAC string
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -40,6 +41,7 @@ type Client struct {
 	ReadTimeout    time.Duration     // net.Conn.SetReadTimeout value for connections, defaults to 2 seconds - overridden by Timeout when that value is non-zero
 	WriteTimeout   time.Duration     // net.Conn.SetWriteTimeout value for connections, defaults to 2 seconds - overridden by Timeout when that value is non-zero
 	TsigSecret     map[string]string // secret(s) for Tsig map[<zonename>]<base64 secret>, zonename must be in canonical form (lowercase, fqdn, see RFC 4034 Section 6.2)
+	TsigVerifySkip bool              // Skip verifying a TSIG RR immediately at read time
```
miekg: Shouldn't a `nil` TsigSecret map convey the same info?

msg.go
```diff
@@ -124,6 +124,7 @@ type Msg struct {
 	Answer   []RR       // Holds the RR(s) of the answer section.
 	Ns       []RR       // Holds the RR(s) of the authority section.
 	Extra    []RR       // Holds the RR(s) of the additional section.
+	RawMsg   []byte     // The original byte stream if this message was decoded from wire format
```
miekg: This is a big change and increases Msg's size *a lot* (doubles?)

client.go
spsholleman: I thought about that and liked it because it's very similar to the logic that is in the Write() call.  But as I mentioned in the PR description I was trying to do it in a manner that doesn't change the default behavior.  By default today if this is nil and you receive a TSIG, that message will not be processed - it's immediately rejected before "serving" it.  If we change the logic as you suggest, that would now be received and processed.

I am fine to change it that way if you are okay changing the default behavior of the package.

msg.go
spsholleman: Yes, it will for incoming messages.  Couple other possibilities:  only do it for TSIG() messages or have a boolean that controls whether it's done or not.  Any of those better?  But that still requires that this field at least be added (which by itself doesn't increase the size) but not used.  That or maybe use the Decorate*() functions and come up with a way to customize the Msg()?

client.go
miekg: I think *not* adding this boolean is cleaner, chaning the default behavior is the price we then pay. So be it.

miekg:
Extending Msg() and Unpack() will be hard sell because we want to keep the API stable.

Why do you want the raw bytes in the first place? I mean this package works really hard to *not* give you the raw bytes (still believe this is one of the reason it is easy to use).

miekg:
I see. I don't have a good solution, adding new fields to Msg is probably a no-go.

That section is a bit weird as well; if you *share* the key you should resign it; otherwise forward it; I imagine have access to the key is problematic in this case.

spsholleman:
> I see. I don't have a good solution, adding new fields to Msg is probably a no-go.

Curious - why is just adding the field a problem?  I get the memory concern (could do this only for TSIG() if you want) and I get issue with changing the API by changing one of the interfaces or something but I don't see why adding something new to the Msg struct is so troublesome.

And again, if you prefer it not here at all that's fine.  I'll just revert that part of the change and we'll have just that one-liner to read w/o TSIG verify and I'll keep it in my fork.

spsholleman:
removed the controversial part of this

bodgit:
Is it worth making a similar change to the corresponding `WriteMsg()` function to be able to turn off automatic TSIG generation?

miekg:
Chaning WriteMsg  looks unaviodable

spsholleman:
Sure, if that will help I will do that.  Wasn't sure if the callback scheme you were discussing in https://github.com/miekg/dns/issues/549 would use this or not.  For my usage I used Write() directly which is why I didn't need to change WriteMsg().  But if there is agreement on this I'll add it quickly.

PR #607, opened 2017-12-19
Add UnpackStrict

lmb:
The function returns an error if the buffer to unpack contained trailing data. The semantics of Unpack are left unchanged.

This fixes a long standing TODO.

### DIFF
--- a/msg.go
+++ b/msg.go
@@ -50,6 +50,7 @@ var (
 	ErrSoa           error = &Error{err: "no SOA"}                             // ErrSOA indicates that no SOA RR was seen when doing zone transfers.
 	ErrTime          error = &Error{err: "bad time"}                           // ErrTime indicates a timing error in TSIG authentication.
 	ErrTruncated     error = &Error{err: "failed to unpack truncated message"} // ErrTruncated indicates that we failed to unpack a truncated message. We unpacked as much as we had so Msg can still be used, if desired.
+	ErrTrailingData  error = &Error{err: "trailing data after message"}
 )
 
 // Id by default, returns a 16 bits random number to be used as a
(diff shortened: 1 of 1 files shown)


lmb:
I realised that this might need an UnpackStructStrict as well, but I'm not sure.

miekg:
This adds a whole new function for a niche feature, so I won't merge this - most likely.

What could be done is having an optional (brr) argument unpackOpt which is a struct that defines how you want errors to be handled.
Making it optional allows the API to be backwards compatible.
Then again, optional argument are evil, so the question comes back: is this so bad that we really want/need this?

lmb:
I see your point, and agree that the optional argument is not the right thing to do. How about adding a function that would allow me to detect trailing data without being too specific? Something like
`UnpackBuffer(msg []byte) (remainder []byte, err error)`.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Add UnpackStrict (#..." ]
>I see your point, and agree that the optional argument is not the right thing to do. How about adding a function that would allow me to detect trailing data without being too specific? Something like
>`UnpackBuffer(msg []byte) (remainder []byte, err error)`.

This could be done, but this seems a bit of an edge case... What would be the 
rational for having this (forever) in the API?

lmb:
The use case is to discard messages which have trailing garbage, which we apparently receive.

lmb:
I was pointed at https://lists.dns-oarc.net/pipermail/dns-operations/2013-April/010037.html by @ogud which at least discusses the problem.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Add UnpackStrict (#..." ]
>I was pointed at https://lists.dns-oarc.net/pipermail/dns-operations/2013-April/010037.html by @ogud which at least discusses the problem.

Ack. But we now accept these message and return the data to the user... the old 
(and maybe dated, be liberal in what you accept). I don't see the benefit of 
being more strict here, becuase we've never done this.

Other than for DNS inspection reasons I can't see why we need to carry a new 
function for this? What are you planning to use this for, just FORMERR?

lmb:
> Other than for DNS inspection reasons I can't see why we need to carry a new function for this? What are you planning to use this for, just FORMERR?

Yes, we want to refuse messages with trailing data. The current API doesn't allow detecting this from what I can tell, so I'd like to add a way.

miekg:
Sorry, but you'll have to carry this patch in a your private fork, I (still) see no reason why would want to add this to the API.

PR #608, opened 2017-12-20
Unpack: return header

miekg:
When we unpack a message and only have the header consider it a valid
message and don't return an error.


andrewtj:
`Msg`'s `Question`, `Answer`, `Ns` and `Extra` should be set to nil in case `Msg` is being reused

miekg:
good point

PR #610, opened 2017-12-21
Don't use untrusted lengths from Header to pre-allocate

### LINKED ISSUE #609: DoS: Msg.Unpack uses untrusted lengths to allocate slices
It is currently possible to make Msg.Unpack allocate several MiB of data per call, since lengths from the header are trusted. Any service using the library to parse packets from an untrusted source is therefore susceptible to DoS: an attacker can overwhelm the GC and force long GC pauses.

This was introduced in #281.

lmb:
We currently use information from a potential attacker to pre-allocate slices for the Question, Answer, etc. sections. This allows an attacker to force allocation of several MiB per parsed Msg.

Instead, don't pre-allocate those slices. append() always allocates in powers of two, which is probably the best we can do.

Fixes #609.

### DIFF
--- a/msg.go
+++ b/msg.go
@@ -612,8 +612,8 @@ func UnpackRR(msg []byte, off int) (rr RR, off1 int, err error) {
 // If we cannot unpack the whole array, then it will return nil
 func unpackRRslice(l int, msg []byte, off int) (dst1 []RR, off1 int, err error) {
 	var r RR
-	// Optimistically make dst be the length that was sent
-	dst := make([]RR, 0, l)
+	// Don't pre-allocate, l may be under attacker control
+	var dst []RR
 	for i := 0; i < l; i++ {
 		off1 := off
 		r, off, err = UnpackRR(msg, off)
@@ -815,9 +815,10 @@ func (dns *Msg) Unpack(msg []byte) (err error) {
 		return ErrTruncated
 	}
 
-	// Optimistically use the count given to us in the header
-	dns.Question = make([]Question, 0, int(dh.Qdcount))
-
+	// Qdcount, Ancount, Nscount, Arcount can't be trusted, as they are
+	// attacker controlled. This means we can't use them to pre-allocate
+	// slices.
+	dns.Question = nil
 	for i := 0; i < int(dh.Qdcount); i++ {
 		off1 := off
 		var q Question


tmthrgd:
Why not just put a cap on the size of the allocation? That way you don’t suffer the performance penalty of lots of allocations and copies, while still preventing an attacker from wastefully allocating lots of memory.

miekg:
A golang slice cap you mean? Think we also had that in the original code.

andrewtj:
@miekg I'm pretty sure @tmthrgd is referring to caping the amount that'll be pre-allocated (`min(250, sectioncount)` or whatever). I'd like to see a bit more discussion before merging this so the goal posts are clearer.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Don't use untrusted..." ]
>@miekg I'm pretty sure @tmthrgd is referring to caping the amount that'll be pre-allocated (`min(250, sectioncount)` or whatever). I'd like to see a bit more discussion before merging this so the goal posts are clearer.

Agreed, although I think this can be merged as-is, and stop the direct bleeding.

doing a such minimum sgtm, but what number? Anything we pick will be wrong in 
some scenario and it is a magic number.

/Miek

--
Miek Gieben

lmb:
I agree with Miek, these numbers will vary based on the query type, e.g. SRV tends to have more RRs than A for example. I have a hunch that the most common queries will be A/AAAA with 0-2 answers, which would mean that most of the time only one allocation happens anyways. I'll try and find some evidence for that.

miekg:
I'll go ahead and merge this.

btw @andrewtj your leaky.go is *interesting*; def. on the trail of a bug here, not sure why we lead though.

andrewtj:
> On 21 Dec 2017, at 8:52 pm, Miek Gieben <notifications@github.com> wrote:
> 
> [ Quoting <notifications@github.com> in "Re: [miekg/dns] Don't use untrusted..." ]
> >@miekg I'm pretty sure @tmthrgd is referring to caping the amount that'll be pre-allocated (`min(250, sectioncount)` or whatever). I'd like to see a bit more discussion before merging this so the goal posts are clearer.
> 
> Agreed, although I think this can be merged as-is, and stop the direct bleeding.
> 

Fair enough.

> doing a such minimum sgtm, but what number? Anything we pick will be wrong in 
> some scenario and it is a magic number.
> 

If you pick a number greater than ~6000 it won't be wrong :p Maybe table that for the aforementioned discussion.

Issue #557, opened 2017-11-08
Move to github org?


miekg:
~~Also see https://github.com/miekg/dns/issues/555~~

~~I propose we drop this; implement dnsutil somewhere else, but not maintained within this package.~~

~~This is backwards incompatible, but easy to fix; copy the function into your own code.~~

andrewtj:
What do you reckon of making a new GitHub organisation? Might be a good place to peel off `dnsutil` and `idn`, as well as any other esoteric or dependency-requiring things. Possibly whatever comes out of the GSS-TSIG discussion for instance.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] proposal: drop dnsu..." ]
>What do you reckon of making a new GitHub organisation? Might be a good place to peel off `dnsutil` and `idn`, as well as any other esoteric or dependency-requiring things. Possibly whatever comes out of the GSS-TSIG discussion for instance.

That's sounds like a good idea. But you move the repo (even with the github 
redirection you break a lot of builds, no?)

miekg:
any ideas for a name, dns is obviously taken. godns is still free, but somewhat dislike the 'god'ns 
(can't hurt to acquire a name before we do something)

miekg:
doesn't preclude dropping it from miekg/dns now though (resurrect it in a new org)

andrewtj:
Yeah I'd assume moving the repo will break things. Maybe issues, etc can be disabled on this repo and it can be be run as a mirror for a while.

I don't have any ideas for a name at the minute but maybe inspiration will strike. Do you want an available .com/.net/.org to go along with the name? And I suppose on that, is moving to a vanity import URL worthwhile? I doubt GitHub is going anywhere but if you wanna snag a namespace, might as well use it I guess.

It look like there isn't too many (public) users of either pkg going by the [dnsutil](https://godoc.org/github.com/miekg/dns/dnsutil?importers) and [idn](https://godoc.org/github.com/miekg/dns/idn?importers) importers. I can probably find the time to drop an issue to each of them pointing back here if you like.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] proposal: drop dnsu..." ]
>Yeah I'd assume moving the repo will break things. Maybe issues, etc can be disabled on this repo and it can be be run as a mirror for a while.

Yep, that might work. We should also move to semver after such a move. Now with 
'dep' this finally makes sense.

[...]

PR #612, opened 2017-12-22
BIND test harness

andrewtj:
This is intended to setup and teardown a local BIND install during tests. It's not quite baked but I think there's enough here that it's worth a look over. If it's merged maybe other people can pick it up and run with it.

BIND is assumed to be available if its executables can be found by `exec.LookPath()`. On Travis `install-bind.bash` is used to compile and install BIND 9.12RC1 built against OpenSSL master which gets us support for ed25519.

The zone served is signed with all the algorithms this package knows about. There's HMAC keys for SHA1, SHA256 and SHA512 and the keys are allowed to update any name below their own name.

The only tests I've added that hit BIND are for exercising AXFRs in `xfr_test.go` which is still behind the `net` build tag. If this is merged in the future the other tests in that file should be converted to hit the local BIND setup so the tag can be removed.

### DIFF
--- a/bind_test.go
+++ b/bind_test.go
@@ -1,0 +1,229 @@
+package dns
+
+import (
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+)
+
+type bindHarnessInfo struct {
+	port       int
+	hmacKeys   map[string]string
+	dnssecKeys map[uint8]string
+}
+
+var bindHarness *bindHarnessInfo
+
+func bindHaveExecutables() bool {
+	bindExecutables := []string{"named", "dnssec-keygen", "dnssec-signzone"}
+	for _, b := range bindExecutables {
+		if _, err := exec.LookPath(b); err != nil {
+			return false
+		}
+	}
+	return true
+}
+
+func bindGenerateDnssecKeys(tempdir string) map[uint8]string {
+	keys := map[uint8]string{}
+	baseArgs := []string{"-q", "-K", tempdir, "-A", "-30s", "-n", "ZONE", "-T", "DNSKEY", "-f", "KSK"}
+	for alg := range AlgorithmToString {
+		args := append(baseArgs, "-a", fmt.Sprintf("%d", alg))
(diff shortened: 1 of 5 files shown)


install-bind.bash
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env bash
+set -xeuo pipefail
+if [ $# -ne 1 ]; then
+    echo "supply target dir"
+    exit 1
+fi
+TARGET_DIR=$1
+BIND_URL="https://ftp.isc.org/isc/bind9/9.12.0rc1/bind-9.12.0rc1.tar.gz"
```
miekg: Why do you compile it? And not just grab the package or other binary?

bind_test.go
```diff
@@ -0,0 +1,229 @@
...
+    file "%s";
+    allow-transfer { any; };
+    update-policy { %s };
+};
+%s
+`, zonefile, updatePolicy, hmacKeyStatements))
+	if err != nil {
+		return nil, nil, err
+	}
+	cmd := exec.Command("named", "-p", "55033", "-g", "-c", conffilePath)
```
miekg: fix port is going to give is trouble

install-bind.bash
andrewtj: To get ed25519 support.

miekg:
We should split this off in another repo and make it specifically do integration tests, maybe push some other bits there as well. I hate to wait for travis to compile bind9 on every push.
This also means we should setup an org #557

bind_test.go
andrewtj: Yep. I put a var on `bindHarnessInfo` for the port but ran out of time to grab a free port, and forgot to add a todo.

andrewtj:
>  I hate to wait for travis to compile bind9 on every push.

If I've understood and followed Travis's docs correctly it should get built once for each variant in the test matrix and then be cached. If that doesn't work it might be better to wait till a binary shows up.

andrewtj:
Closing this for the moment. I'm going to have a go at building a container housing a BIND install with the lot (DNSSEC, GSS-TSIG, SIG0, TSIG) plus whatever's needed for DNS over HTTP, DNS over TLS, and dnscrypt. If and when that's done I'll open another issue.

PR #615, opened 2018-01-04
DNS client: TCP connection reuse

### LINKED ISSUE #614: DNS client: TCP connection reuse for TCP and DNS-over-TLS
Hi, while using the DNS client I have noticed a performance challenge if a TCP or DNS-over-TLS server is used to resolve queries, this compared against an UDP server (that said, 1.5x-2.0x times slower for TCP, and up to 4x times slower for DNS-over-TLS), so in the long term, there is a drastic performance drop for a heavily used DNS client (e.x.: local server proxy/forwarder).

That performance drop is mainly caused by the opening/closing of a _new TCP connection for every DNS request_,  and the required handshake things that are inherit to the protocol (worse for a TCP TLS handshake).

So, as an improvement, it may be good to have an option to reuse an already made TCP connection and send the requests over it, reducing the handshake things over and over again, specially for a DNS-over-TLS server, where a cert is involved.

As a general idea:

- Add a new bool flag to enable/disable the reuse of TCP connections (not valid for UDP) [client.go#L30](https://github.com/miekg/dns/blob/master/client.go#L30)
- Add a new client attribute with the reusable connection [client.go#L30](https://github.com/miekg/dns/blob/master/client.go#L30)
- On every request, check if the reusable connection is NOT nil/closed, and use that connection [client.go#L160](https://github.com/miekg/dns/blob/master/client.go#L160)
- If the connection is nil/closed, make the `Dial` call to establish a new connection [client.go#L162](https://github.com/miekg/dns/blob/master/client.go#L162)
- Avoid closing the connection after the message exchange finishes [client.go#L167](https://github.com/miekg/dns/blob/master/client.go#L167)
- Add an idle timeout to check for inactive connections and automatically close them
- Add a handler to check/capture connection state changes (I don't know if that is possible)

daareiza:
With this PR the DNS client can reuse TCP connections when resolving against TCP/DNS-over-TLS servers, tries to resolve #614.

This is just a very basic approach of what it may look like, I am 100% sure than can be improved in many different ways, community feedback/commits are welcome.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 tags
 test.out
 a.out
+.idea/
--- a/client.go
+++ b/client.go
@@ -10,6 +10,7 @@ import (
 	"io"
 	"net"
 	"strings"
+	"sync"
 	"time"
 )
 
@@ -42,6 +43,15 @@ type Client struct {
 	TsigSecret     map[string]string // secret(s) for Tsig map[<zonename>]<base64 secret>, zonename must be in canonical form (lowercase, fqdn, see RFC 4034 Section 6.2)
 	SingleInflight bool              // if true suppress multiple outstanding queries for the same Qname, Qtype and Qclass
 	group          singleflight
+
+	ReuseTCPConn        bool          // if true reuses "tcp" or "tcp-tls" client TCP connection
+	ReuseTCPIdleTimeout time.Duration // TCP reuse idle timeout, automatically close connection if the connection is not used after specified timeout, defaults to 5 seconds
+
+	cmu  sync.Mutex // reusable connection mutex
+	conn *Conn      // reusable connection
+
+	cimu     sync.RWMutex // reusable connection idle mutex
+	connIdle time.Time    // reusable connection idle timeout
 }
 
 // Exchange performs a synchronous UDP query. It sends the message m to the address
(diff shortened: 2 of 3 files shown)


miekg:
Good stuff. But this might be more complex then we think and it should not be limited to tcp. 
In github.com/coredns/forward I've implemented connection caching - so this is already possible using this library, but I agree having this "by default" would be good.

We should probably morph that code one way or another into here?

daareiza:
Thanks for your comment, Miek. The UDP socket caching/pool behavior is nicely done, but perhaps UDP itself is out of scope for this PR. The main reason for creating this PR was to solve the performance drop noticed while using a DNS-over-TLS upstream server without a UDP fallback, and preventing any ceilings from being reached if used as a forwarder in a high-traffic network.

Perhaps the best thing to do is open a new issue for porting the superior UDP behavior to this project.

If there is any specific coding practice from the coredns/forwarder code you'd like to point out that can relate to tcp/tls connection, I'd definitely be interested.

If you have a spare cycle or two, I'd appreciate any general tips about this PR.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] DNS client: TCP con..." ]
>Thanks for your comment, Miek. The UDP socket caching/pool behavior is nicely done, but perhaps UDP itself is out of scope for this PR. The main reason for creating this PR was to solve the performance drop noticed while using a DNS-over-TLS upstream server without a UDP fallback, and preventing any ceilings from being reached if used as a forwarder in a high-traffic network.
>
>Perhaps the best thing to do is open a new issue for porting the superior UDP behavior to this project.
>
>If there is any specific coding practice from the coredns/forwarder code you'd like to point out that can relate to tcp/tls connection, I'd definitely be interested.
>
>If you have a spare cycle or two, I'd appreciate any general tips about this PR.

I'm afraid that caching connections will open a can of worms...
Like when to discard a connection, or to de-dial when a write fails - which are 
difficult problems to solve.

Having said that, conn caching is a powerful feature that will benefit a lot of 
users. So if we are attempting to do this we should do it all for everyrhing, 
technically boiling down to UDP and TCP.

Code wise this means copying quite a lot from forward.

daareiza:
Closed, I will wait for an official implementation of this feature that hopefully complies with the RFC specification. Thanks Miek for all your time.

PR #617, opened 2018-01-06
Fixes #613

### LINKED ISSUE #613: CPU gets pegged at 100% if a udp listener fails / is closed
**Problem**

The issue is caused by this: https://github.com/miekg/dns/blob/master/server.go#L511

There is an extremely tight CPU-bound loop that the thread gets locked into in the event of any permanent error with `server.readUDP()`. This bug causes excessive user->kernel context switches (due to IO) and pegs a CPU.

**Fix**

The following fix mitigated the issue for us: (line 510ish of server.go)

```go
if err != nil {
	if strings.Contains(err.Error(), "use of closed network connection") {
		return err
	}
	continue
}
```

Unfortunately `internal/poll.ErrNetClosing` is not exported, and I couldn't think of another way to check the error.

A better (and much more efficient) way to do this would be to `continue` only for io timeout errors, and `return err` for all other errors. I'll leave such decisions to you.

twitchyliquid64:
I tested this quickly in production. Can confirm io timeout from deadlines is err.Temporary() == true and use of closed connection is not opError, so it works in prod for me.

### DIFF
--- a/server.go
+++ b/server.go
@@ -509,7 +509,7 @@ func (srv *Server) serveUDP(l *net.UDPConn) error {
 		}
 		srv.lock.RUnlock()
 		if err != nil {
-			if opErr, ok := err.(*net.OpError); ok && opErr.Temporary() {
+			if netErr, ok := err.(net.Error); ok && netErr.Temporary() {
 				continue
 			}
 			return err
--- a/server.go
+++ b/server.go
@@ -509,7 +509,10 @@ func (srv *Server) serveUDP(l *net.UDPConn) error {
 		}
 		srv.lock.RUnlock()
 		if err != nil {
-			continue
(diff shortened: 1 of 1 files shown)


server.go
```diff
@@ -509,7 +509,10 @@ func (srv *Server) serveUDP(l *net.UDPConn) error {
 		}
 		srv.lock.RUnlock()
 		if err != nil {
-			continue
+			if opErr, ok := err.(*net.OpError); ok && opErr.Temporary() {
```
tmthrgd: Wouldn't it be better - and more universal - to cast to the [`net.Error` interface](https://golang.org/pkg/net/#Error) here?

twitchyliquid64: Done.

Issue #577, opened 2017-11-18
Using 'ANY' as class in presentation format


spsholleman:
This works fine
	x, xErr := NewRR("www.miek.nl. 3600 IN TXT \"first\" \"second\"")

but this doesn't:
	x, xErr := NewRR("www.miek.nl. 3600 ANY TXT \"first\" \"second\"")
	if xErr != nil {
		t.Fatalf("failed to parse TXT: %s", xErr)
	}

it fails with the following error:
	dns_test.go:461: failed to parse TXT: dns: unknown RR type: "TXT" at line: 1:26

I assume this should work - or am I confused?

miekg:
[ Quoting <notifications@github.com> in "[miekg/dns] Is (#577)..." ]
>This works fine
>	x, xErr := NewRR("www.miek.nl. 3600 IN TXT \"first\" \"second\"")
>
>but this doesn't:
>	x, xErr := NewRR("www.miek.nl. 3600 ANY TXT \"first\" \"second\"")
>	if xErr != nil {
>		t.Fatalf("failed to parse TXT: %s", xErr)
>	}
>
>it fails with the following error:
>	dns_test.go:461: failed to parse TXT: dns: unknown RR type: "TXT" at line: 1:26
>
>I assume this should work - or am I confused?

I don't know. What section in which RFC makes you think this?

spsholleman:
I was having problems getting the parser to recognize TSIG or TKEY when ANY was used so I came up with a TXT example.  That may not have been the best idea.  I agree the TXT RFC says it should be of IN class but RFC 2845 and 2930 state that TSIG and TKEY should be of class ANY but don't provide any suggested textual representation.  All three return the same error about not recognizing the RR type and it seemed that the parser just doesn't support the ANY class.

I can try and dig into it later but figured others more familiar with the parser might immediately know why.

gibson042:
"ANY" is not a real class mnemonic (cf. [DNS CLASSes](https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#dns-parameters-2)), and we're right to reject it. Per [RFC 3597](https://tools.ietf.org/html/rfc3597#section-5), the proper text representation is "CLASS255".

andrewtj:
That seems like a poor error message if it's an intentional failure due to the CLASS.

It also seems strange to accept CLASS255 since it'll then be presented as ANY and will then fail to parse on a subsequent parse. Eg:

```
package main

import (
	"log"

[...]

PR #618, opened 2018-01-07
ClassANY: don't convert CLASS255 to ANY

miekg:
Class "ANY" is wireformat only. In zonefile you can use CLASS255, but
when String-ing we convert this into "ANY" which is wrong. I.e. this
means we can't read back our own update.

Bit of a kludge to work around this, as I'm not sure we can just remove
ANY from the ClassToString map.

### DIFF
--- a/defaults.go
+++ b/defaults.go
@@ -273,13 +273,11 @@ func (t Type) String() string {
 
 // String returns the string representation for the class c.
 func (c Class) String() string {
-	if c1, ok := ClassToString[uint16(c)]; ok {
-		// Special case ANY, even though it's in the map we should not translate CLASS255 to ANY.
-		if c == 255 {
-			return "CLASS255"
+	if s, ok := ClassToString[uint16(c)]; ok {
+		// Only emit mnemonics when they are unambiguous, specically ANY is in both.
+		if _, ok := StringToType[s]; !ok {
+			return s
 		}
-
-		return c1
 	}
 	return "CLASS" + strconv.Itoa(int(c))
 }
--- a/defaults.go
+++ b/defaults.go
@@ -274,6 +274,11 @@ func (t Type) String() string {
 // String returns the string representation for the class c.
 func (c Class) String() string {
 	if c1, ok := ClassToString[uint16(c)]; ok {
+		// Special case ANY, even though it's in the map we should not translate CLASS255 to ANY.
+		if c == 255 {
+			return "CLASS255"
+		}
+
 		return c1
 	}
 	return "CLASS" + strconv.Itoa(int(c))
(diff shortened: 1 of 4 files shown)


miekg:
I wouldn't say fixes, but ref #577

defaults.go
```diff
@@ -274,6 +274,11 @@ func (t Type) String() string {
 // String returns the string representation for the class c.
 func (c Class) String() string {
 	if c1, ok := ClassToString[uint16(c)]; ok {
+		// Special case ANY, even though it's in the map we should not translate CLASS255 to ANY.
```
gibson042: :+1:

defaults.go
```diff
@@ -274,6 +274,11 @@ func (t Type) String() string {
...
+		if c == 255 {
```
gibson042: Better still (cf. [section 3.2 of RFC 2929](https://tools.ietf.org/html/rfc2929#section-3.2)):
```go
// Output mnemonics only for ranges not reserved by RFC 2929.
// This could also exclude 0 and 0xFFFF ("assignment requires an IETF Standards Action"),
// and maybe also 0xFF00 through 0xFFFE ("Private Use").
if c < 128 || c > 255 {
	if c1, ok := ClassToString[uint16(c)]; ok {
		return c1
	}
}
return "CLASS" + strconv.Itoa(int(c))
```

And we should also to the same for Type.

miekg: But wouldn't reserving by the IETF mean we put it in this ClassToString? The private range might bite people extending this map with a private range class?

What are you looking for, for types? Note that we actually sport a `ANY` RR? /me ducks

gibson042: RFC 2929 reserves CLASS values 128 through 255 "as QCLASSes only" (with only two actually assigned: 254 NONE and 255 ANY), which means that any value in that range will never be valid in a zone file. Similar but weaker reasoning applies to 0 and 0xFFFF, and I'm not sure about the private use range, but I guess I'd trust users to only add to the maps if they really want the values treated as standard.

But maybe none of that really matters, because all we really want is to emit strings that we can read back. Updated suggestion:
```go
func (c Class) String() string {
	if s, ok := ClassToString[uint16(c)]; ok {
		// Only emit mnemonics when they are unambiguous.
		if _, ok := StringToType[s]; !ok {
			return s
		}
	}
	return "CLASS" + strconv.Itoa(int(c))
}
```

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] ClassANY: don't con..." ]
>But maybe none of that really matters, because all we really want is to emit strings that we can read back. Updated suggestion:
>```go
>func (c Class) String() string {
>	if s, ok := ClassToString[uint16(c)]; ok {
>		// Only emit mnemonics when they are unambiguous.
>		if _, ok := StringToType[s]; !ok {
>			return s
>		}
>	}
>	return "CLASS" + strconv.Itoa(int(c))
>}
>```

Done

miekg:
Let's discuss that in a new bug. If we give types priority (as done here), we should be OK to keep using ANY as a TYPE255 alias (how bad this may be from a standards perspective)

gibson042:
It's hard to say if that will be bad... we shouldn't emit strings that we can't parse back in, but "ANY" for TYPE isn't necessarily problematic in and of itself. I guess I'm comfortable with giving this a try.

Issue #619, opened 2018-01-08
dns: short read lock


MikeSchroll:
This morning, having done a new release yesterday with the latest miekg/dns, we began receiving traffic (even traffic we don't consider authorized, and drop) to a dev instance which caused our server code to lock up, and not process any additional incoming queries.

This is the error message:
level=error msg=“unable to listen” address=“REDACTED:53" error=“dns: short read” log_message=“unable to listen” protocol=udp region=FRA serverID=71302

Initially looking into it, we suspected the commit https://github.com/miekg/dns/commit/ac8cd7878c723b698420572df92b293dd041d9ff tied to issue:
https://github.com/miekg/dns/issues/613

We've noticed that this appears to produce errors and lock only on each IP/port we're listening on; and so for instance one server which listens on 20+ IPs/ports had 5 locked from this error. The remaining continued to function.
We've seen this only on UDP, but that may be a factor of it's where most of our traffic comes.

We've initially confirmed this suspicion by rolling back miekg/dns versions, and have not had the error recur. We do not have a way to replicate the error, but we'd like to open this new issue with hopes that others will have some thoughts on the change which is causing this.

tmthrgd:
@MikeSchroll ac8cd78 almost certainly exposed the error you're seeing, *but* that error has always existed, it's simply that before ac8cd78 it was ignored.

The error is coming from https://github.com/miekg/dns/blob/dcdbddd8102b79eeef1329473dffab4aec17ca25/server.go#L636 which is caused by https://github.com/miekg/dns/blob/dcdbddd8102b79eeef1329473dffab4aec17ca25/udp.go#L24 returning `n == 0 && err == nil`.

I'm not familiar enough to say what the correct behaviour here is, or whether this error is safe to ignore, but a simply patch to address this issue is:
```
From 58d0c781273a9fe2e10303c91eb5b4b5e31071fc Mon Sep 17 00:00:00 2001
From: Tom Thorogood <me+github@tomthorogood.co.uk>
Date: Tue, 9 Jan 2018 12:29:43 +1030
Subject: [PATCH] Ignore ErrShortRead (*Server).serveUDP

Fixes #619
---
 server.go | 3 +++
 1 file changed, 3 insertions(+)

diff --git a/server.go b/server.go
index 59135b0f53..171b66ed8b 100644
--- a/server.go
+++ b/server.go
@@ -512,6 +512,9 @@ func (srv *Server) serveUDP(l *net.UDPConn) error {
[...]

PR #622, opened 2018-01-09
Ignore malformed UDP datagrams with incomplete DNS headers

tmthrgd:
See https://github.com/miekg/dns/issues/619#issuecomment-356270075 onwards for rationale.

### DIFF
--- a/server.go
+++ b/server.go
@@ -514,6 +514,10 @@ func (srv *Server) serveUDP(l *net.UDPConn) error {
 			}
 			return err
 		}
+		const msgHeaderLen = 12
+		if len(m) < msgHeaderLen {
+			continue
+		}
 		go srv.serve(s.RemoteAddr(), handler, m, l, s, nil)
 	}
 }
--- a/server.go
+++ b/server.go
@@ -514,8 +514,7 @@ func (srv *Server) serveUDP(l *net.UDPConn) error {
 			}
 			return err
 		}
-		const msgHeaderLen = 12
-		if len(m) < msgHeaderLen {
+		if len(m) < headerSize {
 			continue
 		}
 		go srv.serve(s.RemoteAddr(), handler, m, l, s, nil)


tmthrgd:
@miekg Would you be able to manually run the test you used in https://github.com/miekg/dns/issues/619#issuecomment-356283121 and https://github.com/miekg/dns/issues/619#issuecomment-356283982 to test this pull-request for correctness?

server.go
```diff
@@ -514,6 +514,10 @@ func (srv *Server) serveUDP(l *net.UDPConn) error {
 			}
 			return err
 		}
+		const msgHeaderLen = 12
```
fastest963: I'll leave it up to @miekg but there's a few spots where we just hardcode 12 with a comment (like `compressedLen`). So not sure if we should just hardcode it in the `if` and add a comment above the if instead of defining this const.

tmthrgd:
/CC @twitchyliquid64 (who initially noted the amplification vector).

server.go
tmthrgd: This const is only scoped to the for loop - so it doesn't leak anywhere - and it's definitely what I would consider idiomatic go. It is very much up to @miekg though. My vote is squarely for the const, but a comment would still convey the same information (albeit with slightly less clarity).

miekg: there is a const headerSize

tmthrgd: @miekg I went looking for it, but obviously I didn’t look hard enough. Done.

miekg:
TCP is harder to abuse, but for the sake of symmetry we might need something similar as well.

miekg:
OK. runing coredns with this applied.

miekg:
This behaves similar to the other servers I tried.

miekg:
deb.atoom.net if you want to have a whack

tmthrgd:
@miekg I’m not sure what other implementations do for TCP, but I don’t think this poses for TCP because it a the handshake to verify the source address. I think it would be wrong to ignore any message in the TCP path as it could put the stream in an inconsistent state. That being said, I did find a few bugs in the TCP serving code (including a possible denial of service attack against the server) that I’ll investigate properly tomorrow.

miekg:
I'll merge this tomorrow morning (UTC), unless something really weird shows up in the logs.

twitchyliquid64:
Because of the 3 way handshake and sequence numbers, its not possible to setup a TCP session with a spoofed source IP. You don't need to consider TCP a problem for DNS amplification.

A empty UDP packet is 28 bytes. If an error response is 12 bytes, that means the amplification ratio is 1.42, which is significant.

This PR looks good to me.

PR #623, opened 2018-01-10
Fix TCP Shutdown 'use of closed network connection'

tmthrgd:
The check for `srv.started` being false is in the wrong place, it should be after `Accept` not after `ReadTCP`. If `Shutdown` is called, `serveTCP` will currently return a 'use of closed network connection' error, which is undesired.

This commit mirrors the behaviour of `serveUDP` with respect to `Shutdown`.

### DIFF
--- a/server.go
+++ b/server.go
@@ -460,6 +460,12 @@ func (srv *Server) serveTCP(l net.Listener) error {
 	// deadline is not used here
 	for {
 		rw, err := l.Accept()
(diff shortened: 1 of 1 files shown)


tmthrgd:
@miekg I have a test case to cover this if there's interest in merging it, I'll open another pull request.

miekg:
yes please do

PR #624, opened 2018-01-10
Test that Shutdown does not surface closed errors

tmthrgd:
This test checks that calling `Shutdown` does not cause `ActivateAndServe` (via `serveTCP` and `serveUDP`) to return the underlying 'use of closed network connection' error.

This commit unifies `TestShutdownTCP` with `TestShutdownUDP`. After this commit, both tests will check that `ActivateAndServe` returns a nil error and that `Shutdown` succeeded.

This was previously broken for `serveTCP` (#623).

I didn't add this to `TestShutdownTLS` as the shutdown path is identical to TCP, but I can add it to this pull request if desired.

### DIFF
--- a/server_test.go
+++ b/server_test.go
@@ -55,7 +55,7 @@ func RunLocalUDPServer(laddr string) (*Server, string, error) {
 	return server, l, err
 }
 
-func RunLocalUDPServerWithFinChan(laddr string) (*Server, string, chan struct{}, error) {
+func RunLocalUDPServerWithFinChan(laddr string) (*Server, string, chan error, error) {
 	pc, err := net.ListenPacket("udp", laddr)
 	if err != nil {
 		return nil, "", nil, err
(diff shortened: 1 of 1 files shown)


server_test.go
```diff
@@ -66,11 +66,10 @@ func RunLocalUDPServerWithFinChan(laddr string) (*Server, string, chan struct{},
 	waitLock.Lock()
 	server.NotifyStartedFunc = waitLock.Unlock
 
-	fin := make(chan struct{}, 0)
+	fin := make(chan error, 1)
```
miekg: why did you make this buffered?

tmthrgd: @miekg It needs to be buffered in case fin is never read from. (This can happen if Shutdown fails or the test cases timeout). If it was left unbuffered, the goroutine below would block forever.

miekg: Can we add a comment of this nature?

tmthrgd: Will do.

tmthrgd: @miekg I spent far longer than I care to admit figuring out what to say. Hopefully b0869bd is good.

PR #631, opened 2018-01-25
Fix for CVE-2017-15133 TCP DOS

miekg:
serveTCP calls reader.ReadTCP in the accept loop rather than in
the per-connection goroutine. If an attacker opens a connection
and leaves it idle, this will block the accept loop until the
connection times out (2s by default). During this time no other
incoming connections will succeed, preventing legitimate queries
from being answered.

This commit moves the call to reader.ReadTCP into the per-connection
goroutine. It also adds a missing call to Close whose absence allowed
file-descirptors to leak in select cases.

This attack and fix have no impact on serving UDP queries.

### DIFF
--- a/server.go
+++ b/server.go
@@ -472,11 +472,14 @@ func (srv *Server) serveTCP(l net.Listener) error {
 			}
 			return err
 		}
-		m, err := reader.ReadTCP(rw, rtimeout)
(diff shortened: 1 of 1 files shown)


miekg:
Fixes #631

PR #632, opened 2018-01-27
Server: check client udp buf size

miekg:
If we are returning a large response we now just send it (for udp) even
if the clients advertised buffer will be too small to take it all.
Make the response struct keep track of this udp size and cap returns
that exceed this.

Note this is irrespective of the TC bit in the message; we truncate
regardless.

This fails a test; but the test is wrong.

### DIFF
--- a/server.go
+++ b/server.go
@@ -51,6 +51,7 @@ type response struct {
 	udp            *net.UDPConn      // i/o connection if UDP was used
 	tcp            net.Conn          // i/o connection if TCP was used
 	udpSession     *SessionUDP       // oob data to get egress interface right
+	udpSize        int               // client advertised buf size
 	remoteAddr     net.Addr          // address of the client
 	writer         Writer            // writer to output the raw DNS bits
 }
@@ -564,6 +565,18 @@ Redo:
 			w.tsigRequestMAC = req.Extra[len(req.Extra)-1].(*TSIG).MAC
 		}
 	}
+
+	// If EDNS0 is used use that for size, otherwise we hard cap at go'old 512 bytes.
+	// This value is used when we write the UDP response back to the client.
+	// ignored when we're doing TCP, to signal this to WriteMsg we leave this value at 0.
+	if w.udp != nil {
+		w.udpSize = MinMsgSize
+		opt := req.IsEdns0()
+		if opt != nil && opt.UDPSize() >= MinMsgSize {
+			w.udpSize = int(opt.UDPSize())
+		}
+	}
+
 	h.ServeDNS(w, req) // Writes back to the client
 
 Exit:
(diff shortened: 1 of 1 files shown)


miekg:
cc @grobie

grobie:
I'm not familiar with this code base, but I understand the change to truncate the byte stream returned to the client? Does this mean the last resource record in a response might be incomplete? What about the TC or the rr count headers, or potentially generated signatures?

My plan was to add a `Truncate() error` function to this library, which could be used to ensure the returned message size would not exceed the advertised buffer size by removing Additional, NS and Answer records before writing the message and passing it on to more low-level WriteMsg methods.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Server: check clien..." ]
>I'm not familiar with this code base, but I understand the change to truncate the byte stream returned to the client? Does this mean the last resource record in a response might be incomplete? What about the TC or the rr count headers, or potentially generated signatures?

Yes. The TC and header counters are in the first 12 bytes, so I can't see how 
that could ever be truncated?

>My plan was to add a `Truncate() error` function to this library, which could be used to ensure the returned message size would not exceed the advertised buffer size by removing Additional, NS and Answer records before writing the message and passing it on to more low-level WriteMsg methods.

That might work as well (but is more work). But might be nicer; we still need to 
call it our-selves before writing; so we could do this by default (i.e. when 
done like that truncate() could be made private)

/Miek

--
Miek Gieben

grobie:
I don't understand the goal of this PR. If all messages have been truncated already before being passed down to this response writer, this change is unnecessary. If they haven't, then this change creates invalid messages (breaks at least tsig, and TC and counter headers).

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Server: check clien..." ]
>I don't understand the goal of this PR. If all messages have been truncated already before being passed down to this response writer, this change is unnecessary. If they haven't, then this change creates invalid messages (breaks at least tsig, and TC and counter headers).

This change fixes the problem you saw; sending more bytes then necessary.

TC, headers are not broken at all. TSIG is already broken with TC anyway? 

I'm there good be more lenient ways of doing this, but not for the reason you 
mention here

grobie:
> This change fixes the problem you saw; sending more bytes then necessary.

I'm not sure what statement of mine you're referring to?

I saw CoreDNS and and the Server in miekg/dns not implementing truncation properly. I also saw several cases of `Msg.Len()` in miekg/dns not returning a new byte length after setting `msg.Extra = nil`, which meant the following return path couldn't be reached:

https://github.com/coredns/coredns/blob/449bb1ac6ce1b58cf14caa2daf7f9520a36901ef/request/request.go#L199-L202

Unfortunately, this change doesn't fix either of these issues.

> TC, headers are not broken at all.

This change truncates messages and removes resource records without setting the TC header. This violates the DNS specifcation and can cause all kinds of application issues depending on the use case.

Removing resource records after the ANCOUNT/NSCOUNT/ARCOUNT headers have already been set to different values will cause some clients I have seen to discard the message as invalid.

> TSIG is already broken with TC anyway?

I'm not aware of such limitation, can you link to an explanation why that would be the case? I'd expect a message could be easily sufficiently truncated to make enough room for the tsig record.

miekg:
solved in coredns with https://github.com/coredns/coredns/pull/1543

PR #636, opened 2018-02-02
client: change Exchange signature to match docs

jeloou:


miekg:
I like `address` here, better to update the docs? Could you change to that?

PR #638, opened 2018-02-07
Make it possible to use a custom reader to generate DNSKEY

cesarkuroiwa:
New method (k *DNSKEY) GenerateCustom that receives a custom io.Reader, so that one can choose to use a source of entropy other than rand.Reader

### DIFF
--- a/dnssec_keygen.go
+++ b/dnssec_keygen.go
@@ -10,6 +10,7 @@ import (
 	"math/big"
 
 	"golang.org/x/crypto/ed25519"
+	"io"
 )
 
 // Generate generates a DNSKEY of the given bit size.
@@ -19,6 +20,10 @@ import (
 // The ECDSA algorithms imply a fixed keysize, in that case
 // bits should be set to the size of the algorithm.
 func (k *DNSKEY) Generate(bits int) (crypto.PrivateKey, error) {
+	return k.GenerateCustom(bits, rand.Reader)
+}
+
+func (k *DNSKEY) GenerateCustom(bits int, reader io.Reader) (crypto.PrivateKey, error) {
 	switch k.Algorithm {
 	case DSA, DSANSEC3SHA1:
 		if bits != 1024 {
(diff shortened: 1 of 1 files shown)


miekg:
Sorry, I don't see why this would make sense to add to the library.
I get that you want this, but can't see this being tremendously useful for
anyone else.

On 7 Feb 2018 17:20, "cesarkuroiwa" <notifications@github.com> wrote:

New method (k *DNSKEY) GenerateCustom that receives a custom io.ReaderNew
method (k *DNSKEY) GenerateCustom that receives a custom io.Reader, so that
one can choose to use a source of entropy other than rand.Reader
------------------------------
You can view, comment on, or merge this pull request online at:

  https://github.com/miekg/dns/pull/638
Commit Summary

   - New method (k *DNSKEY) GenerateCustom that receives a custom io.Reader

File Changes

   - *M* dnssec_keygen.go
   <https://github.com/miekg/dns/pull/638/files#diff-0> (16)

Patch Links:

   - https://github.com/miekg/dns/pull/638.patch
   - https://github.com/miekg/dns/pull/638.diff

—
You are receiving this because you are subscribed to this thread.
<https://github.com/miekg/dns/pull/638>, or mute the thread

cesarkuroiwa:
This feature would be really useful for us because we are developing an application that manages a bunch of dnskeys and it is important that the keys are generated using a strong Random Number Generator, other than simply rand.Reader.
Moreover, since the golang standard library makes it possible to use a custom io.Reader to generate keys, I don't see why limit this for dnskeys

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Make it possible to..." ]
>This feature would be really useful for us because we are developing an application that manages a bunch of dnskeys and it is important that the keys are generated using a strong Random Number Generator, other than simply rand.Reader.
>Moreover, since the golang standard library makes it possible to use a custom io.Reader to generate keys, I don't see why limit this for dnskeys

If rand.Reader is not good enough, we should probably switch it instead of 
adding to the API

tmthrgd:
This somewhat feels like a case of the [XY problem](http://xyproblem.info/) to me. @cesarkuroiwa Why do you feel like [`crypto/rand.Reader`](https://golang.org/pkg/crypto/rand/#pkg-variables) is insufficiently 'strong' for you?

It's pretty well established that `/dev/urandom` (either as a character device or exposed via a `getrandom(2)` or `getentropy(2)` syscall) is [the Right Way™](https://www.2uo.de/myths-about-urandom/) to generate cryptographically secure random data and in Golang this is exposed through `crypto/rand.Reader`. It would be a mistake imo to allow callers to change this and shoot themselves in the foot.

miekg:
please rework this to just default to `crypto/rand.Reader` then we can merge.

miekg:
won't merge as-is. Closing here.

PR #647, opened 2018-03-17
WIP: DOH helper functions

miekg:
IETF hackathon

I was trying to make DNS over HTTPs a first class citizen, i.e. create a
dns.Conn and set Net to https (basically just like DNS over TLS), but
this ran into a few snags:

* http/2 has it's own socket pipelining so I can't give it a net.Conn
  and make it run with it. This breaks an assumption in this library.
* The Go http client setup has a function called client.Do which sends
  a reply and waits for a reponse, ala Exchange() in this library. But
  this library uses conn.WriteMsg and conn.ReadMsg as two calls to
  implement that - both of these are also public. So I could fake one of
  these in the case of http/2, i.e. make WriteMsg a noop and do all the
  work in ReadMsg, but I still don't have access to the conn used in the
  http/2 - so that won't fly either.

In the end I ended up with some helper functions that parse from and to
the http/2 request world, but nothing more tangible can be done in this
library. In this regard it is similar to the gRPC support we have in
CoreDNS which is also outside of this library.

### DIFF
--- a/https.go
+++ b/https.go
@@ -1,0 +1,47 @@
+package dns
+
+import (
+	"bytes"
+	"io"
+	"io/ioutil"
+	"net/http"
+)
+
+const mime = "application/dns-udpwireformat"
+
+// MsgToRequest wraps m in a http POST request according to the DNS over HTTPS Spec.
+func MsgToRequest(m *Msg, url string) (*http.Request, error) {
+	out, err := m.Pack()
+	if err != nil {
+		return nil, err
+	}
+	req, err := http.NewRequest("POST", url, bytes.NewReader(out))
+	req.Header.Set("content-type", mime)
+	req.Header.Set("accept", mime)
+
+	return req, nil
+}
+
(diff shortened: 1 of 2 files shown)


https.go
```diff
@@ -0,0 +1,47 @@
...
+const mime = "application/dns-udpwireformat"
+
+// MsgToRequest wraps m in a http POST request according to the DNS over HTTPS Spec.
+func MsgToRequest(m *Msg, url string) (*http.Request, error) {
+	out, err := m.Pack()
+	if err != nil {
+		return nil, err
+	}
+	req, err := http.NewRequest("POST", url, bytes.NewReader(out))
+	req.Header.Set("content-type", mime)
```
tmthrgd: This will cause an allocation. If you use Go’s header case, that’s avoided. So: Content-Type.

https.go
```diff
@@ -0,0 +1,47 @@
...
+	req.Header.Set("accept", mime)
```
tmthrgd: Ditto: Accept.

PR #648, opened 2018-03-18
WIP: TCP: allow out of order TCP processing

miekg:
Kick off another go routine that handles processing a message. This
fails because the response writer is not thread safe.

### DIFF
--- a/server.go
+++ b/server.go
@@ -564,7 +564,7 @@ Redo:
 			w.tsigRequestMAC = req.Extra[len(req.Extra)-1].(*TSIG).MAC
 		}
 	}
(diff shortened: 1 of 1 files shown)


tmthrgd:
@miekg I don't think this is possible or safe without substantially rewriting `serve`. It's certainly true there are many more data-races and edge-cases in this PR.

miekg:
I agree. Also an implicit contract for a reponsewriter is that is will be used to handle the current query; meaning concurrent use of these was never a thing. And note that responseWriters are everywhere. I think in CoreDNS we already have 6 or so...

PR #651, opened 2018-03-19
WIP: DNS-over-HTTP support for Client using x/net/http2

tmthrgd:
This is another quick attempt at adding DNS-over-HTTP support to `Client`, inspired by https://github.com/miekg/dns/pull/647#issue-175710683.

This method uses `x/net/http2` and an already dialed `net.Conn`.

Note: Unlike #649 dial errors are surfaced in `Dial` not `Write`.


MikeSchroll:
What's the status of this?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] WIP: DNS-over-HTTP ..." ]
>What's the status of this?

I rather not merge this but do the correct thing, which is adding a 
http.RoundTripper interface to Client (or have a new Client). However that is 
way more work (maybe)

tmthrgd:
@miekg If you don’t mind the asymmetry (between Client and Conn), I’d suggest supporting it in Exchange* would be the best approach. (@MikeSchroll would that cover your needs?)

It would always be possible to push DoH support further down into Conn (like this and #649 attempted to do) if need be.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] WIP: DNS-over-HTTP ..." ]
>@miekg If you don’t mind the asymmetry (between Client and Conn), I’d suggest supporting it in Exchange* would be the best approach. (@MikeSchroll would that cover your needs?)

No, I don't mind. Roundtripper can retro-actively fix that and maybe make 
dns.Conn obsolete (or needed less)

/Miek

--
Miek Gieben

PR #656, opened 2018-03-27
use a local variable to calculate rtt

clmul:
This makes `Conn` safe for concurrent use if the message doesn't have a TSIG record.

### DIFF
--- a/client.go
+++ b/client.go
@@ -21,8 +21,6 @@ type Conn struct {
 	net.Conn                         // a net.Conn holding the connection
 	UDPSize        uint16            // minimum receive buffer for UDP messages
 	TsigSecret     map[string]string // secret(s) for Tsig map[<zonename>]<base64 secret>, zonename must be in canonical form (lowercase, fqdn, see RFC 4034 Section 6.2)
-	rtt            time.Duration
-	t              time.Time
 	tsigRequestMAC string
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -179,6 +177,7 @@ func (c *Client) exchange(m *Msg, a string) (r *Msg, rtt time.Duration, err erro
 	co.TsigSecret = c.TsigSecret
 	// write with the appropriate write timeout
 	co.SetWriteDeadline(time.Now().Add(c.getTimeoutForRequest(c.writeTimeout())))
+	t := time.Now()
```
tmthrgd: If you moved this up a line, we could avoid calling `time.Now` twice. It’s a small thing, but it does come with overhead.

clmul:
> I don’t know how possible it would ever be to use Conn concurrently, this makes sense to me.

I reconsidered this and I agree that it's better to not use `Conn` concurrently.

client.go
clmul: Fixed.

miekg:
So `dns.Conn` came about because people wanted to use `laddr` and other `Dial` properties which were not exposed by just using Exchange. I think most of these tweaks should just move to `Client`, so we can drop all these exposed bits.

Issue #660, opened 2018-04-02
UDP OOB failures on ARM under QEMU


tmthrgd:
As first noted in https://github.com/miekg/dns/pull/653#issuecomment-377780398, something is wrong on ARM. This was what I discovered in https://github.com/miekg/dns/pull/653#issuecomment-377835878 and https://github.com/miekg/dns/pull/653#issuecomment-377836333.

Kernel and Golang version:
```
$ uname -a
Linux thrax 4.15.13-300.fc27.x86_64 #1 SMP Mon Mar 26 19:06:57 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

$ go version
go version go1.10 linux/amd64
```

Running with:
```
GOARCH=arm go test -c .
qemu-arm-static ./dns.test
```

Gives this line interspersed throughout the output:
> Unsupported ancillary data: 41/50

It also gives this ~interesting~ crash, ~I'm pretty sure something is very wrong with the OOB handling on ARM~:
```
--- FAIL: TestParseDstFromOOB (0.00s)
        udp_test.go:82: failed to parse IPv4 in IPv6: <nil>
        udp_test.go:92: failed to parse IPv6: <nil>
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x20cbb4]

goroutine 587 [running]:
testing.tRunner.func1(0x10984000)
        /usr/lib/golang/src/testing/testing.go:742 +0x264
panic(0x38ced8, 0x5a5268)
        /usr/lib/golang/src/runtime/panic.go:505 +0x204
github.com/tmthrgd/dns/vendor/golang.org/x/net/ipv4.(*ControlMessage).Parse(0x10964060, 0x10ad0040, 0x20, 0x20, 0x0, 0x0)
        /home/tom/go/src/github.com/tmthrgd/dns/vendor/golang.org/x/net/ipv4/control.go:90 +0x2a4
github.com/tmthrgd/dns.parseDstFromOOB(0x10ad0040, 0x20, 0x20, 0x1096ef58, 0x1, 0x1)
        /home/tom/go/src/github.com/tmthrgd/dns/udp.go:68 +0xb4
github.com/tmthrgd/dns.TestParseDstFromOOB(0x10984000)
        /home/tom/go/src/github.com/tmthrgd/dns/udp_test.go:99 +0x318
[...]

PR #661, opened 2018-04-02
Skip UDP OOB tests on unsupported architectures

tmthrgd:
See https://github.com/miekg/dns/issues/660 for rationale and discussion.

### DIFF
--- a/udp_test.go
+++ b/udp_test.go
@@ -5,6 +5,7 @@ package dns
 import (
 	"bytes"
 	"net"
+	"runtime"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


udp_test.go
```diff
@@ -74,6 +75,14 @@ func TestSetUDPSocketOptions(t *testing.T) {
 }
 
 func TestParseDstFromOOB(t *testing.T) {
+	if runtime.GOARCH != "amd64" {
```
miekg: //build line in the file?

miekg: yeah, it needs a new file.... I'll just merge as-is

fastest963:
@tmthrgd can we use the ipv4/ipv6 package to just create the oob data to then test those methods? Would that get around the OS-dependent problem?

tmthrgd:
@fastest963 That was my plan, but the test data puts IPv4 addresses into v6 control messages, which x/net expressly doesn’t support.

fastest963:
@tmthrgd seems useful to do the other test cases just to verify that there isn't any non-amd64 breakage and just separate the test case with ipv4 in ipv6 specifically only on amd64. @miekg what do you think?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Skip UDP OOB tests ..." ]
>@tmthrgd seems useful to do the other test cases just to verify that there isn't any non-amd64 breakage and just separate the test case with ipv4 in ipv6 specifically only on amd64. @miekg what do you think?

Are we testing it wrongly (from the viewpoint of x/net) ? If so, we should just 
update the test.

fastest963:
Well x/net doesn't let you create a control message with a ipv4 address inside of a ipv6 control message, but Linux kernel definitely does that if you enable v4+v6 on an interface, it will send a ipv6 control message no matter if the source IP was ipv4 or not.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Skip UDP OOB tests ..." ]
>@tmthrgd seems useful to do the other test cases just to verify that there isn't any non-amd64 breakage and just separate the test case with ipv4 in ipv6 specifically only on amd64. @miekg what do you think?

sgtm, although I have no strong pref. for a new file with build or the check we 
have now.

PR #662, opened 2018-04-02
Cleanup udp.go somewhat

tmthrgd:
In the course of solving #660, I cleaned up and simplified a few parts of `udp.go`.

### DIFF
--- a/udp.go
+++ b/udp.go
@@ -9,6 +9,24 @@ import (
 	"golang.org/x/net/ipv6"
 )
 
+// This is the required size of the OOB buffer to pass to ReadMsgUDP.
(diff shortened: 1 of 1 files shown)


udp.go
```diff
@@ -9,6 +9,24 @@ import (
 	"golang.org/x/net/ipv6"
 )
 
+// This is the required size of the OOB buffer to pass to ReadMsgUDP.
+var udpOOBSize int
+
+func init() {
```
miekg: As `init`? Maybe slightly neater to just have it as a function?
~~~
var udpOOBSize = func() int {
 // ...
}
~~~

PR #666, opened 2018-04-05
ExchangeWithDialer does not exist anymore

integrii:
ExchangeWithDialer is in the godoc, but its not valid.  This changes it to the appropriate client dialer setting and `c.Exchange` call.


miekg:
thanks!

PR #668, opened 2018-04-10
Fixed len computation when size just goes beyond 14 bits

pierresouchay:
Fixes https://github.com/miekg/dns/issues/663

Since offset of compression has to start within the first 14 bits, we don't want to take it into account before trying to compress but once the length computation has been performed

### DIFF
--- a/msg.go
+++ b/msg.go
@@ -971,7 +971,6 @@ func compressionLenSlice(len int, c map[string]int, rs []RR) int {
 		// track this length, and the global length in len, while taking compression into account for both.
 		x := r.len()
 		l += x
-		len += x
 
 		k, ok := compressionLenSearch(c, r.Header().Name)
 		if ok {
@@ -992,6 +991,7 @@ func compressionLenSlice(len int, c map[string]int, rs []RR) int {
 		if len < maxCompressionOffset {
 			compressionLenHelperType(c, r)
 		}
+		len += x
 	}
 	return l
 }
--- a/length_test.go
+++ b/length_test.go
@@ -172,3 +172,22 @@ func TestMsgCompressLengthLargeRecords(t *testing.T) {
 		t.Fatalf("predicted compressed length is wrong: predicted %s (len=%d) %d, actual %d", msg.Question[0].Name, len(msg.Answer), predicted, len(buf))
 	}
 }
+
+func TestMsgCompressLengthLargeRecordsAllValues(t *testing.T) {
+	msg := new(Msg)
+	msg.Compress = true
+	msg.SetQuestion("redis.service.consul.", TypeSRV)
+	for i := 0; i < 900; i++ {
+		target := fmt.Sprintf("host-redis-%d-%d.test.acme.com.node.dc1.consul.", i/256, i%256)
+		msg.Answer = append(msg.Answer, &SRV{Hdr: RR_Header{Name: "redis.service.consul.", Class: 1, Rrtype: TypeSRV, Ttl: 0x3c}, Port: 0x4c57, Target: target})
+		msg.Extra = append(msg.Extra, &CNAME{Hdr: RR_Header{Name: target, Class: 1, Rrtype: TypeCNAME, Ttl: 0x3c}, Target: fmt.Sprintf("fx.168.%d.%d.", i/256, i%256)})
+		predicted := msg.Len()
+		buf, err := msg.Pack()
+		if err != nil {
+			t.Error(err)
+		}
+		if predicted != len(buf) {
+			t.Fatalf("predicted compressed length is wrong for %d records: predicted %s (len=%d) %d, actual %d", i, msg.Question[0].Name, len(msg.Answer), predicted, len(buf))
+		}
+	}
+}


miekg:
Fun times trying to make 2 diff. algorithms do the same thing: This now fails in travis:
~~~
-- FAIL: TestMsgCompressLengthLargeRecords (0.03s)
	length_test.go:172: predicted compressed length is wrong: predicted my.service.acme. (len=250) 22833, actual 22850
~~~
You PR looks sane btw.

pierresouchay:
Ok, I undestand fully the problem and have a fix:

Basically, the function compressionLenHelper should take as parameter the current len of message to guess whether compression is possible or not:

When you arrive around 16384, lets say 16380, if you plan to add those entries:

let's say you already added those labels:
dom8.example.com

* x1.dom9.example.com => should be compressed fully and re-usable, meaning that dom9.example.com can be re-used later
* long_name.dom9.example.com => will only re-use .example.com later, not dom9.example.com because long_name overflows 16384, thus dom9.example.com won't be accessible to next entries (since it would be outside the 16384 bytes range)

I made a patch, but I am a bit lost regarding the overhead of DNS structure to compute really the overhead of labels...

You probably know the answer better than I do and how to fix my offsets :-)

I you can have a look

Regards

pierresouchay:
@miekg I added a new test that does padding at begin of DNS query to ensure we properly handle all cases, it should fail for now

miekg:
Ah, I see, because `compressionLenHelper` adds the labels that can be used for compression and this can also step over the boundary.

I wonder if all that extra complexity is worth it though, right now `Len()` will over estimate, so this boils down to X missing RRs (compared to `Pack()`)

miekg:
Of course in the range 16K <-> 64K this adds up. hmm

pierresouchay:
@miekg yes... :(

miekg:
Ok `lbs` gives you the index of the dots in the name, so if you have the current len in `l`, then
l + lbs < compressionOffset can be used for compression and not the rest.

I similar fix may or may not be needed in `compressionLenHelperType`.

andrewtj:
> On 11 Apr 2018, at 4:45 pm, Miek Gieben <notifications@github.com> wrote:
> 
> I wonder if all that extra complexity is worth it though, right now Len() will over estimate, so this boils down to X missing RRs (compared to Pack())

This use case would probably be better handled by an API that allowed incrementally building a wire format message. Maybe something to consider for v2?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Fixed len computati..." ]
>
>
>> On 11 Apr 2018, at 4:45 pm, Miek Gieben <notifications@github.com> wrote:
>>
>> I wonder if all that extra complexity is worth it though, right now Len() will over estimate, so this boils down to X missing RRs (compared to Pack())
>
>This use case would probably be better handled by an API that allowed incrementally building a wire format message. Maybe something to consider for v2?

It would, but one of the reason I like the current API is that you can just do 
`ns.Extra = append(ns.Extra, xx)`. Having a lib that surfaces the wireformat is 
entirely different (and probably harder to use).

Now because DNS is silly protocol it is hard to not seep through some of those 
underlaying details.

msg.go
```diff
@@ -962,70 +962,76 @@ func compressedLen(dns *Msg, compress bool) int {
...
 	}
-	return l
+	return lenp - initLen
 }
 
-// Put the parts of the name in the compression map.
-func compressionLenHelper(c map[string]int, s string) {
+// Put the parts of the name in the compression map, return the size added in payload
+func compressionLenHelper(c map[string]int, s string, currentLen int) int {
+	addedSize := 0
```
miekg: I wonder if we can optimize this more; if we have seen `s` here, there shouldn't be a need a spit it at all. WDYT?

miekg: addedLen ?

msg.go
```diff
@@ -962,70 +962,76 @@ func compressedLen(dns *Msg, compress bool) int {
...
 	pref := ""
 	lbs := Split(s)
 	for j := len(lbs) - 1; j >= 0; j-- {
 		pref = s[lbs[j]:]
 		if _, ok := c[pref]; !ok {
-			c[pref] = len(pref)
+			lenAdded := len(pref)
+			numLabelsBefore := len(lbs) - j - 1
+			offsetOfLabel := currentLen + len(s) - len(pref) + numLabelsBefore*2
```
miekg: don't fully understand this yet... why the *2 ?

msg.go
pierresouchay: Yes, probably, split might be reused... but since the review is already big (I mean for the bug we consider), I did prefer to limit as much as possible the changes

pierresouchay: Well, my comment is probably not clear: compressionLenHelper return the size added to global DNS response (thus the addedSize name).

Considering adding "host1.example.com" with a compression map having already:
"host0.example.com" (and thus "example.com" as well as "com")

it is supposed to return 2+ len("host0")  + 2 (pointer to "example.com" label


Do you have a suggestion to have clearer name?

msg.go
pierresouchay: numLabelsBefore*2 => size of label pointers (since their size is 2 bytes => 2 times)

msg.go
pierresouchay: Done in next patch version

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Fixed len computati..." ]
>pierresouchay commented on this pull request.
>
>
>
>>  }
>
>-// Put the parts of the name in the compression map.
>-func compressionLenHelper(c map[string]int, s string) {
>+// Put the parts of the name in the compression map, return the size added in payload
>+func compressionLenHelper(c map[string]int, s string, currentLen int) int {
>+	addedSize := 0
>
>Well, my comment is probably not clear: compressionLenHelper return the size added to global DNS response (thus the addedSize name).
>
>Considering adding "host1.example.com" with a compression map having already:
>"host0.example.com" (and thus "example.com" as well as "com")
>
>it is supposed to return 2+ len("host0")  + 2 (pointer to "example.com" label

Ok, I see where you're heading with this, you're not adding when we shouldn't 
but then the callers does now how much saving there where.
But `addedSizeBytes += 2 + lenAdded` in a loop can't be right (I think).

I need to take a step back here, because I find the logic hard to follow. There 
is also bunch of constants that we should probably name (2 and 6 in the current 
PR).

Let's see: the current code works, unless, precisely on the border of the 14 
bits we have name that can be compressed, but the end of the name overflows the 
14 bits; i.e. we should use any labels before that to compress as well (if we 
haven't seen them).

I.e if "len < 2<<13" and "len+x > 2<<13", we could just check for that case and 
split the name our selves.

Or did you find more issues with the current code?

/Miek

--
Miek Gieben

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Fixed len computati..." ]
>@miekg Ok for the constants...
>
>About the change, I tried several approaches (including adding another function within compressionLenSlice that did check whether currentLen + r.len() did overflow 14bits, but the code was even more unreadable.
>
>The idea I took is that when we add compression, we also add some data in the global payload, so, all function adding compression (compressionLenHelper / compressionLenSearch  should return how much data size they added themselves.
>
>The algorithm is weird, because there are actually 2 parts:
>* one that does decrease the size of x (with the += (1 - k) ) Note that the decrease is weird as well (why += 1 - k  ? => took me a while to understand why)
>* the other that increase it (sound like a more common approach since the size is in bytes)
>
>I first tried to decompose it, but it was not easy, hence the addedSize stuff => but I agree having both ways is crap. (But method for length computation were not present - hence the +6 from nowhere)
>
>Of course, you are the boss, so decide what you really want, but for now, among all my tries, this sounds like the easier to understand

Thanks for your mail. yeah, while thinking about a simple change is occurred 
that this must also be expanded to the compressionLenTypeHelper.

Let me whip up some (internal) branches to understand this code better, so I can 
give a better code review.

pierresouchay:
@miekg Great, I'll update Consul dependency with it as soon as a release is done to re-enable compression in it! Thank you

msg.go
```diff
@@ -962,70 +962,88 @@ func compressedLen(dns *Msg, compress bool) int {
...
-	var l int
+func compressionLenSlice(lenp int, c map[string]int, rs []RR) int {
+	initLen := lenp
 	for _, r := range rs {
 		if r == nil {
 			continue
 		}
-		// track this length, and the global length in len, while taking compression into account for both.
+		// TmpLen is to track len of record at 14bits boudaries, 6 bytes is the raw overhead
+		tmpLen := lenp + 6
```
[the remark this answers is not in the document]
pierresouchay: Don't have access to the code easily from my phone, but I would say the signaling of a new record? (I am not at all an expert in DNS, I just read RFC part related to compression for this patch)

pierresouchay:
@miekg I had not the time to look further, but I think I know why we have to add 6 and not 10 or something similar.

Actually 6 is a constant regarding the start of message, erased at each iteration since we take x (we discard 6 and all added len at each iteration)

What it could mean is just that we already added some len of the DNS message (lets say 4 bytes if you expected the constant to be 10) before entering this section of the code and those bytes are actually supposed to be AFTER this computation. I'll try to have a look tomorrow

pierresouchay:
@miekg I finally go this right I think

I changed the compression map used by Len() to have the same semantics (aka the offset of label) as the one used in Pack() to be able to add unit tests properly.

With this one, I change a few things, but everything is well tested now and "Fake" (aka used by Len()) has the same values as Pack() -> so I am confident compression will work the same way around 14bits

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Fixed len computati..." ]
>@miekg I finally go this right I think
>
>I changed the compression map used by Len() to have the same semantics (aka the offset of label) as the one used in Pack() to be able to add unit tests properly.
>
>With this one, I change a few things, but everything is well tested now and "Fake" (aka used by Len()) has the same values as Pack() -> so I am confident compression will work the same way around 14bits

Ok, thanks. I will take a look after kubecon, so next weekend probably.

pierresouchay:
@miekg have a nice kubecon!

pierresouchay:
@miekg does the patch is OK for you ?

Since now compression map contains the same exact information as the one used for real packing, I am quite confident the result will always be the same since I test many permutations with it in test [TestCompareCompressionMapsForSRV()](https://github.com/miekg/dns/pull/668/files#diff-13b6c7d0ebd5d41bd909bf58ec29bd06R298), we now are for sure that any change in computation is equivalent in both Pack() and Len() since offset are all equal. In order to test this, I had to create a new [packBufferWithCompressionMap()](https://github.com/miekg/dns/pull/668/files#diff-8353741544373e78bb7c8c3f6a0ea2ccR704) function that takes into parameter the compression map so we can compare it easily both maps easily.

The patch now does not take any magic offset anymore and is tested with a very large combination of lengths and offset, so I am pretty confident it works well.

pierresouchay:
@miekg Thank you, I'll be able to fully re-enable compression in Consul then once a new release is created :)

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Fixed len computati..." ]
>@miekg Thank you, I'll be able to fully re-enable compression in Consul then once a new release is created :)

Pushed 1.0.7

Issue #669, opened 2018-05-01
AlgorithmToHash map missing DSA algorithm


saurori:
The mapping found in [dnssec.go](https://github.com/miekg/dns/blob/master/dnssec.go#L74-L83) seems to be missing DSA mapping according to [supported algorithms found at iana](https://www.iana.org/assignments/dns-sec-alg-numbers/dns-sec-alg-numbers.xhtml). I believe `DSA` should map to `crypto.SHA1`

https://github.com/miekg/dns/blob/01d59357d468872339068bcd5d55a00e2463051f/dnssec.go#L74-L83

andrewtj:
I don't think DSA is implemented. Is it something you want to use?

saurori:
I don't need to use it for creating keys, but I have seen it in the wild:

```
dig -t DNSKEY stienriviere.me +dnssec +multi
```

And it would be convenient to be able to map the algorithm to it's hash in order to get `Size()` (used for research purposes).

andrewtj:
It's not implemented for signing or verifying either as I recall. I think the utility here is probably pretty limited (`AlgorithmToString` probably suits most) but as it's only one entry in a map, if it's useful to you, why not open a PR.

saurori:
I'll open a PR, I just wanted to confirm that DSA -> SHA1 mapping was appropriate, in case it was left out intentionally.

tmthrgd:
Adding `DSA` to `AlgorithmToHash` isn’t appropriate because that map is used by `(*RRSIG).Sign` (and `Verify`) to reject unsupported algorithms. In particular the `sign` function doesn’t know what to do for DSA keys.

andrewtj:
@tmthrgd I thought this was harmless as `(*RRSIG).Verify` would fall through with `ErrAlg` while `(*RRSIG).Sign` needs a `crypto.Signer` and so is moot. Have I misread?

tmthrgd:
@andrewtj It looks like you’re right about that. It *should* all fall through to an error case. That does feel a little fragile to me but :shrug:.

andrewtj:
I'm closing this as #670 has been merged.

PR #670, opened 2018-05-02
Add DSA to SHA1 mapping in dnssec AlgorithmToHash

saurori:
Discussion in issue https://github.com/miekg/dns/issues/669


miekg:
I've seen the discussion, but does this still need to be merged?

saurori:
If you feel that DSA should be in the map, then yes.

PR #671, opened 2018-05-03
WIP: DNS-over-HTTPS support for Client.Exchange API

tmthrgd:
This is my third attempt at adding DNS-over-HTTPS support to `Client.`. It supersedes #647, #649 and #651.

This method provides DNS-over-HTTPS support exclusively within the `(*Client).Exchange*` APIs.

It's proving difficult to pipe cancellation and context through the `Client`, and I suspect some refactoring will be needed.

Note: when using DNS-over-HTTPS, the `Dialer`, `TLSConfig` and `*Timeout` (*todo*) fields on `Client` are ignored.


tmthrgd:
This will also address coredns/coredns#1650.

PR #673, opened 2018-05-07
Make MaxTCPQueries configurable

UladzimirTrehubenka:


server.go
```diff
@@ -12,9 +12,6 @@ import (
 	"time"
 )
 
-// Maximum number of TCP queries before we close the socket.
-const maxTCPQueries = 128
```
miekg: We should expose this in the server, and make `0` mean; disable this. 

By removing this the server has no way to kill a client after it things it has enough; Google Public DNS does the same thing btw, so it's not uncommon to have this.

server.go
```diff
@@ -593,8 +592,13 @@ func (srv *Server) serve(w *response) {
 
 	timeout := srv.getReadTimeout()
 
-	// TODO(miek): make maxTCPQueries configurable?
-	for q := 0; q < maxTCPQueries; q++ {
+	q := 0
+	for {
```
tmthrgd: I think this would come out cleaner written as:
```go
for q := 0; q < srv.MaxTCPQueries || srv.MaxTCPQueries == 0; q++ {
```

server.go
```diff
@@ -303,6 +300,8 @@ type Server struct {
 	DecorateReader DecorateReader
 	// DecorateWriter is optional, allows customization of the process that writes raw DNS messages.
 	DecorateWriter DecorateWriter
+	// Maximum number of TCP queries before we close the socket (unlimited if 0).
+	MaxTCPQueries int
```
tmthrgd: I question whether having the default value mean unlimited is best, because it amounts to a behaviour change that could be unexpected. Perhaps it would be best to have 0 mean the current max of 128 and -1 explicitly mean unlimited (maybe even with a `const UnlimitedMaxQueries = -1` for clarity). @miekg thougts?

miekg:
Re: 128, ugly but I agree

server.go
```diff
@@ -13,7 +13,7 @@ import (
 	"time"
 )
 
-// Maximum number of TCP queries before we close the socket.
+// Default maximum number of TCP queries before we close the socket.
 const maxTCPQueries = 128
```
tmthrgd: This should probably be renamed `defaultMaxTCPQueries`, but I can't say I love the name. Maybe it's fine as it is :man_shrugging:.

server.go
```diff
@@ -303,6 +303,8 @@ type Server struct {
 	DecorateReader DecorateReader
 	// DecorateWriter is optional, allows customization of the process that writes raw DNS messages.
 	DecorateWriter DecorateWriter
+	// Maximum number of TCP queries before we close the socket. Default is maxTCPQueries (unlimited if -1).
+	MaxTCPQueries int
```
tmthrgd: I'm a little hesitant about using a magic number here, but it's not blocking to me. (I still like something like `const UnlimitedMaxQueries = -1` simply for the clarity it gives).

PR #675, opened 2018-05-11
do not modify dns.Rcode when packing to wire format

### LINKED ISSUE #674: Do not modify dns.Msg.Rcode when packing message
Currently when calling dns.Msg.Pack on a message with EDNS0 and extended RCode, the rcode is modified during packing: https://github.com/miekg/dns/blob/dfb943b13e55308ee218d33a1dd4f332466ff6c8/msg.go#L717

The issue is that if you were to access dns.Msg.Rcode after packing, a message that would normally contain Rcode BADVERS, will now have NOERROR.

side note: OPT is also modified: https://github.com/miekg/dns/blob/dfb943b13e55308ee218d33a1dd4f332466ff6c8/msg.go#L716

Not sure what would be the best approach here, one solution to keep the rcode untouched across call to Pack, is to simply safe its value in a separate `rcode` integer and use this instead of dns.Msg.Rcode directly.
OPT would still be modified, but at least, the String function would still properly print the rcode.

chantra:
When the message has an EDNS0 option in the additional section and
dns.Msg.Rcode is set to an extended rcode, dns.Msg.PackBuffer() modifies
dns.Msg.Rcode.
If you were to `Pack` the message and log it after, the Rcode would show
NOERROR.

Running the test before the change would error with:

```
=== RUN   TestPackNoSideEffect
--- FAIL: TestPackNoSideEffect (0.00s)
	msg_test.go:51: after pack: Rcode is expected to be BADVERS
```

after fixing dns.Msg.PackBuffer(), all tests are still passing.

Fixes #674

### DIFF
--- a/msg.go
+++ b/msg.go
@@ -698,7 +698,6 @@ func (dns *Msg) PackBuffer(buf []byte) (msg []byte, err error) {
 	var (
 		dh          Header
 		compression map[string]int
-		rcode       int
 	)
 
 	if dns.Compress {
@@ -708,7 +707,6 @@ func (dns *Msg) PackBuffer(buf []byte) (msg []byte, err error) {
 	if dns.Rcode < 0 || dns.Rcode > 0xFFF {
 		return nil, ErrRcode
 	}
-	rcode = dns.Rcode & 0xF
 	if dns.Rcode > 0xF {
 		// Regular RCODE field is 4 bits
 		opt := dns.IsEdns0()
@@ -720,7 +718,7 @@ func (dns *Msg) PackBuffer(buf []byte) (msg []byte, err error) {
 
 	// Convert convenient Msg into wire-like Header.
 	dh.Id = dns.Id
-	dh.Bits = uint16(dns.Opcode)<<11 | uint16(rcode)
+	dh.Bits = uint16(dns.Opcode)<<11 | uint16(dns.Rcode&0xF)
 	if dns.Response {
 		dh.Bits |= _QR
 	}
--- a/msg.go
+++ b/msg.go
@@ -698,6 +698,7 @@ func (dns *Msg) PackBuffer(buf []byte) (msg []byte, err error) {
 	var (
 		dh          Header
 		compression map[string]int
+		rcode       int
 	)
 
 	if dns.Compress {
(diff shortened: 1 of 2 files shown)


msg.go
```diff
@@ -707,19 +708,19 @@ func (dns *Msg) PackBuffer(buf []byte) (msg []byte, err error) {
...
 			return nil, ErrExtendedRcode
 		}
 		opt.SetExtendedRcode(uint8(dns.Rcode >> 4))
-		dns.Rcode &= 0xF
 	}
 
 	// Convert convenient Msg into wire-like Header.
 	dh.Id = dns.Id
-	dh.Bits = uint16(dns.Opcode)<<11 | uint16(dns.Rcode)
+	dh.Bits = uint16(dns.Opcode)<<11 | uint16(rcode)
```
tmthrgd: I think it might be cleaner to just put the `dns.Rcode & 0xF` here in the `uint64(...)`.

msg_test.go
```diff
@@ -26,6 +26,31 @@ var (
...
+			Rrtype: TypeOPT,
+		},
+	}
+	o.SetUDPSize(DefaultMsgSize)
+
+	a.Extra = append(a.Extra, o)
+	a.SetRcode(m, RcodeBadVers)
+
+	if a.Rcode != RcodeBadVers {
+		t.Errorf("before pack: Rcode is expected to be BADVERS")
```
tmthrgd: This check isn’t actually test related and isn’t needed.

msg.go
chantra: I don't have strong feelings. If this is is needed to get this merged (assuming the patch logic is accepted), I will update the PR.

miekg: might be a tad bit cleaner, can you also add a comment above it on why this needed?

msg_test.go
miekg: I would just go with less is more and just only check after Pack()

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] do not modify dns.R..." ]
>
>
>
>>  	}
>
> 	// Convert convenient Msg into wire-like Header.
> 	dh.Id = dns.Id
>-	dh.Bits = uint16(dns.Opcode)<<11 | uint16(dns.Rcode)
>+	dh.Bits = uint16(dns.Opcode)<<11 | uint16(rcode)
>
>@miekg what would be a tad cleaner? using `dns.Rcode & 0xF` inline? or the variable?
>The fact that you are asking for a comment makes me believe you mean the former, but just want to make sure I get you right

yep inline

/Miek

--
Miek Gieben

chantra:
addressed comments

PR #677, opened 2018-05-14
Handle empty NSEC3 salt in scanner

### LINKED ISSUE #676: NSEC3 hashed owner calculation issue
Hi, i'm on `1f2aa4c` and have experienced issues with nsec3 name matching. Let me exemplify.
```
/// one of the NSEC3s returned by the TLD zone to (for example) <microsoft.com IN NS> (with DO flag)
rr, _ := dns.NewRR("CK0POJMG874LJREF7EFN8430QVIT8BSM.com. 86400 IN NSEC3 1 1 0 - CK0Q1GIN43N1ARRC9OSM6QPQR81H5M9A  NS SOA RRSIG DNSKEY NSEC3PARAM")
fmt.Printf("[%v]\n", rr.(*dns.NSEC3).Match("com."))
/// outputs false
fmt.Printf("[%s]\n", dns.HashName("com.", 1, 0, ""))
/// outputs CK0POJMG874LJREF7EFN8430QVIT8BSM
```

The issue, as i see it, is with passing the decorative `-` as salt (meaning, no salt is used) to the hex packer routine.

Best regards!

tmthrgd:
This fixes #676 and should be non-invasive.

/cc @shadrick

### DIFF
--- a/scan_rr.go
+++ b/scan_rr.go
@@ -1255,8 +1255,10 @@ func setNSEC3(h RR_Header, c chan lex, o, f string) (RR, *ParseError, string) {
 	if len(l.token) == 0 || l.err {
 		return nil, &ParseError{f, "bad NSEC3 Salt", l}, ""
 	}
-	rr.SaltLength = uint8(len(l.token)) / 2
-	rr.Salt = l.token
+	if l.token != "-" {
+		rr.SaltLength = uint8(len(l.token)) / 2
+		rr.Salt = l.token
+	}
 
 	<-c
 	l = <-c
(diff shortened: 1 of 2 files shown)


scan_rr.go
```diff
@@ -1321,8 +1323,10 @@ func setNSEC3PARAM(h RR_Header, c chan lex, o, f string) (RR, *ParseError, strin
 	rr.Iterations = uint16(i)
 	<-c
 	l = <-c
-	rr.SaltLength = uint8(len(l.token))
-	rr.Salt = l.token
+	if l.token != "-" {
+		rr.SaltLength = uint8(len(l.token))
```
tmthrgd: As an aside @miekg, is there a reason NSEC3.SaltLength is the byte length and NSEC3Param.SaltLength is the hex length? Is that a bug?

miekg: hmm, looks like it, those fields are identical to rfc 5155

PR #681, opened 2018-05-19
doh: Fix mime type

miekg:
Latest draft uses "application/dns-message".

### DIFF
--- a/client.go
+++ b/client.go
@@ -17,10 +17,12 @@ import (
 	"time"
 )
 
-const dnsTimeout time.Duration = 2 * time.Second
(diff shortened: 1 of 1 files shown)


tmthrgd:
I was going to go at this slightly differently. I want to keep the old and new mime-types so it continues to work with servers implementing an older draft. Content Negotiation will handle picking either mime type.

miekg:
For an older draft... I think we shouldn't bother. If it where standardized
as such we should, but now?

tmthrgd:
@miekg That sounds good to me.

PR #682, opened 2018-05-19
Revert 6f3c0a126ca15bec6b0db0c77e65ed4ad2354ca2

miekg:
Previous behavior was correct. Checking with coredns:

current:

~~~
; <<>> DiG 9.10.3-P4-Debian <<>> +norec +noad +edns=1 +noednsneg soa miek.nl @localhost -p 1043
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: ?256, id: 35480
;; flags: qr; QUERY: 0, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 0
;; Query time: 0 msec
;; SERVER: ::1#1043(::1)
;; WHEN: Sat May 19 16:42:54 BST 2018
;; MSG SIZE  rcvd: 23
~~~

this pr:

~~~
; <<>> DiG 9.10.3-P4-Debian <<>> +norec +noad +edns=1 +noednsneg soa miek.nl @localhost -p 1043
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: BADVERS, id: 25912
;; flags: qr; QUERY: 0, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 0
;; Query time: 0 msec
;; SERVER: ::1#1043(::1)
;; WHEN: Sat May 19 16:51:28 BST 2018
;; MSG SIZE  rcvd: 23
~~~

### DIFF
--- a/edns.go
+++ b/edns.go
@@ -102,12 +102,15 @@ func (rr *OPT) SetVersion(v uint8) {
 
 // ExtendedRcode returns the EDNS extended RCODE field (the upper 8 bits of the TTL).
 func (rr *OPT) ExtendedRcode() int {
-	return int((rr.Hdr.Ttl & 0xFF000000) >> 24)
+	return int((rr.Hdr.Ttl&0xFF000000)>>24) + 15
 }
 
 // SetExtendedRcode sets the EDNS extended RCODE field.
 func (rr *OPT) SetExtendedRcode(v uint8) {
-	rr.Hdr.Ttl = rr.Hdr.Ttl&0x00FFFFFF | (uint32(v) << 24)
+	if v < RcodeBadVers { // Smaller than 16.. Use the 4 bits you have!
+		return
+	}
+	rr.Hdr.Ttl = rr.Hdr.Ttl&0x00FFFFFF | (uint32(v-15) << 24)
 }
 
 // UDPSize returns the UDP buffer size.


tmthrgd:
This is #516 & #517 correct? /cc @m13253 who opened both.

m13253:
I am very sorry to say that neither the old code, nor the new code is standard-compliant.

Let me explain this:
for BADVERS = 16, we should set Rcode = 0, ExRcode = 1.

Behavior          | Rcode                 | ExRcode               | Combine them
------------------|-----------------------|-----------------------|--------------
Bad behavior      | 15 for unknown reason | 15 for unknown reason | 255 (wrong)
"Good" behavior   | always 0              | 16 - 15 = 1           | 16 (correct?)
Standard behavior | 16 >> 4 = 0           | 16 & 0xf = 1          | 16 (correct!)

See that? Both "good" behavior and standard behavior generate 0 and 1. This is true for BADVERS, the only defined code today.

I suggest: Instead of just simply revert to the "good behavior", can we rewrite it to to comply with the standard?

miekg:
ok true enough, except that this method for a better or worse only sees the opt record

m13253:
I agree to your opinion.
And I agree that we should revert this commit.

But how about creating another API besides this to query the combined Rcode?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Revert 6f3c0a126ca1..." ]
>I agree to your opinion.
>And I agree that we should revert this commit.
>
>But how about creating another API besides this to query the combined Rcode?

That's probably the best solution longer term, but hard to discover, right now
someone just sets the RCODE by assigning it to msg.Rcode.

(Although there is m.SetRcode() maybe we can add some smartness there?)

miekg:
Ok, going ahead with merging this and then figuring out some next steps

gibson042:
Current surface area looks like this:
* `func (dns *Msg) SetRcode(request *Msg, rcode int) *Msg` makes `dns` a reply to `request` and sets its `Rcode` without enforcing a value between 0 and 15.
* `func (rr *OPT) SetExtendedRcode(v uint8)` updates the EXTENDED-RCODE bits of `rr.Hdr.Ttl` without really defining `v` (but, being an 8-bit value confined to an OPT RR, `v` logically MUST be interpreted as just the upper bits of the 12-bit RCODE):
  * BADVERS (extended RCODE 16) should be set like (`optRR.SetExtendedRcode(dns.RcodeBadVers >> 4)` or `optRR.SetExtendedRcode(1)`) + `resp.SetRcode(…)` (see below).
  * BADCOOKIE (extended RCODE 23) should be set like (`optRR.SetExtendedRcode(dns.RcodeBadCookie >> 4)` or—again—`optRR.SetExtendedRcode(1)`) + `resp.SetRcode(…)` (see below).
* `func (rr *OPT) ExtendedRcode() int` returns the EXTENDED-RCODE bits of `rr.Hdr.Ttl` as a signed integer between 0 and 255 (but suitable for round-tripping with `SetExtendedRcode` after casting).

If `SetRcode` were limited to the 4 bits in the DNS message header, then proper input would be restricted by something like `& 0x0F`. But that isn't the case; [`func (dns *Msg) packBufferWithCompressionMap`](https://github.com/miekg/dns/blob/3745b9737df4a45eabfb086527be0fa50fb6dcf1/msg.go#L704) calls `opt.SetExtendedRcode(uint8(dns.Rcode >> 4))` to place its 8 upper bits into the OPT RR.

All of which means that the current code as of 6f3c0a126ca15bec6b0db0c77e65ed4ad2354ca2 is correct, but the Rcode and ExtendedRcode semantics are a little strange (full 12-bit RCODE for `Msg`, but just the 8 upper bits for `OPT`). New surface area may be warranted, but the existing code is now functioning correctly for what it's supposed to do.

gibson042:
Well then. As of this merging, the current code is incorrect and the revert needs to be reverted.

miekg:
thanks for the good discussion, I may have misinterpreted "And I agree that we should revert this commit." (@m13253).

How can we expose `opt.SetExtendedRcode(uint8(23 >> 4))` in a sane matter? As docs? Or new set of constants?

m13253:
How about we write two new functions:
```go
func (*Msg) CombinedRcode() int16 {}
func (*Msg) SetCombinedRcode(int16) {}
```
and deprecate our existing `SetRcode / SetExtendedRcode`?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Revert 6f3c0a126ca1..." ]
>How about we write two new functions:
>```go
>func (*Msg) CombinedRcode() int16 {}
>func (*Msg) SetCombinedRcode(int16) {}
>```
>and deprecate our existing `SetRcode / SetExtendedRcode`?

This won't fix existing code and msg.Rcode is exported as well.
I think documenting this now is the best we can do (without breaking backwards 
compatibility).

When we do break backwards compat a fixed `msg.SetRcode()` and the deprecation 
of SetExtendedRcode/ExtendedRcode seem like a good way forward?

PR #683, opened 2018-05-20
Don't use padding in base32 helpers

andrewtj:
The base32 variant NSEC3 uses doesn't have padding. This hasn't been a
problem in practice because SHA1 is the only current NSEC3 hash algorithm
and its output doesn't require padding.

No-pad support was introduced in Go 1.9 which is the oldest release this
package supports.

---

This is a follow up to the discussion in #426.

### DIFF
--- a/msg_helpers.go
+++ b/msg_helpers.go
@@ -141,20 +141,24 @@ func truncateMsgFromRdlength(msg []byte, off int, rdlength uint16) (truncmsg []b
 	return msg[:lenrd], nil
 }
 
+var base32HexNoPadEncoding = base32.HexEncoding.WithPadding(base32.NoPadding)
+
 func fromBase32(s []byte) (buf []byte, err error) {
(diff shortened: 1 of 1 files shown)


tmthrgd:
I threw together a simple benchmark and the call to `WithPadding` has overhead:

```
name        old time/op    new time/op    delta
ToBase32-8     242ns ± 6%     501ns ± 7%  +106.80%  (p=0.000 n=10+8)

name        old alloc/op   new alloc/op   delta
ToBase32-8      128B ± 0%      416B ± 0%  +225.00%  (p=0.000 n=10+10)

name        old allocs/op  new allocs/op  delta
ToBase32-8      2.00 ± 0%      3.00 ± 0%   +50.00%  (p=0.000 n=10+10)
```

The performance overhead goes away entirely if you extract `base32.HexEncoding.WithPadding(base32.NoPadding)` into a global variable.

PR #684, opened 2018-05-21
Require URLs for DOH addresses

tmthrgd:
DoH support now requires the full URL to be passed to `Exchange` rather than just the hostname.

This is to accommodate the (unfortunate) removal of the `.well-known/dns-query` path in draft-ietf-doh-dns-over-https-03, see https://www.ietf.org/rfcdiff?url1=draft-ietf-doh-dns-over-https-02&url2=draft-ietf-doh-dns-over-https-03&difftype=--html.

DNS-over-HTTPS support now works as:
```go
(&Client{Net: "https"}).Exchange(new(Msg), "https://dns.cloudflare.com/dns-query")
```

### DIFF
--- a/client.go
+++ b/client.go
@@ -216,14 +216,11 @@ func (c *Client) exchangeDOH(ctx context.Context, m *Msg, a string) (r *Msg, rtt
 		return nil, 0, err
 	}
 
-	// TODO(tmthrgd): Allow the path to be customised?
-	u := &url.URL{
-		Scheme: "https",
-		Host:   a,
-		Path:   "/.well-known/dns-query",
-	}
-	if u.Port() == "443" {
-		u.Host = u.Hostname()
+	u, err := url.Parse(a)
+	if err != nil {
+		return nil, 0, err
+	} else if u.Scheme != "https" {
+		return nil, 0, fmt.Errorf(`dns: unexpected url scheme %q; expected "https"`, u.Scheme)
 	}
 
 	req, err := http.NewRequest(http.MethodPost, u.String(), bytes.NewReader(p))
--- a/client.go
+++ b/client.go
@@ -12,7 +12,6 @@ import (
 	"io/ioutil"
 	"net"
 	"net/http"
-	"net/url"
 	"strings"
 	"time"
 )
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -216,14 +216,11 @@ func (c *Client) exchangeDOH(ctx context.Context, m *Msg, a string) (r *Msg, rtt
...
-		Scheme: "https",
-		Host:   a,
-		Path:   "/.well-known/dns-query",
-	}
-	if u.Port() == "443" {
-		u.Host = u.Hostname()
+	u, err := url.Parse(a)
+	if err != nil {
+		return nil, 0, err
+	} else if u.Scheme != "https" {
```
miekg: should we just not care? garbage in; garbage out and having a http way of doing things might benefit some local testing?

tmthrgd: I think it’s best to be restrictive, the spec is specifically over HTTPS. Also anyone doing `Exchange(new(Msg), “http://dns.cloudflare/dns-query”)`, will see the request succeeded, but silently leak DNS requests before the redirect kicks in. Also note, there is no existing support for non-HTTPS in HEAD.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Require URLs for DO..." ]
>
>
>
>> @@ -216,14 +216,11 @@ func (c *Client) exchangeDOH(ctx context.Context, m *Msg, a string) (r *Msg, rtt
> 		return nil, 0, err
> 	}
>
>-	// TODO(tmthrgd): Allow the path to be customised?
>-	u := &url.URL{
>-		Scheme: "https",
>-		Host:   a,
>-		Path:   "/.well-known/dns-query",
>-	}
>-	if u.Port() == "443" {
>-		u.Host = u.Hostname()
>+	u, err := url.Parse(a)
>+	if err != nil {
>+		return nil, 0, err
>+	} else if u.Scheme != "https" {
>
>I think it’s best to be restrictive, the spec is specifically over HTTPS. Also anyone doing `Exchange(new(Msg), “http://dns.cloudflare/dns-query”)`, will see the request succeeded, but silently leak DNS requests before the redirect kicks in. Also note, there is no existing support for non-HTTPS in HEAD.

No such a check doesn't provide any guarantees. The proto should also be http/2 
but there is no way to enforce that; on this level we shouldn't care; if the 
remote site doesn't except http we will get an error back anyway.

tmthrgd:
@miekg That feels like a mistake, and I do think the restriction is worthwhile, but I can remove the check if you insist. I’ll change that shortly.

tmthrgd:
The draft has this to say:
```
6.  HTTP Integration

   This protocol MUST be used with the https scheme URI [RFC7230].
```
(https://tools.ietf.org/html/draft-ietf-doh-dns-over-https-08#section-6)

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Require URLs for DO..." ]
>The draft has this to say:
>```
>6.  HTTP Integration
>
>   This protocol MUST be used with the https scheme URI [RFC7230].
>```
>(https://tools.ietf.org/html/draft-ietf-doh-dns-over-https-08#section-6)

Noted. Also note that RFC don't tell you how to implement things directly, they 
tell you how to be conforming to the spec so you can interop. It still think we 
should on care on this level...

tmthrgd:
@miekg Although I believe it be to a mistake, I have removed the https check.

miekg:
Thanks.

I believe latest discussion on DoH draft version is actually moving into this direction as well. Anyway we can always put a check in.

PR #696, opened 2018-06-06
Update vendor with latest versions.

h8liu:
Fix #695

PR #697, opened 2018-06-06
dep ensure -update

miekg:
This is a:
`go get -u` in the master branch
and then in this branch add `dep ensure -update`; which the changes committed first and then the deletes


h8liu:
looks good to me.

miekg:
thanks

merging this (because smaller) instead of #696

PR #699, opened 2018-06-09
Add IsDuplicate function

### LINKED ISSUE #679: Add dns.Equal(r1 RR, r2 RR) bool function
Add an equal function that determines if two rr's are equal from a dns protocol standpoint. This should exclude the TTL, but include the rest of the header and the rdata.

This should make dns.Dedup a lot cheaper and can be used by projects to not add duplicate RRs to a packet.

We can generate most of this function. It should mostly avoid any allocations.

miekg:
The makes checking for duplicates more efficient, as you don't have to
use String() and re-parse elements from it anymore.

Fixes #679


andrewtj:
Is comparing names case insensitively on the agenda for this?

sanitize.go
```diff
@@ -5,6 +5,8 @@ package dns
 // rrs.
 // m is used to store the RRs temporary. If it is nil a new map will be allocated.
 func Dedup(rrs []RR, m map[string]RR) []RR {
+	println("dns: Dedup: this function is deprecated")
```
tmthrgd: The Deprecated doc comment should be added to this too: https://rakyll.org/deprecated/

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Add IsDuplicate fun..." ]
>The Deprecated doc comment should be added to this too: https://rakyll.org/deprecated/

Ah nice. I've removed that printf as it doesn't really belong in this PR.

Also changed the a/aaaa compare to byte for byte as the net.Equal comment has:

func (ip IP) Equal(x IP) bool
    Equal reports whether ip and x are the same IP address. An IPv4 address
    and that same address in IPv6 form are considered to be equal.

Which is not the same as what duplicate does.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Add IsDuplicate fun..." ]
>Is comparing names case insensitively on the agenda for this?

Could point; adding + test

PR #704, opened 2018-06-23
Remove compression from AFSDB

### LINKED ISSUE #521: Don't compress domain names in post-1035 record types
Per [RFC 3597](https://tools.ietf.org/html/rfc3597#section-4):
> servers MUST NOT compress domain names embedded in the RDATA of types that are class-specific or not well-known... it is hereby specified that only the RR types defined in [[RFC1035](https://tools.ietf.org/html/rfc1035)] are to be considered "well-known".

Therefore, `compressionLenHelperType` and `pack`s should _only_ use domain name compression logic in RDATA for
* NS
* SOA
* CNAME
* PTR
* MB
* MD
* MF
* MG
* MINFO
* MR
* MX

miekg:
This type should not compress its rdata.

Fixes #521

### DIFF
--- a/types.go
+++ b/types.go
@@ -330,7 +330,7 @@ func (rr *MX) String() string {
 type AFSDB struct {
 	Hdr      RR_Header
 	Subtype  uint16
-	Hostname string `dns:"cdomain-name"`
+	Hostname string `dns:"domain-name"`
 }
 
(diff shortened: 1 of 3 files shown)


andrewtj:
LGTM

PR #705, opened 2018-06-27
Add tsig on Transfer.Out()

WeipengSun:
I'm doing a project of dns master, I found that the DNS library was not supported adding tsig when using Transfer.Out to reply Axfr or Ixfr request.
My idea is that put the tsig information from request.

### DIFF
--- a/server.go
+++ b/server.go
@@ -51,7 +51,6 @@ type response struct {
 	udp            *net.UDPConn      // i/o connection if UDP was used
 	tcp            net.Conn          // i/o connection if TCP was used
 	udpSession     *SessionUDP       // oob data to get egress interface right
-	remoteAddr     net.Addr          // address of the client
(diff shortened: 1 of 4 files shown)


miekg:
This is done in https://github.com/miekg/dns/blob/master/xfr.go#L229
why doesn't that work for you? What does this fix and what test would fail?

WeipengSun:
When using transfer out to send out an axfr/ixfr result, Envelope can only set Answer section, but not additional section(Tsig RR). function m.IsTsig in ReadMsg always return nil.

miekg:
please add a test.

WeipengSun:
--- FAIL: TestClientRemote (2.00s)
    remote_test.go:14: failed to exchange: read udp 172.17.0.25:58446->176.58.119.54:53: i/o timeout

I doesn't know how to recheck of travis-ci

PR #708, opened 2018-07-11
Add support for truncating a message to fit a given buffer size

grobie:
If DNS messages are too large for a given (UDP) buffer size, the
truncated bit should be set to request clients to retry the request via
a different transport (TCP).

This change introduces a Truncate() function which can be used to handle
the modification of the message. As some basic clients (incorrectly)
don't handle the truncation bit, it aims to return as many records as
possible while still being performant (e.g. not traversing resource
record lists multiple times to find the maximum number of fitting
records).

It ensures to always keep TSIG and OPT records in the response according
to the spec (e.g. RFC 6891 section 7).

@miekg

### DIFF
--- a/msg.go
+++ b/msg.go
@@ -923,58 +923,55 @@ func (dns *Msg) String() string {
 // than packing it, measuring the size and discarding the buffer.
 func (dns *Msg) Len() int { return compressedLen(dns, dns.Compress) }
 
-func compressedLenWithCompressionMap(dns *Msg, compression map[string]int) int {
-	l := 12 // Message header is always 12 bytes
-	for _, r := range dns.Question {
-		compressionLenHelper(compression, r.Name, l)
-		l += r.len()
-	}
-	l += compressionLenSlice(l, compression, dns.Answer)
-	l += compressionLenSlice(l, compression, dns.Ns)
-	l += compressionLenSlice(l, compression, dns.Extra)
-	return l
-}
-
 // compressedLen returns the message length when in compressed wire format
 // when compress is true, otherwise the uncompressed length is returned.
 func compressedLen(dns *Msg, compress bool) int {
-	// We always return one more than needed.
+	lenFn := Len
 	if compress {
-		compression := map[string]int{}
-		return compressedLenWithCompressionMap(dns, compression)
+		lenFn = newCompressionLenFn(dns.Question)
 	}
-	l := 12 // Message header is always 12 bytes
 
-	for _, r := range dns.Question {
-		l += r.len()
+	l := 12 // Message header is always 12 bytes
+	for _, q := range dns.Question {
+		l += q.len()
 	}
 	for _, r := range dns.Answer {
 		if r != nil {
-			l += r.len()
+			l += lenFn(r)
 		}
 	}
 	for _, r := range dns.Ns {
 		if r != nil {
-			l += r.len()
+			l += lenFn(r)
 		}
 	}
 	for _, r := range dns.Extra {
 		if r != nil {
-			l += r.len()
+			l += lenFn(r)
 		}
 	}
 
 	return l
 }
 
-func compressionLenSlice(lenp int, c map[string]int, rs []RR) int {
-	initLen := lenp
-	for _, r := range rs {
+// newCompressionLenFn returns a function which returns the compressed length of
+// a given resource record.
+//
+// Calling the returned length function is not side-effect free, as the name of
+// the record might be added to the compression dictionary, and should only be
+// used for records of the same message.
+func newCompressionLenFn(qs []Question) func(RR) int {
+	c := map[string]int{}
+	l := 12
+	for _, q := range qs {
+		compressionLenHelper(c, q.Name, l)
+		l += q.len()
+	}
+
+	return func(r RR) int {
 		if r == nil {
-			continue
+			return 0
 		}
-		// TmpLen is to track len of record at 14bits boudaries
-		tmpLen := lenp
 
 		x := r.len()
 		// track this length, and the global length in len, while taking compression into account for both.
@@ -985,17 +982,16 @@ func compressionLenSlice(lenp int, c map[string]int, rs []RR) int {
 			x += 1 - k
 		}
 
-		tmpLen += compressionLenHelper(c, r.Header().Name, tmpLen)
+		compressionLenHelper(c, r.Header().Name, l)
 		k, ok, _ = compressionLenSearchType(c, r)
 		if ok {
 			x += 1 - k
 		}
-		lenp += x
-		tmpLen = lenp
-		tmpLen += compressionLenHelperType(c, r, tmpLen)
+		l += x
+		compressionLenHelperType(c, r, l)
 
+		return x
 	}
-	return lenp - initLen
 }
 
 // Put the parts of the name in the compression map, return the size in bytes added in payload
--- a/msg.go
+++ b/msg.go
@@ -1053,6 +1053,86 @@ func compressionLenSearch(c map[string]int, s string) (int, bool, int) {
 	return 0, false, fullSize + len(s)
 }
 
+// Truncate removes resource records from Additional, Ns, and Answer sections to
+// ensure the message doesn't exceed the given response buffer size. It doesn't
+// alter the message if it already fits. If resource records have to be removed
+// from the Answer section, the TC header bit will be set.
+//
+// Compression, EDNS0 and TSIG requirements are taken into account.
+func (dns *Msg) Truncate(size int) {
+	var lenFn = Len
+	if dns.Compress {
+		lenFn = newCompressionLenFn(dns.Question)
+	}
+
+	l := 12 // Message header size in bytes.
+	for _, q := range dns.Question {
+		l += q.len()
+	}
+
+	// OPT and TSIG records must be kept in the response.
+	for _, r := range dns.Extra {
+		switch r.Header().Rrtype {
+		case TypeOPT, TypeSIG:
+			l += lenFn(r)
+		}
+	}
+
+	var n int
+	n, l = maxRRs(dns.Answer, lenFn, l, size)
+	if n < len(dns.Answer) {
+		dns.Answer = dns.Answer[:n]
+		dns.Ns = nil
+		dns.Extra = filterExtra(dns.Extra, 0)
+		dns.Truncated = true
+		return
+	}
+
+	n, l = maxRRs(dns.Ns, lenFn, l, size)
+	if n < len(dns.Ns) {
+		dns.Ns = nil
+		dns.Extra = filterExtra(dns.Extra, 0)
+		return
+	}
+
+	n, l = maxRRs(dns.Extra, lenFn, l, size)
+	if n < len(dns.Extra) {
+		dns.Extra = filterExtra(dns.Extra, n)
+		return
+	}
+}
+
+// maxRRs returns the number of records fitting into the given message size
+// as well as the length of the resulting message including the given initial
+// size.
+func maxRRs(rs []RR, lenFn func(RR) int, initial, size int) (number, length int) {
+	i, l := 0, initial
+	for ; i < len(rs); i++ {
+		if rs[i] == nil {
+			continue
+		}
+		t := lenFn(rs[i])
+		if l+t > size {
+			break
+		}
+		l += t
+	}
+	return i, l
+}
+
+// filterExtra removes all records from the given Additional section above the
+// given n index. It ensures to keep mandatory OPT and SIG records.
+func filterExtra(rs []RR, n int) []RR {
+	for i := 0; i < len(rs); i++ {
+		if i < n || rs[i].Header().Rrtype == TypeOPT || rs[i].Header().Rrtype == TypeSIG {
+			continue
+		}
+		rs = append(rs[:i], rs[i+1:]...)
+		i--
+	}
+	return rs
+}
+
 // Copy returns a new RR which is a deep-copy of r.
 func Copy(r RR) RR { r1 := r.copy(); return r1 }
 
--- a/length_test.go
+++ b/length_test.go
@@ -4,7 +4,6 @@ import (
 	"encoding/hex"
 	"fmt"
 	"net"
-	"reflect"
 	"strings"
 	"testing"
 )
@@ -278,19 +277,12 @@ func TestCompareCompressionMapsForANY(t *testing.T) {
 	for labelSize := 0; labelSize < 63; labelSize++ {
 		msg.SetQuestion(fmt.Sprintf("a%s.service.acme.", strings.Repeat("x", labelSize)), TypeANY)
 
-		compressionFake := make(map[string]int)
-		lenFake := compressedLenWithCompressionMap(msg, compressionFake)
-
-		compressionReal := make(map[string]int)
-		buf, err := msg.packBufferWithCompressionMap(nil, compressionReal)
+		buf, err := msg.packBufferWithCompressionMap(nil, make(map[string]int))
 		if err != nil {
 			t.Fatal(err)
 		}
-		if lenFake != len(buf) {
-			t.Fatalf("padding= %d ; Predicted len := %d != real:= %d", labelSize, lenFake, len(buf))
-		}
-		if !reflect.DeepEqual(compressionFake, compressionReal) {
-			t.Fatalf("padding= %d ; Fake Compression Map != Real Compression Map\n*** Real:= %v\n\n***Fake:= %v", labelSize, compressionReal, compressionFake)
+		if want, got := len(buf), msg.Len(); want != got {
+			t.Fatalf("padding= %d ; Predicted len := %d != real:= %d", labelSize, got, want)
 		}
 	}
 }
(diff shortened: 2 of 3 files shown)


msg.go
```diff
@@ -923,58 +923,55 @@ func (dns *Msg) String() string {
...
-func compressionLenSlice(lenp int, c map[string]int, rs []RR) int {
-	initLen := lenp
-	for _, r := range rs {
+// newCompressionLenFn returns a function which returns the compressed length of
+// a given resource record.
+//
+// Calling the returned length function is not side-effect free, as the name of
+// the record might be added to the compression dictionary, and should only be
+// used for records of the same message.
+func newCompressionLenFn(qs []Question) func(RR) int {
```
miekg: This will only carry 1 question at most, and looking at the code we only use the `q.Name`, so I think we can just use a string here, and drop the range below

msg.go
```diff
@@ -1057,6 +1053,86 @@ func compressionLenSearch(c map[string]int, s string) (int, bool, int) {
...
+			break
+		}
+		l += t
+	}
+	return i, l
+}
+
+// filterExtra removes all records from the given Additional section above the
+// given n index. It ensures to keep mandatory OPT and SIG records.
+func filterExtra(rs []RR, n int) []RR {
```
miekg: I would just name this filterRRs to mirror maxRRs.

msg.go
```diff
@@ -1057,6 +1053,86 @@ func compressionLenSearch(c map[string]int, s string) (int, bool, int) {
...
+		return
+	}
+}
+
+// maxRRs returns the number of records fitting into the given message size
+// as well as the length of the resulting message including the given initial
+// size.
+func maxRRs(rs []RR, lenFn func(RR) int, initial, size int) (number, length int) {
+	i, l := 0, initial
+	for ; i < len(rs); i++ {
```
miekg: this doesn't use binary search, so it's slower than the impl in coredns. Not sure if we need if-then to only do binary search of the number of RRs is larger than X

grobie: The CoreDNS implementation does a binary search by calling Msg.Len() multiple times, so it calculates each answer/authority/additional resource record's len() multiple times. In this implementation the len() of each resource record is calculated at most once.

I don't understand how we could do a binary search here, we have to calculate the length of each record at least once as it's not fixed.

msg.go
grobie: I had that first and passed the RR types as argument, and then thought that's overkill and changed the name. Happy to rename again.

grobie:
While I thought the Len `func(RR) int` interface is a cool way to simplify the code, it seems to come at a performance penalty. I expect due to the usage of a closure and the increased number of function calls. :-( Guess I'll have to find a simpler solution.

andrewtj:
> As some basic clients (incorrectly) don't handle the truncation bit

Which ones?

---

As-is `Truncate()` seems a bit misleading. To my mind a general purpose `Truncate()` would only pop records in complete sets from the additional section. What this currently does seems like application-specific behaviour.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Add support for tru..." ]
>+// maxRRs returns the number of records fitting into the given message size
>+// as well as the length of the resulting message including the given initial
>+// size.
>+func maxRRs(rs []RR, lenFn func(RR) int, initial, size int) (number, length int) {
>+	i, l := 0, initial
>+	for ; i < len(rs); i++ {
>
>The CoreDNS implementation does a binary search by calling Msg.Len() multiple times, so it calculates each answer/authority/additional resource record's len() multiple times. In this implementation the len() of each resource record is calculated at most once.

Ah, yes. And then that `Len()` then just runs in O(n), as does this. So 
optmizing that makes sense. OK

>I don't understand how we could do a binary search here, we have to calculate the length of each record at least once as it's not fixed.

No, I see. Wish we could though. There were some thoughts on caching the wire 
lenght in the message - which can work in special cases; but breaks down when 
you do a `a.Extra = append(a.Extra, x)`, but it might be worth trying something 
that can speed up this case.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Add support for tru..." ]
>> As some basic clients (incorrectly) don't handle the truncation bit
>
>Which ones?
>
>---
>
>As-is `Truncate()` seems a bit misleading. To my mind a general purpose `Truncate()` would only pop records in complete sets from the additional section. What this currently does seems like application-specific behaviour.

A previously suggested name for this was `Fit`. I agree with your RRSet remark 
though; that needs to be taken into account; and then you're getting in DNSSEC 
territory, so we also need to drop any RRSIG if we drop the RRset.

/Miek

--
Miek Gieben

grobie:
> Which ones?

There were reports on CoreDNS when we dropped all answer records at once. Windows/net clients iirc.

> To my mind a general purpose Truncate() would only pop records in complete sets from the additional section. What this currently does seems like application-specific behaviour.

My goal is to provide a general-purpose, RFC compliant Truncate() function, not anything application specific. Though, with all things DNS, it's complicated ...

I looked at the DNS rfcs (especially [RFC 2181: Clarifications to the DNS Specification](https://tools.ietf.org/html/rfc2181)), [BIND](https://www.isc.org/downloads/bind/), and [unbound](https://www.unbound.net/) and mostly modeled the behavior after unbound, given its widespread usage and most permissive implementation.

- [Section 9 of the RFC](https://tools.ietf.org/html/rfc2181#section-9) states "... the results of additional section processing.  In such cases the entire RRSet that will not fit in the response should be omitted,", means we "should" drop whole RRSets from the additional section, and not return partial sets. It's not a MUST though :smile: 
  On the other hand, the RFC explicitly permits partial results in the answer section: "Where TC is set, the partial RRSet that would not completely fit may  be left in the response."
- BIND is super strict, as soon as any record of the authority or answer section doesn't fit, it'll return an empty response only inlcuding OPT records and the TC bit set.
- unbound is a permissive, it will return partial RRsets in the additional section, drop just the authority section without setting the TC bit if the answer section would fit entirely then. However it'll not return partial RRSets in the Answer section, even though the RFCs permits that.

Summary:
- Returning partial RRSets in the Answer section is allowed. Neither bind nor unbound do that. It helps stupid clients not supporting truncation/TCP retry.
- Omitting the Authority section without setting the TC bit is not specified AFAIK. Bind will truncate, unbound not.
- Returning partial RRSets in the Extra section is not advised by the RFC. BIND won't do it, unbound will though.

> A previously suggested name for this was `Fit`.

I'd highly prefer to follow the terminology of the protocol and not invent new names. The whole thing is confusing enough already, having to remember all custom names and their meaning has made the usage of the libraries harder than necessary.

andrewtj:
What's the benefit to returning partial sets in the authority or additional section (with the TC bit unset)? Do the clients that don't understand the TC bit use those sections?

Pruning sets or clearing all the non-ephemeral records seems more useful. With partial sets clients will end up with a partial view, or if they're validating, they'll do extra work just to drop the records.

This still seems like application specific behaviour to me. It's certainly query response specific, if not resolver to client specific. If this is going to land then the name and docs should reflect that imo.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Add support for tru..." ]
>What's the benefit to returning partial sets in the authority or additional section (with the TC bit unset)? Do the clients that don't understand the TC bit use those sections?

Sadly yes, alpine linux doesn't requery using tcp and it want something that 
resembles a sane reply.

>Pruning sets or clearing all the non-ephemeral records seems more useful. With partial sets clients will end up with a partial view, or if they're validating, they'll do extra work just to drop the records.
>
>This still seems like application specific behaviour to me. It's certainly query response specific, if not resolver to client specific. If this is going to land then the name and docs should reflect that imo.

I'm not sure either this function should be in this library. However cutting up 
a response into a chunk that is valid (enough) *and* fits the udp buffer is 
proving a more challenging problem that it looks.

/Miek

--
Miek Gieben

grobie:
Neither this implementation nor the current one in CoreDNS returns partial Authority section. It's all or nothing.

I think the idea behind partial Additional responses was to give clients some hints already, so for example for a Answer SRV RRSet, they could already get half the A RRSet in the additional section. I see the issues with that approach, so we could simply also switch to an all/none RRSet approach here.

> This still seems like application specific behaviour to me. It's certainly query response specific, if not resolver to client specific.

I don't understand yet how the Truncation behavior is application specific. Different applications appear to interpret the RFCs/spec differently, but there is only one concept of truncation in DNS. The complexity of getting Truncation right and the lack of support in this library is what has led to bad/broken behavior in at least 3 server implementations I have come across.

I guess I'm missing something here though, can you give examples how different applications will respond with differently truncated responses @andrewtj @miekg? I had hoped we could iron out all the edge cases and provide a reasonable generic implementation in this library.

The feedback I took away so far and will incorporate:
- Don't return partial RRSets in Additional section
- SIG records need to be taken care of if RRSets get dropped
- Performance needs to be checked

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Add support for tru..." ]
>> This still seems like application specific behaviour to me. It's certainly query response specific, if not resolver to client specific.
>
>I don't understand yet how the Truncation behavior is application specific. 
>Different applications appear to interpret the RFCs/spec differently, but there 
>is only one concept of truncation in DNS. The complexity of getting Truncation 
>right and the lack of support in this library is what has led to bad/broken 
>behavior in at least 3 server implementations I have come across.

I agree, this is surprisingly difficult and having something blessed in this lib 
will help.

>The feedback I took away so far and will incorporate:
>- Don't return partial RRSets in Additional section
>- SIG records need to be taken care of if RRSets get dropped
>- Performance needs to be checked

yes, not simple things, mind you.

I tried caching some lengths in RRs, but that improve things.
If you perf the Scrub function in CoreDNS all lot of time is spend in looking up 
keys in the compression map and doing NextLabel. (bloody compression)

>https://github.com/miekg/dns/pull/708#issuecomment-404502282
/Miek

--
Miek Gieben

andrewtj:
On Thu, Jul 12, 2018 at 10:04 PM, Miek Gieben <notifications@github.com>
wrote:

> [ Quoting <notifications@github.com> in "Re: [miekg/dns] Add support for
> tru..." ]
> >What's the benefit to returning partial sets in the authority or
> additional section (with the TC bit unset)? Do the clients that don't
> understand the TC bit use those sections?
>
> Sadly yes, alpine linux doesn't requery using tcp and it want something
> that
> resembles a sane reply.
>
>
Do you know if that's a WONTFIX? And it actually expects to get everything
from additional, or it's just an example of a client that doesn't
understand the TC bit?

msg.go
```diff
@@ -1057,6 +1053,86 @@ func compressionLenSearch(c map[string]int, s string) (int, bool, int) {
 	return 0, false, fullSize + len(s)
 }
 
+// Truncate removes resource records from Additional, Ns, and Answer sections to
+// ensure the message doesn't exceed the given response buffer size. It doesn't
+// alter the message if it already fits. If resource records have to be removed
+// from the Answer section, the TC header bit will be set.
```
gibson042: Required records are not confined to the Answer section... in referrals and negative responses, they appear in Additional.

gibson042:
> there's code out there that blindly retries queries over TCP even when the TC response came over TCP

Any server that sends truncated responses over TCP is inviting trouble in the first place, since "The TC bit should be set in responses only when an RRSet is required as a part of the response, but could not be included in its entirety" ([RFC 2181](https://tools.ietf.org/html/rfc2181#section-9)).

> The discovery-via-DNS servers out there will probably want to truncate the message and not set the TC bit so that the set doesn't get occluded (for them partial data is much better than no data).

Sending a partial RRSet _without_ setting the TC bit is a pretty gross violation. Can you explain why a server should ever do that?

> this implements one strategy among many for truncating query-response messages. I think the docs need to reflect that.

I agree. But if supporting other strategies is important, maybe the future holds a two-argument `Truncate` that parameterizes it.

andrewtj:
On Fri, Jul 13, 2018 at 12:58 PM, Richard Gibson <notifications@github.com>
wrote:

> there's code out there that blindly retries queries over TCP even when the
> TC response came over TCP
>
> Any server that sends truncated responses over TCP is inviting trouble in
> the first place, since "The TC bit should be set in responses only when an
> RRSet is required as a part of the response, but could not be included in
> its entirety" (RFC 2181 <https://tools.ietf.org/html/rfc2181#section-9>).
>

Agreed, it's unfortunate but servers can end up loading RRSets they can't
serve. Robust clients need to deal with that.

The discovery-via-DNS servers out there will probably want to truncate the
> message and not set the TC bit so that the set doesn't get occluded (for
> them partial data is much better than no data).
>
> Sending a partial RRSet *without* setting the TC bit is a pretty gross
> violation. Can you explain why a server should ever do that?
>

Sure. When I say discovery-via-DNS server (there's probably a better name
for them) I'm referring to the servers that synthesise records for services
registered with them via some orchestration framework or the like. When
they get too many services registered, until they're rebalanced they can
either serve TC which makes a service unavailable, or serve a partial set
which allows the service to remain available.

this implements one strategy among many for truncating query-response
> messages. I think the docs need to reflect that.
>
> I agree. But if supporting other strategies is important, maybe the future
> holds a two-argument Truncate that parameterizes it.
>
>
Maybe, do you have something in mind?

gibson042:
> Sure. When I say discovery-via-DNS server (there's probably a better name for them) I'm referring to the servers that synthesise records for services registered with them via some orchestration framework or the like. When they get too many services registered, until they're rebalanced they can either serve TC which makes a service unavailable, or serve a partial set which allows the service to remain available.

That's a false choice, because serving a partial RRSet with TC=1 is valid (if unusual). So why would TC=1 make a service unavailable? Its meaning is simply "message truncated", as it has been since at least [1987](https://tools.ietf.org/html/rfc1035#page-27).

> > But if supporting other strategies is important, maybe the future holds a two-argument Truncate that parameterizes it.

> Maybe, do you have something in mind?

Off the top of my head, something like `response.Truncate(mtu, dns.TRUNCATE_TO_PARTIAL_RRSET)`.

andrewtj:
On Fri, Jul 13, 2018 at 2:46 PM, Richard Gibson <notifications@github.com>
wrote:

> Sure. When I say discovery-via-DNS server (there's probably a better name
> for them) I'm referring to the servers that synthesise records for services
> registered with them via some orchestration framework or the like. When
> they get too many services registered, until they're rebalanced they can
> either serve TC which makes a service unavailable, or serve a partial set
> which allows the service to remain available.
>
> That's a false choice, because serving a partial RRSet with TC=1 is valid
> (if unusual). So why would TC=1 make a service unavailable? Its meaning is
> simply "message truncated", as it has been since at least 1987
> <https://tools.ietf.org/html/rfc1035#page-27>.
>
>
> This solution came up in light of a resolver that returned SERVFAIL when
it encountered truncated TCP responses. Which resolvers do pass on such
responses?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Add support for tru..." ]
>> tru..." ]
>> >What's the benefit to returning partial sets in the authority or
>> additional section (with the TC bit unset)? Do the clients that don't
>> understand the TC bit use those sections?
>>
>> Sadly yes, alpine linux doesn't requery using tcp and it want something
>> that
>> resembles a sane reply.
>>
>>
>Do you know if that's a WONTFIX? And it actually expects to get everything
>from additional, or it's just an example of a client that doesn't
>understand the TC bit?

Yes, I think they claimed it is not worth the complexity adding this (there 
could be something said for that). But as of RFC 5966 this is not compliant.

andrewtj:
I did not say responding with a SERVFAIL to a TC=1 TCP response was not justified.

Message truncation to suit transport over TCP does happen. Use of the TC bit is not transport specific.

gibson042:
The TC bit 100% _is_ transport specific.

https://tools.ietf.org/html/rfc1035#page-27
> TrunCation - specifies that this message was truncated due to **length greater than that permitted on the transmission channel**.

https://tools.ietf.org/html/rfc1035#section-4.2.1
> Messages carried by UDP are restricted to 512 bytes (not counting the IP or UDP headers). **Longer messages are truncated** and the TC bit is set in the header.

https://tools.ietf.org/html/rfc2181#section-5.1
> The response must be marked as "truncated" if the entire RRSet **will not fit in the response**.

The maximum length of a DNS message transported over TCP is 65535 octets, so setting TC=1 for a message shorter than that is wrong on its face. But it is also wrong even for messages _longer_ than that, because that means the fallback itself has failed ([RFC 2181 §9](https://tools.ietf.org/html/rfc2181#section-9): "_When a DNS client receives a reply with TC set, it should ignore that response, and query again, using a mechanism, such as a TCP connection, that will permit larger replies._"). So simply put, TC=1 has no place in any TCP transport.

andrewtj:
@gibson042 I did not suggest setting TC=1 for a message that has not been truncated to suit its transport. I'm not going to continue discussing whether or not TC=1 messages can be transmitted over TCP. If you do not think this concept exists, setup a BIND playground and then take up your argument with ISC.

gibson042:
If you have evidence that BIND sends TC=1 over a TCP transport, I'd love to see it. Here's evidence that it does not:

<blockquote>
$ dig @127.0.0.1 -p 10053 +short CH version.bind TXT

"9.9.5-3ubuntu0.11-Ubuntu"
<hr>

$ dig @127.0.0.1 -p 10053 +noall +comments +stats   example.com TXT

;; **Truncated, retrying in TCP mode**.
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 50872
;; flags: qr aa rd ra; QUERY: 1, ANSWER: 40, AUTHORITY: 1, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; Query time: 0 msec
;; SERVER: 127.0.0.1#10053(127.0.0.1)
;; WHEN: Fri Jul 13 23:53:55 EDT 2018
;; MSG SIZE  **rcvd: 4586**
<hr>

$ dig @127.0.0.1 -p 10053 +noall +comments +stats **+ignore** example.com TXT

;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 3254
;; flags: qr aa **tc** rd ra ad; QUERY: 1, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; Query time: 0 msec
;; SERVER: 127.0.0.1#10053(127.0.0.1)
;; WHEN: Fri Jul 13 23:53:57 EDT 2018
;; MSG SIZE  **rcvd: 40**
<hr>

$ dig @127.0.0.1 -p 10053 +noall +comments +stats **+tcp +bufsize=4096 +ignore** example.com TXT

;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 64583
;; flags: qr aa rd ra; QUERY: 1, ANSWER: 40, AUTHORITY: 1, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; Query time: 0 msec
;; SERVER: 127.0.0.1#10053(127.0.0.1)
;; WHEN: Fri Jul 13 23:54:04 EDT 2018
;; MSG SIZE  **rcvd: 4586**
</blockquote>

andrewtj:
@gibson042 Use an RRSet that is too large for the transport.

andrewtj:
I'm changing my position on this. @grobie's solution isn't perfect but it's probably good for most use cases, and with the [mentioned changes](https://github.com/miekg/dns/pull/708#issuecomment-404502282) and tighter docs, I don't think there's much harm in having it. This PR isn't progressing in part because of my commentary, so I'm gonna leave it here and say goodbye.

miekg:
will be done differently.

PR #713, opened 2018-07-21
Server - Graceful shutdown (ensure all queries are processed before closing)

fturib:
## Summary

Enhance the Shutdown of the server to ensure that all queries received before shutdown is called are completely handled before closing the connections.

related PR : https://github.com/coredns/coredns/pull/1963
related issue : https://github.com/coredns/coredns/issues/1666

### tracking the incoming msg
It requires a SyncGroup to take into account the thread going into READ wait mode, and ack the handle is finished after either an error or the full service of the query.

After shutdown is started none of the listening thread can go start a new READ process. However the current one has to be continued until either a msg is received, either a timeout is reached.

At shutdown, the process wait that all listening thread are terminated before closing the connection

### Timeout
For those who does not want to wait, it is now possible to specify a ShudownTimeout.
Default value is 0 - which means use the biggest of the ReadTimeout of TCP or UDP (incl idleTimeout).

## Tests

I added a test "testLostQueriesAtShutdownServer" that:
- run a server that handle any incoming queries within a random delay of 100 to 600ms
- run a set of 50 concurrent thread. Each one is sending a query to the server until STOP is asked
- wait 1 sec all thread are running
- ask for STOP the msg, 
- wait 50ms that any started sending msg are received bu Server
- ask for Shutdown of the server
- count all sent msg and all received msg by the 50 threads.
- ensure that sent msg == received msg

This test is run for a each of the 3 protocols : UDP, TCP, TLS

### DIFF
--- a/server.go
+++ b/server.go
@@ -8,7 +8,6 @@ import (
 	"encoding/binary"
 	"io"
 	"net"
-	"runtime"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -486,6 +485,33 @@ func (srv *Server) ActivateAndServe() error {
 	return &Error{err: "bad listeners"}
 }
 
+func (srv *Server) getShutdownTimeout() time.Duration {
+	if srv.ShutdownTimeout != time.Duration(0) {
+		return  srv.ShutdownTimeout
+	}
+
+	// if not defined, the shutdownTimeout should be > to the readTimeout where the thread are blocked waiting for msg
+	// UDP is using only readTimeout
+	timeout := srv.getReadTimeout()
+
+	if srv.Listener != nil {
+		// if TCP listener is in use, the read timeout may be extended with the idleTimeout
+		idleTimeout := tcpIdleTimeout
+		if srv.IdleTimeout != nil {
+			idleTimeout = srv.IdleTimeout()
+		}
+		if idleTimeout > timeout {
+			timeout = idleTimeout
+		}
+	}
+
+	// add a little time to ensure we have time to wait the thread are released
+	timeout = timeout + time.Second
+	return timeout
+
+}
+
+
 // Shutdown shuts down a server. After a call to Shutdown, ListenAndServe and
 // ActivateAndServe will return.
 func (srv *Server) Shutdown() error {
@@ -503,46 +529,29 @@ func (srv *Server) Shutdown() error {
 
 	// and then stop the service : it will wait all queries are served
 
-	timeout := srv.getReadTimeout()
-	idleTimeout := time.Duration(0)
-
 	// we can close the TCP listener (the accept() part)
 	if srv.Listener != nil {
 		srv.Listener.Close()
 
-		// if TCP listener is in use, the read timeout may be extended with the idleTimeout
-		idleTimeout = tcpIdleTimeout
-		if srv.IdleTimeout != nil {
-			idleTimeout = srv.IdleTimeout()
-		}
 	}
 
-	if idleTimeout > timeout {
-		timeout = idleTimeout
-	}
-	timeout = timeout + time.Second
-	if srv.ShutdownTimeout != time.Duration(0) {
-		timeout  = srv.ShutdownTimeout
-	}
 
 	// and then wait all TCPConn and PacketConn finish the on-going work
-	if runtime.GOOS != "windows" {
-		// force connections to close after timeout
-		done := make(chan struct{})
-		go func() {
-			// wait all packet started to be processed are now processed
-			srv.dnsWg.Wait()
-			close(done)
-		}()
+	// force connections to close after timeout
+	done := make(chan struct{})
+	go func() {
+		// wait all packet started to be processed are now processed
+		srv.dnsWg.Wait()
+		close(done)
+	}()
 
-		// Wait for remaining connections to finish or
-		// force them all to close after timeout
-		// use the max of all timeout for UDP, TCP and TCP wait
+	// Wait for remaining connections to finish or
+	// force them all to close after timeout
+	// use the max of all timeout for UDP, TCP and TCP wait
 
-		select {
-		case <-time.After(timeout):
-		case <-done:
-		}
+	select {
+	case <-time.After(srv.getShutdownTimeout()):
+	case <-done:
 	}
 
 	// at this point all Msg received should have been processed
@@ -580,14 +589,6 @@ func (srv *Server) serveTCP(l net.Listener) error {
 			if neterr, ok := err.(net.Error); ok && neterr.Temporary() {
 				continue
 			}
-			// if listener is closed while Accept, we'll get an error
-			// we just not want to notify this error if it is related to shutdown
-			srv.lock.RLock()
-			if !srv.started {
-				srv.lock.RUnlock()
-				return nil
-			}
-			srv.lock.RUnlock()
 			return err
 		}
 		srv.spawnWorker(&response{tsigSecret: srv.TsigSecret, tcp: rw})
--- a/server.go
+++ b/server.go
@@ -391,6 +391,7 @@ func (srv *Server) ListenAndServe() error {
 		srv.UDPSize = MinMsgSize
 	}
 	srv.queue = make(chan *response)
+	defer close(srv.queue)
 	switch srv.Net {
 	case "tcp", "tcp4", "tcp6":
 		a, err := net.ResolveTCPAddr(srv.Net, addr)
@@ -459,6 +460,7 @@ func (srv *Server) ActivateAndServe() error {
 	pConn := srv.PacketConn
 	l := srv.Listener
 	srv.queue = make(chan *response)
+	defer close(srv.queue)
 	if pConn != nil {
 		if srv.UDPSize == 0 {
 			srv.UDPSize = MinMsgSize
@@ -559,7 +561,7 @@ func (srv *Server) ShutdownContext(ctx context.Context) error {
 		err = ctx.Err()
 	}
 
-	srv.closeResources()
+	srv.closeResources(true)
 	return err
 }
 
@@ -574,7 +576,7 @@ func (srv *Server) waitShutdownCompletes() {
 
 }
 
-func (srv *Server) closeResources() {
+func (srv *Server) closeResources(closeShut bool) {
 	srv.lock.Lock()
 	defer srv.lock.Unlock()
 	if srv.Listener != nil {
@@ -585,9 +587,9 @@ func (srv *Server) closeResources() {
 		srv.PacketConn.Close()
 		srv.PacketConn = nil
 	}
-	if srv.queue != nil {
-		close(srv.queue)
-		srv.queue = nil
+	if closeShut {
+		close(srv.shut)
+		srv.shut = nil
 	}
 }
 
@@ -623,7 +625,7 @@ func (srv *Server) serveTCP(l net.Listener) error {
 	}
 	// service is ended - ensure closing the resources
 	srv.waitShutdownCompletes()
-	srv.closeResources()
+	srv.closeResources(false)
 	return reterr
 }
 
@@ -664,7 +666,7 @@ func (srv *Server) serveUDP(l *net.UDPConn) error {
 
 	// service is ended - ensure closing the resources
 	srv.waitShutdownCompletes()
-	srv.closeResources()
+	srv.closeResources(false)
 	return reterr
 }
 
--- a/server.go
+++ b/server.go
@@ -313,10 +313,10 @@ type Server struct {
 	// Shutdown handling
 	lock    sync.RWMutex
 	started bool
-	dnsWg       sync.WaitGroup
 	// track packet processing and allow to wait all are processed before closing
+	dnsWg       sync.WaitGroup
+	// Shutdown timeout. If not indicated, the timeout will be max(readTimeout, idleTimeout) + 1 sec
 	ShutdownTimeout time.Duration
-	// Shutdown timeout. If not indicated, the timeout will be max(readTimeout, idleTimeout)
 
 }
 
@@ -512,8 +512,9 @@ func (srv *Server) getShutdownTimeout() time.Duration {
 }
 
 
-// Shutdown shuts down a server. After a call to Shutdown, ListenAndServe and
-// ActivateAndServe will return.
+// Shutdown shuts down gracefully a server. All in progress queries will be dealt with
+// After a call to Shutdown, ListenAndServe and ActivateAndServe will return when all accepted queries are handled
+// or shutdown timeout occurs
 func (srv *Server) Shutdown() error {
 	srv.lock.Lock()
 	if !srv.started {
@@ -522,44 +523,37 @@ func (srv *Server) Shutdown() error {
 	}
 	// inform the server is no more alive
 	// it will prevent to start read new packet
-	// however the already started READ session will continue.
+	// however the already started READ session will continue until timeout or receiving one msg
 	// it can take to MAX(readTimeout, TCPIdleTimeout) to have all READ session processed
 	srv.started = false
 	srv.lock.Unlock()
 
-	// and then stop the service : it will wait all queries are served
-
-	// we can close the TCP listener (the accept() part)
+	// we can safely close the TCP listener : no more incoming queries will be accepted
 	if srv.Listener != nil {
 		srv.Listener.Close()
-
 	}
+	// for UDP side the Listener is also the Connection so we cannot cloe it here
+	// however, the reading loop will exit as soon as it see the flag srv.started == false
 
 
-	// and then wait all TCPConn and PacketConn finish the on-going work
-	// force connections to close after timeout
+	// wait all queries in progress are processed
 	done := make(chan struct{})
 	go func() {
-		// wait all packet started to be processed are now processed
 		srv.dnsWg.Wait()
 		close(done)
 	}()
-
-	// Wait for remaining connections to finish or
-	// force them all to close after timeout
-	// use the max of all timeout for UDP, TCP and TCP wait
 
 	select {
 	case <-time.After(srv.getShutdownTimeout()):
 	case <-done:
 	}
 
-	// at this point all Msg received should have been processed
-	// or timed out
-
+	// no more msg to send. We can safely cloe the UDP connection
 	if srv.PacketConn != nil {
 		srv.PacketConn.Close()
 	}
+	// on TCP side, the TCP connections are automatically closed at end of function serve(..)
+	// or are handled by the client side (see Hijack())
 	return nil
 }
 
--- a/server.go
+++ b/server.go
@@ -8,6 +8,7 @@ import (
 	"encoding/binary"
 	"io"
 	"net"
+	"runtime"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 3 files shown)


fturib:
@miekg : not sure who is supposed to review here. 
This PR is to replace this one in CoreDNS : https://github.com/coredns/coredns/pull/1963

server.go
```diff
@@ -511,17 +565,25 @@ func (srv *Server) serveTCP(l net.Listener) error {
...
 
 	for {
-		rw, err := l.Accept()
 		srv.lock.RLock()
 		if !srv.started {
 			srv.lock.RUnlock()
 			return nil
 		}
 		srv.lock.RUnlock()
+		rw, err := l.Accept()
```
miekg: why is this moved down?

server.go
```diff
@@ -511,17 +565,25 @@ func (srv *Server) serveTCP(l net.Listener) error {
...
 		if err != nil {
 			if neterr, ok := err.(net.Error); ok && neterr.Temporary() {
 				continue
 			}
+			// if listener is closed while Accept, we'll get an error
+			// we just not want to notify this error if it is related to shutdown
+			srv.lock.RLock()
```
miekg: more locks? And again checking if we are started?

server.go
```diff
@@ -544,20 +606,26 @@ func (srv *Server) serveUDP(l *net.UDPConn) error {
...
 	for {
-		m, s, err := reader.ReadUDP(l, rtimeout)
 		srv.lock.RLock()
+		// check server is still alive
 		if !srv.started {
 			srv.lock.RUnlock()
 			return nil
 		}
+		// and flag that we are processing another packet
+		srv.dnsWg.Add(1)
```
miekg: changes here look sane.

server_test.go
```diff
@@ -50,17 +52,26 @@ func AnotherHelloServer(w ResponseWriter, req *Msg) {
 }
 
 func RunLocalUDPServer(laddr string) (*Server, string, error) {
-	server, l, _, err := RunLocalUDPServerWithFinChan(laddr)
+	server, l, _, err := RunLocalUDPServerWithFinChan(laddr, 0)
```
miekg: Probably makes sense to make a new RunLocalUDPserverWithFinChan function that sets the rdtimeout (or set it after you created it?)

server_test.go
```diff
@@ -50,17 +52,26 @@ func AnotherHelloServer(w ResponseWriter, req *Msg) {
...
 		return nil, "", nil, err
 	}
-	server := &Server{PacketConn: pc, ReadTimeout: time.Hour, WriteTimeout: time.Hour}
+
+	// tune the timeouts. If shutdown is lower than rdtimeout and there is no queries,
+	// then ActivateAndServe will end on an error because connection will be close before timeout of ReadUDP.
+
+	shtimeout := rdtimeout + (1 * time.Second)  // to avoid error return by Server, we need that ShutdownTimeout is > ReadTimeout
+	if rdtimeout == 0 {
+		rdtimeout = 1 * time.Hour
```
miekg: hour?

server_test.go
```diff
@@ -575,6 +586,98 @@ func TestShutdownTCP(t *testing.T) {
 	}
 }
 
+func testLostQueriesAtShutdownServer(t *testing.T, srv *Server, addr string, fin chan error, client *Client) {
```
miekg: this is not tested because test is lowercase

server.go
```diff
@@ -481,14 +487,62 @@ func (srv *Server) Shutdown() error {
...
+	if idleTimeout > timeout {
+		timeout = idleTimeout
+	}
+	timeout = timeout + time.Second
+	if srv.ShutdownTimeout != time.Duration(0) {
+		timeout  = srv.ShutdownTimeout
+	}
+
+	// and then wait all TCPConn and PacketConn finish the on-going work
+	if runtime.GOOS != "windows" {
```
miekg: don't really care this doesnt(?) work on windows. There is no runtime.GOOS in this package, would not like to see it added

server_test.go
fturib: I will change with an option rdtimeout in RunLocalUDPServerWithFinChan as suggested above.
All values were formerly in hours ...

server_test.go
fturib: yes. it is a common function that will be called by 3 test cases (UDP. TCP. TLS)

server.go
fturib: ok. I will remove.

server.go
fturib: why is this move down ? 
=> if Accept() is returning a TCPConn, that means there is most likely an incoming Msg. It is too late to refuse it if the server is stopped - we need to process that msg.
(which means I need to invert also on the process of the TCPConn side)

fturib: hum.. finally, to avoid too much complexity, I will move back as initial.

server.go
fturib: ...and then I can remove this lock.

miekg:
i.e. finishing my comment: If we can't graceful shutdown UDP; we don't need to `dnsWg.Add(1)` in the udp handling? 

And that sucks, cause graceful handling of udp would be the biggest feature here.

fturib:
both UDP, and TCP are graceful shutdown here.
=> There is UT on each of them.

on TCP : can close the Listener BEFORE waiting all query to finish, because its query has it own TCPConn
on UDP : we must close the UDPConn AFTER all queries are done, because we need that UDP Conn to send the reply.

server.go
```diff
@@ -481,14 +513,45 @@ func (srv *Server) Shutdown() error {
...
+
+	select {
+	case <-time.After(srv.getShutdownTimeout()):
+	case <-done:
+	}
+
+	// at this point all Msg received should have been processed
+	// or timed out
+
+	if srv.PacketConn != nil {
```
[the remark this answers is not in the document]
fturib: It is graceful. There is a UT on it.

server.go
```diff
@@ -313,6 +313,11 @@ type Server struct {
 	// Shutdown handling
 	lock    sync.RWMutex
 	started bool
+	dnsWg       sync.WaitGroup
+	// track packet processing and allow to wait all are processed before closing
+	ShutdownTimeout time.Duration
```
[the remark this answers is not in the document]
fturib: I did at first .. 
but most of the UT are running servers for 1 hour as readtimeout => in that case the shutdown nay wait until the end of that timeout.

I am not sure how real case it is to have a long readtimeout ...

I am ok to remove it. 
But for UT I will have to review the timeout values used for read.
What do you think ?

server.go
```diff
@@ -313,6 +313,11 @@ type Server struct {
...
+	// Shutdown timeout. If not indicated, the timeout will be max(readTimeout, idleTimeout)
```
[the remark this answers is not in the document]
fturib: right.

server.go
```diff
@@ -481,14 +513,45 @@ func (srv *Server) Shutdown() error {
...
 		srv.Listener.Close()
+
+	}
+
+
+	// and then wait all TCPConn and PacketConn finish the on-going work
+	// force connections to close after timeout
+	done := make(chan struct{})
+	go func() {
+		// wait all packet started to be processed are now processed
```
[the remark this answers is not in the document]
fturib: will remvoe.

fturib:
the function serveUDP is an infinite loop that 
- listen incoming UDP msg
- ask a worker to handle it.

this loop just stop when it detect the server starts to shutdown. (on !srv.started)

server.go
```diff
@@ -313,6 +313,11 @@ type Server struct {
 	// Shutdown handling
 	lock    sync.RWMutex
 	started bool
+	dnsWg       sync.WaitGroup
+	// track packet processing and allow to wait all are processed before closing
```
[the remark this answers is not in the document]
fturib: oops.

server.go
```diff
@@ -473,6 +478,33 @@ func (srv *Server) ActivateAndServe() error {
...
+
+	// add a little time to ensure we have time to wait the thread are released
+	timeout = timeout + time.Second
+	return timeout
+
+}
+
+
 // Shutdown shuts down a server. After a call to Shutdown, ListenAndServe and
 // ActivateAndServe will return.
```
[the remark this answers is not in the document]
fturib: ok

fturib:
hum .. one msg missed during graceful shutdown in the case of  UT related to TLS.
I need to review what can happen here and if it is UT or a race condition on the wait for in-progress queries.

server.go
```diff
@@ -313,6 +313,11 @@ type Server struct {
 	// Shutdown handling
 	lock    sync.RWMutex
 	started bool
+	// track packet processing and allow to wait all are processed before closing
+	dnsWg       sync.WaitGroup
+	// Shutdown timeout. If not indicated, the timeout will be max(readTimeout, idleTimeout) + 1 sec
+	ShutdownTimeout time.Duration
```
miekg: I don't think we need or want *another* timer in this struct, just us the readTimeout, idleTimeout .

server_test.go
```diff
@@ -575,6 +578,97 @@ func TestShutdownTCP(t *testing.T) {
...
 }
 
+func checkInProgressQueriesAtShutdownServer(t *testing.T, srv *Server, addr string, fin chan error, client *Client) {
+
+	var h = func(w ResponseWriter, req *Msg) {
+		// simulate small delay between 0 to 0.5 sec.
+		time.Sleep(time.Duration((rand.Intn(900))+ 100)*time.Millisecond)
+		HelloServer(w, req)
+	}
+	HandleFunc("example.com", h)
```
tmthrgd: `h` doesn’t need a variable. Just include the `func` in `HandleFunc`.

server_test.go
```diff
@@ -575,6 +578,97 @@ func TestShutdownTCP(t *testing.T) {
...
+
+
+	var sendMsg int64
+	var recvMsg int64
+	var wg sync.WaitGroup
+
+	stop := make(chan bool)
```
tmthrgd: `chan struct{}` by convention.

server_test.go
```diff
@@ -575,6 +578,97 @@ func TestShutdownTCP(t *testing.T) {
 	}
 }
 
+func checkInProgressQueriesAtShutdownServer(t *testing.T, srv *Server, addr string, fin chan error, client *Client) {
```
tmthrgd: This test seems a bit flaky to me (from a quick read over). Is it guaranteed to always succeed?

server.go
fturib: ok.

server_test.go
```diff
@@ -60,7 +62,8 @@ func RunLocalUDPServerWithFinChan(laddr string) (*Server, string, chan error, er
 	if err != nil {
 		return nil, "", nil, err
 	}
-	server := &Server{PacketConn: pc, ReadTimeout: time.Hour, WriteTimeout: time.Hour}
+
+	server := &Server{PacketConn: pc}
```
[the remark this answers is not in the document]
fturib: I was a simplification.

Now with removal of the shutdown timeout, we completely depend of the readTimeout.
We cannot keep at 1 hour else the UT will never ends.

Unless you have a good reason to keep 1 hour  ? (and in that case we need to send msg after shutdown to ensure to trigger the end of the readUDP or readTCP if any.

server_test.go
fturib: It is what I expect.
That is true on my local machine. Not so sure in the Travis CI. Was thinking due to timeout tuning ... 

If you have an advise on how to build better here. I would take it.

fturib: I guess we depend of the time to write a msg on the connection.
to avoid any flakyness due to latency, we would like to be sure all msg are sent before triggering the shutdown.
But to be useful we need to wait the returned msg.
It would need to split the Exchange function and get a signal when the msg is sent.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Server - Graceful s..." ]
>fturib commented on this pull request.
>
>
>
>> @@ -60,7 +62,8 @@ func RunLocalUDPServerWithFinChan(laddr string) (*Server, string, chan error, er
> 	if err != nil {
> 		return nil, "", nil, err
> 	}
>-	server := &Server{PacketConn: pc, ReadTimeout: time.Hour, WriteTimeout: time.Hour}
>+
>+	server := &Server{PacketConn: pc}
>
>I was a simplification.
>
>Now with removal of the shutdown timeout, we completely depend of the readTimeout.
>We cannot keep at 1 hour else the UT will never ends.
>
>Unless you have a good reason to keep 1 hour  ? (and in that case we need to send msg after shutdown to ensure to trigger the end of the readUDP or readTCP if any.

It's a public field, you can just override it in your test?

fturib:
It is too late, you need to override before calling ServeAndActivate().
But I think now it is resolved in latest commit : I added timeouts as params of that function.

server.go
```diff
@@ -473,24 +476,81 @@ func (srv *Server) ActivateAndServe() error {
...
+	}()
+
+	var err error
+	select {
+	case <-done:
+	case <-ctx.Done():
+		err = ctx.Err()
+	}
+
+	// no more msg to send. We can safely cloe the UDP connection
```
miekg: No more messages to send, we can safely close the UPD listener.

server.go
```diff
@@ -473,24 +476,81 @@ func (srv *Server) ActivateAndServe() error {
 	return &Error{err: "bad listeners"}
 }
 
-// Shutdown shuts down a server. After a call to Shutdown, ListenAndServe and
-// ActivateAndServe will return.
+func (srv *Server) getShutdownTimeout() time.Duration {
```
miekg: s/getShutdownTimeout/shutdownTimeout/

server.go
```diff
@@ -473,24 +476,81 @@ func (srv *Server) ActivateAndServe() error {
...
+
+	// add a little time to ensure we have time to wait the thread are released
+	timeout = timeout + time.Second
+	return timeout
+
+}
+
+// Shutdown shuts down gracefully a server. All in progress queries will be dealt with
+// After a call to Shutdown, ListenAndServe and ActivateAndServe will return when all accepted queries are handled
+// or shutdown timeout occurs
```
miekg: "Shutdown will gracefully shut down a server. All in progress queries will be processed after which the server will stop."

When will a non-nil error be returned?

server.go
```diff
@@ -544,20 +604,26 @@ func (srv *Server) serveUDP(l *net.UDPConn) error {
 	rtimeout := srv.getReadTimeout()
 	// deadline is not used here
 	for {
-		m, s, err := reader.ReadUDP(l, rtimeout)
```
miekg: What happens if you leave this line here?

server.go
```diff
@@ -544,20 +601,26 @@ func (srv *Server) serveUDP(l *net.UDPConn) error {
...
 		srv.lock.RLock()
+		// check server is still alive
 		if !srv.started {
 			srv.lock.RUnlock()
 			return nil
 		}
+		// and flag that we are processing another packet
+		srv.dnsWg.Add(1)
 		srv.lock.RUnlock()
+		m, s, err := reader.ReadUDP(l, rtimeout)
```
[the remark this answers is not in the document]
miekg: >  I’m not sure just moving ReadUDP down to here is safe.
me neither

server.go
```diff
@@ -544,20 +604,26 @@ func (srv *Server) serveUDP(l *net.UDPConn) error {
...
 		if err != nil {
 			if netErr, ok := err.(net.Error); ok && netErr.Temporary() {
+				srv.dnsWg.Done() // ack processing the UDP packet
```
miekg: please remove these superfluous comments. dnsWg.Done() is clear enough by itself

server_test.go
```diff
@@ -102,18 +107,18 @@ func RunLocalUDPServerUnsafe(laddr string) (*Server, string, error) {
...
 
 func RunLocalTCPServer(laddr string) (*Server, string, error) {
-	server, l, _, err := RunLocalTCPServerWithFinChan(laddr)
+	server, l, _, err := RunLocalTCPServerWithFinChan(laddr, time.Hour, time.Hour)
 
 	return server, l, err
 }
 
-func RunLocalTCPServerWithFinChan(laddr string) (*Server, string, chan error, error) {
+func RunLocalTCPServerWithFinChan(laddr string, readTimeout time.Duration, writeTimeout time.Duration) (*Server, string, chan error, error) {
```
miekg: This change is not needed you can set these values after the server is created, add shutdownTimeout will be calculated during shutdown.

server_test.go
miekg: this tests needs to be better; it provides little context to actually go and debug when it fails. Currently it is just brute-forcing and hoping for the test. I, however, don't know a good way to test this either.

you want to know that when being shutdown you handle all sent queries... seems the that server's handle func just needs some inside in the server state to determine that any new incoming queries are still within the shutdownTimeout or something?

server_test.go
fturib: yes for shutdownTimeout, but the readTimeout will be used at readUDP() or readTCP() and that happen as soon as ServeAndActivate() is launched.

server.go
fturib: ok

server.go
fturib: I need to have the dnsWg.Add before entering into the ReadUDP,
and I need it to be in the Lock (to ensure that srv.started is true when applying Add.

fturib: I need that the Add happen BEFORE entering the ReadUDP.
and the Add needs to be checked against the srv.started ... 

So I moved the line below the block with Lock.

is there a specific concern ?

At this point of the workflow we do not know if there is a packet coming in or not : the ReadUDP is just waiting for next packet.
So I think that is safe to check the exit before the ReadUDP().

But it is NO MORE safe to test it between the ReadUDP and spanworker ... because we WANT that any coming msg is processed whatever srv.started.

server.go
fturib: for now, never - I respected the behavior that was before.
Do we want to return an error if the stop was triggered by timeout ?

NOTE: that is true for ShutdownContext, but not for Shutdown => would be a change of behavior.

fturib: oops you are right. Before there was an error if the server was not started.
I will fix that to have same behavior as before.

server_test.go
fturib: there is maybe an idea here : I can hold the replies on the HandleFunc until the server is shutdown.
Doing so, all queries will be replied AFTER the shutdown ..

as long as the number of waiting replies is lower than the max workers.

Let me try that. It should be more predictable.

fturib:
hum... for the UT, still work to do !

server_test.go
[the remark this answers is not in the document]
tmthrgd: @fturib Ack, that seems fine to me.

fturib:
for the UT, I focus on the UDP because I reach to have it failed on my local machine.
When the duration of exchange is exactly above 1 sec, then the client fail in "read udp [::1]:54990->[::1]:54983: i/o timeout" 

like if one of Client or Server was not following expected Timeout.

Looks like, for now, on UDP only, it is expected that the reply comes within 1 sec.

fturib:
Oh I did not see msg on rebase. Will push again with a rebase.

fturib:
Integrate with change in server.go from @tmthrgd.
NOTE: I need the dnsWg.Add(..) to be in the Lock - so I extended the function instead of relock the session again.

Now, i think there is some confusion on what is responsible to close the listeners, the queue and when.

I will propose another commit on this part.

tmthrgd:
I have to say I’m finding this PR a bit too complex and finicky to properly follow. I’ve had an idea for a different approach that might work, so let me try that out and I’ll see how far I get.

fturib:
ok.  I put on standby.
Please ping me when your PR is ready.
Thanks.

miekg:
#717 is the new try

PR #714, opened 2018-07-22
Add testcases to validate consistency of packDataNsec

fntlnz:
Signed-off-by: Lorenzo Fontana <lo@linux.com>

### DIFF
--- a/msg_helpers_test.go
+++ b/msg_helpers_test.go
@@ -1,0 +1,108 @@
+package dns
+
+import "testing"
+
+// TestPacketDataNsec tests generated using fuzz.go and with a message pack
+// containing the following bytes: 0000\x00\x00000000\x00\x002000000\x0060000\x00\x130000000000000000000"
+// That bytes sequence created the overflow error and further permutations of that sequence were able to trigger
+// the other code paths.
+func TestPackDataNsec(t *testing.T) {
+	type args struct {
+		bitmap []uint16
+		msg    []byte
+		off    int
+	}
+	tests := []struct {
+		name       string
+		args       args
+		want       int
+		wantErr    bool
+		wantErrMsg string
+	}{
+		{
+			name: "overflow",
+			args: args{
+				bitmap: []uint16{
+					8962, 8963, 8970, 8971, 8978, 8979,
+					8986, 8987, 8994, 8995, 9002, 9003,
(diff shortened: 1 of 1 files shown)


msg_helpers_test.go
```diff
@@ -0,0 +1,104 @@
...
+					0, 19, 48, 48,
+				},
+				off: 48,
+			},
+			wantErr:    true,
+			wantErrMsg: "dns: overflow packing nsec",
+			want:       31,
+		},
+		{
+			name: "disordered nseec bits",
```
miekg: nsec

msg_helpers_test.go
```diff
@@ -0,0 +1,104 @@
+package dns
+
+import "testing"
+
+func TestPackDataNsec(t *testing.T) {
```
miekg: Can you add some comments on how you generated these test cases?

fntlnz: added such comment

msg_helpers_test.go
```diff
@@ -0,0 +1,108 @@
+package dns
+
+import "testing"
+
+// TestPacketDataNsec tests generated using fuzz.go and with a message pack
+// containing the following bytes: 0000\x00\x00000000\x00\x002000000\x0060000\x00\x130000000000000000000"
+// That bytes sequence created the overflow error and further permutations of that sequence were able to trigger
+// the other code paths.
```
miekg: uhm, shouldn't those fixes part of this PR?

fntlnz: Can you elaborate more ?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Add testcases to va..." ]
>fntlnz commented on this pull request.
>
>
>
>> @@ -0,0 +1,108 @@
>+package dns
>+
>+import "testing"
>+
>+// TestPacketDataNsec tests generated using fuzz.go and with a message pack
>+// containing the following bytes: 0000\x00\x00000000\x00\x002000000\x0060000\x00\x130000000000000000000"
>+// That bytes sequence created the overflow error and further permutations of that sequence were able to trigger
>+// the other code paths.
>
>Can you elaborate more ?

you say you saw a crash? If so we should apply a fix?

fntlnz:
I only got a defined error path while passing that input, not a panic or anything worse.

PR #716, opened 2018-07-28
Abstract shutdown checking into seperate function

tmthrgd:
This moves the `srv.started` checking with it's `srv.lock` dance out of `serveUDP` and `serveTCP`, and into a separate function. Once rebased, this might make the changes in #713 slightly cleaner. I believe this is worth doing independently of #713 though.

### DIFF
--- a/server.go
+++ b/server.go
@@ -315,6 +315,13 @@ type Server struct {
 	started bool
 }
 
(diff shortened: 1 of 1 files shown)


miekg:
agreed

PR #717, opened 2018-08-03
Make Shutdown wait for connections to terminate gracefully

tmthrgd:
This is my attempt at an alternate implementation of graceful shutdown as opposed to #713.

`Shutdown` (and now `ShutdownContext`) will wait for all pending handlers to terminate before returning. It also interrupts any blocked goroutines waiting on `net.Conn` reads.

`ListenAndServe` and `ActivateAndServe` will not return until all pending handlers have returned, even if `ShutdownContext` itself returns early.

~I've simply copied the test changes verbatim from #713, but I can't vouch for their accuracy.~

~I can't be certain this is fully functional as I have yet to write my own test case, although it passes all existing tests.~

From the original pull request:
> related PR : coredns/coredns#1963
> related issue : coredns/coredns#1666

/cc @fturib


server.go
```diff
@@ -483,21 +495,48 @@ func (srv *Server) ActivateAndServe() error {
...
 	srv.started = false
 	srv.lock.Unlock()
 
+	if !started {
+		return &Error{err: "server not started"}
+	}
+
 	if srv.PacketConn != nil {
-		srv.PacketConn.Close()
+		srv.PacketConn.SetReadDeadline(aLongTimeAgo) // Unblock reads
```
fturib: that is good idea!

fturib:
NOTE: my UT for shudtdown is doing:
- run a set of parallel client DNS that will send a query before shutdown
- these queries are blocked on Handling side until shutdown is starting
- the reply are sent with a random delay (lower than shutdown timeout)
- we just count that each query sent is coming back w/o communication error.

=> so we verify that all replies sent after shutdown are correctly handled by the servers.

tmthrgd:
I can't reproduce the Travis failure locally. It's only present for TLS on go1.9.x and not tip. (go1.10.x is currently untested, see #718). `EOF` isn't a particularly helpful error to surface from `Exchange`.

Running go1.9.7 locally passes fine (even with different values of `GOMAXPROCS`):
```
go1.9.7 test
PASS
ok      github.com/tmthrgd/dns  7.697s
```

tmthrgd:
I can't reproduce this locally at all. Interestingly after rebasing on #718 (Travis go1.10.x tests), Travis fails for both go1.9.x and go1.10.x, but not for tip. I'll try and see what's changed in `crypto/tls` or `net` since go1.10.3.

---

go1.9.7:
```
$ go1.9.7 test
PASS
ok      github.com/tmthrgd/dns  7.885s
```

go1.10.3:
```
$ go1.10.3 test
PASS
ok      github.com/tmthrgd/dns  7.510s
```

go tip:
```
$ go-tip version
go version devel +5a720d229d Sat Aug 4 16:37:47 2018 +0000 linux/amd64
$ go-tip test
PASS
ok      github.com/tmthrgd/dns  7.408s
```

server.go
```diff
@@ -561,11 +623,30 @@ func (srv *Server) serveUDP(l *net.UDPConn) error {
...
+		})
 	}
+
+	return nil
 }
 
 func (srv *Server) serve(w *response) {
+	defer func() {
+		if w.tcp != nil {
+			srv.lock.Lock()
```
miekg: We now have created a server wide lock that will, sadly, see a lot of contention. Also `defer` is still considered slow

miekg: I'm wonder why the waitGroup is not enough, why another map to keep track of things?

And as alluded, some benchmarking would help - I'm afraid this slows things down a bit.

tmthrgd: @miekg Unfortunately the map is necessary to be able to interrupt the `ReadTCP` call below. Without it, everything will just block until the call times out.

tmthrgd: @miekg While there could be lock contention, I actually don't think it will be significant or an issue. This only occurs in the TCP code path and has no impact on UDP, it will also only be hit once per TCP connection, not per-request.

I could replace the map with a lock-less datastructure, like a linked list perhaps, but that would add complexity; and I don't think `sync.Map` would be appropriate here.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Make Shutdown wait ..." ]
>
>
>
>>  }
>
> func (srv *Server) serve(w *response) {
>+	defer func() {
>+		if w.tcp != nil {
>+			srv.lock.Lock()
>
>@miekg While there could be lock contention, I actually don't think it will be significant or an issue. This only occurs in the TCP code path and has no impact on UDP, it will also only be hit once per TCP connection, not per-request.

ah, that's good to know, and may be worth a comment?

>I could replace the map with a lock-less datastructure, like a linked list perhaps, but that would add complexity; and I don't think `sync.Map` would be appropriate here.

Nah, sync.Map is not something we should be using, I think.

Let me take another look, with this new knowledge.

/Miek

--
Miek Gieben

client_test.go
```diff
@@ -117,7 +117,7 @@ func TestClientTLSSyncV4(t *testing.T) {
 		Certificates: []tls.Certificate{cert},
 	}
 
-	s, addrstr, err := RunLocalTLSServer(":0", &config)
+	s, addrstr, err := RunLocalTLSServer(":0", &config, time.Second, time.Hour)
```
miekg: It should be possible to set these timers after the creation of the server? That would cut down on this PR's size as we don't needlessly need to update a bunch of tests.

server.go
```diff
@@ -22,6 +23,10 @@ const idleWorkerTimeout = 10 * time.Second
 // Maximum number of workers
 const maxWorkersCount = 10000
 
+// aLongTimeAgo is a non-zero time, far in the past, used for
+// immediate cancelation of network operations.
+var aLongTimeAgo = time.Unix(1, 0)
```
miekg: If your system time is set to 1970 this won't work. I know this sounds a bit far-fetched, but maybe we should set it at least earlier than epoch? like 1965 or someting?

server.go
miekg: maybe drop the defer

server_test.go
```diff
@@ -575,6 +579,96 @@ func TestShutdownTCP(t *testing.T) {
 	}
 }
 
+func checkInProgressQueriesAtShutdownServer(t *testing.T, srv *Server, addr string, client *Client) {
```
miekg: We need to figure this one out. I'm fine with merging this *without* this test and then having a thought. This will either be flaky or hard to debug when it fails.

server.go
tmthrgd: This is the value used by the standard library: https://github.com/golang/go/blob/43f54c8d2e3bddfc6ad7887286eb6564986cb6ad/src/net/http/http.go#L21-L23.

server.go
tmthrgd: I merged the defer into the existing defer for the TCP path and removed it for the UDP path.

server_test.go
tmthrgd: @miekg I can do that, but it does seem to be a little risky. It's very hard to debug at the moment, I really can't figure out why `TestInProgressQueriesAtShutdownTLS` is failing for go1.10 but not go1.11 and only on Travis. Without a solid test, I don't feel completely confident in this PR.

tmthrgd:
The race detector is being stupid here, `testShutdownNotify` is not being written concurrently as the tests are running in serial. I'm not sure what the best way to fix that is.

server_test.go
```diff
@@ -575,6 +578,107 @@ func TestShutdownTCP(t *testing.T) {
...
+	}
+}
+
+func checkInProgressQueriesAtShutdownServer(t *testing.T, srv *Server, addr string, client *Client) {
+	const requests = 100
+
+	HandleFunc("example.com.", func(w ResponseWriter, req *Msg) {
+		// wait the signal to reply
+		testShutdownNotify.L.Lock()
+		testShutdownNotify.Wait()
```
tmthrgd: (Note to self: `for srv.isStarted()`).

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Make Shutdown wait ..." ]
>
>
>
>> @@ -575,6 +579,96 @@ func TestShutdownTCP(t *testing.T) {
> 	}
> }
>
>+func checkInProgressQueriesAtShutdownServer(t *testing.T, srv *Server, addr string, client *Client) {
>
>@miekg I can do that, but it does seem to be a little risky. It's very hard to debug at the moment, I really can't figure out why `TestInProgressQueriesAtShutdownTLS` is failing for go1.10 but not go1.11 and only on Travis. Without a solid test, I don't feel completely confident in this PR.

Ok, lets leave it in, until we discover a better way to test.

server.go
```diff
@@ -489,23 +501,56 @@ func (srv *Server) ActivateAndServe() error {
 // Shutdown shuts down a server. After a call to Shutdown, ListenAndServe and
 // ActivateAndServe will return.
 func (srv *Server) Shutdown() error {
+	return srv.ShutdownContext(context.Background())
```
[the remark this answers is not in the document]
tmthrgd: Ack, will leave as is. It's a pretty simple change to go back to the old behaviour if people complain.

tmthrgd:
*(Wowee, that's a lot of commits to get here).*

I'm fairly confident in the state of this pull request *and* the test cases. I believe they should be testing the desired behaviour and shouldn't prove flaky at all (fingers crossed).

@miekg If you're happy with this, it should be good to merge.

miekg:
\o/ Thanks for working on this, appreciated!

@fturib care to take look as well?

server_test.go
```diff
@@ -599,61 +595,68 @@ func checkInProgressQueriesAtShutdownServer(t *testing.T, srv *Server, addr stri
...
 	}
 
-	// wait that all go routines are stopped and report their numbers
-	wg.Wait()
-
-	// now check we received ALL the msg sent
-	if sendMsg != recvMsg {
-		t.Errorf("sent %v msgs to the server, but only %v msgs were returned", sendMsg, recvMsg)
+	if eg.Wait() != nil {
+		t.Fatalf("conn.ReadMsg error: %v", eg.Wait())
```
fturib: Good to know this google.ord.../sync for errGroup !
I agree that it is a better organization for the test, and ensure all steps are synced as expected.

server_test.go
```diff
@@ -595,10 +602,7 @@ func checkInProgressQueriesAtShutdownServer(t *testing.T, srv *Server, addr stri
 			t.Errorf("ResponseWriter.WriteMsg error: %s", err)
 		}
 	})
-	defer func() {
-		testShutdownNotify = nil
-		HandleRemove("example.com.")
-	}()
+	defer HandleRemove("example.com.")
```
fturib: why remove the "nil-ing" of testShutdownNotify ? 
this nil status is still the one that is tested in Server.shutdownContext ...

server_test.go
```diff
@@ -799,19 +799,6 @@ func TestShutdownUDP(t *testing.T) {
 	}
 }
 
-func TestShutdownUDPWithContext(t *testing.T) {
```
fturib: Why removing this test ?

server_test.go
tmthrgd: @fturib The nil-test in `ShutdownContext` is there because `testShutdownNotify` isn't populated except when testing, it's not relevant to the actual test. When I used a per-test `chan` before, I was tripped the race detector. Now I'm using a `sync.Cond` (set only in `init`) which doesn't need to be replaced, and can't be without tripped the race detector. It's also side-effect free, so there's no harm with `Broadcast` being called in `ShutdownContext`.

server_test.go
tmthrgd: @fturib A few reasons, it was UDP only and didn't have a TCP and TLS mirror; and that I don't think it really added much over `TestInProgressQueriesAtShutdown*` which really exercised the shutdown code. What was your intent behind it?

server_test.go
fturib: ok.

server_test.go
fturib: This code is testing the shutdown when there is no dns activity, but in a case where the deadline of context was > than readTimeout.
I guess that is also tested in TestShutdownUDP.

server_test.go
```diff
@@ -10,6 +11,8 @@ import (
 	"sync"
 	"testing"
 	"time"
+
+	"golang.org/x/sync/errgroup"
```
miekg: hmmm this adds a dep, just for a test. Do we really need it?

tmthrgd: Without doing these network operations in parallel the test runs for several minutes. That being said, I can probably throw together a simple replacement that should work without added dependencies. I’ve sketched something that’s 20 lines that should work, I’ll try it out when I get to my computer.

tmthrgd: It would basically end up as a reimplementation of x/sync/errgroup. I’d rather just use that.

miekg:
Ok. The I'll stop the nits.

Lgtm. Feel free to merge.

bjhaid:
Hi, when will should we expect a coredns release containing this PR? Thanks!

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Make Shutdown wait ..." ]
>Hi, when will should we expect a coredns release containing this PR? Thanks!

next release, most likely.

bjhaid:
Thanks for the quick response! Do you have an idea on the date/timing of the next release?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Make Shutdown wait ..." ]
>Thanks for the quick response! Do you have an idea on the date/timing of the next release?

We're due for a new release, last one was 3 weeks

bjhaid:
Nice! I will keep my eyes open

bjhaid:
ping

PR #719, opened 2018-08-05
Simplify unlocking dance in *Serve functions

tmthrgd:
This simplifies the two calls to `srv.lock.Unlock` in both `ListenAndServe` and `ActivateAndServe`, and removes the need to relock the mutex before the return.

It also fixes the case where one of either `serveUDP` or `serveTCP` panics causing `Unlock` to be called twice and that panic clobbering the original panic message. *(Note this has not been actually observed to happen).*

### DIFF
--- a/server.go
+++ b/server.go
@@ -367,10 +367,17 @@ func (srv *Server) spawnWorker(w *response) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


miekg:
busy at work, trusting you, @tmthrgd, here a bit.

PR #720, opened 2018-08-10
Add tsig support on Transfer.Out() and add some xfr testcase

WeipengSun:
I'm doing a project of dns master, I found that the DNS library was not supported adding tsig when using Transfer.Out to reply Axfr or Ixfr request.
Transfer.Out calls ResponseWriter.WriteMsg, only when Msg sets Tsig RR, the message sent will carry Tsig.


WeipengSun:
@miekg Tsig support of Transfer.Out

Issue #725, opened 2018-08-16
maxWorkersCount


larytet:
Hi,

The constant maxWorkersCount is 10000.   The constant is not exported. My understanding of the code is that the application is not expected to serve more than maxWorkersCount DNS requests simultaneously. In other words the application can not have more than maxWorkersCount pending requests. 
If,  for example, it takes 2s on average for the application to figure out a correct response the application will serve 5K requests/s.

Is my understanding correct?

tmthrgd:
See my comment on https://github.com/miekg/dns/issues/722#issuecomment-412932068:
> Also to clarify, `maxWorkersCount` is poorly named. It doesn’t actually control the maximum number of workers that can run, but the maximum that are allowed to stay idle before being torn down. If that limit is reached, the server will just keep spawning workers (goroutines) for each new request, they’ll just only be used once.

This line is where new goroutines are spawned
https://github.com/miekg/dns/blob/208cd1e89ec4a671e7ae6891f8c94cfe75399570/server.go#L366

and `maxWorkersCount` is only checked here
https://github.com/miekg/dns/blob/208cd1e89ec4a671e7ae6891f8c94cfe75399570/server.go#L330

after `srv.serve` has been called, but before the worker waits idling.

larytet:
My bad. I I did not pay enough attention that there is a call to srv.serve(w) before the cleanup.

tmthrgd:
@larytet Not a problem, I hope I've cleared that all up for you.

Issue #722, opened 2018-08-10
Configure the maximum number of workers


cesarkuroiwa:
Is there a way to limit the maximum number of workers a server can spawn.
I see on the code that the limit is maxWorkersCount = 10000, but how can I change it, other than editing the code?

Thanks

miekg:
[ Quoting <notifications@github.com> in "[miekg/dns] Configure the maximum n..." ]
>Is there a way to limit the maximum number of workers a server can spawn.
>I see on the code that the limit is maxWorkersCount = 10000, but how can I change it, other than editing the code?

you can't. And I'm really opposed to exporting this as a thing you can tweak.

tmthrgd:
@cesarkuroiwa What do you want to limit the maximum number of workers for?

cesarkuroiwa:
The thing is I have a DNS server that basically just have to answer a bunch of zone transfer once in a while. Every 5 minutes or so, it gets about 10k requests for (A|I)XFR almost at once, and we're seeing its memory usage spike at those moments.

I am inclined to believe that, since those requests come all at once, the server spawns a lot of goroutines to handle each request, which may cause the spike in memory usage. At some moments, I've the server with about 1000 goroutines, but the actual number might be higher than that.

Some zone transfers are considerably large (about 1gb), I've seen the server get close to 26gb of memory (it is running on a machine with 32gb of RAM)

So, that's why I was thinking that, maybe, putting a limit to the number of spawned servers might limit the memory usage, while also decreasing the performance, which wouldn't be very critical in our case.

tmthrgd:
@cesarkuroiwa Would you mind trying the following (or something similar):

```go
...
var h dns.Handler = ... // (existing handler)
limit := make(chan struct{}, 500)
limited := dns.HandlerFunc(func(w dns.ResponseWriter, r *dns.Msg) {
    limit <- struct{}{}
    defer func() { <-limit }()
    h.ServeDNS(w, r)
})
srv := &dns.Server{Handler: limit}
...
```

That should limit the number of concurrent dns requests being handled to 500. (I typed this out on mobile so apologise if there are typos).

tmthrgd:
Also to clarify, `maxWorkersCount` is poorly named. It doesn’t actually control the maximum number of workers that can run, but the maximum that are allowed to stay idle before being torn down. If that limit is reached, the server will just keep spawning workers (goiroutines) for each new request, they’ll just only be used once.

larytet:
@tmthrgd 
[...]

PR #726, opened 2018-08-16
Improve documentation of maxWorkersCount const

tmthrgd:
The meaning of this const has come up in both #722 and #725, so it's clearly non-obvious. Hopefully this wording will clarify things.

/cc @UladzimirTrehubenka


miekg:
Lgtm

PR #728, opened 2018-08-17
[tls] Carry TLS state within (possibly) response writer

### LINKED ISSUE #711: Accessing tls.ConnectionState from ResponseWriter
The `ResponseWriter` currently exposes some information about the connection, such as the `LocalAddr()` of the server and the `RemoteAddr()` of the client.

Given that the DNS server supports TLS connections, it may be helpful to access the `tls.ConnectionState` of the TLS connection. Handlers can utilize this information for logging purposes, or to craft a different response if it is known that the request was made over a secure connection.

I suggest adding the following method to `ResponseWriter`

```
// ConnectionState returns the the tls.ConnectionState information associated with
// the request, if it was made over a TLS connection. Otherwise, returns nil
TLSConnectionState() *tls.ConnectionState
```

Would be happy to send a PR if there are no objections to extending the ResponseWriter interface. Any comments/objections?

/cc @chantra

chantra:
This allows a server to make decision wether or not the link used to
connect to the DNS server is using TLS.
This can be used by the handler for instance to (but not limited to):
- log that the request was TLS vs TCP
- craft specific responsed knowing that the link is secured
- return custom answers based on client cert (if provided)
...

Fixes #711

### DIFF
--- a/server.go
+++ b/server.go
@@ -56,6 +56,12 @@ type ResponseWriter interface {
 	Hijack()
 }
 
+// A TLSResponse interface is used by a DNS Handler to access TLS connection state
+// when available.
+type TLSResponse interface {
+	TLS() *tls.ConnectionState
+}
+
 type response struct {
 	msg            []byte
 	hijacked       bool // connection has been hijacked by handler
@@ -784,6 +790,18 @@ func (w *response) Close() error {
 		e := w.tcp.Close()
 		w.tcp = nil
 		return e
+	}
+	return nil
+}
+
+// TLS implements the TLSResponse.TLS method.
+func (w *response) TLS() *tls.ConnectionState {
+	if w.tcp == nil {
+		return nil
+	}
+	if v, ok := w.tcp.(*tls.Conn); ok {
+		t := v.ConnectionState()
+		return &t
 	}
 	return nil
 }
--- a/server.go
+++ b/server.go
@@ -56,10 +56,10 @@ type ResponseWriter interface {
 	Hijack()
 }
 
-// A ConnectionState interface is used by a DNS Handler to access TLS connection state
+// A ConnectionStater interface is used by a DNS Handler to access TLS connection state
 // when available.
-type ConnectionState interface {
-	ConnectionState() tls.ConnectionState
+type ConnectionStater interface {
+	ConnectionState() *tls.ConnectionState
 }
 
 type response struct {
@@ -794,10 +794,11 @@ func (w *response) Close() error {
 	return nil
 }
 
-// ConnectionState() implements the ConnectionState.ConnectionState() interface.
-func (w *response) ConnectionState() (t tls.ConnectionState) {
+// ConnectionState() implements the ConnectionStater.ConnectionState() interface.
+func (w *response) ConnectionState() *tls.ConnectionState {
 	if v, ok := w.tcp.(*tls.Conn); ok {
-		return v.ConnectionState()
+		t := v.ConnectionState()
+		return &t
 	}
-	return
+	return nil
 }
--- a/server.go
+++ b/server.go
@@ -59,7 +59,7 @@ type ResponseWriter interface {
 // A ConnectionState interface is used by a DNS Handler to access TLS connection state
 // when available.
 type ConnectionState interface {
-	ConnectionState() *tls.ConnectionState
+	ConnectionState() tls.ConnectionState
 }
 
 type response struct {
@@ -795,10 +795,9 @@ func (w *response) Close() error {
 }
 
 // ConnectionState() implements the ConnectionState.ConnectionState() interface.
-func (w *response) ConnectionState() *tls.ConnectionState {
+func (w *response) ConnectionState() (t tls.ConnectionState) {
 	if v, ok := w.tcp.(*tls.Conn); ok {
-		t := v.ConnectionState()
-		return &t
+		return v.ConnectionState()
 	}
-	return nil
+	return
 }
--- a/server.go
+++ b/server.go
@@ -796,9 +796,6 @@ func (w *response) Close() error {
 
 // TLS implements the TLSResponse.TLS method.
 func (w *response) TLS() *tls.ConnectionState {
-	if w.tcp == nil {
-		return nil
-	}
 	if v, ok := w.tcp.(*tls.Conn); ok {
 		t := v.ConnectionState()
 		return &t
--- a/server.go
+++ b/server.go
@@ -56,10 +56,10 @@ type ResponseWriter interface {
 	Hijack()
 }
 
-// A TLSResponse interface is used by a DNS Handler to access TLS connection state
+// A ConnectionState interface is used by a DNS Handler to access TLS connection state
 // when available.
-type TLSResponse interface {
-	TLS() *tls.ConnectionState
+type ConnectionState interface {
+	ConnectionState() *tls.ConnectionState
 }
 
 type response struct {
@@ -794,8 +794,8 @@ func (w *response) Close() error {
 	return nil
 }
 
-// TLS implements the TLSResponse.TLS method.
-func (w *response) TLS() *tls.ConnectionState {
+// ConnectionState() implements the ConnectionState.ConnectionState() interface.
+func (w *response) ConnectionState() *tls.ConnectionState {
 	if v, ok := w.tcp.(*tls.Conn); ok {
 		t := v.ConnectionState()
 		return &t
--- a/server.go
+++ b/server.go
@@ -796,7 +796,10 @@ func (w *response) Close() error {
 
 // ConnectionState() implements the ConnectionStater.ConnectionState() interface.
 func (w *response) ConnectionState() *tls.ConnectionState {
-	if v, ok := w.tcp.(*tls.Conn); ok {
+	type tlsConnectionStater interface {
+		ConnectionState() tls.ConnectionState
+	}
+	if v, ok := w.tcp.(tlsConnectionStater); ok {
 		t := v.ConnectionState()
 		return &t
 	}
--- a/server_test.go
+++ b/server_test.go
@@ -11,11 +11,6 @@ import (
 	"time"
 )
 
-const (
-	TLSOffMsg = "TLS: off"
-	TLSOnMsg  = "TLS: on"
-)
-
 func HelloServer(w ResponseWriter, req *Msg) {
 	m := new(Msg)
 	m.SetReply(req)
@@ -51,18 +46,6 @@ func AnotherHelloServer(w ResponseWriter, req *Msg) {
 
 	m.Extra = make([]RR, 1)
 	m.Extra[0] = &TXT{Hdr: RR_Header{Name: m.Question[0].Name, Rrtype: TypeTXT, Class: ClassINET, Ttl: 0}, Txt: []string{"Hello example"}}
-	w.WriteMsg(m)
-}
-
-func TLSStateServer(w ResponseWriter, req *Msg) {
-	m := new(Msg)
-	m.SetReply(req)
-	txt := TLSOffMsg
-	if tlsState := w.(TLSResponse).TLS(); tlsState != nil {
-		txt = TLSOnMsg
-	}
-	m.Extra = make([]RR, 1)
-	m.Extra[0] = &TXT{Hdr: RR_Header{Name: m.Question[0].Name, Rrtype: TypeTXT, Class: ClassINET, Ttl: 0}, Txt: []string{txt}}
 	w.WriteMsg(m)
 }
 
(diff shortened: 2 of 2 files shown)


chantra:
@miekg here is an attempt at solving #711 . Feedbacks welcome!

server_test.go
```diff
@@ -49,6 +54,18 @@ func AnotherHelloServer(w ResponseWriter, req *Msg) {
 	w.WriteMsg(m)
 }
 
+func TLSStateServer(w ResponseWriter, req *Msg) {
+	m := new(Msg)
+	m.SetReply(req)
+	txt := TLSOffMsg
+	m.Extra = make([]RR, 1)
```
tmthrgd: Move this line (`m.Extra`) after the if statement.

server.go
```diff
@@ -787,3 +793,15 @@ func (w *response) Close() error {
 	}
 	return nil
 }
+
+// TLS implements the TLSResponse.TLS method.
+func (w *response) TLS() *tls.ConnectionState {
+	if w.tcp == nil {
```
tmthrgd: This check isn’t needed: https://play.golang.org/p/Thjve6Gl3xl

server.go
```diff
@@ -56,6 +56,12 @@ type ResponseWriter interface {
 	Hijack()
 }
 
+// A TLSResponse interface is used by a DNS Handler to access TLS connection state
```
miekg: I would just rename these a bit:
* TLSResponse -> ConnectionStater
* TLS() -> ConnectState()

server.go
```diff
@@ -787,3 +793,12 @@ func (w *response) Close() error {
 	}
 	return nil
 }
+
+// TLS implements the TLSResponse.TLS method.
```
miekg: implements the ... *interface*

server.go
```diff
@@ -787,3 +793,12 @@ func (w *response) Close() error {
...
+func (w *response) TLS() *tls.ConnectionState {
+	if v, ok := w.tcp.(*tls.Conn); ok {
+		t := v.ConnectionState()
+		return &t
```
miekg: why does we need to take the address here?

server_test.go
```diff
@@ -197,13 +216,26 @@ func TestServing(t *testing.T) {
...
+	txt = r.Extra[0].(*TXT).Txt[0]
+	if txt != TLSOffMsg {
+		t.Error("unexpected result for tlsstate.examepl.net.", txt, "!= "+TLSOffMsg)
+	}
 }
 
 func TestServingTLS(t *testing.T) {
 	HandleFunc("miek.nl.", HelloServer)
 	HandleFunc("example.com.", AnotherHelloServer)
+	HandleFunc("tlsstate.example.net.", TLSStateServer)
```
miekg: Adding your test to this testcase is a bit unfortunate and inflates the PR also. If possible can you just make a small TLS test server that gets 2 reqs. UDP and TCP+TLS and see if you can access in the connection state *in* the handler ,so you don't need to do this TXT record dance?

server.go
chantra: I do not have strong opinions. This was just mimicking the HTTP.Request/HTTP.Response strucs which have a TLS *tls.Connection state attribute, and have this set to nil when not a TLS connection.
I will need to change the signature to  (tls.ConnectionState, err) if we do not want to use address.

server_test.go
chantra: ok, will fix this.

server.go
miekg: But ConnectionState() doesn't (for some reason), would be good to just mimic that.

chantra: I suppose this is because `tls.ConnectionState` is done on a TLS connection, while in our case, we wont know until we can do a type assertion.
I have pushed a diff on top of the current stack. Now, the only way to find it is not a Connectionstate from a TLs connection would be to check that connectionstate.Version == 0

chantra:
@miekg id the last commit addressing your last concern? Would you prefer an interface like:
```go
func (w *response) ConnectionState() (tls.ConnectionState, error) {
}
```
instead?

tmthrgd:
@chantra Having it return an error would be a mistake, there isn't a valid error condition except non-TLS which is better signalled with a `nil` pointer.

The right approach is either:

- `ConnectionState() *tls.ConnectionState` being implemented on all `response`s (read: even those without TLS) with `cs, ok := rw.(dns.ConnectionState); ok && cs != nil` being the check for TLS; or,
- Making `ConnectionState` an optional-interface that's only implemented for TLS based `response`s, with `_, ok := rw.(dns.ConnectionState); ok` being the check for TLS.

server.go
```diff
@@ -56,6 +56,12 @@ type ResponseWriter interface {
 	Hijack()
 }
 
+// A ConnectionState interface is used by a DNS Handler to access TLS connection state
+// when available.
+type ConnectionState interface {
```
tmthrgd: I think idiomatic Go would be for this to be called `ConnectionStater` (see for example [http.Hijacker](https://golang.org/pkg/net/http/#Hijacker) and [http.Pusher](https://golang.org/pkg/net/http/#Pusher)).

server_test.go
```diff
@@ -259,6 +259,99 @@ func TestServingTLS(t *testing.T) {
 	}
 }
 
+// TestServingTLSConnectionState tests that we only can access
+// tls.ConnectionState under a DNS query handled by a TLS DNS server.
+// This test will sequentially create a TLS, UDP and TCP server, attach a custom
+// handler which will set a testing error if tls.ConnectionState is available
+// when it is not expected, or the other way around.
+func TestServingTLSConnectionState(t *testing.T) {
+
```
tmthrgd: nit: newline isn't needed.

server_test.go
```diff
@@ -259,6 +259,99 @@ func TestServingTLS(t *testing.T) {
...
+	handlerResponse := "Hello example"
```
tmthrgd: nit: this doesn't need to be a variable, just inline it.

server_test.go
```diff
@@ -259,6 +259,99 @@ func TestServingTLS(t *testing.T) {
...
+	// tlsHandlerTLS is a HandlerFunc that can be set to expect or not TLS
+	// connection state.
+	tlsHandlerTLS := func(tlsExpected bool) func(ResponseWriter, *Msg) {
+		return func(w ResponseWriter, req *Msg) {
+			m := new(Msg)
+			m.SetReply(req)
+			tlsFound := true
+			if connState := w.(ConnectionState).ConnectionState(); connState.Version == 0 {
```
tmthrgd: I think `ConnectionState` has to be returned to a pointer, it doesn't seem at all idiomatic or obvious to have to compare `Version == 0`. (cc @miekg).

server_test.go
```diff
@@ -259,6 +259,99 @@ func TestServingTLS(t *testing.T) {
...
+	}
+
+	s, addrstr, err := RunLocalTLSServer(":0", &config)
+	if err != nil {
+		t.Fatalf("unable to run test server: %v", err)
+	}
+	defer s.Shutdown()
+
+	// TLS DNS query
+	c := new(Client)
```
tmthrgd: This should be replaced with:
```go
c := &Client{
	Net:    "tcp-tls",
	TLSConfig: &tls.Config{
		InsecureSkipVerify: true,
	},
}
```

server_test.go
```diff
@@ -259,6 +259,99 @@ func TestServingTLS(t *testing.T) {
...
+
+	// TCP DNS Server
+	s, addrstr, err = RunLocalTCPServer(":0")
+	if err != nil {
+		t.Fatalf("unable to run test server: %v", err)
+	}
+	defer s.Shutdown()
+
+	// TCP DNS query
+	c = new(Client)
```
tmthrgd: ditto.

```go
c := &Client{Net: "tcp"}
```

server_test.go
miekg: I don't know why the std lib choose not to do so, but I like to mimic that as close as possible, i.e. return a non-pointer.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] [tls] Carry TLS sta..." ]
>
>
>
>> @@ -787,3 +793,12 @@ func (w *response) Close() error {
> 	}
> 	return nil
> }
>+
>+// TLS implements the TLSResponse.TLS method.
>+func (w *response) TLS() *tls.ConnectionState {
>+	if v, ok := w.tcp.(*tls.Conn); ok {
>+		t := v.ConnectionState()
>+		return &t
>
>I suppose this is because `tls.ConnectionState` is done on a TLS connection, while in our case, we wont know until we can do a type assertion.
>I have pushed a diff on top of the current stack. Now, the only way to find it is not a Connectionstate from a TLs connection would be to check that connectionstate.Version == 0

Hmm, yeah I can see this getting tricky, but comparing to an empty connection 
state would work? `tls.ConnectionState{}` ? I see the ease of having a pointer 
here (and also the disconnect from the stdlib, where TLS is always true for 
connections). But just an empty struct would work, no?

server.go
chantra: oh, I thought the original **r** from @miekg 's comment was a typo :s.

chantra:
> But just an empty struct would work, no?

this is what I am doing in the latest commit right?

```go
func (w *response) ConnectionState() (t tls.ConnectionState) {
	if v, ok := w.tcp.(*tls.Conn); ok {
		return v.ConnectionState()
	}
	return
}
```

But then, there is no good way for the caller to know it is empty (other than looking at the version being equal to 0.... which is not a valid TLS version, which is abusing the fact of having an intimate knowledge of the struct itself).

tmthrgd:
I think a better comparison is with the [http.Request.TLS](https://golang.org/pkg/net/http/#Request.TLS) field, that uses a pointer: `TLS *tls.ConnectionState`.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] [tls] Carry TLS sta..." ]
>I think a better comparison is with the [http.Request.TLS](https://golang.org/pkg/net/http/#Request.TLS) field, that uses a pointer: `TLS *tls.ConnectionState`.

Ah, yes, that makes totally sense. Then a pointer is fine.

chantra:
On Sun, Sep 9, 2018 at 12:42 PM Miek Gieben <notifications@github.com>
wrote:

> [ Quoting <notifications@github.com> in "Re: [miekg/dns] [tls] Carry TLS
> sta..." ]
> >I think a better comparison is with the [http.Request.TLS](
> https://golang.org/pkg/net/http/#Request.TLS) field, that uses a pointer:
> `TLS *tls.ConnectionState`.
>
> Ah, yes, that makes totally sense. Then a pointer is fine.
>
> Ok, i Will move back to pointers, address the bits and rename to
ConnectionState**r** whenever I get to a computer.

chantra:
seems the test that failed is transient and unrelated to my change:

```
--- FAIL: TestTCPRtt (2.02s)
	issue_test.go:20: dial tcp 8.8.4.4:53: i/o timeout
```

tmthrgd:
@chantra I just restarted the Travis job. That failure comes up from time to time, but it shouldn't be related.

server.go
```diff
@@ -787,3 +793,12 @@ func (w *response) Close() error {
 	}
 	return nil
 }
+
+// ConnectionState() implements the ConnectionStater.ConnectionState() interface.
+func (w *response) ConnectionState() *tls.ConnectionState {
+	if v, ok := w.tcp.(*tls.Conn); ok {
```
tmthrgd: I think we should avoid using a concrete type here and, like [x/net/http2](https://github.com/golang/net/blob/922f4815f713f213882e8ef45e0d315b164d705c/http2/server.go#L386), use [an interface](https://github.com/golang/net/blob/922f4815f713f213882e8ef45e0d315b164d705c/http2/http2.go#L332). Perhaps something like:
```go
func (w *response) ConnectionState() *tls.ConnectionState {
	type tlsConnectionStater interface {
		ConnectionState() tls.ConnectionState
	}
	if v, ok := w.tcp.(tlsConnectionStater); ok {
		t := v.ConnectionState()
		return &t
	}
	return nil
} 
```

That way the `*tls.Conn` can be wrapped by a custom listener.

chantra: ok, I am fine with that. @miekg @tmthrgd is there any other feedback on this diff you want me to push in?

server.go
```diff
@@ -787,3 +793,15 @@ func (w *response) Close() error {
 	}
 	return nil
 }
+
+// ConnectionState() implements the ConnectionStater.ConnectionState() interface.
+func (w *response) ConnectionState() *tls.ConnectionState {
+	type tlsConnectionStater interface {
```
miekg: I'm not 100% sure what @tmthrgd meant here, but this looks weird to me? Are we now back in the discussion on why we're returning a pointer to "connectionState"?

chantra:
I think the point is to expect an interface that inplements
ConnectionState() so it could be a wrapper around tls.Conn which is used
instead if an actual tls.Conn.
We are still returning a pointer though.

PR #736, opened 2018-09-08
Add SO_REUSEPORT support

### LINKED ISSUE #654: SO_REUSEPORT support
Just dropping this here. I tested SO_REUSEPORT on go-dns using:

https://gist.github.com/chantra/dfab011267be22989b25134537963e5c

To test, I patched `reflect` with https://gist.github.com/chantra/abc8edf0c4cdce2e5aaf8c0d5473dd79

## Perf test
```
GOPATH=~/go go run reflect/reflect.go  -soreuseport 0 -cpu 8
```
~88k

```
GOPATH=~/go go run reflect/reflect.go  -soreuseport 0 -cpu 4
```
~92k

```
GOPATH=~/go go run reflect/reflect.go  -soreuseport 4 -cpu 8
```
~225k

```
GOPATH=~/go go run reflect/reflect.go  -soreuseport 8 -cpu 8
```
~221k

```
GOPATH=~/go go run reflect/reflect.go  -soreuseport 4 -cpu 4
```
~150k

```
GOPATH=~/go go run reflect/reflect.go  -soreuseport 8 -cpu 4
```
~150k

So there is ways to squeeze more out of the service but there is some fine tuning that needs to be done between GOMAXPROCS and the number of sockets.

## known issue

The empty address is not supported, e.g passing `:5353` as an IP/Port to bind to would fail and report error:
```
address family not supported by protocol
```

see https://github.com/libp2p/go-reuseport/issues/7

tmthrgd:
This pull request adds `SO_REUSEPORT` support to the server.

This fixes #654.

/cc @chantra @bortzmeyer


server_go111.go
```diff
@@ -0,0 +1,43 @@
+// +build go1.11,!windows
```
tmthrgd: @miekg The full list of build tags taken from the `x/sys/unix` package is: `aix darwin dragonfly freebsd linux netbsd openbsd solaris`. I've gone with `!windows` for now, because it matches `udp.go`. This neglects `plan9` though, so I'm not sure what you'd prefer here: the full list of just not-windows.

server_go_not111.go
```diff
@@ -0,0 +1,23 @@
...
+
+package dns
+
+import "net"
+
+const supportsReuseport = false
+
+func listenTCP(network, addr string, reuseport bool) (net.Listener, error) {
+	if reuseport {
+		// TODO(tmthrgd): return an error?
```
tmthrgd: @miekg What would your preferred behaviour here be? Ignore the reuseport bool or return an error?

server_go111.go
miekg: fine

server_go_not111.go
miekg: on the fence..., but... ignore?

tmthrgd:
@miekg Here are some benchmarks: https://gist.github.com/tmthrgd/66ec98db2596214676e5255edc7206bc. They were run with `go run reflect.go -soreuseport N` with `N = 0, 6 & 12` on a 12-thread i7-8700k CPU. They show an increase from 54000 q/s to 313000 q/s.

I'm against exposing these even if they are useful for CoreDNS. They're not exactly the most portable and I wouldn't be surprised if people both in and out of this package wanted more socket options. If CoreDNS wants the same, it should copy these files.

server.go
```diff
@@ -312,6 +314,9 @@ type Server struct {
 	DecorateWriter DecorateWriter
 	// Maximum number of TCP queries before we close the socket. Default is maxTCPQueries (unlimited if -1).
 	MaxTCPQueries int
+	// Whether to set the SO_REUSEPORT socket option, allowing multiple listeners to be bound to a single address.
+	// It is only supported on go1.11+ and when using ListenAndServe.
+	Reuseport bool
```
tmthrgd: @miekg Are you happy with naming this `Reuseport bool`?

server_go_not111.go
tmthrgd: Ack.

server.go
miekg: sgtm, maybe ReusePort (to nit about it)

otherwise this is all lgtm, feel free to merge.

PR #738, opened 2018-09-08
Use ReadMsgUDP and WriteMsgUDP on windows

tmthrgd:
This addresses the following three TODOs (@fastest963) all in `udp_windows.go`:
```go
// TODO(fastest963): Once go1.10 is released, use ReadMsgUDP.
```

```go
// TODO(fastest963): Once go1.10 is released, use WriteMsgUDP. 
```

```go
// TODO(fastest963): Once go1.10 is released and we can use *MsgUDP methods
// use the standard method in udp.go for these.
```

I haven't tested this on a Windows machine (and Travis doesn't have Windows support), so this should probably be manually tested before merging.


miekg:
All in favor, can test either, because who runs windows these days.

Issue #742, opened 2018-09-19
dns.ParseZone() parsing error leading to nil pointer dereference


tr3ee:
### What did I do?

PoC.go
```golang
package main

import (
	"fmt"
	"strings"

	"github.com/miekg/dns"
)

func main() {
	r := strings.NewReader(` Ta 0 0 0`)
	for x := range dns.ParseZone(r, "", "") {
		if x.Error != nil {
			fmt.Println(x.Error)
		} else if x == nil {
			// This should never be reached
			fmt.Println("x == NIL")
		} else {
			// Do something with x.RR
			fmt.Println(x.RR.String())
		}
	}
	return
}
```

### What happened?

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x65a762]

goroutine 1 [running]:
main.main()
	/path/to/gopath/src/fuzz/dns/worker/main.go:20 +0x122
exit status 2
```

tr3ee:
@miekg 
I will take a look this tomorrow or a few day : )

tr3ee:
@miekg 
A PR #745  has been sent, please review it

tmthrgd:
#745 didn't have the magic commit message that would have closed this issue.

PR #745, opened 2018-09-22
Fix issue #742

tr3ee:
Fix ParseZone() parsing error #742 
 - fix error comparison in setTA() in scan_rr.go
 - add testcase to make sure everything is ok

### DIFF
--- a/scan_test.go
+++ b/scan_test.go
@@ -79,3 +79,13 @@ func TestParseZoneInclude(t *testing.T) {
 		}
 	}
 }
+
+func TestParseTA(t *testing.T) {
+	rr, err := NewRR(` Ta 0 0 0`)
+	if err != nil {
+		t.Fatalf("expected no error, but got %s", err)
+	}
+	if rr == nil {
+		t.Fatal(`expected a normal RR, but got nil`)
+	}
+}
--- a/scan_rr.go
+++ b/scan_rr.go
@@ -1665,9 +1665,9 @@ func setTA(h RR_Header, c chan lex, o, f string) (RR, *ParseError, string) {
 		return nil, &ParseError{f, "bad TA DigestType", l}, ""
 	}
 	rr.DigestType = uint8(i)
-	s, e, c1 := endingToString(c, "bad TA Digest", f)
-	if e != nil {
-		return nil, e.(*ParseError), c1
+	s, err, c1 := endingToString(c, "bad TA Digest", f)
+	if err != nil {
+		return nil, err, c1
 	}
 	rr.Digest = s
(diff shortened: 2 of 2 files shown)


scan_test.go
```diff
@@ -79,3 +79,20 @@ func TestParseZoneInclude(t *testing.T) {
 		}
 	}
 }
+
+func TestParseZoneTA(t *testing.T) {
+	r := strings.NewReader(` Ta 0 0 0`)
```
miekg: This should also break with NewRR() which leads to a shorter testcase

tr3ee:
```Go
func TestParseZoneTA(t *testing.T) {
	rr, err := dns.NewRR(` Ta 0 0 0`)
	if err != nil {
		t.Fatalf("expected no error, but got %s", err)
	}
	if rr == nil {
		t.Fatalf(`expected a normal RR, but got nil`)
	}
}
```
what about this then?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Fix issue #742 (#74..." ]
>tr3ee commented on this pull request.
>
>
>
>> @@ -79,3 +79,20 @@ func TestParseZoneInclude(t *testing.T) {
> 		}
> 	}
> }
>+
>+func TestParseZoneTA(t *testing.T) {
>+	r := strings.NewReader(` Ta 0 0 0`)
>
>```Go
>func TestParseZoneTA(t *testing.T) {
>	rr, err := dns.NewRR(` Ta 0 0 0`)
>	if err != nil {
>		t.Fatalf("expected no error, but got %s", err)
>	}
>	if rr == nil {
>		t.Fatalf(`expected a normal RR, but got nil`)
>	}
>}
>```
>what about this then?

yep. Except the name can just be TestParseTA and the second Fatalf can just be 
Fatal

tr3ee:
> [ Quoting <notifications@github.com> in "Re: [miekg/dns] Fix issue #742 (#74..." ]
> tr3ee commented on this pull request. > @@ -79,3 +79,20 @@ func TestParseZoneInclude(t *testing.T) { } } } + +func TestParseZoneTA(t *testing.T) { +	r := strings.NewReader(` Ta 0 0 0`) ```Go func TestParseZoneTA(t *testing.T) { rr, err := dns.NewRR(` Ta 0 0 0`) if err != nil { t.Fatalf("expected no error, but got %s", err) } if rr == nil { t.Fatalf(`expected a normal RR, but got nil`) } } ``` what about this then?
> yep. Except the name can just be TestParseTA and the second Fatalf can just be Fatal

updated : )

Issue #748, opened 2018-09-25
"dns: overflow unpacking uint16" on (dns.ResponseWriter).TsigStatus()


thyth:
I'm having trouble verifying TSIGs on requests when running a DNS server built upon this library: when the TSIG is present, calling TsigStatus() on the dns.ResponseWriter produces the error `dns: overflow unpacking uint16` where I expect `nil`.

I am also seeing this behavior in a case based upon the sample code listed in the docs @ https://godoc.org/github.com/miekg/dns#hdr-TRANSACTION_SIGNATURE -- with the exact code I'm using at this gist: https://gist.github.com/thyth/28115553537dbff36093cce962185b04

When running that gist, this is my output (from a fresh Ubuntu 16.04 environment). Seeing the same issue on macOS (with same Go version) too:
```
ubuntu@chroniton:~/scratch$ sudo ./scratch
panic: dns: overflow unpacking uint16

goroutine 21 [running]:
main.handleRequest(0x761840, 0xc000154090, 0xc000166000)
	/home/ubuntu/scratch/scratch.go:42 +0x40f
github.com/miekg/dns.HandlerFunc.ServeDNS(0x7279c8, 0x761840, 0xc000154090, 0xc000166000)
	/home/ubuntu/go/src/github.com/miekg/dns/server.go:112 +0x44
github.com/miekg/dns.(*ServeMux).ServeDNS(0xc000088c30, 0x761840, 0xc000154090, 0xc000166000)
	/home/ubuntu/go/src/github.com/miekg/dns/server.go:238 +0x62
github.com/miekg/dns.(*Server).serveDNS(0xc000112000, 0xc000154090)
	/home/ubuntu/go/src/github.com/miekg/dns/server.go:773 +0x140
github.com/miekg/dns.(*Server).serve(0xc000112000, 0xc000154090)
	/home/ubuntu/go/src/github.com/miekg/dns/server.go:725 +0x210
github.com/miekg/dns.(*Server).worker(0xc000112000, 0xc000154090)
	/home/ubuntu/go/src/github.com/miekg/dns/server.go:356 +0x4d
created by github.com/miekg/dns.(*Server).spawnWorker
	/home/ubuntu/go/src/github.com/miekg/dns/server.go:396 +0x86

ubuntu@chroniton:~/scratch$ go version
go version go1.11 linux/amd64

ubuntu@chroniton:~/scratch$ uname -a
[...]

PR #749, opened 2018-09-25
Bugfix for miekg/dns#748

thyth:
w.msg was being prematurely cleared prior to use by TsigVerify in server.go

Analysis in miekg/dns#748

### DIFF
--- a/server.go
+++ b/server.go
@@ -741,6 +741,7 @@ func (srv *Server) serveDNS(w *response) {
 	if w.udp != nil && cap(w.msg) == srv.UDPSize {
 		srv.udpPool.Put(w.msg[:srv.UDPSize])
 	}
+	tmpMsg := w.msg
 	w.msg = nil
 	if err != nil { // Send a FormatError back
 		x := new(Msg)
@@ -756,7 +757,7 @@ func (srv *Server) serveDNS(w *response) {
 	if w.tsigSecret != nil {
 		if t := req.IsTsig(); t != nil {
 			if secret, ok := w.tsigSecret[t.Hdr.Name]; ok {
-				w.tsigStatus = TsigVerify(w.msg, secret, "", false)
+				w.tsigStatus = TsigVerify(tmpMsg, secret, "", false)
 			} else {
 				w.tsigStatus = ErrSecret
 			}
@@ -764,6 +765,7 @@ func (srv *Server) serveDNS(w *response) {
 			w.tsigRequestMAC = req.Extra[len(req.Extra)-1].(*TSIG).MAC
 		}
 	}
+	tmpMsg = nil
 
 	handler := srv.Handler
 	if handler == nil {
--- a/server.go
+++ b/server.go
@@ -735,21 +735,23 @@ func (srv *Server) serve(w *response) {
 	}
 }
 
-func (srv *Server) serveDNS(w *response) {
-	req := new(Msg)
-	err := req.Unpack(w.msg)
+func (srv *Server) disposeBuffer(w *response) {
 	if w.udp != nil && cap(w.msg) == srv.UDPSize {
 		srv.udpPool.Put(w.msg[:srv.UDPSize])
 	}
-	tmpMsg := w.msg
 	w.msg = nil
+}
+
+func (srv *Server) serveDNS(w *response) {
+	req := new(Msg)
+	err := req.Unpack(w.msg)
 	if err != nil { // Send a FormatError back
 		x := new(Msg)
 		x.SetRcodeFormatError(req)
 		w.WriteMsg(x)
-		return
 	}
-	if !srv.Unsafe && req.Response {
+	if err != nil || !srv.Unsafe && req.Response {
+		srv.disposeBuffer(w)
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


server.go
```diff
@@ -741,6 +741,7 @@ func (srv *Server) serveDNS(w *response) {
 	if w.udp != nil && cap(w.msg) == srv.UDPSize {
 		srv.udpPool.Put(w.msg[:srv.UDPSize])
 	}
+	tmpMsg := w.msg
```
tmthrgd: This is wrong, the `srv.udpPool.Put` above will cause data corruption. As a quick fix, move the entire
```go
if w.udp != nil && cap(w.msg) == srv.UDPSize {
	srv.udpPool.Put(w.msg[:srv.UDPSize])
}
w.msg = nil
```
block to immediately after the TSIG verification.

tmthrgd: I’d like something that doesn’t drop buffers at the returns above as a more premenent fix though.

tmthrgd:
This is more or less what I’d propose the fix be:
```go
func (w *response) disposeBuffer() {
	if w.udp != nil && cap(w.msg) == srv.UDPSize {
		srv.udpPool.Put(w.msg[:srv.UDPSize])
	}
	w.msg = nil
}

func (srv *Server) serveDNS(w *response) {
	req := new(Msg)
	err := req.Unpack(w.msg)
	if err != nil { // Send a FormatError back
		x := new(Msg)
		x.SetRcodeFormatError(req)
		w.WriteMsg(x)
	}
	if err != nil || !srv.Unsafe && req.Response {
		w.disposeBuffer()
		return
	}

	w.tsigStatus = nil
	if w.tsigSecret != nil {
		if t := req.IsTsig(); t != nil {
			if secret, ok := w.tsigSecret[t.Hdr.Name]; ok {
				w.tsigStatus = TsigVerify(w.msg, secret, "", false)
			} else {
				w.tsigStatus = ErrSecret
			}
			w.tsigTimersOnly = false
			w.tsigRequestMAC = req.Extra[len(req.Extra)-1].(*TSIG).MAC
		}
	}

	w.disposeBuffer()

	handler := srv.Handler
	if handler == nil {
		handler = DefaultServeMux
	}

	handler.ServeDNS(w, req) // Writes back to the client
}
```

thyth:
Agreed. I didn't initially understand that the semantics of a sync.Pool "get" is like a "malloc" and "put" is like a "free". Calling out the intent explicitly with a function named "disposeBuffer" makes the intent clearer too.

I've made it a function on the server rather than the message though, since it needs to refer to the server's pool, and UDPSize fields.

I'll add a TSIG query creation/validation unit test to go along with this.

server_test.go
```diff
@@ -999,6 +999,60 @@ func TestServerReuseport(t *testing.T) {
...
+			if status == nil {
+				// *Msg r has an TSIG record and it was validated
+				m.SetTsig("test.", HmacMD5, 300, time.Now().Unix())
+				handlerError = nil
+			} else {
+				// *Msg r has an TSIG records and it was not valided
+				handlerError = fmt.Errorf("invalid TSIG: %v", status)
+			}
+		} else {
+			handlerError = fmt.Errorf("missing TSIG")
```
tmthrgd: You’re allowed to simply call `t.Error` (though not `t.Fatal`) from here.

server_test.go
```diff
@@ -999,6 +1019,61 @@ func TestServerReuseport(t *testing.T) {
...
+			if status == nil {
+				// *Msg r has an TSIG record and it was validated
+				m.SetTsig("test.", HmacMD5, 300, time.Now().Unix())
+				close(handlerError)
+			} else {
+				// *Msg r has an TSIG records and it was not valided
+				handlerError <- fmt.Errorf("invalid TSIG: %v", status)
+			}
+		} else {
+			handlerError <- fmt.Errorf("missing TSIG")
```
tmthrgd: You’re allowed to directly call `t.Error` (though not `t.Fatal`) from here.

server_test.go
```diff
@@ -59,6 +59,26 @@ func RunLocalUDPServer(laddr string) (*Server, string, error) {
 	return server, l, err
 }
 
+func RunLocalUDPServerWithSecrets(laddr string, secrets map[string]string) (*Server, string, error) {
```
tmthrgd: I don’t like adding this. Unless it trips the race detector, why not set `TsigSecret` in `TestServerRoundtripTsig`.

thyth: It tripped the race detector. I couldn't see a way to set TsigSecret under a lock without adding this.

tmthrgd: I thought it would (I had the same problem in #717). I have a solution that won't blow up the diff though!

Replace:
`func RunLocalUDPServerWithFinChan(laddr string) (*Server, string, chan error, error)`
with
`func RunLocalUDPServerWithFinChan(laddr string, opts ...func(*Server)) (*Server, string, chan error, error)`.

You can then invoke each `opts` before the `ActivateAndServe` goroutine is started, and a closure can be passed in to change fields on `Server`. Because this is a `variadic` function, none of the existing calls have to be updated.

/cc @miekg for his thoughts.

thyth: That's ultra clever. Implementing it that way.

PR #751, opened 2018-09-26
Replace unpackTxtString with identical unpackString

tmthrgd:
These two functions were identical (sans-variable names) before I optimized `unpackString` in 5debfeec63 (#746).

This will improve the performance of it's only caller `unpackTxt` and is covered by the test and benchmark added in 5debfeec63.

### DIFF
--- a/msg.go
+++ b/msg.go
@@ -512,45 +512,12 @@ func unpackTxt(msg []byte, off0 int) (ss []string, off int, err error) {
 	off = off0
 	var s string
 	for off < len(msg) && err == nil {
-		s, off, err = unpackTxtString(msg, off)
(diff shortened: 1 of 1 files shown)


miekg:
NIce. a long standing annoyance that I shrugged off years ago.

miekg:
go tip looks off in travis.

tmthrgd:
@miekg Hmm, I've seen that exact same thing happen for a few other Travis builds. I'll just restart it now and see what happens. It seems to just randomly hang before running `go test` sometimes.

tmthrgd:
Master has passed the Travis build (https://travis-ci.org/miekg/dns/builds/433359542) fine so it had to have been transient.

PR #754, opened 2018-09-26
Rework and optimise ServeMux

tmthrgd:
This is several changes to `ServeMux` in one PR:
- The zero `ServeMux` is now usable, `NewServeMux` doesn't need to be called.
- A series of documentation changes.
- Changes to make `ServeMux.ServeDNS` more readable.
- Performance improvements, and simplification, in `ServeMux.match`. Note this is complicated somewhat by ^1.

```
name                   old time/op    new time/op    delta
MuxMatch/lowercase-12     108ns ± 3%      68ns ± 2%   -36.94%  (p=0.000 n=10+10)
MuxMatch/uppercase-12     111ns ± 3%     120ns ± 1%    +7.34%  (p=0.000 n=9+10)

name                   old alloc/op   new alloc/op   delta
MuxMatch/lowercase-12     32.0B ± 0%      0.0B       -100.00%  (p=0.000 n=10+10)
MuxMatch/uppercase-12     32.0B ± 0%     32.0B ± 0%      ~     (all equal)

name                   old allocs/op  new allocs/op  delta
MuxMatch/lowercase-12      1.00 ± 0%      0.00       -100.00%  (p=0.000 n=10+10)
MuxMatch/uppercase-12      1.00 ± 0%      1.00 ± 0%      ~     (all equal)
```

^1: `strings.ToLower` contains one extra allocation in the uppercase case that will be fixed by [CL 137575](https://go-review.googlesource.com/c/go/+/137575). Until that lands, I've reimplemented an ASCII-only `ToLower` that avoids the allocation. This complicates the diff and only improves the case where `q` (the domain name) has uppercase characters. @miekg If this is uncommon, I'd be very happy to revert that part of this PR, but it will temporarily worsen the uppercase performance.

### DIFF
--- a/serve_mux.go
+++ b/serve_mux.go
@@ -5,12 +5,14 @@ import (
 	"sync"
 )
 
-// ServeMux is an DNS request multiplexer. It matches the
-// zone name of each incoming request against a list of
-// registered patterns add calls the handler for the pattern
-// that most closely matches the zone name. ServeMux is DNSSEC aware, meaning
-// that queries for the DS record are redirected to the parent zone (if that
-// is also registered), otherwise the child gets the query.
+// ServeMux is an DNS request multiplexer. It matches the zone name of
+// each incoming request against a list of registered patterns add calls
+// the handler for the pattern that most closely matches the zone name.
+//
+// ServeMux is DNSSEC aware, meaning that queries for the DS record are
+// redirected to the parent zone (if that is also registered), otherwise
+// the child gets the query.
+//
 // ServeMux is also safe for concurrent access from multiple goroutines.
 //
 // The zero ServeMux is empty and ready for use.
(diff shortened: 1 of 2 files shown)


tmthrgd:
[CL 137575](https://go-review.googlesource.com/c/go/+/137575) has now been merged as golang/go@14e7f174c1, meaning it will appear in go1.12. That should be released at the start of February next year ([see here](https://pocketgophers.com/go-release-timeline/#go1.12)).

serve_mux.go
```diff
@@ -81,25 +109,26 @@ func (mux *ServeMux) HandleRemove(pattern string) {
...
+// are redirected to the parent zone (if that is also registered),
+// otherwise the child gets the query.
+//
+// If no handler is found, or there is no question, a standard SERVFAIL
+// message is returned
+func (mux *ServeMux) ServeDNS(w ResponseWriter, req *Msg) {
 	var h Handler
-	if len(request.Question) < 1 { // allow more than one question
-		h = failedHandler()
+	if len(req.Question) >= 1 { // allow more than one question
```
miekg: this may be a breaking change that the current code depending on this cannot deal with. Also qsection > 1 isn't really defined, so we don't have to cater for it.

PR #755, opened 2018-09-26
Fix unpackString bug: 127 DEL is unprintable

tmthrgd:
This case previously differed from `UnpackDomainName` in msg.go and both `sprintTxtOctet` and `appendTXTStringByte` in types.go and was incorrect.

32 is ASCII space and 126 is ASCII tilde, with 127 being DEL. This was obviously meant to be `>=` but this is consistent with the format of types.go.

### DIFF
--- a/msg_helpers.go
+++ b/msg_helpers.go
@@ -275,7 +275,7 @@ func unpackString(msg []byte, off int) (string, int, error) {
 		case b == '"' || b == '\\':
 			s.WriteByte('\\')
 			s.WriteByte(b)
-		case b < 32 || b > 127: // unprintable
+		case b < ' ' || b > '~': // unprintable
 			var buf [3]byte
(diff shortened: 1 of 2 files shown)


msg_helpers.go
```diff
@@ -275,7 +275,7 @@ func unpackString(msg []byte, off int) (string, int, error) {
 		case b == '"' || b == '\\':
 			s.WriteByte('\\')
 			s.WriteByte(b)
-		case b < 32 || b > 127: // unprintable
+		case b < ' ' || b > '~': // unprintable
```
miekg: I think using integer values is slightly better, as this signals "This is ASCII",not some Go unicode

miekg: inconsistency for the win ! let's leave the chars here then as well.

PR #759, opened 2018-09-26
Run go fmt on package

tmthrgd:
This is go fmt from go1.11 and so it picks up the new map formatting heuristic.

See golang/go@542ea5ad91.


miekg:
On mobile. Fmt change only feel free to merge

PR #762, opened 2018-09-27
Run dep ensure -update

tmthrgd:
This uses the latest dep as of golang/dep@3c04147ed4.

While we're here, add the missing constraints to Gopkg.toml.

Various packages where missing from vendor after #717 and #736.

I'd like to follow this up with an attempt to transition to go.mod.


tmthrgd:
Abandoned, see cef340ab2c08ba17fdfa1ed4fab6890c40beb122.

Issue #756, opened 2018-09-26
Travis build hangs


tmthrgd:
The Travis build has reached `go test` and hung before any tests have run several times now before being terminated by Travis CI.

It's happened recently in:
[job 433357210](https://travis-ci.org/miekg/dns/jobs/433357210)—go 1.10—as part of #754, and
[job 433352922](https://travis-ci.org/miekg/dns/jobs/433352922)—go tip—as part of #751.

This failure is transient and the test will pass after being reset. This should be investigated.

This is an example of log output:
```
travis_time:start:32705ed3
$ go test $TESTS


No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

The build has been terminated
```

tmthrgd:
This seems to be becoming quite common. I've also seen it now in:
[job 433469286](https://travis-ci.org/miekg/dns/jobs/433469286)—go tip—*again* as part of #754, and
[job 433557704](https://travis-ci.org/miekg/dns/jobs/433557704)—go 1.11—as part of #755.

tmthrgd:
I don’t see how it could be. Tests are always run in the same order, and each run is preceded by the name of the test being printed. The graceful shutdown tests were nowhere near the start. When it hangs, nothing is printed. It’s hard to say, but it feels like go test is hanging before running any tests at all. I don’t understand how either.

tmthrgd:
Is there anyway to get a core dump when Travis terminates a build in this way?

miekg:
Not that I know off. Usually when this happens it showed up locally with a
for loop to run the the tests, but YMMV because travis

tmthrgd:
It looks like it might be related to graceful shutdown (#717) in particular that `Shutdown` now waits indefinitely (`goroutine 743`). You can see that `goroutine 744` is stuck waiting in `net.(*UDPConn).ReadMsgUDP` which should have been interrupted.

tmthrgd:
I was on the right track with #774. I augmented the UDP code with logging in #761 and lo and behold:
```
2018/10/05 16:37:36 0xc0000a2618 in ShutdownContext
2018/10/05 16:37:36 0xc0000a2618 SetReadDeadline: 1970-01-01 00:00:01 +0000 UTC
[...]

PR #763, opened 2018-09-27
Make Server.Shutdown not wait for graceful shutdown

tmthrgd:
In b0dc93d276 the behaviour of `Shutdown` was changed to wait indefinitely for graceful shutdown. This reverts it to it's previous behaviour of terminating immediately.

`ShutdownContext` is now the only wait to perform a graceful shutdown.

This doesn't technically solve the problem as to why connections were getting stuck in `net.(*UDPConn).ReadMsgUDP`, but it makes `Shutdown` safe to use again.

Updates #756

### DIFF
--- a/server.go
+++ b/server.go
@@ -403,14 +403,29 @@ func (srv *Server) ActivateAndServe() error {
 	return &Error{err: "bad listeners"}
 }
 
+var cancelledCtx = func() context.Context {
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel()
+	return ctx
+}()
+
 // Shutdown shuts down a server. After a call to Shutdown, ListenAndServe and
 // ActivateAndServe will return.
+//
+// Unlike ShutdownContext, Shutdown will not wait for connections to
+// gracefully terminate.
 func (srv *Server) Shutdown() error {
-	return srv.ShutdownContext(context.Background())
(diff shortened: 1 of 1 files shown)


miekg:
I would rather find the root cause here; considering the amount work that went into this. I need to get some coredns work out of the way, but I'll try to take a look.

tmthrgd:
Graceful shutdown will still work using the `ShutdownContext` call in this case. I’ll keep looking into why the read isn’t being interrupted.

tmthrgd:
@miekg Another option similar to this, is to make `Shutdown` have a timeout by default (perhaps say one minute) so it can't block forever (which is the problem at the moment). If someone wants to wait forever, they can still use `ShutdownContext` with `context.Background()`.

miekg:
(should look at the code)

But how can we have that problem? Why is it blocked when we initiated a
shutdown, at some point the queue should be empty and we can stop?

tmthrgd:
@miekg For the shutdown to be graceful, we have to wait until all the inflight requests are finished before closing the UDP connection. `SetReadDeadline` is used to unblock reads (exactly as is done in the standard library), but for some reason it doesn't always unblock and the read never finishes causing the whole thing to block and timeout.

tmthrgd:
@miekg Ohhhhhh I think I've figured it out. Let's see if I can fix it properly.

PR #765, opened 2018-09-28
Rollback PR #738 because it breaks compatibility with Windows

ibvrubleg:
Fixes this issue of the CoreDNS: https://github.com/coredns/coredns/issues/2138

### DIFF
--- a/udp.go
+++ b/udp.go
@@ -1,3 +1,5 @@
+// +build !windows
+
 package dns
 
 import (
--- a/udp_test.go
+++ b/udp_test.go
@@ -1,3 +1,5 @@
+// +build linux,!appengine
+
 package dns
 
 import (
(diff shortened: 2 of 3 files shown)


miekg:
seems harmless enough to revert. @tmthrgd any thoughts here?

tmthrgd:
I had wanted #738 tested before it merged, which would have caught this.

`SetControlMessage` just isn’t implemented on Windows:
https://github.com/golang/net/blob/e514e69ffb8bc3c76a71ae40de0118d794855992/ipv4/control_windows.go

I’ll open an issue against x/net/{ipv4,ipv6} for that, but this should be merged in the meantime. I’m not sure from my limited searching that windows even supports control messages.

tmthrgd:
I've filed a bug upstream as golang/go#27966.

PR #768, opened 2018-10-02
NSEC type bitmap packing bug

### LINKED ISSUE #741: NSEC type bitmap bug
Hi

I believe there is bug in the way NSEC RR is packed into wireformat
when compression is turned ON.

The problem occurs when the first label of the NSEC ownername is long enough
(about 20 bytes is enough) and it is encoded with compression on the
wireformat of the DNS message.

On the func 'packDomainName' in msg.go, the msg is filled up until the
first '.' is found, and then, if compression is available, the pointer
is added, but the extra bytes of the label that were filled previously
are left there (they don't get zeroed). What happens next, is that on
func 'packDataNsec', the bitmap part of the msg is filled with the
operator '|=', on a byte might already be set.

I don't know if I made myself clear enough, but if any doubts remain, I'd be happy to help.

tmthrgd:
This clears the tainted part of the message before writing the compression pointer.

This includes a test case ported from the original issue.

Fixes #741

/cc @cesarkuroiwa

### DIFF
--- a/msg.go
+++ b/msg.go
@@ -302,6 +302,12 @@ func packDomainName(s string, msg []byte, off int, compression map[string]int, c
 	}
 	// If we did compression and we find something add the pointer here
 	if pointer != -1 {
+		// Clear the msg buffer after the pointer location, otherwise
+		// packDataNsec writes the wrong data to msg.
+		tainted := msg[nameoffset:off]
+		for i := range tainted {
+			tainted[i] = 0
+		}
 		// We have two bytes (14 bits) to put the pointer in
 		// if msg == nil, we will never do compression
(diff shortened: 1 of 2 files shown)


msg.go
```diff
@@ -302,6 +302,12 @@ func packDomainName(s string, msg []byte, off int, compression map[string]int, c
 	}
 	// If we did compression and we find something add the pointer here
 	if pointer != -1 {
+		// Clear the msg buffer after the pointer location, otherwise
+		// packDataNsec writes the wrong data to msg.
+		tainted := msg[nameoffset:off]
```
miekg: would copy() be applicable here?

tmthrgd: I don't see how, we just need to zero all the bits after the compression pointer.

Issue #766, opened 2018-09-28
Panic on udpsession


semihalev:
I got the panic two times this week. How can i handle this?

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x93e4db]

goroutine 129363104 [running]:
github.com/miekg/dns.(*SessionUDP).RemoteAddr(...)
	/build/go/src/github.com/miekg/dns/udp.go:34
github.com/miekg/dns.(*response).RemoteAddr(0xc0025ebdd0, 0xc0001d6a80, 0xc000952d72)
	/build/go/src/github.com/miekg/dns/server.go:773 +0x5b

miekg:
[ Quoting <notifications@github.com> in "[miekg/dns] Panic on udpsession (#7..." ]
>I got the panic two times this week. How can i handle this?
>
>panic: runtime error: invalid memory address or nil pointer dereference
>[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x93e4db]
>
>goroutine 129363104 [running]:
>github.com/miekg/dns.(*SessionUDP).RemoteAddr(...)
>	/build/go/src/github.com/miekg/dns/udp.go:34
>github.com/miekg/dns.(*response).RemoteAddr(0xc0025ebdd0, 0xc0001d6a80, 0xc000952d72)
>	/build/go/src/github.com/miekg/dns/server.go:773 +0x5b

on linux? With master?

semihalev:
Yes on linux with master.

tmthrgd:
@semihalev Would you mind supplying the full panic?

This looks like days corruption to me. Are you, directly or indirectly, using `unsafe` or Cgo?

tmthrgd:
Actually scratch that, I know exactly what’s happening. You’re calling ‘(*response).Close` somewhere that sets `(*response).tcp` to `nil`, and *then* calling `(*response).RemoteAddr` which assumes that one of `tcp` or `udpSession` is set.

This needs to be fixed properly, but if you stop doing that it will all work.

semihalev:
The panic log very small, there is one more function. I don't have `unsafe` or Cgo.

Yes I'm calling `Close` method on defer. I will check this but the main problem looks like , `(*SessionUDP).raddr` is nil

tmthrgd:
[...]

PR #769, opened 2018-10-02
Explicitly panic after (*response).Close

tmthrgd:
Calling any of `Write`, `LocalAddr` or `RemoteAddr` after calling `Close` on a TCP connection will cause a panic like the following:
```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x93e4db]

goroutine 129363104 [running]:
github.com/miekg/dns.(*SessionUDP).RemoteAddr(...)
/build/go/src/github.com/miekg/dns/udp.go:34
github.com/miekg/dns.(*response).RemoteAddr(0xc0025ebdd0, 0xc0001d6a80, 0xc000952d72)
/build/go/src/github.com/miekg/dns/server.go:773 +0x5b
```

This pull request adds explicit panics in each of these functions. An alternative is to make `LocalAddr` and `RemoteAddr` to return valid results after `Close`, but this is more explicit at least until such a decision is made.

Updates #766

/cc @semihalev

### DIFF
--- a/server.go
+++ b/server.go
@@ -754,7 +754,7 @@ func (w *response) Write(m []byte) (int, error) {
 		n, err := io.Copy(w.tcp, bytes.NewReader(m))
 		return int(n), err
 	default:
-		panic("Write called after Close")
+		panic("dns: Write called after Close")
 	}
 }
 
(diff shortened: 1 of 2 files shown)


server.go
```diff
@@ -753,24 +753,33 @@ func (w *response) Write(m []byte) (int, error) {
 
 		n, err := io.Copy(w.tcp, bytes.NewReader(m))
 		return int(n), err
+	default:
+		panic("Write called after Close")
```
miekg: for panic it's customary to prefix it them with the package: `panic("dns: write called after Close")`

tmthrgd: I don't know how I missed that, I always prefix errors and panics in my own packages. :man_facepalming:

PR #770, opened 2018-10-02
Run dep ensure -update

tmthrgd:
This uses the latest dep as of golang/dep@3c04147ed4.

While we're here, add the missing constraints to Gopkg.toml.

Various packages where missing from vendor after #717 and #736.

I'd like to follow this up with an attempt to transition to go.mod. vendoring and dep is a very messy solution.

This is a second attempt after #762 because it conflicted after cef340ab2c08ba17fdfa1ed4fab6890c40beb122. @miekg Can you explain what happened in that commit?


miekg:
Merge whatever works. I previously saw compile errors that looked dep
related. Didn't found the inclination to delve deeper.

PR #771, opened 2018-10-02
Dirty hack to fix Travis build failures

tmthrgd:
This is a somewhat dirty hack to stop the Travis builds from failing. It's only intended to be temporary and is an alternative to #763, although #763 may be worth merging regardless.

Updates #756


miekg:
do we still want this?

If so feel free to merge

PR #774, opened 2018-10-04
Avoid overriding aLongTimeAgo read deadline

### LINKED ISSUE #756: Travis build hangs

tmthrgd:
The `(*Server).readTCP` and `(*Server).readUDP` methods both call `SetReadDeadline`. If they race with `ShutdownContext`, they can override the `aLongTimeAgo` read deadline that is set in `ShutdownContext`. That will cause `ShutdownContext` to hang or timeout.

~:crossed_fingers: Let's see if this works.~

Updates #756
Updates #763

### DIFF
--- a/server.go
+++ b/server.go
@@ -667,10 +667,10 @@ func (srv *Server) serveDNS(w *response) {
 
 func (srv *Server) readTCP(conn net.Conn, timeout time.Duration) ([]byte, error) {
 	if srv.isStarted() {
-		// If we race with ShutdownContext, a distant read
-		// deadline will be set to unblock the read below.
-		// We must not override it, otherwise ShutdownContext
-		// might hang.
+		// If we race with ShutdownContext, the read deadline may
+		// have been set in the distant past to unblock the read
+		// below. We must not override it, otherwise we may block
+		// ShutdownContext.
 		conn.SetReadDeadline(time.Now().Add(timeout))
 	}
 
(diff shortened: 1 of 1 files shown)


tmthrgd:
Damn, didn't fix it. @miekg This is still worth merging anyhow as it is still a race and it should have been in the original PR.

server.go
```diff
@@ -666,7 +666,14 @@ func (srv *Server) serveDNS(w *response) {
 }
 
 func (srv *Server) readTCP(conn net.Conn, timeout time.Duration) ([]byte, error) {
-	conn.SetReadDeadline(time.Now().Add(timeout))
+	if srv.isStarted() {
+		// If we race with ShutdownContext, a distant read
+		// deadline will be set to unblock the read below.
```
tmthrgd: I don’t like the ‘a distant read deadline’ wording, I’ll update it.

miekg:
:/
was just about to say "nice find"! But then I saw your other comments.

/lgtm

PR #775, opened 2018-10-04
Only test coverage against latest stable go

tmthrgd:
**DO NOT MERGE**: This is a work in progress.

The coverage Travis builds are much slower (2x on average) than the non-coverage builds. The coverage results also shouldn't diverge that significantly between go versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,16 @@
 language: go
 sudo: false
+
 go:
   - 1.10.x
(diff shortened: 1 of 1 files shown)


tmthrgd:
`go: tip` is the slowest build so we're only getting one minute faster with this PR. I'll revisit this once travis-ci/gimme#151 lands which should substantially speed up `go: tip`.

PR #776, opened 2018-10-05
Allows larger offset in $GENERATE

growler:
Although BIND9 documentations does not specify the possible size of offset
in $GENERATE clause, it clearly says that range must be a positive integer
between 0 and (2^31)-1. Moreover, BIND perfectly supports large offsets
which might be really handy when you're dealing with large intranets. I.e.
consider following case

```
$GENERATE 0-255 dhcp-${0,4,d}   A 10.0.0.$
$GENERATE 0-255 dhcp-${256,4,d} A 10.0.1.$
$GENERATE 0-255 dhcp-${512,4,d} A 10.0.2.$
...
```

This change removes offset size check and introduces check that
0 >= (start + offset) and (end + offset) < (2^31)-1.

### DIFF
--- a/generate.go
+++ b/generate.go
@@ -107,6 +107,8 @@ BuildRR:
 					mod, offset, err = modToPrintf(s[j+2 : j+2+sep])
 					if err != nil {
 						return err.Error()
+					} else if start + offset < 0 || end + offset > 1<<31-1 {
+						return "bad offset in $GENERATE"
 					}
 					j += 2 + sep // Jump to it
 				}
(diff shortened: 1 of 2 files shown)


growler:
(sorry for repeated merge request, I think I messed up the first one)

We're using miekg/dns for a tool auto-deploying internal zones from gitlab repository, so I believe this change is useful while being totally harmless. Please let me know if anything else should be fixed (I'll try to not mess it up this time). Thank you!

PR #780, opened 2018-10-07
Hold srv.lock while calling SetReadDeadline

### LINKED ISSUE #756: Travis build hangs

tmthrgd:
It seems I was very much on the right track with #774, but the race was still there. This *should* finally solve the problem and make graceful shutdown more reliable and not hang like before.

Like I did in #774, I'm going to run the Travis build a few times and see if anything trips.

~This does introduce a lock where there previously wasn't so it could negatively affect performance, that being said, only a read lock is held during the critical path which shouldn't cause any lock congestion.~ (Edit: no it doesn’t, `isShutdown` already held the same lock).

Fixes #756
Closes #761
Closes #763
Closes #771
Updates #774

### DIFF
--- a/server.go
+++ b/server.go
@@ -416,13 +416,12 @@ func (srv *Server) Shutdown() error {
 // to terminate.
 func (srv *Server) ShutdownContext(ctx context.Context) error {
 	srv.lock.Lock()
-	started := srv.started
-	srv.started = false
-	srv.lock.Unlock()
-
-	if !started {
+	if !srv.started {
+		srv.lock.Unlock()
 		return &Error{err: "server not started"}
 	}
+
+	srv.started = false
 
 	if srv.PacketConn != nil {
 		srv.PacketConn.SetReadDeadline(aLongTimeAgo) // Unblock reads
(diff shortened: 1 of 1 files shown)


miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Hold srv.lock while..." ]
>@miekg I've run the build multiple times now and I cannot reproduce the original failure. :crossed_fingers: This looks like The Fix™.

Nice!

Issue #786, opened 2018-10-11
Memory Leak in github.com/miekg/dns.parseZone


w8rbt:
I'm using coredns 1.2.2 with the built-in file plugin (https://github.com/coredns/coredns/tree/master/plugin/file) to serve a few RFC 1035 master zone files. coredns uses github.com/miekg/dns. I'm also using the more resent version of Go (1.11.1). 

In this configuration, coredns memory usage grows steadily and crashes after about 18 hours with an OOM error with a server that contains 2GB of memory. Here's a snippet from the coredns log file showing the crash:

2018/10/10 13:56:17 [INFO] CoreDNS-1.2.2
2018/10/10 13:56:17 [INFO] linux/amd64, go1.11.1, 8a9c6174
...
 1360 fatal error: runtime: out of memory
 1361 
 1362 runtime stack:
 1363 runtime.throw(0x192f945, 0x16)
 1364     /home/ubuntu/go1.11/src/runtime/panic.go:608 +0x72
 1365 runtime.sysMap(0xc064000000, 0x4000000, 0x29ef978)
 1366     /home/ubuntu/go1.11/src/runtime/mem_linux.go:156 +0xc7
 1367 runtime.(*mheap).sysAlloc(0x29d59a0, 0x4000000, 0x29d59b8, 0x7fd3ecb43a28)
 1368     /home/ubuntu/go1.11/src/runtime/malloc.go:619 +0x1c7
 1369 runtime.(*mheap).grow(0x29d59a0, 0x1, 0x0)
 1370     /home/ubuntu/go1.11/src/runtime/mheap.go:920 +0x42
 1371 runtime.(*mheap).allocSpanLocked(0x29d59a0, 0x1, 0x29ef988, 0x0)
 1372     /home/ubuntu/go1.11/src/runtime/mheap.go:848 +0x337
 1373 runtime.(*mheap).alloc_m(0x29d59a0, 0x1, 0x430008, 0xc000044500)
 1374     /home/ubuntu/go1.11/src/runtime/mheap.go:692 +0x119
 1375 runtime.(*mheap).alloc.func1()
 1376     /home/ubuntu/go1.11/src/runtime/mheap.go:759 +0x4c
 1377 runtime.(*mheap).alloc(0x29d59a0, 0x1, 0xc000010008, 0x432339)
 1378     /home/ubuntu/go1.11/src/runtime/mheap.go:758 +0x8a
[...]

Issue #568, opened 2017-11-16
go routine leak in scan.go?


miekg:
from https://github.com/coredns/coredns/issues/1233

~~~
goroutine profile: total 27
9 @ 0x3e6c8 0x3e790 0x14124 0x13ea8 0x224060 0x6fd2c
#	0x22405f	github.com/miekg/dns.zlexer+0x347 /home/miek/g/src/github.com/miekg/dns/scan.go:602
~~~

This is code that calls `NewRR` which starts a goroutine to parse, maybe that one can be made not to return?

Highly speculative atm, btw.

miekg:
Number of leaks has 1:1 mapping with seeing the following record:
`mplus.ims.vodafone.com. 1866 NAPTR 10 100 S SIP+D2U  _sip._udp.mplus.ims.vodafone.com.`

miekg:
According to the spec S and SIP+D2U need to be enclosed in quotes, see https://www.ietf.org/rfc/rfc2915.txt

https://dns.google.com/resolve?name=mplus.ims.vodafone.com.&type=NAPTR
{"Status": 0,"TC": false,"RD": true,"RA": true,"AD": false,"CD": false,"Question":[ {"name": "mplus.ims.vodafone.com.","type": 35}],"Answer":[ {"name": "mplus.ims.vodafone.com.","type": 35,"TTL": 1985,"data": "10 100 S SIPS+D2T  _sips._tcp.mplus.ims.vodafone.com."},{"name": "mplus.ims.vodafone.com.","type": 35,"TTL": 1985,"data": "10 100 S SIP+D2U  _sip._udp.mplus.ims.vodafone.com."},{"name": "mplus.ims.vodafone.com.","type": 35,"TTL": 1985,"data": "10 100 S SIP+D2T  _sip._tcp.mplus.ims.vodafone.com."}]}

miekg:
So I think this is happening:
We crap out of scan_rr.go because the record is wrong: https://github.com/miekg/dns/blob/master/scan_rr.go#L690
But the `chan lex` is not closed (that happens in the zlexer function with a `defer`), so that stays alive, and the scanText also doesn't see a error, so from inside `zlexer` everything looks OK - hence goroutine leak.

This is true for all RRs btw, just showed up in this NAPTR.

miekg:
Closed via #570

PR #792, opened 2018-10-14
Eliminate lexer goroutines

tmthrgd:
This is a pretty large pull request, but it's ultimately rather simple. The `zlexer` and `klexer` goroutines are eliminated and replaced with a new type with a simple chan-like API. This eliminates the goroutines and lexer chans, which will reduce memory usage.

The `scan_rr.go` changes aren't very interesting as they're mostly just find-and-replace.

Errors such as #568/#570 should now be impossible from the `zlexer` code, although similar goroutine leaks can occur from the `ParseZone` code, and in fact have in #786, which needs to be fixed later.

Note: This pull request contains an API change. `io.Reader` errors that were previously ignored in `zlexer` and `klexer` are now surfaced in `ParseZone` and `(*DNSKEY).ReadPrivateKey` respectively.

Also note: `testRR` now panics on error, because that tripped me up multiple times while building this.

```
name       old time/op    new time/op    delta
NewRR-12     15.5µs ± 7%     6.3µs ± 1%  -59.24%  (p=0.000 n=10+10)
ReadRR-12    15.2µs ± 0%     6.7µs ± 1%  -56.14%  (p=0.000 n=7+10)

name       old alloc/op   new alloc/op   delta
NewRR-12     5.22kB ± 0%    0.81kB ± 0%  -84.52%  (p=0.000 n=10+10)
ReadRR-12    5.11kB ± 0%    1.82kB ± 0%  -64.44%  (p=0.000 n=9+10)

name       old allocs/op  new allocs/op  delta
NewRR-12       23.0 ± 0%      17.0 ± 0%  -26.09%  (p=0.000 n=10+10)
ReadRR-12      23.0 ± 0%      19.0 ± 0%  -17.39%  (p=0.000 n=10+10)
```

Updates #786
Updates coredns/coredns#2194


tmthrgd:
@miekg Almost all the memory allocated in NewRR (all but 0.98KB per run) is from [`bufio.NewScanner`](https://github.com/golang/go/blob/389e942745ddd7eef44b71571c463b0dfc3dcac2/src/bufio/bufio.go#L60-L63) and it's default [4KiB buffer](https://github.com/golang/go/blob/389e942745ddd7eef44b71571c463b0dfc3dcac2/src/bufio/bufio.go#L18). That seems excessive to me.

miekg:
J*sus, your're fast. Let me review :)

bufio.NewScanner seems a big hit indeed.

tmthrgd:
I've addressed the `bufio.Reader` allocation by special-casing `io.ByteReader` and reducing the buffer size where it's needed. We read byte-by-byte and any type that implements `io.ByteReader` shouldn't need the extra buffering anyway. That drastically cuts down on the memory use and gives a nice speed-up.

> J*sus, your're fast. Let me review :)

It's a worthwhile Sunday afternoon. :grin:

tmthrgd:
Hold off merging, I screwed up the case of comments inside braces I think. I need to review that part a little closer. It’s not covered by any tests at the moment.

dnssec_keyscan.go
```diff
@@ -219,66 +209,140 @@ func parseKey(r io.Reader, file string) (map[string]string, error) {
...
-	var l lex
-	str := "" // Hold the current read text
-	commt := false
-	key := true
-	x, err := s.tokenText()
-	defer close(c)
-	for err == nil {
-		l.column = s.position.Column
-		l.line = s.position.Line
+func (kl *klexer) Next() (lex, bool) {
```
miekg: The comment above it not true anymore.

tmthrgd: Good catch, I'll remove the comment from both of them.

tmthrgd:
@miekg What should the behaviour of this code be?
```go
func TestParseBracedComments(t *testing.T) {
	r := strings.NewReader(`name. IN SOA  a6.nstld.com. hostmaster.nic.name. (
		203362132 ; serial
		5m        ; refresh (5 minutes)
		5m        ; retry (5 minutes)
		2w        ; expire (2 weeks)
		300       ; minimum (5 minutes)
	) ; y
	. 3600000  IN  NS ONE.MY-ROOTS.NET. ; x`)
	for r := range ParseZone(r, "", "") {
		t.Log(r.Comment)
	}
}
```

Without the `; y` comment it prints:
```
=== RUN   TestParseBracedComments
--- PASS: TestParseBracedComments (0.00s)
    parse_test.go:352:
    parse_test.go:352:
```

With the `; y` comment it prints:
```
=== RUN   TestParseBracedComments
--- PASS: TestParseBracedComments (0.00s)
    parse_test.go:352: ; serial ; refresh (5 minutes) ; retry (5 minutes) ; expire (2 weeks) ; minimum (5 minutes) ; y
    parse_test.go:352:
```

It loses the `; x` comment because it's at the end of a line and it loses the comment in the braces if there isn't a comment after the braces. Does it matter and is this intentional?

miekg:
The "returning comments" seemed like a must have feature at the time, but I doubt it is useful. There is also no spec that says what it should do, so anything goes.

tmthrgd:
I was hoping I could rip out all the comment handling in a follow-up pull request, because it's somewhat messy and dicey, but unfortunately it's relied upon by [barnybug/cli53](https://github.com/barnybug/cli53/blob/d6f1c68a757d81dd3e8b3ecbb7caf82109232592/bind.go#L15-L50).

tmthrgd:
@miekg Unless there's anything you want me to change, I believe this is now done and ready to merge.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Eliminate lexer gor..." ]
>I was hoping I could rip out all the comment handling in a follow-up pull request, because it's somewhat messy and dicey, but unfortunately it's relied upon by [barnybug/cli53](https://github.com/barnybug/cli53/blob/d6f1c68a757d81dd3e8b3ecbb7caf82109232592/bind.go#L15-L50).

yeah, stuff like this is why I added it, at the time pool.ntp.org wanted such a 
feature.

/Miek

--
Miek Gieben

miekg:
coredns test suite also runs OK using this pr.

LGTM

tmthrgd:
@miekg Shall I merge?

miekg:
Go ahead

On 15 Oct 2018 08:06, "Tom Thorogood" <notifications@github.com> wrote:

@miekg <https://github.com/miekg> Shall I merge?

—
You are receiving this because you were mentioned.

<https://github.com/miekg/dns/pull/792#issuecomment-429731155>, or mute the
thread

PR #799, opened 2018-10-21
ZoneParser api change

miekg:

### DIFF
--- a/generate.go
+++ b/generate.go
@@ -80,7 +80,7 @@ func (zp *ZoneParser) generate(l lex) (RR, bool) {
 		lex:  &l,
 	}
 	zp.sub = NewZoneParser(r, zp.origin, zp.file)
-	zp.sub.includeDepth, zp.sub.includeAllowed = zp.includeDepth, zp.includeAllowed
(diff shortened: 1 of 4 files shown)


miekg:
Slight change here to just make things public  as this remove a Setter function.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] ZoneParser api chan..." ]
>Why remove the setter function? I don’t see the motive here.

It's not needed, why have it?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] ZoneParser api chan..." ]
>Why remove the setter function? I don’t see the motive here.

won't die on this hill though; but in general is you can just make fields 
public you have a slightly smaller API and loose these 1 line get/set functions.

miekg:
If we're doing an include path it might be worth making the allowIncludes a
string specifying what to allow. / is the entire fs, '.' is relative to the
zonefile and empty signals not allowing includes.

miekg:
On second thought this api is fine and we also support 1035 you just have to flip a bit. Sorry for the noise.

PR #800, opened 2018-10-21
Revert doh

miekg:
After seeing https://github.com/semihalev/sdns *is not* using any of the build in DoH stuff I became more convinced we should not attempt to do things in this library.

### DIFF
--- a/client.go
+++ b/client.go
@@ -490,6 +490,7 @@ func (c *Client) ExchangeContext(ctx context.Context, m *Msg, a string) (r *Msg,
 	}
 	// not passing the context to the underlying calls, as the API does not support
 	// context. For timeouts you should set up Client.Dialer and call Client.Exchange.
+	// TODO(tmthrgd,miekg): this is a race condition.
 	c.Dialer = &net.Dialer{Timeout: timeout}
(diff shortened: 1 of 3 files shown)


miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Revert doh (#800)..." ]
>I’m not sure I see the connection, I can’t see anywhere that https://github.com/semihalev/sdns has a DOH *client* which is what is implemented in this library. Instead I can only see a DOH server.

It's not a full solution and burdens us with supporting it; it's a http 
protocol, not a DNS one. We could potentialy (and finally) start using the godns 
org and make a doh package or something?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Revert doh (#800)..." ]
>I think I'm fairly ambivalent to it at this point, so if you want to rip it out that's fine by me. I think it would need some more work to be entirely spec-compliant anyway. I do like the idea of some sort of DOH companion package.
>
>This LGTM, aside from the race condition comment that should be retained.

I'll fix that.

I share your feelings about DoH. 

I need to implement DoH in the plugin/forward anyway, would be interesting see 
how that looks.

Issue #802, opened 2018-10-27
Code sets socket to SO_REUSEPORT whether system supports it or not


stuartnelson3:
This option depends on the underlying system supporting it, in addition to being compiled with go1.11. Support should be checked before setting the option.

See discussion here:

https://github.com/coredns/coredns/issues/2239

miekg:
[ Quoting <notifications@github.com> in "[miekg/dns] Code unconditionally se..." ]
>This option depends on the underlying system supporting it, in addition to being compiled with go1.11. Support should be checked before setting the option.
>
>See discussion here:
>
>https://github.com/coredns/coredns/issues/2239

we're not using this code, we're setting it ourselves in coredns.

stuartnelson3:
That explains why I couldn't find where it was set :sweat_smile:

stuartnelson3:
From the PR:
>I'm not sure where you want this to live, or if it's on the user to know they can use SO_REUSEPORT or not.

Seeing as this is being set by users of the library, not the library itself, perhaps this check belongs elsewhere.

stuartnelson3:
Closing as the value is configurable by the user

PR #803, opened 2018-10-27
Read SO_REUSEPORT on socket before setting

stuartnelson3:
See if we're able to read the value from the
socket. If that fails, skip setting the value on
the port.

Note: I don't have any machines with an old enough kernel lying around to test that it fails correctly, but perhaps someone out there could jump in and give it a try.

Given this comment: https://github.com/miekg/dns/issues/802#issuecomment-433618478
I'm not sure where you want this to live, or if it's on the user to know they can use SO_REUSEPORT or not.

### DIFF
--- a/listen_go111.go
+++ b/listen_go111.go
@@ -5,6 +5,7 @@ package dns
 
 import (
 	"context"
+	"fmt"
 	"net"
 	"syscall"
 
@@ -16,6 +17,11 @@ const supportsReusePort = true
 func reuseportControl(network, address string, c syscall.RawConn) error {
 	var opErr error
 	err := c.Control(func(fd uintptr) {
+		_, err := unix.GetsockoptInt(int(fd), unix.SOL_SOCKET, unix.SO_REUSEPORT)
+		if err != nil {
+			opErr = fmt.Errorf("could not get SO_REUSEPORT: %s", err.Error())
+			return
+		}
 		opErr = unix.SetsockoptInt(int(fd), unix.SOL_SOCKET, unix.SO_REUSEPORT, 1)
 	})
 	if err != nil {


miekg:
Might be simpler to always do this, instead of the conditional compilation? Would there be any downsides to that?

stuartnelson3:
Based on this comment: https://github.com/miekg/dns/issues/654#issuecomment-375315602

It sounds like this behavior is only supported in go1.11+?

listen_go111.go
```diff
@@ -16,6 +17,11 @@ const supportsReusePort = true
 func reuseportControl(network, address string, c syscall.RawConn) error {
 	var opErr error
 	err := c.Control(func(fd uintptr) {
+		_, err := unix.GetsockoptInt(int(fd), unix.SOL_SOCKET, unix.SO_REUSEPORT)
+		if err != nil {
+			opErr = fmt.Errorf("could not get SO_REUSEPORT: %s", err.Error())
+			return
```
chantra: shouldn't this explicitly return opErr?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Read SO_REUSEPORT o..." ]
>
>
>
>> @@ -16,6 +17,11 @@ const supportsReusePort = true
> func reuseportControl(network, address string, c syscall.RawConn) error {
> 	var opErr error
> 	err := c.Control(func(fd uintptr) {
>+		_, err := unix.GetsockoptInt(int(fd), unix.SOL_SOCKET, unix.SO_REUSEPORT)
>+		if err != nil {
>+			opErr = fmt.Errorf("could not get SO_REUSEPORT: %s", err.Error())
>+			return
>
>shouldn't this explicitly return opErr?

this code already has a toggle to allow you to use it, there is no direct reason 
to tweak this. In coredns we ignore the error here.

stuartnelson3:
Closing, as users are allowed to opt-in to this code.

PR #804, opened 2018-10-28
nsec3 cover problems

semihalev:
I got a NX record (219.50.238.170.virbl.dnsbl.bit.nl) but looks like not covering in `(*NSEC3).Cover()` 

I checked on http://dnsviz.net/d/219.50.238.170.virbl.dnsbl.bit.nl/dnssec/ wildcard hash covered and valid. 

RFC 5155 section 8.4 (https://tools.ietf.org/html/rfc5155#section-8.4) described this.

Can you check it?

### DIFF
--- a/nsecx.go
+++ b/nsecx.go
@@ -63,6 +63,8 @@ func (rr *NSEC3) Cover(name string) bool {
 	}
 
 	nextHash := rr.NextDomain
+
+	// if empty interval found, try cover wildcard hashes so nameHash shouldn't match with ownerHash
 	if ownerHash == nextHash && nameHash != ownerHash { // empty interval
 		return true
 	}
--- a/nsecx.go
+++ b/nsecx.go
@@ -64,6 +64,9 @@ func (rr *NSEC3) Cover(name string) bool {
 
 	nextHash := rr.NextDomain
 	if ownerHash == nextHash { // empty interval
+		if nameHash < nextHash {
+			return true
+		}
 		return false
 	}
 	if ownerHash > nextHash { // end of zone
--- a/nsecx.go
+++ b/nsecx.go
@@ -64,10 +64,7 @@ func (rr *NSEC3) Cover(name string) bool {
 
 	nextHash := rr.NextDomain
 	if ownerHash == nextHash { // empty interval
-		if nameHash < nextHash {
-			return true
-		}
-		return false
+		return true
 	}
 	if ownerHash > nextHash { // end of zone
 		if nameHash > ownerHash { // covered since there is nothing after ownerHash
--- a/nsecx.go
+++ b/nsecx.go
@@ -63,7 +63,7 @@ func (rr *NSEC3) Cover(name string) bool {
 	}
 
 	nextHash := rr.NextDomain
-	if ownerHash == nextHash { // empty interval
+	if ownerHash == nextHash && nameHash != ownerHash { // empty interval
 		return true
 	}
 	if ownerHash > nextHash { // end of zone
--- a/nsecx_test.go
+++ b/nsecx_test.go
@@ -112,6 +112,18 @@ func TestNsec3(t *testing.T) {
 			name:   "asd.com.",
 			covers: false,
 		},
+		{ // empty interval close to next
+			rr: &NSEC3{
+				Hdr:        RR_Header{Name: "2n1tb3vairuobl6rkdvii42n9tfmialp.com."},
+				Hash:       1,
+				Flags:      1,
+				Iterations: 5,
+				Salt:       "F10E9F7EA83FC8F3",
+				NextDomain: "2N1TB3VAIRUOBL6RKDVII42N9TFMIALP",
+			},
+			name:   "asdc.com.",
+			covers: true,
+		},
 		{ // name hash is before owner hash, not covered
 			rr: &NSEC3{
 				Hdr:        RR_Header{Name: "3V62ULR0NRE83V0RJA2VJGTLIF9V6RAB.com."},
--- a/nsecx_test.go
+++ b/nsecx_test.go
@@ -110,6 +110,18 @@ func TestNsec3(t *testing.T) {
 				NextDomain: "2N1TB3VAIRUOBL6RKDVII42N9TFMIALP",
 			},
 			name:   "asd.com.",
+			covers: false,
+		},
+		{ // empty interval wildcard
+			rr: &NSEC3{
+				Hdr:        RR_Header{Name: "2n1tb3vairuobl6rkdvii42n9tfmialp.com."},
+				Hash:       1,
+				Flags:      1,
+				Iterations: 5,
+				Salt:       "F10E9F7EA83FC8F3",
+				NextDomain: "2N1TB3VAIRUOBL6RKDVII42N9TFMIALP",
+			},
+			name:   "*.asd.com.",
 			covers: true,
 		},
 		{ // name hash is before owner hash, not covered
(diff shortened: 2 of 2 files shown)


semihalev:
I found another record from net.sg, like the patch doesn't cover this problem. Owner and next hash  same but `namehash > nexthash` in this patch I covered `namehash < nexthash`  so the whole cover problems only be solve like this; 

```diff
if ownerHash == nextHash { // empty interval
-	return false
+	return true
}
```

```shell
$ dig ns starhub.net.sg +dnssec @NS2.CUHK.EDU.HK

; <<>> DiG 9.10.6 <<>> ns starhub.net.sg +dnssec @NS2.CUHK.EDU.HK
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 4496
;; flags: qr rd; QUERY: 1, ANSWER: 0, AUTHORITY: 4, ADDITIONAL: 3
;; WARNING: recursion requested but not available

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags: do; udp: 4096
;; QUESTION SECTION:
;starhub.net.sg.			IN	NS

;; AUTHORITY SECTION:
starhub.net.sg.		3600	IN	NS	pricus.starhub.net.sg.
starhub.net.sg.		3600	IN	NS	secdns.starhub.net.sg.
0V30QR0IPJLHVM43EML2DT3SULFGI7I8.net.sg. 900 IN	NSEC3 1 1 20 A4A4B9F8AC97A77B 0V30QR0IPJLHVM43EML2DT3SULFGI7I8  NS SOA RRSIG DNSKEY NSEC3PARAM
0V30QR0IPJLHVM43EML2DT3SULFGI7I8.net.sg. 900 IN	RRSIG NSEC3 8 3 900 20181128092242 20181029092242 47088 net.sg. H9EODbGuH4yRANlsfEgZBsdPOml7aUWXGebq8tVX8rdgEVLMFY7Uloid zggzGlHio63E+7TkhjmZIGwY6i8RYaoKdFdfJ7qvGnEgWzjF2I/DUwHB Is6A94aR60efZCDhqZWOgxqARiYA9bk4whhOtlvgwQQAM7norboZc7w/ h+g=

;; ADDITIONAL SECTION:
pricus.starhub.net.sg.	3600	IN	A	203.116.1.93
secdns.starhub.net.sg.	3600	IN	A	203.116.1.94

;; Query time: 496 msec
;; SERVER: 137.189.6.21#53(137.189.6.21)
;; WHEN: Mon Oct 29 15:07:26 +03 2018
;; MSG SIZE  rcvd: 371
```

semihalev:
http://dnsviz.net/d/starhub.net.sg/dnssec/ again the record secured and valid.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] nsec3 cover fix (#8..." ]
>I found another record from net.sg, like the patch doesn't cover this problem. Owner and next hash  same but `namehash > nexthash` in this patch I covered `namehash < nexthash`  so the whole cover problems only be solve like this;

IIRC the namespace is circular. I don't have the time to deep dive back into 
5155, but it seems you fix is too simple?

semihalev:
okey, I updated final patch.

miekg:
So you have an empty interval and that's OK because? Sorry nsec3 has been a while, can you help me understand why this fix is OK? Is there more text from 5155 that may help here?

semihalev:
let me try clear here with examples first;
```shell
$ dig 219.50.238.170.virbl.dnsbl.bit.nl +dnssec +multiline
; <<>> DiG 9.10.6 <<>> 219.50.238.170.virbl.dnsbl.bit.nl +dnssec +multiline @127.0.0.1
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 25713
;; flags: qr rd ra ad; QUERY: 1, ANSWER: 0, AUTHORITY: 4, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags: do; udp: 1536
;; QUESTION SECTION:
;219.50.238.170.virbl.dnsbl.bit.nl. IN A

;; AUTHORITY SECTION:
virbl.dnsbl.bit.nl.	10 IN SOA nsauth1.bit.nl. hostmaster.bit.nl. (
				2017020102 ; serial
				28800      ; refresh (8 hours)
				7200       ; retry (2 hours)
				604800     ; expire (1 week)
				600        ; minimum (10 minutes)
				)
ctj988dgude30juq42eq1v9sdr5li56h.virbl.dnsbl.bit.nl. 10	IN NSEC3 1 0 10 6E1285 (
				CTJ988DGUDE30JUQ42EQ1V9SDR5LI56H
				NS SOA RRSIG DNSKEY NSEC3PARAM )
```
you can see that, owner and next hash same, so `(*NSEC3).Cover()` return false immediately so we are saying this is unsecure and not cover. when i query in Google or Cloudflare their response, this is OK and secured because returned with AuthenticatedData flag. They are only using AD flag with secured responses. You can test it. Also in dnsviz.net, this record secured and covered. You can check here: http://dnsviz.net/d/219.50.238.170.virbl.dnsbl.bit.nl/dnssec/ 

5155 we try understand everything but I firstly tested in different public platforms. I can cancel this request but in resolver side I have too many errors about this.

semihalev:
This is the pdns implementation;

https://github.com/PowerDNS/pdns/blob/master/pdns/validate.cc#L30

they `return true` also they added another condition and final like this in our case;

```go
if ownerHash == nextHash && nameHash != ownerHash {
	return true
}
```

miekg:
Ok, can you put some kind of comment in the code, e.g. why the `nameHash != ownerHash` needed?

semihalev:
I put short comment.

nsecx.go
```diff
@@ -63,8 +63,10 @@ func (rr *NSEC3) Cover(name string) bool {
 	}
 
 	nextHash := rr.NextDomain
-	if ownerHash == nextHash { // empty interval
-		return false
+
+	// if empty interval found, try cover wildcard hashes so nameHash shouldn't match with ownerHash
```
miekg: Maybe:
If empty interval is found, ...
And then I'm lost; what is the second check proving here? `nameHash` is not equal to `ownerHash` because there may be a wildcard?

[I think it's important to get this right, because NSEC3 is so complex]

miekg: No, be more specific on what this checks does in the context of RFC 5155 and be a bit verbose about it. What does this check do and why is it needed (in what nsec3 corner case)

semihalev: RFC 5155 is not crystal clear about this. I read whole document but I couldn't explain correctly. I checked other implementations, like PERL libs, PowerDNS etc and debugged NSEC3 records and found this solution. PowerDNS has same implementation and there is no comment.

https://github.com/PowerDNS/pdns/blob/master/pdns/validate.cc#L30

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] nsec3 cover problem..." ]
>semihalev commented on this pull request.
>
>
>
>> @@ -63,8 +63,10 @@ func (rr *NSEC3) Cover(name string) bool {
> 	}
>
> 	nextHash := rr.NextDomain
>-	if ownerHash == nextHash { // empty interval
>-		return false
>+
>+	// if empty interval found, try cover wildcard hashes so nameHash shouldn't match with ownerHash
>
>RFC 5155 is not crystal clear about this. I read whole document but I couldn't explain correctly. I checked other implementations, like PERL libs, PowerDNS etc and debugged NSEC3 records and found this solution. PowerDNS has same implementation and there is no comment.
>
>https://github.com/PowerDNS/pdns/blob/master/pdns/validate.cc#L30

Ok, let's not rat hole this further then.

PR #808, opened 2018-11-03
Fix TCP connection tracking memory leak

### LINKED ISSUE #766: Panic on udpsession

### LINKED ISSUE #807: Server leaks memory on every TCP connection
Hello,

These insertions into `srv.conns`:

https://github.com/miekg/dns/blob/6ae357d393978566133a91aa1d4207b4f67da2c1/server.go#L496

are never released when the connection is closed:

https://github.com/miekg/dns/blob/6ae357d393978566133a91aa1d4207b4f67da2c1/server.go#L575-L585

The preceding call to `w.Close()` nils out `w.tcp`; the `delete()` has no effect, so the map grows without bound.

I use this library by way of [coredns/coredns](https://github.com/coredns/coredns).  Here is what the leak looks like after a quick local test (y-axis is RSS bytes):

screen shot 2018-11-03 at 04 23 29

I then applied this patch and tried again:

```
diff --git a/server.go b/server.go
index 4b4ec33c..ac655c4e 100644
--- a/server.go
+++ b/server.go
@@ -573,14 +573,14 @@ func (srv *Server) serve(w *response) {
        }

        defer func() {
-               if !w.hijacked {
-                       w.Close()
-               }
-
                srv.lock.Lock()
                delete(srv.conns, w.tcp)
                srv.lock.Unlock()

+               if !w.hijacked {
+                       w.Close()
+               }
+
                w.wg.Done()
        }()

```

Results:

screen shot 2018-11-03 at 04 26 02

tmthrgd:
This refactors how `response.Close` works. This fixes both the memory leak observed in #807 and a similar memory leak that would occur if `Close` was called manually by the handler. The `checkInProgressQueriesAtShutdownServer` test now checks the tracking map is empty after `ShutdownContext` returns.

Along for the ride are some changes to which `response` methods ~panic-after-`Close`~ (edit: fail-after-`Close`), this fixes #766 and allows `LocalAddr` and `RemoteAddr` to work after `Close`.

`Close` also now returns an error if called multiple times.

Fixes #807
Fixes #766
Updates #769

### DIFF
--- a/server.go
+++ b/server.go
@@ -82,6 +82,7 @@ type ConnectionStater interface {
 
 type response struct {
 	msg            []byte
+	closed         bool // connection has been closed
 	hijacked       bool // connection has been hijacked by handler
 	tsigTimersOnly bool
 	tsigStatus     error
@@ -749,6 +750,10 @@ func (w *response) WriteMsg(m *Msg) (err error) {
 
 // Write implements the ResponseWriter.Write method.
 func (w *response) Write(m []byte) (int, error) {
+	if w.closed {
+		panic("dns: Write called after Close")
+	}
+
 	switch {
 	case w.udp != nil:
 		n, err := WriteToSessionUDP(w.udp, m, w.udpSession)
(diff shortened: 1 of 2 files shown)


server.go
```diff
@@ -728,6 +729,10 @@ func (srv *Server) readUDP(conn *net.UDPConn, timeout time.Duration) ([]byte, *S
 
 // WriteMsg implements the ResponseWriter.WriteMsg method.
 func (w *response) WriteMsg(m *Msg) (err error) {
+	if w.closed {
+		panic("dns: WriteMsg called after Close")
```
miekg: not a superfan of all these panics

miekg: better IHMO to just return Err{err: "text"} on these

tmthrgd: Ack.

Issue #709, opened 2018-07-12
Add CachedLen()


miekg:
Think aloud here...

PR #708 tries to fit a message in a buffer; this is expensive because of compression when calling the Len() function on a message.

Len() in general is expansive because it goes over the entire message looking for compression points, so O(n) and it needs a map to store all that info, so it generates garbage as well.

What if we cached the length and map in Msg... If you are just adding RR(sets) we only need to count the new ones. This only works if you layer and message correctly; i.e answer first and authority and then a potential additional. Just would make Len() a lot cheaper. Of course it can be called `Len()` so we need something else, `CachedLen()`.

I think this can even work if you remove RRs, at least if you start at the end; the only difficulty there is know how long the RR will be because it may have been compressed (meaning you need to do some lookups). Removing from the ends works because compression pointers can only point forward.

tmthrgd:
In the context of #708 and after the merging of #833, `msgLenWithCompressionMap` is a really simple function so it should be possible to implement message truncation really efficiently without needing a `CachedLen` method or to store any state on `Msg`.

What would a good algorithm for truncating messages be? Is the one in [coredns](https://github.com/coredns/coredns/blob/bf67c9134c47dd1e0655a49b57a9113a6089ff48/request/request.go#L237) suitable? I'll have a go at implementing this if so.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Add CachedLen() (#7..." ]
>In the context of #708 and after the merging of #833, `msgLenWithCompressionMap` is a really simple function so it should be possible to implement message truncation really efficiently without needing a `CachedLen` method or to store any state on `Msg`.
>
>What would a good algorithm for truncating messages be? Is the one in [coredns](https://github.com/coredns/coredns/blob/bf67c9134c47dd1e0655a49b57a9113a6089ff48/request/request.go#L237) suitable? I'll have a go at implementing this if so.

Yes, that was the best I could come up with.
Also note: https://github.com/coredns/coredns/pull/2333/files


/Miek

--
Miek Gieben

tmthrgd:
I'm not sure how I can incorporate `SizeAndDo` because it relies on having access to the request message:

https://github.com/coredns/coredns/blob/bf67c9134c47dd1e0655a49b57a9113a6089ff48/request/request.go#L194-L196

tmthrgd:
Ignore that, I figured it out and I'll send a PR shortly.

miekg:
[...]

PR #820, opened 2018-11-26
Use map[string]struct{} for compression map in Len

tmthrgd:
`map[string]int` requires 8 bytes per entry to store the unused position information.

~I couldn't get any clean benchmarks from this, but `map[string]struct{}` definitely uses less memory than `map[string]int`.~

This does unfortunately make the related test cases slightly weaker as they no longer test that the `Len` code is finding compression points in the correct spot.

Updates #709.

Edit: I was able to get some better benchmarks that show a very notable improvement in memory use.

```
$ benchstat {old,new}.bench
name                       old time/op    new time/op    delta
MsgLength-12                  329ns ± 0%     328ns ± 0%   -0.27%  (p=0.040 n=9+9)
MsgLengthNoCompression-12    9.37ns ± 1%    9.34ns ± 2%     ~     (p=0.099 n=10+10)
MsgLengthPack-12             2.26µs ± 9%    2.36µs ±23%     ~     (p=0.780 n=9+10)
MsgLengthMassive-12          63.1µs ± 8%    53.0µs ± 3%  -15.89%  (p=0.000 n=10+10)

name                       old alloc/op   new alloc/op   delta
MsgLength-12                  32.0B ± 0%     32.0B ± 0%     ~     (all equal)
MsgLengthNoCompression-12     0.00B          0.00B          ~     (all equal)
MsgLengthPack-12               896B ± 0%      896B ± 0%     ~     (all equal)
MsgLengthMassive-12          20.2kB ± 0%    14.8kB ± 0%  -26.74%  (p=0.000 n=8+9)

name                       old allocs/op  new allocs/op  delta
MsgLength-12                   1.00 ± 0%      1.00 ± 0%     ~     (all equal)
MsgLengthNoCompression-12      0.00           0.00          ~     (all equal)
MsgLengthPack-12               8.00 ± 0%      8.00 ± 0%     ~     (all equal)
MsgLengthMassive-12             138 ± 0%       138 ± 0%     ~     (all equal)
```

### DIFF
--- a/compress_generate.go
+++ b/compress_generate.go
@@ -101,7 +101,7 @@ Names:
 
 	// compressionLenHelperType - all types that have domain-name/cdomain-name can be used for compressing names
 
-	fmt.Fprint(b, "func compressionLenHelperType(c map[string]int, r RR, initLen int) int {\n")
+	fmt.Fprint(b, "func compressionLenHelperType(c map[string]struct{}, r RR, initLen int) int {\n")
 	fmt.Fprint(b, "currentLen := initLen\n")
 	fmt.Fprint(b, "switch x := r.(type) {\n")
 	for _, name := range domainTypes {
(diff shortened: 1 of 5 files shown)


miekg:
I would have sworn this length was somehow needed, guess not.

miekg:
sorry, went to the wrong way through the PRs, created a merge conflict /o\

miekg:
lgtm

tmthrgd:
These were going to create a merge conflict regardless of ordering. I've manually rebased.

PR #818, opened 2018-11-26
Improve PackDomainName performance

tmthrgd:
`PackDomainName` will now only allocate exactly once<sup>†</sup> and only if the domain name is escaped.

~This PR also includes a few changes that are unrelated to `PackDomainName`, but follow the same pattern as changes to that function.~ Edit: I'm pulling these changes out into separate PRs.

This PR also fixes two bugs (5547fd63a0ca32a7aa0db0b0cff51fb96fa4df6c & 6aa05940d56cc6879e519112fcf68b9b486b1d6f) related to the compression map.

<details>
<summary>$ benchstat {old,new}.bench</summary>
<pre>
$ benchstat {old,new}.bench
name                            old time/op    new time/op    delta
MsgLength-12                       338ns ± 0%     294ns ± 1%   -13.15%  (p=0.000 n=10+10)
MsgLengthNoCompression-12         9.68ns ± 1%    9.60ns ± 0%    -0.83%  (p=0.006 n=10+9)
MsgLengthPack-12                  2.30µs ±22%    1.36µs ±25%   -41.01%  (p=0.000 n=10+10)
PackDomainName-12                  205ns ± 6%     117ns ± 1%   -42.71%  (p=0.000 n=10+9)
UnpackDomainName-12                166ns ±14%     156ns ± 8%    -6.02%  (p=0.030 n=9+9)
UnpackDomainNameUnprintable-12     154ns ± 5%     151ns ± 5%      ~     (p=0.126 n=10+9)
Copy-12                            872ns ±14%     798ns ±16%    -8.49%  (p=0.028 n=9+10)
PackA-12                          41.5ns ± 1%    39.3ns ± 3%    -5.33%  (p=0.000 n=9+10)
UnpackA-12                         230ns ±13%     237ns ±16%      ~     (p=0.411 n=9+10)
PackMX-12                         75.2ns ± 1%    67.2ns ± 1%   -10.67%  (p=0.000 n=10+9)
UnpackMX-12                        270ns ±12%     277ns ±20%      ~     (p=0.436 n=10+10)
PackAAAAA-12                      40.7ns ± 1%    38.7ns ± 2%    -5.07%  (p=0.000 n=9+10)
UnpackAAAA-12                      221ns ±13%     222ns ±16%      ~     (p=0.719 n=9+10)
PackMsg-12                        1.51µs ±15%    0.89µs ± 2%   -40.90%  (p=0.000 n=10+10)
[...]

PR #826, opened 2018-11-27
Use NextLabel in compressionLenHelper

tmthrgd:
This avoids the allocation of Split and should have slightly better performance.

This was pulled out of #818 and is reflected in the benchmarks there.


tmthrgd:
Hold off on this for now. I'm working on replacing `compressionLenHelper` entirely and this will only cause difficulty merging.

PR #828, opened 2018-11-27
Use strings.EqualFold rather than strings.ToLower

tmthrgd:
`strings.ToLower` may allocate and will do more work than `strings.EqualFold`.

### DIFF
--- a/dnssec.go
+++ b/dnssec.go
@@ -401,7 +401,7 @@ func (rr *RRSIG) Verify(k *DNSKEY, rrset []RR) error {
 	if rr.Algorithm != k.Algorithm {
 		return ErrKey
 	}
-	if strings.ToLower(rr.SignerName) != strings.ToLower(k.Hdr.Name) {
+	if !strings.EqualFold(rr.SignerName, k.Hdr.Name) {
 		return ErrKey
 	}
 	if k.Protocol != 3 {
--- a/sig0.go
+++ b/sig0.go
@@ -167,7 +167,7 @@ func (rr *SIG) Verify(k *KEY, buf []byte) error {
 	}
 	// If key has come from the DNS name compression might
 	// have mangled the case of the name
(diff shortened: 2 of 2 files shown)


miekg:
Doesn't equalfold assume utf8? Which isnt true for DNS. Do/should we care?

tmthrgd:
It does, but `strings.ToLower` does Unicode case-folding anyway.

>  ToLower returns a copy of the string s with all Unicode letters mapped to their lower case.

If we really want to stick to ASCII, we'd probably need to go around and replace all the calls to `strings.ToLower` and `strings.ToUpper` with our own ASCII-only implementations (in a separate PR).

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Use strings.EqualFo..." ]
>It does, but `strings.ToLower` does Unicode case-folding anyway.
>
>>  ToLower returns a copy of the string s with all Unicode letters mapped to their lower case.
>
>If we really want to stick to ASCII, we'd probably need to go around and replace all the calls to `strings.ToLower` and `strings.ToUpper` with our own ASCII-only implementations (in a separate PR).

I rather not go that route; I think we can assume ASCII input here, even IDN is 
converted back to ASCII; to EqualFold should be good. We might need some extra 
documentation to start tracking these decisions.

tmthrgd:
Shall I merge this as is? I don't think these changes would be appropriate places for documentation as there are a host of `strings.ToLower` and `strings.ToUpper` calls throughout that also have an ASCII-only requirement. Documenting this somewhere can be part of another PR/issue.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Use strings.EqualFo..." ]
>Shall I merge this as is? I don't think these changes would be appropriate places for documentation as there are a host of `strings.ToLower` and `strings.ToUpper` calls throughout that also have an ASCII-only requirement. Documenting this somewhere can be part of another PR/issue.

Yes, go ahead.

PR #830, opened 2018-11-27
Add MsgAcceptFunc

miekg:
Generalize the srv.Unsafe and make it pluggeable. Also add a default
pass function that allows to discard of a malformed DNS message very
early on.

Note this makes srv.Unsafe a noop - wondering if we should remove it out right as the next release will be breaking anyway.

### DIFF
--- a/acceptfunc.go
+++ b/acceptfunc.go
@@ -1,0 +1,46 @@
+package dns
+
+// MsgAcceptFunc is used early in the server code to accept or reject a message with RcodeFormatError.
+// There are to booleans to be returned, once signaling the rejection and another to signal if
+// a reply is to be send back (you want to prevent DNS ping-pong and not reply to a response for instance).
+type MsgAcceptFunc func(dh Header) (accept bool, respond bool)
+
+// DefaultMsgAcceptFunc checks the request and will reject if:
+//
+// * isn't a request (don't respond in that case).
+// * opcode isn't OpcodeQuery or OpcodeNotify
+// * Zero bit isn't zero
+// * has more than 1 question in the question section
+// * has more than 0 RRs in the Answer section
+// * has more than 0 RRs in the Authority section
+// * has more than 2 RRs in the Additional section
+var DefaultMsgAcceptFunc = defaultMsgAcceptFunc
+
+var defaultMsgAcceptFunc = func(dh Header) (bool, bool) {
+	if isResponse := dh.Bits&_QR != 0; isResponse {
+		return false, false
+	}
+
+	// Don't allow dynamic updates, because then the sections can contain a whole bunch of RRs.
+	opcode := int(dh.Bits>>11) & 0xF
+	if opcode != OpcodeQuery && opcode != OpcodeNotify {
+		return false, true
+	}
+
+	if isZero := dh.Bits&_Z != 0; isZero {
+		return false, true
+	}
+	if dh.Qdcount != 1 {
+		return false, true
+	}
+	if dh.Ancount != 0 {
+		return false, true
+	}
+	if dh.Nscount != 0 {
+		return false, true
+	}
+	if dh.Arcount > 2 {
+		return false, true
+	}
+	return true, true
+}
--- a/acceptfunc.go
+++ b/acceptfunc.go
@@ -3,7 +3,7 @@ package dns
 // MsgAcceptFunc is used early in the server code to accept or reject a message with RcodeFormatError.
 // There are to booleans to be returned, once signaling the rejection and another to signal if
 // a reply is to be send back (you want to prevent DNS ping-pong and not reply to a response for instance).
-type MsgAcceptFunc func(dh Header) (accept bool, respond bool)
+type MsgAcceptFunc func(dh Header) MsgAcceptAction
 
 // DefaultMsgAcceptFunc checks the request and will reject if:
 //
@@ -14,33 +14,42 @@ type MsgAcceptFunc func(dh Header) (accept bool, respond bool)
 // * has more than 0 RRs in the Answer section
 // * has more than 0 RRs in the Authority section
 // * has more than 2 RRs in the Additional section
-var DefaultMsgAcceptFunc = defaultMsgAcceptFunc
+var DefaultMsgAcceptFunc MsgAcceptFunc = defaultMsgAcceptFunc
 
-var defaultMsgAcceptFunc = func(dh Header) (bool, bool) {
+// MsgAcceptAction represents the action to be taken.
+type MsgAcceptAction int
+
+const (
+	MsgAccept MsgAcceptAction = iota
+	MsgReject
+	MsgIgnore
+)
+
+var defaultMsgAcceptFunc = func(dh Header) MsgAcceptAction {
 	if isResponse := dh.Bits&_QR != 0; isResponse {
-		return false, false
+		return MsgIgnore
 	}
 
 	// Don't allow dynamic updates, because then the sections can contain a whole bunch of RRs.
 	opcode := int(dh.Bits>>11) & 0xF
 	if opcode != OpcodeQuery && opcode != OpcodeNotify {
-		return false, true
+		return MsgReject
 	}
 
 	if isZero := dh.Bits&_Z != 0; isZero {
-		return false, true
+		return MsgReject
 	}
 	if dh.Qdcount != 1 {
-		return false, true
+		return MsgReject
 	}
 	if dh.Ancount != 0 {
-		return false, true
+		return MsgReject
 	}
 	if dh.Nscount != 0 {
-		return false, true
+		return MsgReject
 	}
 	if dh.Arcount > 2 {
-		return false, true
+		return MsgReject
 	}
-	return true, true
+	return MsgAccept
 }
--- a/acceptfunc.go
+++ b/acceptfunc.go
@@ -1,8 +1,7 @@
 package dns
 
 // MsgAcceptFunc is used early in the server code to accept or reject a message with RcodeFormatError.
-// There are to booleans to be returned, once signaling the rejection and another to signal if
-// a reply is to be send back (you want to prevent DNS ping-pong and not reply to a response for instance).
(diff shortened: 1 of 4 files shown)


passfunc.go
```diff
@@ -0,0 +1,38 @@
+package dns
+
+// PassFunc is used early in the server code to accept or reject a message. When rejecting the returned boolean is false.
+type PassFunc func(dh Header) (rcode int, ok bool)
```
tmthrgd: I can't say I really like the name. Though a better name doesn't immediately spring to mind.

passfunc.go
```diff
@@ -0,0 +1,38 @@
...
+// DefaultPassFunc returns false when the request:
+//
+// * isn't a request
+// * opcode isn't OpcodeQuery or OpcodeNotify
+// * Zero bit isn't zero
+// * has more than 1 question in the question section
+// * has more than 0 RRs in the Answer section
+// * has more than 0 RRs in the Authority section
+// * has more than 2 RRs in the Additional section
+var DefaultPassFunc = func(dh Header) (int, bool) {
```
tmthrgd: This should be:
```go
var DefaultPassFunc = defaultPassFunc

func defaultPassFunc(dh Header) (int, bool) {
	// ...
```

server.go
```diff
@@ -631,13 +634,36 @@ func (srv *Server) disposeBuffer(w *response) {
...
 func (srv *Server) serveDNS(w *response) {
 	req := new(Msg)
-	err := req.Unpack(w.msg)
-	if err != nil { // Send a FormatError back
-		x := new(Msg)
-		x.SetRcodeFormatError(req)
-		w.WriteMsg(x)
+	dh, off, err := unpackMsgHdr(w.msg, 0)
+	req.setHdr(dh)
+
```
tmthrgd: Nit: Lose this newline.

server.go
```diff
@@ -631,13 +634,36 @@ func (srv *Server) disposeBuffer(w *response) {
...
+	if err != nil {
+		// Let client hang, they are sending crap; any reply can be used to amplify.
+		return
+	}
+
+	// Response MUST be handled separately, because we can't reply to these messages, if we
+	// do we can ping-ping ourselves to death.
+	response := dh.Bits&_QR != 0
```
tmthrgd: `isResponse` would be more idiomatic go.

server_test.go
```diff
@@ -66,6 +66,7 @@ func RunLocalUDPServerWithFinChan(laddr string, opts ...func(*Server)) (*Server,
 		return nil, "", nil, err
 	}
 	server := &Server{PacketConn: pc, ReadTimeout: time.Hour, WriteTimeout: time.Hour}
+	server.init()
```
tmthrgd: `ActivateAndServe` already calls `Server.init` so these don't have to be explicitly added.

server.go
```diff
@@ -203,9 +203,6 @@ type Server struct {
 	IdleTimeout func() time.Duration
 	// Secret(s) for Tsig map[<zonename>]<base64 secret>. The zonename must be in canonical form (lowercase, fqdn, see RFC 4034 Section 6.2).
 	TsigSecret map[string]string
-	// Unsafe instructs the server to disregard any sanity checks and directly hand the message to
-	// the handler. It will specifically not check if the query has the QR bit not set.
-	Unsafe bool
```
tmthrgd: Is `Unsafe` widely used? What will the breakage of this be like?

server.go
```diff
@@ -631,13 +634,36 @@ func (srv *Server) disposeBuffer(w *response) {
...
+		return
+	}
+
+	// Response MUST be handled separately, because we can't reply to these messages, if we
+	// do we can ping-ping ourselves to death.
+	response := dh.Bits&_QR != 0
+	if response {
+		return
+	}
+
```
tmthrgd:
```go
req := new(Msg)
req.setHdr(dh)
```

should be moved right here. `req` isn't used before now.

tmthrgd:
@miekg Is there a specific use-case for this in mind?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Add PassFunc (#830)..." ]
>@miekg Is there a specific use-case for this in mind?

yes, we need to return early in the server, this avoids a DoS. Also needs to be 
pluggeable for people using `Unsafe` and dynamic updates and mdns, so yes def. 
see a need here.

server.go
miekg: good catch. done.

server_test.go
miekg: done

miekg:
As both the ErrTruncated and this will be backwards incompatible I like to do a `1.1.0` release (and fix up the CoreDNS side as well)

miekg:
Hmm. Unsafe allowed replying to responses... I'm not sure if that will ever
be valid except for mdns

miekg:
This will see some more changes. And I'll change the name to MsgAcceptFunc

miekg:
OK, done here. Note I've made the function mandatory, but not checking for `==nil`

acceptfunc.go
```diff
@@ -0,0 +1,46 @@
+package dns
+
+// MsgAcceptFunc is used early in the server code to accept or reject a message with RcodeFormatError.
+// There are to booleans to be returned, once signaling the rejection and another to signal if
+// a reply is to be send back (you want to prevent DNS ping-pong and not reply to a response for instance).
+type MsgAcceptFunc func(dh Header) (accept bool, respond bool)
```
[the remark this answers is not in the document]
miekg: done. Although I dropped Accept from each of these.

miekg:
I'll done. Only conflicts left *sigh*

msg.go
```diff
@@ -778,30 +778,8 @@ func (dns *Msg) packBufferWithCompressionMap(buf []byte, compression map[string]
...
-	dns.Truncated = dh.Bits&_TC != 0
-	dns.RecursionDesired = dh.Bits&_RD != 0
-	dns.RecursionAvailable = dh.Bits&_RA != 0
-	dns.Zero = dh.Bits&_Z != 0
-	dns.AuthenticatedData = dh.Bits&_AD != 0
-	dns.CheckingDisabled = dh.Bits&_CD != 0
-	dns.Rcode = int(dh.Bits & 0xF)
-
+func (dns *Msg) unpack(dh Header, msg []byte, off int) (err error) {
+	dns.setHdr(dh)
```
tmthrgd: `dns.setHdr(dh)` should be moved out of here and into `(*Msg).Unpack`. The code in `serveDNS` has already called `setHdr` at this point.

server.go
```diff
@@ -630,14 +633,34 @@ func (srv *Server) disposeBuffer(w *response) {
...
 	req := new(Msg)
-	err := req.Unpack(w.msg)
-	if err != nil { // Send a FormatError back
-		x := new(Msg)
-		x.SetRcodeFormatError(req)
-		w.WriteMsg(x)
+	req.setHdr(dh)
+
+	if action := srv.MsgAcceptFunc(dh); action != MsgAccept {
```
tmthrgd: `switch srv.MsgAcceptFunc(dh)`?

server_test.go
```diff
@@ -102,6 +102,7 @@ func RunLocalTCPServerWithFinChan(laddr string) (*Server, string, chan error, er
 	}
 
 	server := &Server{Listener: l, ReadTimeout: time.Hour, WriteTimeout: time.Hour}
+	server.init()
```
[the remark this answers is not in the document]
tmthrgd: @miekg You missed this.

msg.go
miekg: Not sure what you mean here. Msg.Unpack can be called independently from serveDNS

server.go
```diff
@@ -630,14 +633,36 @@ func (srv *Server) disposeBuffer(w *response) {
...
-	err := req.Unpack(w.msg)
-	if err != nil { // Send a FormatError back
-		x := new(Msg)
-		x.SetRcodeFormatError(req)
-		w.WriteMsg(x)
+	req.setHdr(dh)
+
+	switch srv.MsgAcceptFunc(dh) {
+	case MsgAccept:
+
```
tmthrgd: Nit: These newlines between switch cases is inconsistent with elsewhere in the library and idiomatic go.

msg.go
tmthrgd: `serveDNS` calls `unpackMsgHdr`, then `setHdr` and then `unpack`. `unpack` then calls `setHdr` again.

Unpack should instead be:
```go
func (dns *Msg) Unpack(msg []byte) (err error) {
	dh, off, err := unpackMsgHdr(msg, 0)
	if err != nil {
		return err
	}

	dns.setHdr(dh)
	return dns.unpack(dh, msg, off)
}
```

miekg:
ok .ptal

tmthrgd:
(@miekg Is there a reason you've been merging with a merge commit rather than squashing and merging? It's made the git history slightly messier).

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Add MsgAcceptFunc (..." ]
>(@miekg Is there a reason you've been merging with a merge commit rather than squashing and merging? It's made the git history slightly messier).

uh no, should be squash.

/Miek

--
Miek Gieben

server.go
bboreham: WeaveDNS was reimplemented and that code removed in June 2015.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Add MsgAcceptFunc (..." ]
>> @@ -203,9 +203,6 @@ type Server struct {
> 	IdleTimeout func() time.Duration
> 	// Secret(s) for Tsig map[<zonename>]<base64 secret>. The zonename must be in canonical form (lowercase, fqdn, see RFC 4034 Section 6.2).
> 	TsigSecret map[string]string
>-	// Unsafe instructs the server to disregard any sanity checks and directly hand the message to
>-	// the handler. It will specifically not check if the query has the QR bit not set.
>-	Unsafe bool
>
>WeaveDNS was reimplemented and that code removed in June 2015.

lol. Thanks for the ping back.

PR #835, opened 2018-11-28
Fix "too many compression points" for valid message

### LINKED ISSUE #834: "too many compression pointers"
The following test case fails:
```go
func TestPackUnpackManyCompressionPointers(t *testing.T) {
	m := new(Msg)
	m.Compress = true
	m.SetQuestion("example.org.", TypeNS)

	for domain := "a."; len(domain) < maxDomainNameWireOctets; domain += "a." {
		m.Answer = append(m.Answer, &NS{Hdr: RR_Header{Name: domain, Rrtype: TypeNS, Class: ClassINET}, Ns: "example.org."})

		b, err := m.Pack()
		if err != nil {
			t.Fatalf("Pack failed for %q and %d records with: %v", domain, len(m.Answer), err)
		}

		var m2 Msg
		if err := m2.Unpack(b); err != nil {
			t.Fatalf("Unpack failed for %q and %d records with: %v", domain, len(m.Answer), err)
		}
	}
}
```

with:
```
--- FAIL: TestPackUnpackManyCompressionPointers (0.00s)
    msg_test.go:276: Unpack failed for "a.a.a.a.a.a.a.a.a.a.a.a." and 12 records with: dns: too many compression pointers
```

We should be able to unpack every message we pack IMHO, and this is a semantically valid message. The problem is that each label (intentionally) compresses into the previous one forming a long chain of compressed labels. This trips the following arbitrary limit in `UnpackDomainName`:

https://github.com/miekg/dns/blob/7ae05cdcf8241aef158afa93d26863126796a302/msg.go#L402-L404

This limit is necessary to prevent infinite loops as `UnpackDomainName` doesn't restrict pointers to earlier offsets in the message. There are two solutions I can see here, one is to accept any compression pointer provided it heads backwards. The other is to raise the limit to the maximum any valid name could be. (`maxDomainNameWireOctets / 2`?)

tmthrgd:
The comment should explain my rational for sizing the constant. The added test case now passes.

Fixes #834

### DIFF
--- a/msg.go
+++ b/msg.go
@@ -24,6 +24,18 @@ import (
 const (
 	maxCompressionOffset    = 2 << 13 // We have 14 bits for the compression pointer
 	maxDomainNameWireOctets = 255     // See RFC 1035 section 2.3.4
+
+	// This is the maximum number of compression pointers that should occur in a
+	// semantically valid message. Each label in a domain name must be at least one
+	// octet and is separated by a period. The root label won't be represented by a
+	// compression pointer to a compression pointer, hence the -2 to exclude the
+	// smallest valid root label.
+	//
+	// It is possible to construct a valid message that has more compression pointers
+	// than this, and still doesn't loop, by pointing to a previous pointer. This is
+	// not something a well written implementation should ever do, so we ignore that
+	// possibility.
+	maxCompressionPointers = (maxDomainNameWireOctets+1)/2 - 2
 )
 
 // Errors defined in this package.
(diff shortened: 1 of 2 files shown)


tmthrgd:
We should perhaps hoist the `maxLen` check higher up in `UnpackDomainName` because it may be possible to cause a very large allocation before it fails later on.

msg.go
```diff
@@ -24,6 +24,18 @@ import (
...
+	// This is the maximum number of compression pointers that should occur in a
+	// semantically valid message. Each label in a domain name must be at least one
+	// octet and is separated by a period. The root label won't be represented by a
+	// compression pointer to a compression pointer, hence the -2 to exclude the
+	// smallest valid root label.
+	//
+	// It is possible to construct a valid message that has more compression pointers
+	// than this, and still doesn't loop, by pointing to a previous pointer. This is
+	// not something a well written implementation should ever do, so we ignore that
+	// possibility.
```
miekg: but this can be constructed and put on the wire, no? Are we discarding those messages. I.e. from this text it looks like we're introducing a dos?

tmthrgd: Is that clearer now? I can't figure out the best wording for it.

PR #839, opened 2018-11-28
Bail early from UnpackDomainName when name is too long

### LINKED ISSUE #836: UnpackDomainName DOS potential, large allocations
The following code shows that `UnpackDomainName` will allocate a lot of memory when unpacking potentially malicious domain names.

```go
func TestNameCompressionAllocations(t *testing.T) {
	name := []byte("\x3Faaabbbcccdddeeefffggghhhiiijjjkkklllmmmnnnooopppqqqrrrssstttuuu\xC0\x00")

	allocs, mem := allocsPerRun(10, func() {
		_, _, err := UnpackDomainName(name, 0)
		if e, ok := err.(*Error); err != ErrLongDomain && (!ok || e.err != "too many compression pointers") {
			t.Fatalf("UnpackDomainName failed with: %v", err)
		}
	})
	t.Error(allocs, mem)
}

// allocsPerRun is like testing.AllocsPerRun but also reports
// the heap allocations made in bytes.
func allocsPerRun(runs int, f func()) (allocs, total float64) {
	defer runtime.GOMAXPROCS(runtime.GOMAXPROCS(1))

	// Warm up the function
	f()

	// Measure the starting statistics
	var memstats runtime.MemStats
	runtime.ReadMemStats(&memstats)
	mallocs := 0 - memstats.Mallocs
	talloc := 0 - memstats.TotalAlloc

	// Run the function the specified number of times
	for i := 0; i < runs; i++ {
		f()
	}

	// Read the final statistics
	runtime.ReadMemStats(&memstats)
	mallocs += memstats.Mallocs
	talloc += memstats.TotalAlloc

	// Average the mallocs over the runs (not counting the warm-up).
	// We are forced to return a float64 because the API is silly, but do
	// the division as integers so we can ask if AllocsPerRun()==1
	// instead of AllocsPerRun()<2.
	return float64(mallocs / uint64(runs)), float64(talloc / uint64(runs))
}
```

For the current maximum number of compression points, 10, this generates 1936B (~1.8KiB) of garbage before returning an error. After the proposed raising in #835, this generates 36112B (~35.3KiB) of garbage before returning an error.

This could be fixed by checking the maximum allowed length after each label has been unpacked. That may subtle change the semantics of `UnpackDomainName` though. As this will only principally effect the compression pointer case, which already returns an empty string instead of the name, perhaps we could just special case it there.

/cc @miekg

tmthrgd:
This PR drastically reduces the amount of garbage created in `UnpackDomainName` for certain malicious names.

The wire formatted name
```go
"\x3Faaabbbcccdddeeefffggghhhiiijjjkkklllmmmnnnooopppqqqrrrssstttuuu\xC0\x00"
```
would previously generate 1936B of garbage before #835, and 36112B of garbage since #835, before returning the `"too many compression pointers"` error, while after this PR it will generate just 384B of garbage before returning the error.

Related #835
Fixes #836

### DIFF
--- a/msg.go
+++ b/msg.go
@@ -371,7 +371,7 @@ Loop:
 			if off+c > lenmsg {
 				return "", lenmsg, ErrBuf
 			}
-			budget -= c + 1 // +1 for the trailing period
+			budget -= c + 1 // +1 for the label separator
 			if budget <= 0 {
 				return "", lenmsg, ErrLongDomain
 			}
--- a/msg.go
+++ b/msg.go
@@ -423,11 +423,11 @@ Loop:
 	if ptr == 0 {
 		off1 = off
 	}
-	if len(s) == 0 {
-		s = []byte(".")
-	}
 	if budget <= 0 || budget > maxDomainNameWireOctets { // handle overflow
 		return "", lenmsg, ErrLongDomain
+	}
+	if len(s) == 0 {
+		return ".", off1, nil
 	}
 	return string(s), off1, nil
 }
--- a/msg.go
+++ b/msg.go
@@ -352,7 +352,7 @@ func UnpackDomainName(msg []byte, off int) (string, int, error) {
 	s := make([]byte, 0, 64)
 	off1 := 0
 	lenmsg := len(msg)
-	maxLen := maxDomainNameWireOctets
+	budget := maxDomainNameWireOctets
 	ptr := 0 // number of pointers followed
 Loop:
 	for {
@@ -371,14 +371,13 @@ Loop:
 			if off+c > lenmsg {
 				return "", lenmsg, ErrBuf
 			}
+			budget -= c + 1 // +1 for the trailing period
 			for j := off; j < off+c; j++ {
 				switch b := msg[j]; b {
 				case '.', '(', ')', ';', ' ', '@':
 					fallthrough
 				case '"', '\\':
 					s = append(s, '\\', b)
-					// presentation-format \X escapes add an extra byte
-					maxLen++
 				default:
 					if b < 32 || b >= 127 { // unprintable, use \DDD
 						var buf [3]byte
@@ -388,8 +387,6 @@ Loop:
 							s = append(s, '0')
 						}
 						s = append(s, bufs...)
-						// presentation-format \DDD escapes add 3 extra bytes
-						maxLen += 3
 					} else {
 						s = append(s, b)
 					}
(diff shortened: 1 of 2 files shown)


msg.go
```diff
@@ -427,10 +427,7 @@ Loop:
 		off1 = off
 	}
 	if len(s) == 0 {
-		s = []byte(".")
-	} else if len(s) >= maxLen {
-		// error if the name is too long, but don't throw it away
-		return string(s), lenmsg, ErrLongDomain
+		return ".", off1, nil
 	}
 	return string(s), off1, nil
```
semihalev:
```suggestion
	return *(*string)(unsafe.Pointer(&s)), off1, nil
```

another string cast costly, double allocate here, with this change can stop another allocation. This function from `strings.Builder` and discussed here [https://github.com/golang/go/issues/25484#issuecomment-391415660](url)

miekg: there is probably lower hanging fruit (elsewhere) than to micro optimize these allocations. This is borderline unreadable...

tmthrgd: @semihalev That's out of scope for this PR. I actually suspect that change would make things worse in this function because I think `s := make([]byte, 0, 64)` is being stack allocated in most cases. Whereas if you were to make that change, it would cause `s` to escape and be heap allocated.

I've tried switching this (`UnpackDomainName`) over to `strings.Builder` several times now, but always seem to end up with worse performance. (You can see 5debfeec63c371d5806bb64f4553078717b2eef5 & 0d29b283ac0f967dd3a02739bf26a22702210d7a where I've switched other parts of the code to use `strings.Builder`).

msg.go
```diff
@@ -427,10 +427,7 @@ Loop:
 		off1 = off
 	}
 	if len(s) == 0 {
-		s = []byte(".")
-	} else if len(s) >= maxLen {
-		// error if the name is too long, but don't throw it away
-		return string(s), lenmsg, ErrLongDomain
```
miekg: maybe update the doc comment about this behavior?

tmthrgd: @miekg Would you prefer a comment on `UnpackDomainName` or where the `if budget <= 0` return is?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Bail early from Unp..." ]
>
>
>
>> @@ -427,10 +427,7 @@ Loop:
> 		off1 = off
> 	}
> 	if len(s) == 0 {
>-		s = []byte(".")
>-	} else if len(s) >= maxLen {
>-		// error if the name is too long, but don't throw it away
>-		return string(s), lenmsg, ErrLongDomain
>
>@miekg Would you prefer a comment on `UnpackDomainName` or where the `if budget <= 0` return is?

function level doc about this edge case. Current doc is rather sparse.

msg.go
tmthrgd: How is that?

Issue #841, opened 2018-11-28
Escaped names compression length mismatch


tmthrgd:
The following test case fails due to differences in escaping between `PackDomainName` and `Len`:
```go
func TestMsgCompressionEscapingCompressionMap(t *testing.T) {
	msg := new(Msg)
	msg.Compress = true
	msg.SetQuestion("www.example.org.", TypeA)
	msg.Answer = append(msg.Answer, &NS{Hdr: RR_Header{Name: "ex\\097mple.org.", Rrtype: TypeNS, Class: ClassINET}, Ns: "ns.example.org."})

	predicted := msg.Len()
	buf, err := msg.Pack()
	if err != nil {
		t.Error(err)
	}
	if predicted != len(buf) {
		t.Fatalf("predicted compressed length is wrong: predicted %d, actual %d", predicted, len(buf))
	}
}
```

with:
```
--- FAIL: TestMsgCompressionEscapingCompressionMap (0.00s)
    length_test.go:398: predicted compressed length is wrong: predicted 61, actual 50
```

The problem is that `\\097` is a perfectly valid escaping of `a`, `PackDomainName` handles this by putting the unescaped name (only the first label is unescaped) into the compression map, while the `Len` code puts the escaped name in the compression map. (This isn't effected by #833).

I'm not sure what the best way to fix this is, fixing this in `Len` would (I think) require extra allocations, and fixing this in `PackDomainName` is trickier than it might seem.

This is one of the corner cases I mentioned in https://github.com/miekg/dns/issues/652#issuecomment-442426951.

miekg:
[ Quoting <notifications@github.com> in "[miekg/dns] Escaped names compressi..." ]
>This is one of the corner cases I mentioned in https://github.com/miekg/dns/issues/652#issuecomment-442426951.

Ack.

I'm all for normalizing this and not making this horrible (internal type 
that we convert from and to?). It's clear that the current impl. is not the best 
one and we can keep on finding non-working corner cases.

Which may be harder than it sounds...

PR #833, opened 2018-11-27
Properly calculate compressed message lengths

tmthrgd:
This entirely replaces the rather hacky and very confusing dance done in `compressionLenSlice`, with a new approach that is far more like how packing is performed. It fixes a number of different bugs (some listed at the end).

The new code has two allocations that aren't actually needed, but are hard to avoid. (`MsgLength` & `MsgLengthMassive` should have 0 and 9 allocations respectively). The problem is that `compression map[string]struct{}` is passed to `RR.len` where `RR` is an interface. This interacts badly with Golang's escape analysis (see golang/go#19361 and <https://npat-efault.github.io/programming/2016/10/10/escape-analysis-and-interfaces.html>), forcing the map to be heap allocated instead of stack allocated. It is possible to force these to be devirtualised by using a massive type switch statement, this worsens performance overall for `MsgLengthMassive` while eliminating the loss for `MsgLength`.

The performance loss in `MsgLengthNoCompression` & `MsgLengthOnlyQuestion` is (I believe) due to the call to `domainNameLen` which can't be inlined currently. Though this is unfortunate, the overhead is necessary as it fixes a bug where `Len` could previously be off by one for uncompressed messages. (See the existing `TestMsgLength2` that is now fixed). Inlining changes are coming to go1.12 so this can be revisited then and it may be possible to recover this performance loss.

Some of the benchmarks below (like `LenLabels` and `MuxMatch`) are just noise (although I don't quite understand how they survived through benchstat and ten runs), but they could come down to the compiler laying out code differently and cache interactions. TLDR; they're a distraction.

When taken along with #820, `MsgLengthMassive` is now twice as fast, with half the allocated bytes and 92% fewer allocations.

<details>
<summary>$ benchstat {old,new}.bench</summary>
<pre>
$ benchstat {old,new}.bench
name                            old time/op    new time/op    delta
MsgLength-12                       336ns ± 0%     545ns ± 6%   +62.24%  (p=0.001 n=6+8)
MsgLengthNoCompression-12         9.95ns ± 3%   26.64ns ± 2%  +167.79%  (p=0.000 n=10+10)
MsgLengthPack-12                  2.20µs ±15%    2.30µs ±23%      ~     (p=0.315 n=10+10)
MsgLengthMassive-12               54.0µs ± 5%    31.8µs ± 6%   -41.04%  (p=0.000 n=10+9)
[...]

PR #842, opened 2018-11-28
Put escaped names into compression map

tmthrgd:
After this PR, the escaped form of the name will be inserted into the compression map rather than the partially unescaped form. This is slightly non-optimal as certain names won't compress even though they're identical, this could be improved in future, but it's not a bug as compression is purely optional.

This only partially addresses #841, the rest of it will be fixed in a follow up PR once #833 has been merged.

Updates #841


tmthrgd:
(@miekg You did a regular merge, not squash, here too).

PR #844, opened 2018-11-29
Reduce allocations in UnpackDomainName by better sizing slice

tmthrgd:
This simple change greatly improves the performance and memory consumption of `UnpackDomainName`. It's perfectly safe to size `s` this large as it never escapes and will be stack allocated.

After this PR, `UnpackDomainName` will never be the source of memory garbage.

In terms of number of allocations and allocated memory, this is absolutely the best possible results, *i.e.* `UnpackDomainName` can never be more efficient with regards to memory usage.

```
name                                   old time/op    new time/op    delta
UnpackDomainName-12                       169ns ±13%     167ns ±19%     ~     (p=0.955 n=10+10)
UnpackDomainNameUnprintable-12            150ns ±13%     160ns ± 6%   +6.61%  (p=0.015 n=10+10)
UnpackDomainNameLongest-12               1.45µs ±30%    0.74µs ± 7%  -48.74%  (p=0.000 n=10+9)
UnpackDomainNameLongestUnprintable-12    6.05µs ±16%    3.21µs ± 7%  -47.00%  (p=0.000 n=10+10)

name                                   old alloc/op   new alloc/op   delta
UnpackDomainName-12                       64.0B ± 0%     64.0B ± 0%     ~     (all equal)
UnpackDomainNameUnprintable-12            48.0B ± 0%     48.0B ± 0%     ~     (all equal)
UnpackDomainNameLongest-12                 640B ± 0%      256B ± 0%  -60.00%  (p=0.000 n=10+10)
UnpackDomainNameLongestUnprintable-12    2.94kB ± 0%    1.02kB ± 0%  -65.22%  (p=0.000 n=10+10)

name                                   old allocs/op  new allocs/op  delta
UnpackDomainName-12                        1.00 ± 0%      1.00 ± 0%     ~     (all equal)
UnpackDomainNameUnprintable-12             1.00 ± 0%      1.00 ± 0%     ~     (all equal)
UnpackDomainNameLongest-12                 3.00 ± 0%      1.00 ± 0%  -66.67%  (p=0.000 n=10+10)
UnpackDomainNameLongestUnprintable-12      5.00 ± 0%      1.00 ± 0%  -80.00%  (p=0.000 n=10+10)
```

/cc @semihalev (for https://github.com/miekg/dns/issues/836#issuecomment-442413416 & https://github.com/miekg/dns/pull/839#discussion_r237073104)

### DIFF
--- a/msg.go
+++ b/msg.go
@@ -36,6 +36,9 @@ const (
 	// not something a well written implementation should ever do, so we leave them
 	// to trip the maximum compression pointer check.
 	maxCompressionPointers = (maxDomainNameWireOctets+1)/2 - 2
+
+	// This is the maximum length of a domain name in presentation format.
+	maxDomainNamePresentationLength = 61*4 + 1 + 63*4 + 1 + 63*4 + 1 + 63*4 + 1
 )
 
 // Errors defined in this package.
@@ -372,7 +375,7 @@ func isRootLabel(s string, bs []byte, off, end int) bool {
 // When an error is encountered, the unpacked name will be discarded
 // and len(msg) will be returned as the offset.
 func UnpackDomainName(msg []byte, off int) (string, int, error) {
-	s := make([]byte, 0, 64)
+	s := make([]byte, 0, maxDomainNamePresentationLength)
 	off1 := 0
 	lenmsg := len(msg)
 	budget := maxDomainNameWireOctets
--- a/msg.go
+++ b/msg.go
@@ -37,7 +37,12 @@ const (
 	// to trip the maximum compression pointer check.
 	maxCompressionPointers = (maxDomainNameWireOctets+1)/2 - 2
 
-	// This is the maximum length of a domain name in presentation format.
+	// This is the maximum length of a domain name in presentation format. The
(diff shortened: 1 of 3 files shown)


msg.go
```diff
@@ -36,6 +36,9 @@ const (
 	// not something a well written implementation should ever do, so we leave them
 	// to trip the maximum compression pointer check.
 	maxCompressionPointers = (maxDomainNameWireOctets+1)/2 - 2
+
+	// This is the maximum length of a domain name in presentation format.
+	maxDomainNamePresentationLength = 61*4 + 1 + 63*4 + 1 + 63*4 + 1 + 63*4 + 1
```
miekg: could you explain the calculation in the comment? I can reason back with the 63 (a bit), but not reconstruct it myself.

tmthrgd: Ack, this does need a better explanation. I was working off mostly https://blogs.msdn.microsoft.com/oldnewthing/20120412-00/?p=7873/ because it's actually harder to figure out than it would seem.

tmthrgd: Done. I think that should be clear enough.

miekg:
thanks, looks good.

PR #846, opened 2018-11-29
Use a table lookup for escaping unprintable bytes

tmthrgd:
This PR replaces the `strconv.AppendInt` based unprintable octet escaping with pre-calculated table-lookups. This is also much simpler. It also eliminates the duplication between `writeEscapedByte` and `UnpackDomainName`.

I know I just landed #845, but I didn't realise just what an improvement could be had. `strconv.AppendInt` has a table-lookup optimisation for numbers `< 100`, which is all `UnpackDomainNameLongestUnprintable` has, so for names that have unprintable octets `> 126` this should be an even greater performance improvement.

These benchmarks were taken after #844 and show a very big improvement (memory is the same):
```
name                                   old time/op    new time/op    delta
UnpackDomainName-12                       125ns ± 5%     126ns ± 7%     ~     (p=0.807 n=10+10)
UnpackDomainNameUnprintable-12            138ns ± 2%     102ns ± 5%  -26.24%  (p=0.000 n=10+10)
UnpackDomainNameLongest-12                576ns ± 5%     522ns ± 6%   -9.31%  (p=0.000 n=10+10)
UnpackDomainNameLongestUnprintable-12    2.70µs ± 2%    1.40µs ±10%  -48.09%  (p=0.000 n=10+10)
SprintName-12                             220ns ± 1%     180ns ± 1%  -18.04%  (p=0.000 n=10+10)
SprintTxtOctet-12                         183ns ± 5%     136ns ± 9%  -25.51%  (p=0.000 n=10+10)
SprintTxt-12                              245ns ± 1%     203ns ± 2%  -17.07%  (p=0.000 n=10+10)
```

**Updated**: Here are the benchmark numbers if `maxUnprintableLabel` uses unprintable characters `> 126` and takes the slower path in `strconv.AppendInt`, it shows even greater performance improvements (`UnpackDomainNameLongest` included for reference):
```
name                                   old time/op    new time/op    delta
UnpackDomainNameLongest-12                563ns ± 5%     533ns ± 6%   -5.29%  (p=0.005 n=10+10)
UnpackDomainNameLongestUnprintable-12    4.50µs ± 0%    1.58µs ± 7%  -64.82%  (p=0.000 n=8+10)
```

### DIFF
--- a/msg.go
+++ b/msg.go
@@ -406,10 +406,7 @@ Loop:
 					s = append(s, '\\', b)
 				default:
 					if b < ' ' || b > '~' { // unprintable, use \DDD
-						var buf [3]byte
-						bufs := strconv.AppendInt(buf[:0], int64(b), 10)
-						s = append(s, '\\', '0', '0', '0')
-						copy(s[len(s)-len(bufs):], bufs)
+						s = append(s, escapeByte(b)...)
 					} else {
 						s = append(s, b)
 					}
--- a/msg_helpers.go
+++ b/msg_helpers.go
@@ -275,7 +275,7 @@ func unpackString(msg []byte, off int) (string, int, error) {
 			s.WriteByte('\\')
 			s.WriteByte(b)
 		case b < ' ' || b > '~': // unprintable
-			writeEscapedByte(&s, b)
+			s.WriteString(escapeByte(b))
 		default:
 			s.WriteByte(b)
 		}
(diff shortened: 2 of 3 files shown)


tmthrgd:
#844, #845 and this PR all taken together provide a `-74.90%` time/op delta for `UnpackDomainNameLongestUnprintable`. That is, all three of them together improve the performance of that benchmark by nearly 75%.

miekg:
> hat is, all three of them together improve the performance of that benchmark by nearly 75%.

holy cow

Issue #522, opened 2017-09-20
Return error when encountering a disallowed compression pointer


gibson042:
Per [RFC 3597](https://tools.ietf.org/html/rfc3597#section-4):
> Receiving servers MUST decompress domain names in RRs of well-known type, and SHOULD also decompress RRs of type RP, AFSDB, RT, SIG, PX, NXT, NAPTR, and SRV... Future specifications for new RR types that contain domain names within their RDATA MUST NOT allow the use of name compression for those names

Therefore, `unpack` should return an error for any record containing a compression pointer unless it is one of the following types:
* NS
* SOA
* CNAME
* PTR
* MB
* MD
* MF
* MG
* MINFO
* MR
* MX
* RP
* AFSDB
* RT
* SIG
* PX
* NXT
* NAPTR
* SRV

miekg:
Think there is even a comment somewhere saying we're too lenient (compress to many types). Note that we still need to deal with these types coming in compressed

tmthrgd:
RFC 3597 doesn't actually prohibit decompressing names outside that list.

The key paragraph is the following:
```
   Receiving servers MUST decompress domain names in RRs of well-known
   type, and SHOULD also decompress RRs of type RP, AFSDB, RT, SIG, PX,
   NXT, NAPTR, and SRV (although the current specification of the SRV RR
   in [RFC2782] prohibits compression, [RFC2052] mandated it, and some
   servers following that earlier specification are still in use).
```

Note decompressing well-known types is a MUST, and decompressing the above list is a SHOULD. It makes no mention of what to do when encountering compression pointers in other types. This is implementation specific and it makes equal sense to accept or prohibit them IMHO.

The change required to implement this is very simple if @miekg wants that approach.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Return error when e..." ]
>RFC 3597 doesn't actually prohibit decompressing names outside that list.
>
>The key paragraph is the following:
>```
>   Receiving servers MUST decompress domain names in RRs of well-known
>   type, and SHOULD also decompress RRs of type RP, AFSDB, RT, SIG, PX,
>   NXT, NAPTR, and SRV (although the current specification of the SRV RR
>   in [RFC2782] prohibits compression, [RFC2052] mandated it, and some
>   servers following that earlier specification are still in use).
>```
>
[...]

PR #847, opened 2018-11-30
Stop compressing names in RT records

tmthrgd:
Although [RFC 1183](https://tools.ietf.org/html/rfc1183#section-3.3) allows names in the RT record to be compressed with:
```
   The concrete encoding is identical to the MX RR.
```

But [RFC 3597](https://tools.ietf.org/html/rfc3597#section-4) specifically prohibits compressing names in any record not defined in [RFC 1035](https://tools.ietf.org/html/rfc1035#section-3.2).

**Note:** I had to modify the two generated files by hand as they weren't picking up my changes to types.go for some reason. There shouldn't be any changes running `go generate` later though.

Updates #522

### DIFF
--- a/types.go
+++ b/types.go
@@ -353,7 +353,7 @@ func (rr *X25) String() string {
 type RT struct {
 	Hdr        RR_Header
 	Preference uint16
-	Host       string `dns:"domain-name"`
+	Host       string `dns:"domain-name"` // RFC 3597 prohibits compressing records not defined in RFC 1035.
 }
 
 func (rr *RT) String() string {
--- a/types.go
+++ b/types.go
@@ -353,7 +353,7 @@ func (rr *X25) String() string {
 type RT struct {
 	Hdr        RR_Header
 	Preference uint16
-	Host       string `dns:"cdomain-name"`
+	Host       string `dns:"domain-name"`
 }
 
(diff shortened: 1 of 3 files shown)


types.go
```diff
@@ -353,7 +353,7 @@ func (rr *X25) String() string {
 type RT struct {
 	Hdr        RR_Header
 	Preference uint16
-	Host       string `dns:"cdomain-name"`
+	Host       string `dns:"domain-name"`
```
tmthrgd: This probably should be documented either on this field or on the RR.

Something like:
```
// RFC 3597 prohibits compressing records not defined in RFC 1035.
```

tmthrgd: @miekg Thoughts/

miekg:
you need to 'go install' at the right moment, because the generating code looks there for the new definition instead of the current dir with the new types.go

types.go
miekg: Yeah, just leaving the comment on the Host field is fine.

tmthrgd:
@miekg Well that worked, I had no idea, how odd.

types.go
tmthrgd: @miekg Done.

miekg:
Lgtm

PR #848, opened 2018-11-30
Prohibit compression pointers outside of RFC 3597

### LINKED ISSUE #522: Return error when encountering a disallowed compression pointer

tmthrgd:
I'm not sure if this is the approach we should take, but @gibson042 is quite adamant we should. I'm putting this up here for discussion.

**Note:** This still needs tests.

Fixes #522


tmthrgd:
@gibson042 Yep, compressed names in record headers will still be decompressed.

miekg:
so instead DoS if you receive a message with such a record and pointer?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Prohibit compressio..." ]
>@miekg I'm not sure I follow. Only an implementation that's violating the spec would ever trip this check and it won't have any effect on other messages. It's not any different than receiving a message with the reserved `0x80` or `0x40` in it, or that's too long.

but I can create such a message with ease and put on the wire, no?

tmthrgd:
I'm not sure I follow, creating a message with compressed names in any of the types prohibited in this PR is an explicit spec violation. It's by definition a malformed message. Maybe a broken implementation might do it, but they might do a lot of things.

tmthrgd:
This PR just returns an error like for any other malformed name, there is no panic anywhere.

Oh wait I see where you're confused, the panic is in `msg_generate.go` and is purely a sanity check to make sure we don't ever start accidentally compressing names we're not allowed to. It's there solely to catch someone adding a <code>\`dns:"cdomain-name"\`</code> tag to a record that isn't permitted to be compressed.

This definitely seems like a judgement call to me, but see https://github.com/miekg/dns/issues/522#issuecomment-443067656. I tend to agree that there could be very subtle security implications that are not yet clear. IMHO, this change will not break a single correct client/server and is simple enough that it's worth inclusion.

miekg:
On Sat, 1 Dec 2018, 10:39 Tom Thorogood <notifications@github.com wrote:

> This PR just returns an error like for any other malformed name, there is
> no panic anywhere.
>
> Oh wait I see where you're confused, the panic is in msg_generate.go and
> is purely a sanity check to make sure we don't ever start accidentally
> compressing names we're not allowed to. It's there solely to catch someone
> adding a `dns:"cdomain-name"` tag to a record that isn't permitted to be
> compressed.
>
> Ah sorry missed that.

This definitely seems like a judgement call to me, but see #522 (comment)

miekg:
Sorry, DNS screwed this up with "be liberal  what you accept". These
messages are trivially accepted and it's clear how to do validation with
DNSSEC, so I don't share those concerns.

tmthrgd:
Ack, will close. "Be liberal with what you accept," has caused far more problems than it's worth.

PR #854, opened 2018-12-01
Add a message truncation implementation

### LINKED ISSUE #709: Add CachedLen()

tmthrgd:
This PR adds a message truncation implementation. It aims to be performant and only walks the message exactly once.

This probably needs better tests. The current ones were taken from CoreDNS. (@miekg is that license compatible with miekg/dns? I presume so).

`BenchmarkMsgTruncate` is this PR and `BenchmarkRequestScrub` is the [`Scrub` method from CoreDNS](https://github.com/coredns/coredns/blob/bf67c9134c47dd1e0655a49b57a9113a6089ff48/request/request.go#L237):
```
BenchmarkMsgTruncate-12           200000              9230 ns/op            2462 B/op          7 allocs/op
BenchmarkRequestScrub-12           30000             52770 ns/op           16783 B/op         44 allocs/op
```

Closes #708
Fixes #709

/cc @grobie (for #708)

### DIFF
--- a/msg_truncate.go
+++ b/msg_truncate.go
@@ -51,7 +51,7 @@ func (dns *Msg) Truncate(size int) {
 
 	compression := make(map[string]struct{})
 
-	l = 12 // Message header is always 12 bytes
+	l = headerSize
 	for _, r := range dns.Question {
 		l += r.len(l, compression)
 	}
@@ -72,7 +72,7 @@ func (dns *Msg) Truncate(size int) {
 	}
 
 	// According to RFC 2181, the TC bit should only be set if not all
-	// of the answer RRs cannot be included.
+	// of the answer RRs can be included in the response.
 	dns.Truncated = len(dns.Answer) > numAnswer
 
 	dns.Answer = dns.Answer[:numAnswer]
--- a/msg_truncate.go
+++ b/msg_truncate.go
@@ -43,7 +43,10 @@ func (dns *Msg) Truncate(size int) {
 	if edns0 != nil {
 		// Account for the OPT record that gets added at the end,
 		// by subtracting that length from our budget.
-		size -= 12 // OPT record length.
+		//
+		// The EDNS(0) OPT record must have the root domain and
+		// it's length is thus unaffected by compression.
+		size -= Len(edns0)
 
 		// Remove the OPT record and handle it separately.
 		//
--- a/msg_truncate.go
+++ b/msg_truncate.go
@@ -1,7 +1,7 @@
 package dns
 
-// Scrub truncates the reply message so that it will fit the requested
-// buffer size.
+// Truncate ensures the reply message will fit into the requested buffer
+// size by removing records that exceed the requested size.
 //
 // It will first check if the reply fits without compression and then with
 // compression. If it won't fit with compression, Scrub then walks the
@@ -15,7 +15,7 @@ package dns
 // record, if present, and is transport specific otherwise. dns.MinMsgSize
 // should be used for UDP requests without an OPT record, and
 // dns.MaxMsgSize for TCP requests without an OPT record.
-func (dns *Msg) Scrub(size int) {
+func (dns *Msg) Truncate(size int) {
 	if dns.IsTsig() != nil {
 		// To simplify this implementation, we don't perform
 		// scrubbing on responses with a TSIG record.
--- a/msg_truncate.go
+++ b/msg_truncate.go
@@ -39,7 +39,7 @@ func (dns *Msg) Truncate(size int) {
 
 	dns.Compress = true
 
-	edns0 := dns.IsEdns0()
+	edns0 := dns.popEdns0()
 	if edns0 != nil {
 		// Account for the OPT record that gets added at the end,
 		// by subtracting that length from our budget.
@@ -47,12 +47,6 @@ func (dns *Msg) Truncate(size int) {
 		// The EDNS(0) OPT record must have the root domain and
 		// it's length is thus unaffected by compression.
 		size -= Len(edns0)
-
-		// Remove the OPT record and handle it separately.
-		//
-		// TODO(tmthrgd): IsEdns0 checks more than just the
-		// last record. This could remove the wrong record.
-		dns.Extra = dns.Extra[:len(dns.Extra)-1]
 	}
 
 	compression := make(map[string]struct{})
@@ -86,6 +80,7 @@ func (dns *Msg) Truncate(size int) {
 	dns.Extra = dns.Extra[:numExtra]
 
 	if edns0 != nil {
+		// Add the OPT record back onto the additional section.
 		dns.Extra = append(dns.Extra, edns0)
 	}
 }
--- a/msg_truncate.go
+++ b/msg_truncate.go
@@ -18,7 +18,7 @@ package dns
 func (dns *Msg) Truncate(size int) {
 	if dns.IsTsig() != nil {
 		// To simplify this implementation, we don't perform
-		// scrubbing on responses with a TSIG record.
+		// truncation on responses with a TSIG record.
 		return
 	}
 
--- a/scrub.go
+++ b/scrub.go
@@ -22,6 +22,14 @@ func (dns *Msg) Scrub(size int, req *Msg) {
 		return
 	}
 
+	// RFC 6891 mandates that the payload size in an OPT record
+	// less than 512 bytes must be treated as equal to 512 bytes.
+	//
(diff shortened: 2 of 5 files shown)


tmthrgd:
This is actually quite similar to #708 which I didn't even realise until just now.

tmthrgd:
One notable difference between this PR and #708, is that this PR will always compress the response if it won't fit, while #708 will respect the `dns.Compress` field.

scrub.go
```diff
@@ -0,0 +1,119 @@
...
+	o.Hdr.Rrtype = TypeOPT
+	o.SetVersion(0)
+	o.Hdr.Ttl &^= 0xffff // clear flags
+	o.SetDo(reqEDNS0.Do())
+	o.SetUDPSize(reqEDNS0.UDPSize())
+
+	dns.Extra = append(dns.Extra, o)
+}
+
+func (dns *Msg) truncateLoop(rrs []RR, size, l int, compression map[string]struct{}) (int, int) {
```
tmthrgd: Note to self: This shouldn’t be a method on `*Msg`.

msg_truncate.go
```diff
@@ -0,0 +1,106 @@
...
+	for _, r := range dns.Question {
+		l += r.len(l, compression)
+	}
+
+	var numAnswer int
+	if l < size {
+		l, numAnswer = truncateLoop(dns.Answer, size, l, compression)
+	}
+
+	var numNS int
```
miekg: truncating NS section... I wonder if that is a) ever needed [yes it can be artificially inflated] b) described in the RFCs somewhere?

msg_truncate.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+}
+
+func truncateLoop(rrs []RR, size, l int, compression map[string]struct{}) (int, int) {
+	for i, r := range rrs {
+		if r == nil {
+			continue
+		}
+
+		l += r.len(l, compression)
```
miekg: is it possible to call len less?

tmthrgd: It's called at most twice per-record. Once to calculate the original uncompressed length, and then once more* to calculate the compressed length. (Until it reaches the end or exceeds the size limit). I don't see anyway this could be more efficient.

msg_truncate.go
tmthrgd: I don't really think there is any other way to do this. Because the compression map is, and must, be created incrementally, it's not possible to truncate the answer section but not the NS section and still have it fit. You'd either drop far too many answer records, or always exceed the size limit.

You have to walk the message record-by-record and answers go before NS, the only way to not-truncate the NS section, while still being accurate, would be the costly approach in CoreDNS.

That being said, I'll scour the RFCs and see if I can find any guidance here.

tmthrgd: One option if you don't think the authority section should/can be truncated is to remove the uncompressed length of the authority section from the budget and include them all. This would mean that more answer records are excluded than is actually required.

gibson042: [RFC 1035](https://tools.ietf.org/html/rfc1035#section-6.2):
> When a response is so long that truncation is required, the truncation should start at the end of the response and work forward in the datagram.  Thus if there is any data for the authority section, the answer section is guaranteed to be unique.

There are definitely clients that would break if a truncated message included a section after an incomplete one.

tmthrgd: @gibson042 Thank you for that quote, that’s what I had been thinking. The approach in this PR is correct.

miekg:
I'll rereview this later. Just wanted to add that this algo is more efficient that what we currently do in coredns as we call Len() on the entire message *less*, but looking at it *per RR* we can Len() more. Here we just Len() each RR once

tmthrgd:
@miekg I don’t think there’s anything left to add to this.

markdingo:
I know this issue is close, so maybe I should raise a new one, but it would be convenient if msg.Truncate() returned a bool saying that it did truncate the message. We cannot rely on msg.Truncate because that only applies of the Answer slice was truncated (as per the DNS spec).

I think that functional change would be upward compatible.

My use case is solely for reporting purposes. I want to know whether a message has been truncated so as to track ideal message buffer sizes.

tmthrgd:
@markdingo What’s stopping you from doing exactly that now?

```go
rrcount := len(msg.Answer) + len(msg.Ns) + len(msg.Extra)
msg.Truncate(size)
truncated := len(msg.Answer) + len(msg.Ns) + len(msg.Extra) < rrcount
```

markdingo:
> @markdingo What’s stopping you from doing exactly that now?
> 
> ```go
> rrcount := len(msg.Answer) + len(msg.Ns) + len(msg.Extra)
> msg.Truncate(size)
> truncated := len(msg.Answer) + len(msg.Ns) + len(msg.Extra) < rrcount
> ```

As I understand msg_truncate.go, the message could be compressed to fit the size without changing the RR count.

My code fragment is wrong in that regard and should be ignored.

If I was being fussy, what I actually want to know is how much msg.Len() changes due to msg.Truncate() without having to make a subsequent call to msg.Len(). Also I guess abstractly, looking inside the msg to try and detect the effect of the function seems to be presumptive about what Truncate() might be capable of doing.

I admit these are not strong arguments, it was mainly that Truncate() could notice stuff "in passing" as it performs its job and could provide the caller with potentially useful info "for free".

My actual application code right now is:

old := msg.Len()
msg.Truncate()
diff := old - msg.Len()
if diff != 0 {
...

markdingo:
On a related front, I think the logic behind setting the Truncate flag is a bit simplistic.

The code comment says "According to RFC 2181, the TC bit should only be set if not all of the answer RRs can be included in the response." and the code does precisely this. But that's not what RFC2181 says. It says TC should be set if a "required" RRSet doesn't fit, regardless of whether that required RRSet is in the answer section or not.

Admittedly that whole section in RFC2181 is pretty muddled, but there is at least one use-case where TC should be set when the additional section is truncated. That's when the additional section contains referral/glue RRs for a sub-domain. By way of example, if we imagine a query to an example.net name server for sub.example.net/NS which results in:

;; QUESTION SECTION:
;sub.example.net.               IN      NS

;; AUTHORITY SECTION:
sub.example.net.        172800  IN      NS      c.sub.example.net.
sub.example.net.        172800  IN      NS      a.sub.example.net.
sub.example.net.        172800  IN      NS      b.sub.example.net.

;; ADDITIONAL SECTION:
c.sub.example.net.       172800  IN      A       ...
a.sub.example.net.       172800  IN      A       ...
b.sub.example.net.       172800  IN      A       ...

If just the additional section is truncated by msg.Truncate() then the client will fail to resolve the sub-domain as it won't see TC=1 to re-query with TCP and trying to resolve the sub NS address via the parent doesn't work since the address RRs are glue. You can try this for yourself with one of the TLDs such as .com. Imagine the additional section for google.com/NS is truncated with TC=0. You'll never be able to resolve google.com as you never see the NS address glue and can't get at it directly via the .com servers.

The problem here is that msg.Truncate() really has no idea whether non-Answer RRSets are "required" or not since it's just a generic function that is oblivious to the caller's intentions or role.

I'm not sure what the right approach is. Either the description of msg.Truncate() should accurately reflect that it's setting of TC is only a guide and the caller really has to make the final TC=1 decision. Alternatively it could follow the approach that PowerDNS takes the next release which is to set TC on *any* truncation. This doesn't disagree with RFC2181 and it is a safer approach.

And, if you think about it, setting TC=1 only with an Answer truncation vs any truncation is mostly just a corner-case optimisation so forgoing this optimisation is unlikely to be harmful so the powerDNS approach isn't terrible.

Maybe not knowing what is "required" as per RFC2181 explains why some of the main DNS packages force the matter by wiping out *all* sections and setting TC=1 when any truncation is needed? They never give the client a chance to use a partial answer.

Anyway, there is at least one use-case where relying on msg.Truncate() to set TC correctly seems to be risky.

PR #856, opened 2018-12-01
Simplify TKEY presentation format

### LINKED ISSUE #855: TKEY "presentation" format
The TKEY's String method is a bit whacky. As I'm working the NULL RR; it probably makes sense to just
prefix this should with a `;` instead of this contraption.
~~~
// TKEY has no official presentation format, but this will suffice.
func (rr *TKEY) String() string {
	s := "\n;; TKEY PSEUDOSECTION:\n"
	s += rr.Hdr.String() + " " + rr.Algorithm + " " +
		strconv.Itoa(int(rr.KeySize)) + " " + rr.Key + " " +
		strconv.Itoa(int(rr.OtherLen)) + " " + rr.OtherData
	return s
}
~~~

miekg:
Just put add ";" in front of it, instead of the whole pseudo option
text.

Fixes #855

### DIFF
--- a/types.go
+++ b/types.go
@@ -1050,8 +1050,7 @@ type TKEY struct {
 
 // TKEY has no official presentation format, but this will suffice.
 func (rr *TKEY) String() string {
-	s := "\n;; TKEY PSEUDOSECTION:\n"
-	s += rr.Hdr.String() + " " + rr.Algorithm + " " +
+	s := ";" + rr.Hdr.String() + " " + rr.Algorithm + " " +
 		strconv.Itoa(int(rr.KeySize)) + " " + rr.Key + " " +
 		strconv.Itoa(int(rr.OtherLen)) + " " + rr.OtherData
 	return s
--- a/types.go
+++ b/types.go
@@ -1050,9 +1050,16 @@ type TKEY struct {
 
 // TKEY has no official presentation format, but this will suffice.
 func (rr *TKEY) String() string {
-	s := ";" + rr.Hdr.String() + " " + rr.Algorithm + " " +
-		strconv.Itoa(int(rr.KeySize)) + " " + rr.Key + " " +
-		strconv.Itoa(int(rr.OtherLen)) + " " + rr.OtherData
+	s := ";" + rr.Hdr.String() +
+		" " + rr.Algorithm +
+		" " + TimeToString(rr.Inception) +
+		" " + TimeToString(rr.Expiration) +
+		" " + strconv.Itoa(int(rr.Mode)) +
+		" " + strconv.Itoa(int(rr.Error)) +
+		" " + strconv.Itoa(int(rr.KeySize)) +
+		" " + rr.Key +
+		" " + strconv.Itoa(int(rr.OtherLen)) +
+		" " + rr.OtherData
 	return s
 }
 


types.go
```diff
@@ -1050,8 +1050,7 @@ type TKEY struct {
 
 // TKEY has no official presentation format, but this will suffice.
 func (rr *TKEY) String() string {
-	s := "\n;; TKEY PSEUDOSECTION:\n"
-	s += rr.Hdr.String() + " " + rr.Algorithm + " " +
+	s := ";" + rr.Hdr.String() + " " + rr.Algorithm + " " +
 		strconv.Itoa(int(rr.KeySize)) + " " + rr.Key + " " +
 		strconv.Itoa(int(rr.OtherLen)) + " " + rr.OtherData
```
gibson042: This seems to be missing about half of the [RDATA fields](https://tools.ietf.org/html/rfc2930#section-2)... my advice would be to adopt the [approach from BIND](https://gitlab.isc.org/isc-projects/bind9/blob/master/lib/dns/rdata/generic/tkey_249.c#L116) as a de facto standard:
```
$algorithm.as.domain.name. $inception_as_decimal $expiration_as_decimal $mode $error $key_size (
	$key_data_as_base64 ) $other_size ( $other_data_as_base64 )
```

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Simplify TKEY prese..." ]
>> @@ -1050,10 +1050,16 @@ type TKEY struct {
>
> // TKEY has no official presentation format, but this will suffice.
> func (rr *TKEY) String() string {
>-	s := "\n;; TKEY PSEUDOSECTION:\n"
>-	s += rr.Hdr.String() + " " + rr.Algorithm + " " +
>-		strconv.Itoa(int(rr.KeySize)) + " " + rr.Key + " " +
>-		strconv.Itoa(int(rr.OtherLen)) + " " + rr.OtherData
>+	s := ";" + rr.Hdr.String() +
>+		" " + rr.Algorithm +
>+		" " + TimeToString(rr.Inception) +
>+		" " + TimeToString(rr.Expiration) +
>
>This will format the inception and expiration fields in ISO 8601 basic format (e.g., `20181202152006`) rather than the POSIX timestamp (e.g., `1543764006`) employed by BIND... is that divergence intentional?

mostly because that function exists and it wraps so the posix timestamp won't 
work after 2038. It also mimic RRSIGs which employ the same thing essentially.

gibson042:
> so the posix timestamp won't work after 2038. It also mimic RRSIGs which employ the same thing essentially.

I misspoke by referring to the BIND format as POSIX; it's really just an unsigned decimal integer that coincides with POSIX timestamps until 32-bit rollover. But regardless, rollover is an equivalent issue for both formats because all 32-bit timestamp fields in RDATA are subject to RFC 1982 ring/serial number arithmetic—and RRSIGs support both timestamp representations anyway (cf. [RFC 4034](https://tools.ietf.org/html/rfc4034#section-3.2)).

I don't have particularly strong feelings here because I consider it likely that an official presentation format would also support both, but I wanted to make sure that any divergence from BIND in this respect was intentional.

miekg:
Ack. 

I don't see any particular appetite on the various lists for standardizing this presentation format though.

PR #859, opened 2018-12-02
Avoid setting the Rdlength field when packing records

### LINKED ISSUE #858: coredns test failures with master and not 1.1.0
I'm seeing test failures in the CoreDNS test suite; they look dependent on the miekg/dns version used. *master* breaks, v1.1.0 works.

tmthrgd:
This PR fixes potential data-races when packing messages that are shared between goroutines. Records will no longer be mutated during packing.

`PackRR` will continue to set the `Rdlength` field for compatibility. `PackRR` is called in three remaining places in this library.

---

`rawSignatureData` which operates on a copy of the `RR`:
https://github.com/miekg/dns/blob/778aa4f83dde7ea59f1cc8a94e51e2ce19e76914/dnssec.go#L728

`TsigGenerate` which operates on a `new(TSIG)` record that is not passed in by it's caller:
https://github.com/miekg/dns/blob/778aa4f83dde7ea59f1cc8a94e51e2ce19e76914/tsig.go#L137

`(*SIG).Sign` which already modifies it's receiver:
https://github.com/miekg/dns/blob/778aa4f83dde7ea59f1cc8a94e51e2ce19e76914/sig0.go#L40

---

Fixes #858


msg.go
```diff
@@ -619,23 +619,33 @@ func intToBytes(i *big.Int, length int) []byte {
 // PackRR packs a resource record rr into msg[off:].
 // See PackDomainName for documentation about the compression.
 func PackRR(rr RR, msg []byte, off int, compression map[string]int, compress bool) (off1 int, err error) {
-	return packRR(rr, msg, off, compressionMap{ext: compression}, compress)
+	headerEnd, off1, err := packRR(rr, msg, off, compressionMap{ext: compression}, compress)
+	if err == nil {
```
miekg: doesn't this leave the same race?

tmthrgd: No it doesn’t, because `PackRR` isn’t called by the packing code which calls the internal `packRR`. This can be removed of course, but I can imagine someone might be relying on it.

PR #860, opened 2018-12-02
Don't the rdlength when packing a RR.

miekg:
This introduces a race when multiple goroutines are handling the same RR.
The rdlength is now mostly 0, except the RRs that came in via the wire,
i.e. NewRR never set it either.

We might consider setting it to zero for from-the-wire-RRs as well for
consistency.


miekg:
@tmthrgd I like this simpler PR slightly more than #859

tmthrgd:
@miekg IMHO #859 is a better approach. For one, removing `rawSetRdlength` makes the implementation simpler and slightly faster. (I didn’t bother to run benchmarks because this was about correctness, but I can).

tmthrgd:
After #859 there will be a single place where names are unpacked, `UnpackDomainName` meaning less opportunity for bugs. Quite honestly, I didn’t even know `rawSetRdlength` existed and I would have overlooked it making any correctness change to name unpacking.

PR #866, opened 2018-12-11
Document SetExtendedRcode use

miekg:
Not obvious how to use this function, so a bunch of words on how this
should work.


miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Document SetExtende..." ]
>tmthrgd requested changes on this pull request.
>
>I actual don’t think this is right once I took a closer look at the code. If I understand it correctly, consumers should only set `m.Rcode`, which will handle this while packing, and avoid calling this function at all.

So we kept it only for backwards compat reasons? We should then noop it and add 
a line that it will be removed at some point.

miekg:
just setting the rcode works. Closing

Issue #867, opened 2018-12-17
Reject nscount > 0?


cesarkuroiwa:
https://github.com/miekg/dns/blob/450ab7d57fb96d2a1d85cc507d30a91d00185dd4/acceptfunc.go#L47

Regarding the DefaultMsgAccept function, I am not sure it would be appropriate to consider Nscount > 0 a reason to reject the request. An IXFR request could have a SOA record in the NS section.

Even though I know I could simply use a custom MsgAcceptFunc, I'm just considering if it'd be appropriate to be the default behaviour

miekg:
Prolly make sense to allow that. Which RFC is this?

cesarkuroiwa:
[RFC 1995](https://tools.ietf.org/html/rfc1995)

miekg:
Sgtm.. please reference section 3 of that rfc in your PR

PR #868, opened 2018-12-17
Don't reject Nscount > 0

cesarkuroiwa:
IXFR request could have a SOA RR in the NS section
RFC 1995, section 3: https://tools.ietf.org/html/rfc1995

Ref issue #867

### DIFF
--- a/acceptfunc.go
+++ b/acceptfunc.go
@@ -44,9 +44,11 @@ var defaultMsgAcceptFunc = func(dh Header) MsgAcceptAction {
 	if dh.Ancount != 0 {
 		return MsgReject
 	}
-	if dh.Nscount != 0 {
-		return MsgReject
-	}
+	// IXFR request could have a SOA RR in the NS section
+	// RFC 1995, section 3: https://tools.ietf.org/html/rfc1995
+	// if dh.Nscount != 0 {
+	//	return MsgReject
+	// }
 	if dh.Arcount > 2 {
 		return MsgReject
 	}


acceptfunc.go
```diff
@@ -44,9 +44,11 @@ var defaultMsgAcceptFunc = func(dh Header) MsgAcceptAction {
 	if dh.Ancount != 0 {
 		return MsgReject
 	}
-	if dh.Nscount != 0 {
-		return MsgReject
-	}
+	// IXFR request could have a SOA RR in the NS section
+	// RFC 1995, section 3: https://tools.ietf.org/html/rfc1995
+	// if dh.Nscount != 0 {
```
miekg: only one can be there, no?

cesarkuroiwa: Yes, I believe you are correct.
Already updated the code

acceptfunc.go
```diff
@@ -44,7 +44,9 @@ var defaultMsgAcceptFunc = func(dh Header) MsgAcceptAction {
 	if dh.Ancount != 0 {
 		return MsgReject
 	}
-	if dh.Nscount != 0 {
+	// IXFR request could have one SOA RR in the NS section
+	// RFC 1995, section 3: https://tools.ietf.org/html/rfc1995
```
tmthrgd: There's no need to include the URL. There are only two places in the library that do that, everywhere else just references the RFC number.

cesarkuroiwa: OK, fixed

PR #870, opened 2018-12-19
Update WriteMsg to return (int, error)

stuartnelson3:
Discussed in https://github.com/coredns/coredns/issues/2317

This will be integrated into coredns in a follow-up pr in that repo.


stuartnelson3:
Yeah, the whole thing becomes a very big mess if the two libraries don't agree on the interface. It didn't take long to update, if you're ever interested in making this change.

PR #872, opened 2018-12-30
Reduce allocations in ReverseAddr

tmthrgd:
The IPv6 change in this PR is really trivial, while the IPv4 change is less so. The IPv4 code is quite similar to the IPv6 code now.

```
name                old time/op    new time/op    delta
ReverseAddr/IP4-12     175ns ± 5%     140ns ±13%  -19.94%  (p=0.000 n=10+10)
ReverseAddr/IP6-12     364ns ±14%     218ns ±22%  -40.23%  (p=0.000 n=10+10)

name                old alloc/op   new alloc/op   delta
ReverseAddr/IP4-12     51.0B ± 0%     48.0B ± 0%   -5.88%  (p=0.000 n=10+10)
ReverseAddr/IP6-12      176B ± 0%       96B ± 0%  -45.45%  (p=0.000 n=10+10)

name                old allocs/op  new allocs/op  delta
ReverseAddr/IP4-12      3.00 ± 0%      2.00 ± 0%  -33.33%  (p=0.000 n=10+10)
ReverseAddr/IP6-12      3.00 ± 0%      2.00 ± 0%  -33.33%  (p=0.000 n=10+10)
```

### DIFF
--- a/defaults.go
+++ b/defaults.go
@@ -244,9 +244,16 @@ func ReverseAddr(addr string) (arpa string, err error) {
 	if ip == nil {
 		return "", &Error{err: "unrecognized address: " + addr}
 	}
-	if ip.To4() != nil {
-		return strconv.Itoa(int(ip[15])) + "." + strconv.Itoa(int(ip[14])) + "." + strconv.Itoa(int(ip[13])) + "." +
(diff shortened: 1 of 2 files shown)


tmthrgd:
I should add, I doubt these are performance critical, but I saw them manually reviewing the code.

PR #873, opened 2018-12-30
Stop using packDomainName in IsDomainName

tmthrgd:
I'm not sure if this is worth doing, but @miekg mentioned it in https://github.com/miekg/dns/pull/818#issuecomment-442355376. `IsDomainName` is documented as being incredibly liberal, so it may be possible to make it simpler at the expense of strictness.

### DIFF
--- a/defaults.go
+++ b/defaults.go
@@ -164,8 +164,8 @@ func (dns *Msg) IsEdns0() *OPT {
 // the number of labels.  When false is returned the number of labels is not
 // defined.  Also note that this function is extremely liberal; almost any
 // string is a valid domain name as the DNS is 8 bit protocol. It checks if each
-// label fits in 63 characters, but there is no length check for the entire
-// string s. I.e.  a domain name longer than 255 characters is considered valid.
+// label fits in 63 characters and that the entire name will fit into the 255
+// octet wire format limit.
 func IsDomainName(s string) (labels int, ok bool) {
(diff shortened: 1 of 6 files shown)


defaults.go
```diff
@@ -166,8 +167,68 @@ func (dns *Msg) IsEdns0() *OPT {
 // label fits in 63 characters, but there is no length check for the entire
 // string s. I.e.  a domain name longer than 255 characters is considered valid.
```
miekg: This comment is not true anymore

defaults.go
```diff
@@ -166,8 +167,68 @@ func (dns *Msg) IsEdns0() *OPT {
...
+
+	// Each dot ends a segment of the name. Except for escaped dots (\.), which
+	// are normal dots.
+
+	var (
+		off    int
+		begin  int
+		wasDot bool
+	)
+	for i := 0; i < len(s); i++ {
```
miekg: Maybe refer to packDomainName and that this code is a (almost) copy of that routine there.

miekg:
Think this will be a very welcome change. Not allocating a buffer here should a (huge?) win.

tmthrgd:
I don't know if this will be a huge win performance wise as that buffer is only allocated when the name has escape sequences since #818. There will be a performance improvement when the name does have escape sequences though.

defaults.go
tmthrgd: I actually don't think this was ever true.

PR #877, opened 2018-12-30
Move scanner comment handling out of scanRR

tmthrgd:
This changes how comments are handled by the scanning code. They're now handled solely by `zlexer` and `ZoneParser`. Now comments no longer infest scan_rr.go in sometimes odd ways. There should be no observable behaviour changes.


miekg:
ah, nice, makes all the function returns a lot cleaner.

Issue #793, opened 2018-10-15
scan_rr.go should be generated


miekg:
All, but `scan_rr`'s per RR functions are generated from the structs in `types.go`. There is not actually a reason why this can't happen for these as well. Would cut back on maintenance and typos in this file.

tmthrgd:
I took a brief shot at this and it's not the easiest thing to do. So far I've got quite a few unfortunate special-cases, because most of the tags aren't about the presentation format. I think this would be worth doing, but it might be an undertaking.

A small tool I developed that might be helpful is [here](https://gist.github.com/tmthrgd/917b25210ebe45e3df75d0f5bbfb6dac). It sorts all the functions and global map keys in a file, which minimises the diff as the functions would move around otherwise. (It should be used after yanking the not to be generated stuff out of the file).

miekg:
Interesting. We could just set aside the hard ones and generate the ones we
can as a first step. Getting the nuances right for all might be too hard
for all records.

miekg:
this might be too lofty, let's close for now

PR #879, opened 2019-01-03
Be stricter when parsing zone files

tmthrgd:
**DO NOT MERGE**: This is still very much a work in progress.

This is an attempt to make the `ZoneParser` & `zlexer` stricter. Currently they're unintentionally quite permissive. For example both of the following records parse despite being invalid:
```
example.com. 42 IN SOA ns1.example.com."hostmaster.example.com."1 86400 60 86400 3600
nl. IN NSEC3PARAM 1 0 5""
```

This also has the benefit of eliminating the `ZoneParser` state machine which leaves much simpler and easier to understand code.

This still has lots of rough edges and corner cases that need to be addressed, but it should be ready for a least a curious once over.

### DIFF
--- a/generate.go
+++ b/generate.go
@@ -53,7 +53,7 @@ func (zp *ZoneParser) generate(l lex) (RR, bool) {
 		return zp.setParseError("bad range in $GENERATE range", l)
 	}
 
-	zp.c.Next() // _BLANK
+	zp.c.Expect(zBlank)
 
 	// Create a complete new string, which we then parse again.
 	var s string
--- a/generate.go
+++ b/generate.go
@@ -53,7 +53,9 @@ func (zp *ZoneParser) generate(l lex) (RR, bool) {
 		return zp.setParseError("bad range in $GENERATE range", l)
 	}
 
-	zp.c.Expect(zBlank)
+	if l, _ := zp.c.Expect(zBlank); l.err {
+		return zp.setParseError("bad data in $GENERATE directive", l)
+	}
 
 	// Create a complete new string, which we then parse again.
 	var s string
--- a/parse_test.go
+++ b/parse_test.go
@@ -341,7 +341,7 @@ func TestParseDirectiveMisc(t *testing.T) {
 
 func TestNSEC(t *testing.T) {
 	nsectests := map[string]string{
-		"nl. IN NSEC3PARAM 1 0 5 30923C44C6CBBB8F":                                                                                                 "nl.\t3600\tIN\tNSEC3PARAM\t1 0 5 30923C44C6CBBB8F",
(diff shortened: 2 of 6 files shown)


tmthrgd:
This should probably wait for #793. After that issue is solved the scan_re.go will be cleaner and simpler to reason about and change.

gibson042:
I haven't reviewed the code changes, but I strongly support this effort. I'd specifically like to see the "missing RDATA" (e.g., `example.com. 60 AAAA`) cases fixed, so please make sure to reject those if that is not already the case.

miekg:
Where do you see those?

gibson042:
> Where do you see those?

It looks like the majority have been fixed, but there's still an issue at end of file:
```go
$ cat main.go
package main

import (
	"fmt"
	"github.com/miekg/dns"
	"strings"
)

func main() {
	fmt.Printf("; miekg/dns version: %v\n", dns.Version)

	zone := `
example.com. 3600 IN SOA ns.example.net. hostmaster.example.com. 1 86400 60 86400 3600
example.com. 3600 NS ns.example.net.

; Dude, where's my RDATA?
example.com. 60 AAAA
`
	for token := range dns.ParseZone(strings.NewReader(zone), "", "") {
		if token.Error != nil {
			fmt.Printf("; ERROR: %v\n", token.Error)
		} else {
			fmt.Println(token.RR.String())
		}
	}
}

$ go run main.go
; miekg/dns version: 1.1.3
example.com.	3600	IN	SOA	ns.example.net. hostmaster.example.com. 1 86400 60 86400 3600
example.com.	3600	IN	NS	ns.example.net.
example.com.	60	IN	AAAA	
```

miekg:
Good that you bring this up, because of dynamic updates this needs to be
supported.

gibson042:
Dynamic updates are no excuse for accepting invalid zone files in my opinion, but shouldn't they be processed from _wire_ format anyway? Why would they ever hit this code at all?

gibson042:
I also found a new issue when exploring missing RDATA parses: #901

gibson042:
@miekg Can you clarify what you're claiming "is a valid use case" and what "needs full support"? I'm not talking about invoking `NewRR` or generating presentation format, but about _parsing_ presentation format—and all RRs in a zone file MUST have RDATA.

If you think that supporting DNS UPDATE as defined by [RFC 2136](https://tools.ietf.org/html/rfc2136) and subsequently updated requires **parsing** presentation format RRs with no RDATA, then you are mistaken because that protocol is defined in terms of wire format.

tmthrgd:
I have no time to work on this right now so I'm going to close it for the time being. Hopefully I can resurrect this eventually. I happen to really like the `ZoneParser.Next` changes included.

PR #883, opened 2019-01-03
Use an interface method for IsDuplicate

tmthrgd:
This is simpler and more inline with other RR methods and eliminates the `isDuplicateRdata` switch statement. This also eliminates a panic that could occur if a record was lying about it's RRType (see `TestDuplicateWrongRrtype`).

### DIFF
--- a/dns.go
+++ b/dns.go
@@ -44,7 +44,7 @@ type RR interface {
 	// pack packs an RR into wire format.
 	pack(msg []byte, off int, compression compressionMap, compress bool) (headerEnd int, off1 int, err error)
 
-	// isDuplicate returns whether two RRs have duplicate RDATA.
+	// isDuplicate returns whether the two RRs are duplicates.
 	isDuplicate(r2 RR) bool
 }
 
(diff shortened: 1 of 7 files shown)


dns.go
```diff
@@ -81,6 +84,8 @@ func (h *RR_Header) len(off int, compression map[string]struct{}) int {
 	return l
 }
 
+func (r1 *RR_Header) isDuplicate(r2 RR) bool { return false }
```
miekg: maybe nicer to implement the code that checks the header here instead of in the IsDuplicate (in duplicate.go)?

tmthrgd: I had that before a3d7eba50825d546074d81084e639bbecf9cbb57, but changed it back to this approach. I’m happy to revert a3d7eba50825d546074d81084e639bbecf9cbb57 if you’d like.

tmthrgd: The diff would like https://github.com/miekg/dns/compare/master...tmthrgd:a3d7eba%5E if I do revert that commit.

miekg: why did you change it back? I think here is better, but would like to know the reason you reverted that?

tmthrgd: @miekg Are you happy with how it is now?

PR #885, opened 2019-01-03
Move RR header packing out of generated code

tmthrgd:
The header will be packed into the `msg` buffer before the generated `pack()` function is called. This simplifies the generated code and eliminates the `headerEnd` return value.


miekg:
lgtm

PR #886, opened 2019-01-04
Use an interface method for parsing zone file records

tmthrgd:
This is #884 but for zone file parsing.

This eliminates the `typeToparserFunc` map which should offer a small speedup to program init time. This also simplifies the code.

### DIFF
--- a/dns.go
+++ b/dns.go
@@ -50,6 +50,11 @@ type RR interface {
 	// This will only be called on a new and empty RR type with only the header populated. It
 	// will only be called if the record's RDATA is non-empty.
 	unpack(msg []byte, off int) (off1 int, err error)
+
+	// parse parses an RR from zone file format.
+	//
+	// This will only be called on a new and empty RR type with only the header populated.
+	parse(c *zlexer, origin, file string) *ParseError
 }
 
 // RR_Header is the header all DNS resource records share.
@@ -95,6 +100,10 @@ func (h *RR_Header) pack(msg []byte, off int, compression compressionMap, compre
 
 func (h *RR_Header) unpack(msg []byte, off int) (int, error) {
 	panic("dns: internal error: unpack should never be called on RR_Header")
+}
+
+func (h *RR_Header) parse(c *zlexer, origin, file string) *ParseError {
+	panic("dns: internal error: parse should never be called on RR_Header")
 }
 
 // ToRFC3597 converts a known RR to the unknown RR representation from RFC 3597.
--- a/edns.go
+++ b/edns.go
@@ -88,6 +88,10 @@ func (rr *OPT) len(off int, compression map[string]struct{}) int {
 	return l
 }
 
+func (rr *OPT) parse(c *zlexer, origin, file string) *ParseError {
+	return &ParseError{file, "OPT record cannot be represented in a zone file", c.LastToken()}
+}
+
 // return the old value -> delete SetVersion?
 
 // Version returns the EDNS version used. Only zero is defined.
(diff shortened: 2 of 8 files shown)


tmthrgd:
I eliminated the `Variable` field by only slurping if the last token wasn't `zNewline` or `zEOF`. I'm pretty confident this will work successfully, but it might ultimately be a little bit fragile and laxer than intended. The error check in `setRR` after the `parse` method call should ensure that only an intended `zNewline` or `zEOF` will be allowed to reach this check.

tmthrgd:
One other thing to note, it was previously possible to represent an ANY, NULL, OPT or TSIG record in a zone file by using the RFC 3597 syntax. This is no longer possible and will instead return an error. If this is a problem I can restore that behaviour for one or all of those records.

miekg:
Re: ANY, NULL, OPT or TSIG  

I think that should be allowed, because you can always do this with TYPEXXX syntax, according the RFC both TYPEXXX and the actual type should be allowed.

That ANY, OPT, and TSIG can't be in a zonefile is not relevant.

tmthrgd:
So currently the following test:
```
func TestParseSyntheticRecord(t *testing.T) {
	// This tests the parsing of synthetic records that can't appear
	// in zone files and do not have a defined presentation format.

	for _, tc := range [...]struct {
		name, s string
		typ     reflect.Type
	}{
		{"ANY", "example.org. 360 TYPE255 \\# 0", reflect.TypeOf((*ANY)(nil))},
		{"NULL", "example.org. 360 NULL \\# 4 deadbeef", reflect.TypeOf((*NULL)(nil))},
		{"OPT", "example.org. 360 OPT \\# 0", reflect.TypeOf((*OPT)(nil))},
		{"TSIG", "example.org. 360 TSIG \\# 0", reflect.TypeOf((*TSIG)(nil))},
	} {
		rr, err := NewRR(tc.s)
		if err != nil {
			t.Errorf("error parsing %s record: %v", tc.name, err)
			continue
		}
		if reflect.TypeOf(rr) != tc.typ {
			t.Errorf("invalid type when parsing %s record: wanted %v, got %v", tc.name, tc.typ, reflect.TypeOf(rr))
		}
	}
}
```

produces this output:
```
--- FAIL: TestParseSyntheticRecord (0.00s)
    scan_test.go:170: invalid type when parsing ANY record: wanted *dns.ANY, got *dns.RFC3597
    scan_test.go:170: invalid type when parsing NULL record: wanted *dns.NULL, got *dns.RFC3597
    scan_test.go:170: invalid type when parsing OPT record: wanted *dns.OPT, got *dns.RFC3597
    scan_test.go:170: invalid type when parsing TSIG record: wanted *dns.TSIG, got *dns.RFC3597
```

Each record is parsed as an `*RFC3597`, are you saying this is the intended behaviour? I can try and replicate it if so. I could also work them to produce the proper `RR` types, by unpacking the RFC3597 data.

tmthrgd:
It currently produces this output with this PR:
```
--- FAIL: TestParseSyntheticRecord (0.00s)
    scan_test.go:166: error parsing ANY record: dns: ANY record cannot be represented in a zone file: " " at line: 1:25
    scan_test.go:166: error parsing NULL record: dns: NULL record cannot be represented in a zone file: " " at line: 1:22
    scan_test.go:166: error parsing OPT record: dns: OPT record cannot be represented in a zone file: " " at line: 1:21
    scan_test.go:166: error parsing TSIG record: dns: TSIG record cannot be represented in a zone file: " " at line: 1:22
```

miekg:
I think returning them as 3597 is the intended behaviour. Don't know if the
rfc actually says something about this situation though.

tmthrgd:
@miekg They'll now parse and be returned as `*RFC3597` as before.

PR #896, opened 2019-01-04
Test for escaped dots in IsFqdn

### LINKED ISSUE #851: Escaped name cheats FQDN check
The following functions don't account for an escaped dot when determining FQDN status:

`packDomainName`:
https://github.com/miekg/dns/blob/778aa4f83dde7ea59f1cc8a94e51e2ce19e76914/msg.go#L204-L223

`IsFqdn`:
https://github.com/miekg/dns/blob/778aa4f83dde7ea59f1cc8a94e51e2ce19e76914/defaults.go#L192-L199

`SplitDomainName` which is documented as such:
https://github.com/miekg/dns/blob/778aa4f83dde7ea59f1cc8a94e51e2ce19e76914/labels.go#L5-L21

`toAbsoluteName`:
https://github.com/miekg/dns/blob/778aa4f83dde7ea59f1cc8a94e51e2ce19e76914/scan.go#L1238-L1257

Of these: `IsFqdn` could be a problem, `SplitDomainName` is documented as ignoring them, but `packDomainName` is causing a mismatch with the length code in this case.

A domain name such as `"example.org\."` is, based on our escaping, not a FQDN, but is treated as such by the functions above. This may be a sufficiently 'edgey' edge-case that we should just ignore it. Thoughts @miekg?

tmthrgd:
This catches situations where the final dot in a name is actually escaped. This was causing problems when packing these domain names.

Fixes #851

### DIFF
--- a/defaults.go
+++ b/defaults.go
@@ -253,7 +253,18 @@ func IsMsg(buf []byte) error {
 
 // IsFqdn checks if a domain name is fully qualified.
 func IsFqdn(s string) bool {
-	return strings.HasSuffix(s, ".")
+	s2 := strings.TrimSuffix(s, ".")
+	if s == s2 {
+		return false
+	}
+
+	i := strings.LastIndexFunc(s2, func(r rune) bool {
+		return r != '\\'
+	})
+
+	// Test whether we have an even number of escape sequences before
+	// the dot or none.
+	return (len(s2)-i)%2 != 0
 }
 
 // IsRRset checks if a set of RRs is a valid RRset as defined by RFC 2181.
--- a/labels_test.go
+++ b/labels_test.go
@@ -174,6 +174,42 @@ func TestIsDomainName(t *testing.T) {
 	}
 }
 
+func TestIsFqdnEscaped(t *testing.T) {
+	for s, expect := range map[string]bool{
+		".":                  true,
+		"\\.":                false,
+		"\\\\.":              true,
+		"\\\\\\.":            false,
+		"\\\\\\\\.":          true,
+		"a.":                 true,
+		"a\\.":               false,
+		"a\\\\.":             true,
+		"a\\\\\\.":           false,
+		"ab.":                true,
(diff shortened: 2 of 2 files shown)


defaults.go
```diff
@@ -253,7 +253,18 @@ func IsMsg(buf []byte) error {
...
+		return false
+	}
+
+	i := strings.LastIndexFunc(s2, func(r rune) bool {
+		return r != '\\'
+	})
+
+	// Test whether we have an even number of escape sequences before
+	// the dot or none.
+	return (len(s2)-i)%2 != 0
```
tmthrgd: I'm not 100% sure I've got the logic right here, but it passes every test I can throw at it.

tmthrgd: @miekg That's exactly what this does and there's no allocation. `TrimSuffix` will remove the dot if it exists by slicing `s`, `s == s2` checks if there was a dot, `strings.LastIndexFunc` finds the last non-backslash in the name, and then `(len(s2)-i)%2` calculates if it's an even or odd number of backslashes.

defaults.go
```diff
@@ -253,7 +253,18 @@ func IsMsg(buf []byte) error {
 
 // IsFqdn checks if a domain name is fully qualified.
 func IsFqdn(s string) bool {
-	return strings.HasSuffix(s, ".")
+	s2 := strings.TrimSuffix(s, ".")
```
miekg: I still don't really like the changes. Having a new string and then comparing the entire thing again, while we're only interested in the last bit.

tmthrgd: `strings.TrimSuffix` only slices the string, it doesn’t create a new one. Do you have any alternative in mind?

miekg: how is that not a new string? https://play.golang.org/p/d4-FIn187I6

miekg: ok cool. Thanks for the explainer.

lgtm to me then

PR #897, opened 2019-01-06
Sort scan_rr.go methods alphabetically

tmthrgd:
I'm not sure whether this is worth merging or not. I sorted each method in scan_rr.go alphabetically using [this tool](https://gist.github.com/tmthrgd/917b25210ebe45e3df75d0f5bbfb6dac). This should make diffs smaller later when we try to generate scan_rr.go, for now it remains as hand-written as always.

This was done with a tool and should be entirely cosmetic.

Updates #793


miekg:
This might aid in readability (a bit), but there is no automatic way to keep enforcing this, so it will deteriorate with time.

miekg:
it works if we do something like #793 (but that's harder)

miekg:
Yeah. It's not edited often. Feel free to merge

tmthrgd:
@miekg Actually I'm going to leave this for now. I might try going straight for generating it first, because this is going to make `git blame` harder to reason with.

PR #902, opened 2019-01-22
fix OPT Record deep copy 

fturib:
While debugging CoreDNS, I found that the dns.Msg.Copy() was partially invalid for OPT records.
Because OPT is []EDNS0 and EDNS0 is an interface (a pointer...).

- Added a Copy() method on the EDNS0 interface
- modify the generation of OPT.Copy() (file types_generated)
- add a UT in Msg that show failing before the change and correct after (although testing only EDNS0_LOCAL sub type).

### DIFF
--- a/edns.go
+++ b/edns.go
@@ -159,6 +159,8 @@ type EDNS0 interface {
 	unpack([]byte) error
 	// String returns the string representation of the option.
 	String() string
+	// Copy return deep copy of this EDNS0 information
+	Copy() EDNS0
 }
 
 // EDNS0_NSID option is used to retrieve a nameserver
@@ -190,6 +192,7 @@ func (e *EDNS0_NSID) pack() ([]byte, error) {
 func (e *EDNS0_NSID) Option() uint16        { return EDNS0NSID } // Option returns the option code.
 func (e *EDNS0_NSID) unpack(b []byte) error { e.Nsid = hex.EncodeToString(b); return nil }
 func (e *EDNS0_NSID) String() string        { return e.Nsid }
+func (e *EDNS0_NSID) Copy() EDNS0           { return &EDNS0_NSID{e.Code, e.Nsid} }
 
 // EDNS0_SUBNET is the subnet option that is used to give the remote nameserver
 // an idea of where the client lives. See RFC 7871. It can then give back a different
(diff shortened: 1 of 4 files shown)


fturib:
I committed the file ztypes.go that is supposed to be generated.
Maybe it is not anymore.

NOTE: on my local machines, if I generate it, the NULL type that implements RR interface has missing methods (like Header()).

tmthrgd:
@fturib You’ll need to run `go install github.com/miekg/dns` before running `go generate` to have the generated file be correct.

fturib:
@tmthrgd : Thanks for the info. I just did it, and finally get the same zfiles.go than the one I committed. 
I have one question though : this `go generate` is not applied before launching the CI ? (why should I have to commit the ztypes.go ?)

tmthrgd:
@fturib `go generate` is designed to be run by the developer and then checked into the repository.

fturib:
@tmthrgd : Thanks for info. 
The correct file is checked-in.

miekg:
I note we have
~~~
func (rr *OPT) copy() RR {
	Option := make([]EDNS0, len(rr.Option))
	copy(Option, rr.Option)
	return &OPT{rr.Hdr, Option}
}
~~~

Why is this not sufficient?

fturib:
I though of it, but what I need really is to make a copy of the OPT. 
As it's copy() is not public, the way I used is to duplicate the function copy of OPT in my code in REWRITE of CoreDNS. I will not be able if the EDNS0.Copy() is not public
Then I will have to copy the whole msg, to get only a copy of the OPT ...

My preferred way would be to make public the Copy of RR ...

types_generate.go
```diff
@@ -244,6 +244,14 @@ func main() {
 					splits := strings.Split(t, ".")
 					t = splits[len(splits)-1]
 				}
+				if t == "EDNS0" {
+					// need to use a Copy for each element
```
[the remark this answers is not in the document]
miekg: Put a comment above the switch:
```
// For the EDNS0 interface (used in the OPT RR)  iwe need to call the copy method on each element.
```

miekg:
having a public copy is a diff. discussion. Copy for specific rrs is also not public; so don't force that change in this pr.

The test is good to highlight the issue, but doesn't add much (and tests a very narrow usecase), so I think that can be removed.

fturib:
OK. 
Updated w/o the UT and with Copy() changed to copy()

fturib:
NOTE: I reworked the the CoreDNS PR that required this fix (and a public Copy).
this fix is no more required by that PR.

However, I think it would be good that the msg.Copy() is consistent and deep copy complete (including copy of EDNS0).

edns.go
```diff
@@ -159,6 +159,8 @@ type EDNS0 interface {
 	unpack([]byte) error
 	// String returns the string representation of the option.
 	String() string
+	// Copy return deep copy of this EDNS0 information
```
miekg: `// Copy returns a deep-copy of the option.`

fturib: ok.

fturib:
pushed with all review comments addressed - I believe.

PR #919, opened 2019-03-06
Add checks on data length for A and AAAA records

No-ops:
Fixes panic when parsing A or AAAA records with no data

### DIFF
--- a/format_test.go
+++ b/format_test.go
@@ -4,26 +4,7 @@ import (
 	"testing"
 )
 
-func TestFieldEmptyAData(t *testing.T) {
-	rr := &A{
-		Hdr:  RR_Header{},
-		A: nil,
-	}
-
-	res := Field(rr, 1)
-	if res != "" {
-		t.Errorf("expected empty string but got %v", res)
-	}
-}
-
-func TestFieldEmptyAAAAData(t *testing.T) {
-	rr := &AAAA{
-		Hdr:  RR_Header{},
-		AAAA: nil,
-	}
-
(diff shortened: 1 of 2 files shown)


miekg:
do you have a short test that triggers this?

No-ops:
Good point. Added tests for these cases. 

The tests are a bit simplified but if we parse a dns packet with empty A/AAAA data we will get the same results.

format_test.go
```diff
@@ -0,0 +1,29 @@
+package dns
+
+import (
+	"testing"
+)
+
+func TestFieldEmptyAData(t *testing.T) {
+	rr := &A{
+		Hdr:  RR_Header{},
```
tmthrgd: one nit: These tests could just be `Field(new(A), 1)` and `Field(new(AAAA), 1)`. It would probably be worth merging them into a single test case.

miekg: +1, can you make that change? Then I'll merge.

No-ops: I don't fully agree. I specifically didn't want to test the nil header for the because I think that should go in a separate test if that should be tested.

I made two separate tests because they test two different flows in the Field function.

format_test.go
```diff
@@ -0,0 +1,10 @@
+package dns
+
+import (
+	"testing"
+)
+
+func TestFieldEmptyAOrAAAAData(t *testing.T) {
+	Field(new(A), 1)
```
tmthrgd: I meant keep the `!= “”` checks as well. Apologies if that wasn’t clear.

PR #920, opened 2019-03-07
corrected default value of Server.MsgAcceptFunc as documented

jinmeiib:
the description states

but actually defaultMsgAcceptFunc (not visible outside of the package) was used.

At least the description and code should be consistent.  The proposed change assumes the description is correct.


tmthrgd:
I thought about making this exact change previously, but decided against it. I think the existing behaviour is actually preferable as global variables tend to lead to strange behaviour.

That said there’s nothing technically wrong with this change and it seems more consistent with the original intent behind this variable.

PR #922, opened 2019-03-07
[WIP/Demo] Anti-burst worker mechanism

fturib:
Related to issue : https://github.com/coredns/coredns/issues/2593

This code intend to demonstrate how we could avoid memory spike in case of burst of incoming request.

See description in [this comment ](https://github.com/coredns/coredns/issues/2593#issuecomment-470307464)

See corresponding [results with prometheus graphics here](https://github.com/coredns/coredns/issues/2593#issuecomment-470636374)

> Adding an "anti-burst" mechanism in Server.worker(...), avoiding to create to much extra go-routines, ensure that CoreDNS is running properly and delivers its 20 kqps:

> * limit the total workers (go-routines) to 1000
> * limit the total re-usable workers to 500 (in fact, it seems that 100 is enough)
> * recycle the workers after 10k msg served
> * when cannot add a worker because already 1000 is running... just drop a few msg by adding a Sleep of 10ms .. to let other routines running. [we are losing those msg anyway ...]

### DIFF
--- a/server.go
+++ b/server.go
@@ -9,6 +9,7 @@ import (
 	"encoding/binary"
 	"errors"
 	"io"
+	"math/rand"
 	"net"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


miekg:
this wasn't the root cause, so closing.

PR #923, opened 2019-03-08
ZoneParser: error on parsing an IPv6 address in an A record

tomfitzhenry:
And vice versa for IPv4 with AAAA.

The implementation of isIPv6 is inspired by https://github.com/golang/go/blob/e341bae08d75611adea6566c1d01c1e3a0de57f9/src/net/ip.go#L678-L681 .

### DIFF
--- a/scan_rr.go
+++ b/scan_rr.go
@@ -133,7 +133,8 @@ func (rr *A) parse(c *zlexer, o, f string) *ParseError {
 	}
 
 	rr.A = net.ParseIP(l.token)
-	if rr.A == nil || l.err {
+	isIPv6 := strings.Contains(l.token, ":")
+	if rr.A == nil || isIPv6 || l.err {
 		return &ParseError{f, "bad A A", l}
 	}
 	return slurpRemainder(c, f)
@@ -146,7 +147,8 @@ func (rr *AAAA) parse(c *zlexer, o, f string) *ParseError {
 	}
 
 	rr.AAAA = net.ParseIP(l.token)
-	if rr.AAAA == nil || l.err {
+	isIPv6 := strings.Contains(l.token, ":")
+	if rr.AAAA == nil || !isIPv6 || l.err {
 		return &ParseError{f, "bad AAAA AAAA", l}
 	}
 	return slurpRemainder(c, f)
--- a/scan_rr.go
+++ b/scan_rr.go
@@ -133,8 +133,12 @@ func (rr *A) parse(c *zlexer, o, f string) *ParseError {
 	}
 
 	rr.A = net.ParseIP(l.token)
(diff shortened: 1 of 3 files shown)


scan_rr.go
```diff
@@ -133,7 +133,8 @@ func (rr *A) parse(c *zlexer, o, f string) *ParseError {
 	}
 
 	rr.A = net.ParseIP(l.token)
-	if rr.A == nil || l.err {
+	isIPv6 := strings.Contains(l.token, ":")
```
miekg: A better way (and use in other places in this lib) is to use rr.A.To4() and check if that's `nil` or not

scan_test.go
```diff
@@ -121,6 +121,59 @@ func TestZoneParserIncludeDisallowed(t *testing.T) {
...
+1.example.org. 600 IN AAAA ::1
+2.example.org. 600 IN AAAA ::FFFF:127.0.0.1`
+
+	wantRRs := []*AAAA{
+		&AAAA{Hdr: RR_Header{Name: "1.example.org."}, AAAA: net.IPv6loopback},
+		&AAAA{Hdr: RR_Header{Name: "2.example.org."}, AAAA: net.ParseIP("::FFFF:127.0.0.1")},
+	}
+
+	wantIdx := 0
+	zp := NewZoneParser(strings.NewReader(zone), "", "")
```
miekg: NewRR used the same code; these two tests can be collapsed to just checking the new code works, by testing
~~~
1.example.org. 600 IN AAAA ::1
2.example.org. 600 IN AAAA ::FFFF:127.0.0.1`
"1.bad.example.org. 600 IN A ::1",
"2.bad.example.org. 600 IN A ::FFFF:127.0.0.1",
"3.bad.example.org. 600 IN AAAA 127.0.0.1",
~~~

scan_rr.go
tomfitzhenry: I considered that, but that causes test failures because `net.ParseIP("::FFFF:127.0.0.1").To4()` is non-nil.

scan_test.go
tomfitzhenry: Done.

scan_rr.go
tomfitzhenry: https://github.com/asaskevich/govalidator/pull/100 is another project having observed the same issue.

miekg:
But that gives you the info you need, no?

scan_rr.go
miekg: https://play.golang.org/p/zIBXCABhfax
seems checking for colons is the only way :(

miekg: please add a comment saying this

tomfitzhenry: Done.

PR #924, opened 2019-03-08
[WIP/Demo] Use constant workers count

UladzimirTrehubenka:
Created just to compare results against https://github.com/miekg/dns/pull/922

### DIFF
--- a/server.go
+++ b/server.go
@@ -66,6 +66,8 @@ type ConnectionStater interface {
 }
 
 type response struct {
+	msg []byte
+
 	closed         bool // connection has been closed
 	hijacked       bool // connection has been hijacked by handler
 	tsigTimersOnly bool
@@ -201,6 +203,12 @@ type Server struct {
 	// By default DefaultMsgAcceptFunc will be used.
 	MsgAcceptFunc MsgAcceptFunc
 
+	// UDP packet or TCP connection queue
+	queue chan *response
+
+	// Number of workers, if set to zero - use spawn goroutines instead
+	Workers int
+
 	// Shutdown handling
 	lock     sync.RWMutex
 	started  bool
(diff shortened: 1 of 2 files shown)


miekg:
not confident merging anything here right now; we probably need a rethink and some kind of design fleshed out in an issue.

server.go
```diff
@@ -294,6 +309,17 @@ func makeUDPBuffer(size int) func() interface{} {
 func (srv *Server) init() {
 	srv.queue = make(chan *response)
 
+	const workersCount = 100
+	for i := 0; i < workersCount; i++ {
+		go func() {
+			for msg := range srv.queue {
+				msg.wg.Add(1)
```
tmthrgd: You can’t call `Add` from inside a goroutine because will end up concurrent with `Wait`.

server.go
```diff
@@ -294,6 +309,17 @@ func makeUDPBuffer(size int) func() interface{} {
 func (srv *Server) init() {
 	srv.queue = make(chan *response)
 
+	const workersCount = 100
```
tmthrgd: I’m very much against this kind of magic tuning variable. It’s impossible to pick a suitable value. Too small and performance is artificially limited. Too high and you’ll still get the OOM and crashes as before. What’s worse is that the correct value isn’t even static and depends on what the system is otherwise doing.

server.go
```diff
@@ -294,6 +309,17 @@ func makeUDPBuffer(size int) func() interface{} {
...
+	for i := 0; i < workersCount; i++ {
+		go func() {
```
tmthrgd: These goroutines are unnamed and won’t appear clearly in traces (`Server.init.func1`). They should be a seperate named function or method.

PR #927, opened 2019-03-08
Trying to revert the worker code

miekg:
this is an attempt to revert 98a1ef45. Left with a few conflicts and not
100% what to bring back because the reverted commit is a bit old.

But half way there, at least.

### DIFF
--- a/server.go
+++ b/server.go
@@ -12,22 +12,15 @@ import (
 	"net"
 	"strings"
 	"sync"
-	"sync/atomic"
 	"time"
 )
 
 // Default maximum number of TCP queries before we close the socket.
 const maxTCPQueries = 128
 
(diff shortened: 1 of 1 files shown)


miekg:
probably makes more sense to go forward; i.e. just delete the current code, then to actually revert this older commit. I'll try that as well.

miekg:
also want to remove w.msg from `response`, because its not part of the response. This triggers question about disposeBuffer

tmthrgd:
I wouldn’t try a `git revert` on this because it has changed a lot since then. If we ditch the worker code, we can go back to some of the cleaner function signatures that were there before the workers. In that case the message buffer and can be passed directly because it is needed up until `disposeBuffer`.

PR #928, opened 2019-03-08
Manually revert worker's goroutine 98a1ef45

miekg:
Manually revert the worker model.

### DIFF
--- a/server.go
+++ b/server.go
@@ -12,25 +12,11 @@ import (
 	"net"
 	"strings"
 	"sync"
-	"sync/atomic"
 	"time"
 )
 
 // Default maximum number of TCP queries before we close the socket.
 const maxTCPQueries = 128
-
-// The maximum number of idle workers.
-//
-// This controls the maximum number of workers that are allowed to stay
-// idle waiting for incoming requests before being torn down.
-//
-// If this limit is reached, the server will just keep spawning new
-// workers (goroutines) for each incoming request. In this case, each
-// worker will only be used for a single request.
-const maxIdleWorkersCount = 10000
-
-// The maximum length of time a worker may idle for before being destroyed.
-const idleWorkerTimeout = 10 * time.Second
 
 // aLongTimeAgo is a non-zero time, far in the past, used for
 // immediate cancelation of network operations.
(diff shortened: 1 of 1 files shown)


miekg:
ok,  much cleaner than #927, but failing a test; probably missed a `defer` somewhere that did a cleanup

miekg:
ok, failing because of wait group stuff

server.go
```diff
@@ -477,9 +408,7 @@ func (srv *Server) serveTCP(l net.Listener) error {
 		srv.NotifyStartedFunc()
 	}
 
-	var wg sync.WaitGroup
```
tmthrgd: This `sync.WaitGroup` is still needed for the graceful shutdown code.

server.go
```diff
@@ -522,9 +446,7 @@ func (srv *Server) serveUDP(l *net.UDPConn) error {
 		reader = srv.DecorateReader(reader)
 	}
 
-	var wg sync.WaitGroup
```
tmthrgd: Ditto here.

server.go
```diff
@@ -570,8 +489,6 @@ func (srv *Server) serve(w *response) {
 	if w.udp != nil {
 		// serve UDP
 		srv.serveDNS(w)
-
-		w.wg.Done()
```
tmthrgd: Ditto here.

server.go
```diff
@@ -583,8 +500,6 @@ func (srv *Server) serve(w *response) {
 		srv.lock.Lock()
 		delete(srv.conns, w.tcp)
 		srv.lock.Unlock()
-
-		w.wg.Done()
```
tmthrgd: Ditto here too.

tmthrgd:
I can cleanup some of these functions and get `msg` off of `response` in a follow-up PR. The server code was notably cleaner pre-worker.

miekg:
all done. I'll to the msg untangling as well on top of this one.

miekg:
#929 remove `msg` from response, this leads to a `nil` being passed to `serve()` which is somewhat ugly, but I can remember how that did look.

tmthrgd:
@mielg Can this PR be retitled so it’s more obvious in the git history?

PR #929, opened 2019-03-09
Move msg []byte out of response

miekg:

### DIFF
--- a/server.go
+++ b/server.go
@@ -12,25 +12,11 @@ import (
 	"net"
 	"strings"
 	"sync"
-	"sync/atomic"
 	"time"
 )
 
 // Default maximum number of TCP queries before we close the socket.
 const maxTCPQueries = 128
-
-// The maximum number of idle workers.
-//
-// This controls the maximum number of workers that are allowed to stay
-// idle waiting for incoming requests before being torn down.
-//
-// If this limit is reached, the server will just keep spawning new
-// workers (goroutines) for each incoming request. In this case, each
-// worker will only be used for a single request.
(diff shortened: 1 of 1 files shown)


server.go
```diff
@@ -542,15 +542,15 @@ func (srv *Server) serve(w *response, wg *sync.WaitGroup) {
...
 
-func (srv *Server) disposeBuffer(w *response) {
-	if w.udp != nil && cap(w.msg) == srv.UDPSize {
-		srv.udpPool.Put(w.msg[:srv.UDPSize])
+func (srv *Server) disposeBuffer(udp bool, m []byte) {
+	if udp && cap(m) == srv.UDPSize {
+		srv.udpPool.Put(m[:srv.UDPSize])
 	}
-	w.msg = nil
+	m = nil
```
tmthrgd: This isn’t needed anymore. It won’t block GC of the buffer as it’s only acessable during this function call.

tmthrgd:
https://github.com/miekg/dns/pull/928#issuecomment-471159380
> #929 remove `msg` from response, this leads to a `nil` being passed to `serve()` which is somewhat ugly, but I can remember how that did look.

Originally it looked something like this, then I went and split up the code into two seperate `serveTCP` and `serveUDP` methods that were substantially cleaner. The worker code forced this all back through a single method. I’ll try and find my old PRs that cleaned this up.

server.go
miekg: the `= nil` part you mean?

miekg:
#930 because bases this of the *reverty* branch exceeded my git skills

PR #930, opened 2019-03-09
Remove msg member from response

miekg:
Remove the `msg []byte` member from the response as it does not belong
there.

### DIFF
--- a/server.go
+++ b/server.go
@@ -66,8 +66,8 @@ type ConnectionStater interface {
 	ConnectionState() *tls.ConnectionState
 }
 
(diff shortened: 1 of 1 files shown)


miekg:
@tmthrgd merge this? Or do you have bigger plans?

tmthrgd:
@miekg I'm just about to open a PR that splits up `Server.serve` and should be cleaner. It get's rid of the `msg` field and is closer to the pre-worker seperation.

miekg:
ok, let's abandon this

PR #933, opened 2019-03-10
Split Server.serve into separate TCP and UDP methods

tmthrgd:
This restores the pre-worker method split of `Server.serveTCPConn` and `Server.serveUDPPacket`. This also removes the `msg` field from the `response` struct. It also reduces some duplication in `Server.serveDNS`.

Closes #930

### DIFF
--- a/server.go
+++ b/server.go
@@ -551,12 +551,6 @@ func (srv *Server) serveUDPPacket(wg *sync.WaitGroup, m []byte, u *net.UDPConn, 
 	wg.Done()
 }
 
-func (srv *Server) disposeBuffer(m []byte, w *response) {
-	if w.udp != nil && cap(m) == srv.UDPSize {
-		srv.udpPool.Put(m[:srv.UDPSize])
-	}
-}
-
 func (srv *Server) serveDNS(m []byte, w *response) {
 	dh, off, err := unpackMsgHdr(m, 0)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


server.go
```diff
@@ -545,15 +538,21 @@ func (srv *Server) serve(w *response, wg *sync.WaitGroup) {
 	wg.Done()
 }
 
-func (srv *Server) disposeBuffer(w *response) {
-	if w.udp != nil && cap(w.msg) == srv.UDPSize {
-		srv.udpPool.Put(w.msg[:srv.UDPSize])
+// Serve a new UDP request.
+func (srv *Server) serveUDPPacket(wg *sync.WaitGroup, m []byte, u *net.UDPConn, s *SessionUDP) {
+	w := &response{tsigSecret: srv.TsigSecret, udp: u, udpSession: s}
+	if srv.DecorateWriter != nil {
```
miekg: this can be put in init as well?

tmthrgd: This is used to set a field on the `response` rather than the `Server` like the other ones did.

That being said, we could add a `srv.DecorateWriter = func(w Writer) Writer { return w }` and similar for `DecorateReader` in `Server.init`, though I'm not sure how clean that would end up being.

miekg: ah, true, my mistake, I just saw `srv` and assumed.

PR #934, opened 2019-03-10
Use net.Buffers for writing TCP message

tmthrgd:
This is much cleaner than the existing code and it avoids copying the message buffer. This should see a (small?) performance improvement for TCP connections, while possibly being a (small?) performance loss for TLS connections.

I want to similarly clean up the TCP reading code, but I'm not quite sure how best to handle the subtle error dance.

### DIFF
--- a/client.go
+++ b/client.go
@@ -3,7 +3,6 @@ package dns
 // A client implementation.
 
 import (
-	"bytes"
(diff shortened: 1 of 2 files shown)


miekg:
looks good.

PR #935, opened 2019-03-10
Simplify TCP reading

tmthrgd:
The previous code is quite complicated and has some very subtle error behaviour, while this new code is far simpler and more obvious while behaving very similarly.

While there are a select number of subtle changes in error behaviour, I don't believe it will be noticeable in normal use or that anyone would be relying on it. It ultimately wasn't particularly helpful.

It's very late here and this code was quite complicated, so it may be worth running the CoreDNS tests against this PR (which also includes the TCP writing changes from master) before merging.

### DIFF
--- a/client.go
+++ b/client.go
@@ -227,9 +227,7 @@ func (co *Conn) ReadMsgHeader(hdr *Header) ([]byte, error) {
 		}
 
 		p = make([]byte, length)
-		if n, err = io.ReadFull(co.Conn, p); err != nil {
-			return nil, err
-		}
+		n, err = io.ReadFull(co.Conn, p)
 	default:
 		if co.UDPSize > MinMsgSize {
 			p = make([]byte, co.UDPSize)
--- a/client.go
+++ b/client.go
@@ -219,18 +219,17 @@ func (co *Conn) ReadMsgHeader(hdr *Header) ([]byte, error) {
 		n   int
 		err error
 	)
-
-	switch t := co.Conn.(type) {
+	switch co.Conn.(type) {
 	case *net.TCPConn, *tls.Conn:
-		r := t.(io.Reader)
-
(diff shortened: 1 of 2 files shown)


miekg:
thanks, looks good to me. Now figuring how to test a branch with go modules

miekg:
Ran so some local tests, looks all OK. Easiest seems to merge, up the version and send a PR to coredns.

tmthrgd:
@miekg From following the go modules discussions on the issue tracker, you should be able to add a `replace` directive to the checked out CoreDNS `go.mod` file to point to a local copy of miekg/dns. Otherwise the merge, up version approach seems fine to me.

client.go
```diff
@@ -219,18 +219,15 @@ func (co *Conn) ReadMsgHeader(hdr *Header) ([]byte, error) {
...
-	switch t := co.Conn.(type) {
+	switch co.Conn.(type) {
 	case *net.TCPConn, *tls.Conn:
-		r := t.(io.Reader)
-
-		// First two bytes specify the length of the entire message.
-		l, err := tcpMsgLen(r)
-		if err != nil {
+		var length uint16
+		if err := binary.Read(co.Conn, binary.BigEndian, &length); err != nil {
```
miekg: is there any way to test this for short reads, i.e. see comments in tcpMsgLen where you only get 1 byte on this read?

miekg:
ah, (need to write this down somewhere), I symlinked g/pkg/mod/... to another copy of this lib

client.go
tmthrgd: `binary.Read` uses `io.ReadFull` internally which handles this case. That being said, I could try and build a test using `testing/iotest`.

miekg:
I think that would be useful, cause that behaviour was really wtf moment
for me

PR #937, opened 2019-03-13
Use for range loops instead of manual for loops

tmthrgd:
This replaces the use of manual `for i := 0; i < len(...); i++` loops with simpler and more idiomatic `for range` loops.

`Msg.CopyTo` had a slight reorganisation of it's logic which leaves it much simpler: 5bab01b0a655730908f4523a5b2510d80a515167.

### DIFF
--- a/clientconfig.go
+++ b/clientconfig.go
@@ -68,14 +68,10 @@ func ClientConfigFromReader(resolvconf io.Reader) (*ClientConfig, error) {
 			}
 
 		case "search": // set search path to given servers
-			c.Search = make([]string, len(f)-1)
-			for i := 0; i < len(c.Search); i++ {
-				c.Search[i] = f[i+1]
-			}
+			c.Search = append([]string(nil), f[1:]...)
 
 		case "options": // magic options
-			for i := 1; i < len(f); i++ {
-				s := f[i]
+			for _, s := range f[1:] {
 				switch {
 				case len(s) >= 6 && s[:6] == "ndots:":
 					n, _ := strconv.Atoi(s[6:])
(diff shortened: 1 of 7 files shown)


clientconfig.go
```diff
@@ -68,14 +68,10 @@ func ClientConfigFromReader(resolvconf io.Reader) (*ClientConfig, error) {
 			}
 
 		case "search": // set search path to given servers
-			c.Search = make([]string, len(f)-1)
-			for i := 0; i < len(c.Search); i++ {
-				c.Search[i] = f[i+1]
-			}
+			c.Search = append([]string(nil), f[1:]...)
```
miekg: `string(nil)`? can't use `[]string{}` ?

dnssec.go
```diff
@@ -556,9 +556,9 @@ func (k *DNSKEY) publicKeyRSA() *rsa.PublicKey {
 	pubkey := new(rsa.PublicKey)
 
 	var expo uint64
-	for i := 0; i < int(explen); i++ {
+	for _, v := range keybuf[keyoff:modoff] {
```
miekg: found that `explen` give some extra documentation here, add comment that this loops is executed `explen` times?

tmthrgd: @miekg Can do. I’ll document it tomorrow when I’m back at my computer.

clientconfig.go
tmthrgd: Idiomatic go is to use `nil` here. It needs the type of the slice, but not the actual slice. It’s straight out of the wiki: https://github.com/golang/go/wiki/SliceTricks#copy.

dnssec.go
tmthrgd: @miekg How is that comment?

Issue #685, opened 2018-05-21
Transfer In and Out need some tests and polish


jeffbean:
Based on trying to unit test the Transfer In I ran into issues when doing so. The surrounding code to Transfer looks to be known to be a work in progress with TODO's. `Think we need to away to stop the transfer` 

I am getting i/o read timeouts based on the tests seen here. https://raw.githubusercontent.com/jeffbean/dns/dffc41a423f896c0b7cf9da6def2b561890495c4/xfr_test.go

I have not invested to much time looking at solutions but the test code shows  how flaky the `transfer.In` is without better coordination of the connection. This issue will serve as a tracking task for myself as well on work produced to fix these issues around flaky results.

My forked branch can be found here https://github.com/jeffbean/dns/tree/adding-xfr-tests 

and to reproduce the timeouts run `go test ./... -count 50`

miekg:
[ Quoting <notifications@github.com> in "[miekg/dns] Transfer In and Out nee..." ]
>Based on trying to unit test the Transfer In I ran into issues when doing so. The surrounding code to Transfer looks to be known to be a work in progress with TODO's. `Think we need to away to stop the transfer`
>
>I am getting i/o read timeouts based on the tests seen here. https://raw.githubusercontent.com/jeffbean/dns/dffc41a423f896c0b7cf9da6def2b561890495c4/xfr_test.go
>
>I have not invested to much time looking at solutions but the test code shows  how flaky the `transfer.In` is without better coordination of the connection. This issue will serve as a tracking task for myself as well on work produced to fix these issues around flaky results.
>
>My forked branch can be found here https://github.com/jeffbean/dns/tree/adding-xfr-tests
>
>and to reproduce the timeouts run `go test ./... -count 50`

Haven't run that tests, but some cleanups and possibly fixes are welcome of 
course.

Note that in my use of CoreDNS as a secondary this seems to work quite well.

jeffbean:
I think it might even be the testing setup using Transfer itself as the server response. I will try to work on a PR for `xfr.go` itself to get it reliably testable. It works fine in production so just getting testability is important as well.

miekg:
This issue looks stale, so I'll close it soon.

miekg:
we've merged a thing referencing this, closing

PR #939, opened 2019-03-14
Set TSIG on Transfer Out

NullOranje:
Requests sent via `Transfer.Out()` do not currently check if the request has TSIG record, so those responses are sent without expected TSIG record(s).  This follows the pattern described in the documentation and adds a check to `q *Msg` for a TSIG record and, if present, sets the TSIG on the outgoing message(s).

### DIFF
--- a/xfr.go
+++ b/xfr.go
@@ -198,11 +198,16 @@ func (t *Transfer) Out(w ResponseWriter, q *Msg, ch chan *Envelope) error {
 		r.Authoritative = true
 		// assume it fits TODO(miek): fix
 		r.Answer = append(r.Answer, x.RR...)
+		if tsig := q.IsTsig(); tsig != nil {
+			if w.TsigStatus() == nil {
+				r.SetTsig(tsig.Hdr.Name, tsig.Algorithm, tsig.Fudge, time.Now().Unix())
+			}
+		}
 		if err := w.WriteMsg(r); err != nil {
 			return err
 		}
+		w.TsigTimersOnly(true)
 	}
-	w.TsigTimersOnly(true)
 	return nil
 }
 
--- a/xfr.go
+++ b/xfr.go
@@ -198,10 +198,8 @@ func (t *Transfer) Out(w ResponseWriter, q *Msg, ch chan *Envelope) error {
 		r.Authoritative = true
 		// assume it fits TODO(miek): fix
 		r.Answer = append(r.Answer, x.RR...)
-		if tsig := q.IsTsig(); tsig != nil {
-			if w.TsigStatus() == nil {
-				r.SetTsig(tsig.Hdr.Name, tsig.Algorithm, tsig.Fudge, time.Now().Unix())
-			}
+		if tsig := q.IsTsig(); tsig != nil && w.TsigStatus() == nil {
+			r.SetTsig(tsig.Hdr.Name, tsig.Algorithm, tsig.Fudge, time.Now().Unix())
 		}
 		if err := w.WriteMsg(r); err != nil {
 			return err
--- a/xfr_test.go
+++ b/xfr_test.go
@@ -1,7 +1,6 @@
 package dns
 
 import (
-	"fmt"
 	"net"
 	"sync"
 	"testing"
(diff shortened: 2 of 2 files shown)


NullOranje:
Certainly.  `xfr.go` needs tests generally.  #685 is tracking this, but I think that issue stale.  I'll put something together for this PR.  It would be good to revisit the issue of tests for this generally.

xfr.go
```diff
@@ -198,11 +198,16 @@ func (t *Transfer) Out(w ResponseWriter, q *Msg, ch chan *Envelope) error {
 		r.Authoritative = true
 		// assume it fits TODO(miek): fix
 		r.Answer = append(r.Answer, x.RR...)
+		if tsig := q.IsTsig(); tsig != nil {
```
tmthrgd: Can you merge these two `if`-statements? So: `tsig != nil && w.TsigStatus() == nil`.

xfr_test.go
```diff
@@ -0,0 +1,195 @@
...
+var (
+	tsigSecret = map[string]string{"axfr.": "so6ZGir4GPAqINNh9U5c3A=="}
+	xfrSoa     = `miek.nl.	0	IN	SOA	linode.atoom.net. miek.miek.nl. 2009032802 21600 7200 604800 3600`
+	xfrA       = `x.miek.nl.	1792	IN	A	10.0.0.1`
+	xfrMX      = `miek.nl.	1800	IN	MX	1	x.miek.nl.`
+	testData   = axfrZoneData()
+)
+
+func axfrZoneData() []RR {
+	soa, _ := NewRR(xfrSoa)
```
tmthrgd: Use `testRR` instead of `NewRR`here.

xfr_test.go
```diff
@@ -0,0 +1,195 @@
...
+	"testing"
+	"time"
+)
+
+var (
+	tsigSecret = map[string]string{"axfr.": "so6ZGir4GPAqINNh9U5c3A=="}
+	xfrSoa     = `miek.nl.	0	IN	SOA	linode.atoom.net. miek.miek.nl. 2009032802 21600 7200 604800 3600`
+	xfrA       = `x.miek.nl.	1792	IN	A	10.0.0.1`
+	xfrMX      = `miek.nl.	1800	IN	MX	1	x.miek.nl.`
+	testData   = axfrZoneData()
```
tmthrgd: I don’t think this needs to be a seperate function and should be renamed. `xfrTestData = []RR{...}` and then `xfrSoa = testRR("...")`.

xfr_test.go
```diff
@@ -0,0 +1,195 @@
...
+
+	s, addrstr, err := RunLocalTCPServerWithTsig(":0", tsigSecret)
+	if err != nil {
+		t.Fatalf("unable to run test server: %s", err)
+	}
+	defer s.Shutdown()
+
+	testCases := []struct {
+		name string
+		tsig map[string]string
```
tmthrgd: This field isn’t used for anything and the two sub-tests are identical.

xfr_test.go
```diff
@@ -0,0 +1,195 @@
...
```
tmthrgd: Ditto here.

xfr_test.go
```diff
@@ -0,0 +1,195 @@
...
+				records = append(records, rr)
+			}
+		}
+
+		if len(records) != len(testData) {
+			t.Fatalf("bad axfr: expected %v, got %v", records, testData)
+		}
+
+		for i := range records {
+			if records[i].String() != testData[i].String() {
```
tmthrgd: I don’t think this is the proper way to compare RRs.

NullOranje: I suspect not.  Would it make more sense to use `isDuplicate()` instead?

```
for i := range records {
	if !records[i].isDuplicate(xfrTestData[i]) {
		t.Fatalf("bad axfr: expected %v, got %v", records, xfrTestData)
	}
}
```

tmthrgd: Yes it would, but use the exported `IsDuplicate` function and not the I exported method.

xfr_test.go
```diff
@@ -0,0 +1,165 @@
...
+		for msg := range c {
+			if msg.Error != nil {
+				t.Fatal(msg.Error)
+			}
+			for _, rr := range msg.RR {
+				records = append(records, rr)
+			}
+		}
+
+		if len(records) != 4 {
```
tmthrgd: Don’t hardcore the 4 here, it’s better with `len(xfrTestData)`.

xfr_test.go
NullOranje: Thanks.  I wasn't tracking the exported version.

miekg:
I'll take another look, but afaict this can just be merged.

PR #942, opened 2019-03-18
Remove pointless cast in Conn.Read

tmthrgd:
This was accidentally added in #935, though I'm not quite sure why I ended up with this cast.


miekg:
client.go has another one of these, as does server.go (at least grep shows this)

tmthrgd:
@miekg The only line I see is the `net.Buffers` one, but the `WriteTo` method returns an `int64` so that cast is needed.

PR #947, opened 2019-03-19
Add constants for missing registered EDNS option codes

gibson042:
https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#dns-parameters-11

Value | Name | Status | Reference
-- | -- | -- | --
13 | CHAIN | Standard | [[RFC7901](https://tools.ietf.org/html/rfc7901)]
14 | edns-key-tag | Optional | [[RFC8145](http://www.iana.org/go/rfc8145)]

### DIFF
--- a/edns.go
+++ b/edns.go
@@ -71,10 +71,11 @@ func (rr *OPT) String() string {
 			s += "\n; DS HASH UNDERSTOOD: " + o.String()
 		case *EDNS0_N3U:
 			s += "\n; NSEC3 HASH UNDERSTOOD: " + o.String()
-		case *EDNS0_LOCAL:
-			s += "\n; LOCAL OPT: " + o.String()
 		case *EDNS0_PADDING:
 			s += "\n; PADDING: " + o.String()
+
+		case *EDNS0_LOCAL:
+			s += "\n; LOCAL OPT: " + o.String()
 		}
 	}
 	return s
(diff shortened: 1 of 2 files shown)


edns.go
```diff
@@ -69,10 +71,11 @@ func (rr *OPT) String() string {
 			s += "\n; DS HASH UNDERSTOOD: " + o.String()
 		case *EDNS0_N3U:
 			s += "\n; NSEC3 HASH UNDERSTOOD: " + o.String()
-		case *EDNS0_LOCAL:
```
tmthrgd: Why are this file being rearranged? It doesn’t have anything to do with the constants.

msg_helpers.go
```diff
@@ -474,6 +474,8 @@ Option:
 		}
 		edns = append(edns, e)
 		off += int(optlen)
+
+	// Process any unrecognized option code as a local/experimental option.
```
tmthrgd: I don’t think this comment adds much.

gibson042: I consider it important to note because `EDNS_LOCAL` at a glance _looks_ similar to the other `EDNS0_*` structures, but is in fact a generic container that can contain anything, including options that have been standardized but are not yet handled by this library.

edns.go
gibson042: Some cleanup in anticipation of introducing `EDNS0_CHAIN` and `EDNS0_KEYTAG` types (specifically, always defining and dealing with the generic fallback type last). The changes are completely separable from introducing the new constants, though, and in fact already have their own commit. I can revert that aspect upon request if you'd rather not see it here.

miekg:
Where is the code that implements the options?

gibson042:
It has not yet been written; this just introduces the constants so consuming code can use e.g. `option.Code == dns.EDNS0KEYTAG` rather than `option.Code == 14`. I believe the constants alone provide sufficient value to land by replacing opaque numbers with descriptive names, but if you insist on the types and behavior as well then it may be a while because I don't think I have the time for them right now.

miekg:
Yes, it will be better do this in one Go instead of giving the impression we've implemented it, without providing the pack and unpack code.

miekg:
If you're unable to do this, I'll close this PR.

PR #950, opened 2019-03-27
Add Windows to Travis build

tmthrgd:
This is a WIP.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,11 @@ go:
   - 1.12.x
   - tip
 
+matrix:
+  include:
+    - os: windows
+      go: 1.12.x
+
 before_install:
   # don't use the miekg/dns when testing forks
   - mkdir -p $GOPATH/src/github.com/miekg


miekg:
Hmm, not a fan. This requires me to wait longer for an OS that I haven't bought and is closed source for any change that I do

tmthrgd:
@miekg I'm not sure I follow what your concern is here. The builder is only marginally slower than the existing tip builder. This repo also contains Windows-specific code that has broken inadvertently in the past.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Add Windows to Trav..." ]
>@miekg I'm not sure I follow what your concern is here. The builder is only marginally slower than the existing tip builder. This repo also contains Windows-specific code that has broken inadvertently in the past.

I can't fix windows bugs and I still don't really care about closed source OSes. 
If someone cares about this OS, they can fork, setup testing and feed us fixes.

PR #957, opened 2019-04-24
Handle all net.Conn connections correctly

Pepijnvi:
This change allows alternative net.Conn implementations such as in memory connections based on net.Pipe for implementation testing.

### DIFF
--- a/client.go
+++ b/client.go
@@ -256,21 +256,20 @@ func (co *Conn) Read(p []byte) (n int, err error) {
 		return 0, ErrConnEmpty
 	}
 
-	switch co.Conn.(type) {
-	case *net.TCPConn, *tls.Conn:
-		var length uint16
-		if err := binary.Read(co.Conn, binary.BigEndian, &length); err != nil {
-			return 0, err
-		}
-		if int(length) > len(p) {
-			return 0, io.ErrShortBuffer
-		}
-
-		return io.ReadFull(co.Conn, p[:length])
+	if _, ok := co.Conn.(net.PacketConn); ok {
+		// UDP connection
+		return co.Conn.Read(p)
 	}
 
-	// UDP connection
-	return co.Conn.Read(p)
+	var length uint16
+	if err := binary.Read(co.Conn, binary.BigEndian, &length); err != nil {
+		return 0, err
+	}
+	if int(length) > len(p) {
+		return 0, io.ErrShortBuffer
+	}
+
+	return io.ReadFull(co.Conn, p[:length])
 }
 
 // WriteMsg sends a message through the connection co.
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -299,7 +301,9 @@ func (co *Conn) WriteMsg(m *Msg) (err error) {
 // Write implements the net.Conn Write method.
 func (co *Conn) Write(p []byte) (n int, err error) {
 	switch co.Conn.(type) {
-	case *net.TCPConn, *tls.Conn:
+	case *net.UDPConn:
+		break
+	case net.Conn:
 		if len(p) > MaxMsgSize {
 			return 0, &Error{err: "message too large"}
```
miekg: actually this length check makes sense regardless of transport used.

client.go
```diff
@@ -225,12 +232,7 @@ func (co *Conn) ReadMsgHeader(hdr *Header) ([]byte, error) {
 		p = make([]byte, length)
 		n, err = io.ReadFull(co.Conn, p)
 	default:
-		if co.UDPSize > MinMsgSize {
-			p = make([]byte, co.UDPSize)
-		} else {
-			p = make([]byte, MinMsgSize)
-		}
-		n, err = co.Read(p)
+		fmt.Errorf("unknown connection type: %T", co.Conn)
```
miekg: can this even happen?

client.go
Pepijnvi: Agreed moved upwards.

client.go
Pepijnvi: Unlikely as most connection will be catched by net.Conn anyway made that the default.

client.go
```diff
@@ -298,12 +298,14 @@ func (co *Conn) WriteMsg(m *Msg) (err error) {
...
-	case *net.TCPConn, *tls.Conn:
-		if len(p) > MaxMsgSize {
-			return 0, &Error{err: "message too large"}
-		}
+	if len(p) > MaxMsgSize {
+		return 0, &Error{err: "message too large"}
+	}
 
+	switch co.Conn.(type) {
+	case *net.UDPConn:
```
tmthrgd: This should instead be:
```go
if u, ok := co.Conn.(*net.UDPConn); ok {
    return u.Write(p)
}

l := make([]byte, 2)
...
```

client.go
```diff
@@ -216,21 +216,21 @@ func (co *Conn) ReadMsgHeader(hdr *Header) ([]byte, error) {
 		err error
 	)
 	switch co.Conn.(type) {
-	case *net.TCPConn, *tls.Conn:
+	case *net.UDPConn:
```
tmthrgd: I’d say replace this `switch` statement with an `if`/`else`.

tmthrgd: Switch the `*net.UDPConn` typecast with `net.PacketConn`.

miekg:
think this handles all comments and can be merged

miekg:
ping @Pepijnvi are you able to make these last change(s)? tx

PR #964, opened 2019-05-09
Add support for the Proposed DNS XPF RR

tomsanbear:
As described in this RFC draft [(IETF Draft)](https://tools.ietf.org/html/draft-bellis-dnsop-xpf-04), the DNS XPF RR implementation adds meta information about the original destination and source to the DNS request. Something that I made a decision on was the Record Type, I went with the option code that Wireshark is currently mapping the record to (65422). This would need updating if/when the RFC goes through. 

Another reference implementation that was used as inspiration was from the PowerDNS dnsdist feature that was added.

Adding this here as I'd like to get this feature supported in CoreDNS eventually.

Feedback appreciated!

### DIFF
--- a/duplicate_generate.go
+++ b/duplicate_generate.go
@@ -57,7 +57,7 @@ func main() {
 			continue
 		}
 
-		if name == "PrivateRR" || name == "OPT" {
+		if name == "PrivateRR" || name == "OPT" || name == "XPF" {
 			continue
 		}
 
--- a/types.go
+++ b/types.go
@@ -249,6 +249,25 @@ func (rr *XPF) parse(c *zlexer, origin, file string) *ParseError {
 	panic("dns: internal error: parse should never be called on XPF")
 }
 
+func (rr *XPF) len(off int, compression map[string]struct{}) int {
+	l := rr.Hdr.len(off, compression)
+	l += 1 // IpVersion
+	l += 1 // Protocol
+	switch rr.IpVersion {
+	case 4:
+		l += 4 // SrcAddr
+		l += 4 // DestAddr
+
+	case 6:
+		l += 8 // SrcAddr
+		l += 8 // DestAddr
+	}
+	l += 2 // SrcPort
+	l += 2 // DestPort
+	//
+	return l
+}
+
 // ANY is a wildcard record. See RFC 1035, Section 3.2.3. ANY
 // is named "*" there.
 type ANY struct {
--- a/types.go
+++ b/types.go
@@ -106,6 +106,7 @@ const (
 
 	TypeTA       uint16 = 32768
 	TypeDLV      uint16 = 32769
+	TypeXPF      uint16 = 65422
 	TypeReserved uint16 = 65535
 
 	// valid Question.Qclass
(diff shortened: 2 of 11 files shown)


xpf.go
```diff
@@ -0,0 +1,81 @@
...
+
+import (
+	"fmt"
+	"net"
+)
+
+// XPF RR. See https://tools.ietf.org/html/draft-bellis-dnsop-xpf-04
+type XPF struct {
+	Hdr  RR_Header
+	Data XPF_Data `dns:"xpf"` // This is seperated out to make this more compatible with the code generation
```
tmthrgd: I don’t like the use of a seperate structure like this. Aside from the address fields, this seems easy to parse with the existing generated code.

I think a better approach would be something like `dns:"ip-family:IpVersion"` (bikeshed away; "a/aaaa:IpVersion" maybe) and then add support for that tag to the generator.

duplicate_generate.go
```diff
@@ -57,7 +57,7 @@ func main() {
 			continue
 		}
 
-		if name == "PrivateRR" || name == "OPT" {
+		if name == "PrivateRR" || name == "OPT" || name == "XPF" {
```
tmthrgd: This generator will work fine for this type. All it needs is to be updated to understand that a new ‘possibly A or AAAA’ field type should be compared as an IP address like `dns:"a"`.

xpf.go
```diff
@@ -0,0 +1,80 @@
+package dns
+
+import (
+	"fmt"
+	"net"
+)
+
+// XPF RR. See https://tools.ietf.org/html/draft-bellis-dnsop-xpf-04
+type XPF struct {
```
tmthrgd: Once you switch to adding a new field type to the generators, this should just go in `types.go`.

xpf.go
```diff
@@ -0,0 +1,80 @@
...
+	Hdr  RR_Header
+	Data XPF_Data `dns:"xpf"` // This is seperated out to make this more compatible with the code generation
+}
+
+// XPF_Data encapsulates the xpf data for custom deserialization
+type XPF_Data struct {
+	IpVersion   uint8
```
tmthrgd: This should be `IPVersion`.

xpf.go
```diff
@@ -0,0 +1,80 @@
...
+		return false
+	}
+	if xpf_data.DestPort != xpf_data2.DestPort {
+		return false
+	}
+	return true
+}
+
+func (rr *XPF) String() string {
+	return fmt.Sprintf("%v Source=%v:%v Destination=%v:%v", rr.Hdr.String(), rr.Data.SrcAddress, rr.Data.SrcPort, rr.Data.DestAddress, rr.Data.DestPort)
```
tmthrgd: This should be consistent with `NULL`, `TSIG` and `OPT`. It should output something prefixed by a comment so it can be passed back into the `ZoneParser` (`NewRR`) without causing an error.

types.go
```diff
@@ -106,6 +106,7 @@ const (
 
 	TypeTA       uint16 = 32768
 	TypeDLV      uint16 = 32769
+	TypeXPF      uint16 = 65422
```
miekg: this can only be added once a code has been allocated. This is I believe in the private range; if you want this to work now in your code you'll need to fiddle with a PrivateRR.

Has a request been sent to IANA for a code allocation point?

miekg:
The only thing weird in this record are the address fields, this can potentially just be split into two members, depending on the value of family:

~~~
IPVersion uint8
Protocol uint8 
SrcA net.ip `dns:"a"`
SrcAAAA net.ip `dns:"aaaa"`
DstA net.ip `dns:"a"`
DstAAAA net.ip `dns:"aaaa"`
SrcPort uint16
DstPort uint16                     
~~~

This also means the addresses can be assigned similar to `A` and `AAAA` records

types.go
tomsanbear: I'm not all too sure about the state of the proposal. It's something that PowerDNS seemed to have been trying to push through as they did the initial draft. 

I'm waiting to hear back from the original author(s) whether this was something that just got left out to dry, or if it's in limbo with IANA/IETF.

miekg: Once you have a draft you can allocate a code. The process it relatively light weight nowadays, you need to fill out a form and give that to iana

tomsanbear:
Alright, given the situation of this type not having an official RRType Code, my option for integrating this into any external code for the time being is via the PrivateRR type, until such a time that the draft is accepted, or another proposal with a similar feature set is introduced. 

Just some context on what I am hoping to achieve with this... I'm mostly looking to get this functionality added to the Coredns via either the Forward plugin (since I can control dropping the record) or a custom plugin. I've got a use case that requires information about the client request being passed to the upstream DNS server.

If you have any other suggestion in regards to direction with this that'd be great, otherwise, I'll close this PR for now.

miekg:
Ok sgtm

adding a new record should not require any code changes unless you need a) presentation format b) extra member fields in an easy way

CoreDNS will just work because it will treat it as a RFC3597  - unknown RR, which will deliver the payload as base64 encoded bytes

PR #966, opened 2019-05-10
packDataOpt: fix overflow in packing opt

chantra:
An option needs a minimum of 4 bytes that contains OPTION-CODE and OPTION-LENGTH
The code was checkign off + 3 > len(msg) instead of off + 4

### DIFF
--- a/fuzz_test.go
+++ b/fuzz_test.go
@@ -1,0 +1,73 @@
+package dns
+
+import (
+	"net"
+	"testing"
+)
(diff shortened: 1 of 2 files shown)


chantra:
cc @miekg

chantra:
@miekg anything you need me to address before merging this? Once it is merged, I will be able to rebase #967 on top of it which should take care of some of the comments you had.

PR #967, opened 2019-05-10
Fix crash nsec

chantra:


chantra:
cc @miekg

chantra:
BTW, this contains the diff from #966 . The reason I included it in this diff too is because both those diff are writing to this newly created `fuzz_test.go` .

fuzz_test.go
```diff
@@ -0,0 +1,109 @@
...
+// TestCrashNSEC tests generated using fuzz.go and with a message pack
+// containing the following bytes:
+// "0000\x00\x00000000\x00\x00/00000" +
+// "0\x00\v\x00#\b00000\x00\x00\x00\x00\x00\x1a000" +
+// "000\x00\x00\x00\x00\x1a000000\x00\x00\x00\x00\x1a0" +
+// "00000\x00\v00\a0000000\x00"
+// That byte sequence, when Unpack() and subsequential Pack() created a
+// panic: runtime error: slice bounds out of range
+// github.com/miekg/dns.(*Msg).packBufferWithCompressionMap(0xc0000d4000, 0x0, 0x0, 0x0, 0x0, 0x0, 0x100, 0x14, 0x14e80b, 0xbf2d4654d501a3c8, ...)
+// /Users/chantra/go/src/github.com/miekg/dns/msg.go:868 +0x13a8
```
tmthrgd: Don’t include this path or the `packBufferWithCompress` line above it. They’ll both get out of date really quickly.

types.go
```diff
@@ -854,14 +854,20 @@ func (rr *NSEC) String() string {
...
-		if uint32(window) != lastwindow {
-			l += 1 + 32
+		length := (t-window*256)/8 + 1
+		if window > lastwindow && lastlength != 0 { // New window, jump to the new offset
+			l += int(lastlength) + 2
+			lastlength = 0
 		}
-		lastwindow = uint32(window)
+		if window < lastwindow || length < lastlength {
+			continue // FIXME: packDataNsec would return Error{err: "nsec bits out of order"}
```
tmthrgd: What do you propose should be done here? `len` can’t and shouldn’t return an error, instead returning some sort of best effort length is appropriate as I believe this code already does.

fuzz_test.go
```diff
@@ -0,0 +1,109 @@
...
+)
+
+// TestPackDataOpt tests generated using fuzz.go and with a message pack
+// containing the following bytes:
+// "0000\x00\x00000000\x00\x00/00000" +
+// "0\x00\v\x00#\b00000000\x00\x00)000" +
+// "000\x00\x1c00\x00\x0000\x00\x01000\x00\x00\x00\b" +
+// "\x00\v\x00\x02\x0000000000"
+// That bytes sequence created the overflow error.
+func TestPackDataOpt(t *testing.T) {
```
miekg: could you move this to #966 ?

fuzz_test.go
miekg: +1 on removing here. Potentially put it in the commit mesg

msg_helpers.go
```diff
@@ -495,7 +495,7 @@ Option:
 func packDataOpt(options []EDNS0, msg []byte, off int) (int, error) {
 	for _, el := range options {
 		b, err := el.pack()
-		if err != nil || off+3 > len(msg) {
+		if err != nil || off+4 > len(msg) {
```
miekg: could you move this to #966 as well?

types.go
miekg: yeah, we can't error here, because we just want a length.

So the trigger is that the window count is out of order (has been a while, since this code was written)? I mean is the actual lenght calculation wrong, and the pack/unpack code a bit too liberal?

chantra: I have little knowledge of NSEC RR. As far as I can tell this will return the actual length of the packet, valid or not, and packing will take care of erroring at a later stage. What really matters is that len() never return a length smaller than what could be used during packing, or we could overflow the buffer. I think it is fine to just drop the FIXME.

fuzz_test.go
chantra: This is already in #966. This PR is made of 2 diffs, the one from dataopt (#966) and the nsec one. Because fuzz_test.go was newly created, it would conflict. It will be easier to just rebase the last commotion top of 966 once it is landed (which is what I did in this diff). Essentially what only matters in this PR is the top diff.

chantra:
So only ed9ef049f0c8a061bffe8ea2dac3c2b9014284a1 matters for this PR but I don’t know if a simple way to handle stacked diff and independent PR in GH. The only reason those 2 diffs are dependent is because of the creation of fuzz_test.go .
I will address the comments about the comment probably tomorrow when I get in front of a computer.

chantra:
updated be9f4bcf8e6fb61b6347fa1f360709bbbdd1d34c

Once #966 id merged, I will be able to rebase this diff and remove the one it is stacked on.

miekg:
#996 is merged.

chantra:
Great! I will update this diff. Thanks

>

chantra:
ok @miekg and @tmthrgd this PR is rebased on top of master, there is only 1 commit now, and I believe the original comments were addressed.

miekg:
thank you! Merging

Issue #968, opened 2019-05-12
DNSSEC: remove old(er) algorithms


miekg:
See https://tools.ietf.org/html/rfc6944 and https://indico.dns-oarc.net/event/31/contributions/688/attachments/650/1067/matthijs-mekking-oarc30-unsupported-algorithms.pdf (what does BIND9 do) on what algorithms should be used.

This means `RSAMD5` should not be supported going forward, and probably a couple of others. I'm in favour of actually returning an "Unsupported Algorithm" error in such cases.

tmthrgd:
I 100% support this. Old algorithms should be removed when possible. If all other major software has removed support for RSAMD5, then we should too and hopefully that won’t break anyone. The others removed in BIND9 should probably be removed as well.

miekg:
closed via 9c315c51 and 0930b62a

PR #970, opened 2019-05-19
DNSSEC: remove deprecated algorithm RSAMD5 (RFC 6944)

miekg:
This removes RSAMD5 as an algorithm you can use. BIND also has
deprecated *all* DSA algos which is more involved can removes more
helper codes as well, so that should be done in a new PR.

This adds a ErrDeprecated signalling the DNSSEC algorithm is removed and
can't be used.

See #968

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,6 +152,7 @@ Example programs can be found in the `github.com/miekg/exdns` repository.
 * 6844 - CAA record
 * 6891 - EDNS0 update
 * 6895 - DNS IANA considerations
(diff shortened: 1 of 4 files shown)


tmthrgd:
Is there any real need for a specific error type? I would think that the existing error for unknown algorithms (BadAlg?) would be appropriate here.

miekg:
Didn't see and obvious candidate, but BadAlg reuse sgtm

PR #972, opened 2019-05-20
Remove DSA* algorithms

miekg:
This follows BIND9 and removed support for the DSA family of algorithms.
Any DNSSEC implementation should consider those zones using it,
insecure.

### DIFF
--- a/parse_test.go
+++ b/parse_test.go
@@ -1212,7 +1212,6 @@ func TestNewPrivateKey(t *testing.T) {
 		{ECDSAP384SHA384, 384},
 		{RSASHA1, 1024},
 		{RSASHA256, 2048},
-		{DSA, 1024},
 		{ED25519, 256},
 	}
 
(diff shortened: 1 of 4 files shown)


parse_test.go
```diff
@@ -1227,7 +1229,11 @@ func TestNewPrivateKey(t *testing.T) {
 		key.Algorithm = algo.name
 		privkey, err := key.Generate(algo.bits)
 		if err != nil {
-			t.Fatal(err)
+			if !algo.deprecated {
+				t.Fatal(err)
+			}
+			// deprecated algo, this must fail
+			continue
 		}
```
tmthrgd: If you want to keep testing these, there should be a separate block for `err == nil` like:

```go
} else {
  if algo.deprecated {
    t.Errorf(...)
  }
}
```

miekg: i'll remove this; it's testing something else entirely (actually)

PR #973, opened 2019-05-20
[nsec3] fix crash in nsec3 packing

chantra:
Both NSEC and NSEC3 use the same logic to pack the bitmap.
CSYNC.pack also appear to make use of `packDataNsec` so I am giving it
the same treatment by moving the logic in a helper function and making
all those types `len` call use that function.

### DIFF
--- a/fuzz_test.go
+++ b/fuzz_test.go
@@ -105,3 +105,41 @@ func TestCrashNSEC(t *testing.T) {
 		t.Fatalf("expected length of %d, got %d", expectedLength, l)
 	}
 }
+
+// TestCrashNSEC3 tests generated using fuzz.go and with a message pack
+// containing the following bytes:
+// "0000\x00\x00000000\x00\x00200000" +
+// "0\x00\v0000\x00\x00#\x0300\x00\x00\x00\x1a000" +
+// "000\x00\v00\x0200\x00\x03000\x00"
+// That byte sequence, when Unpack() and subsequential Pack() created a
+// panic: runtime error: slice bounds out of range
+// which was attributed to the fact that NSEC3 RR length computation was
+// different (and smaller) then within NSEC3.pack (which relies on
+// packDataNsec).
+func TestCrashNSEC3(t *testing.T) {
+	compression := make(map[string]struct{})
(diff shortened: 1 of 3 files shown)


chantra:
This is similar to #967 . The length computation logic is now refactored in a function that is used by all types that I could find that made use of `packDataNsec`

fuzz_test.go
```diff
@@ -105,3 +105,40 @@ func TestCrashNSEC(t *testing.T) {
...
+// TestCrashNSEC3 tests generated using fuzz.go and with a message pack
+// containing the following bytes:
+// "0000\x00\x00000000\x00\x00200000" +
+// "0\x00\v0000\x00\x00#\x0300\x00\x00\x00\x1a000" +
+// "000\x00\v00\x0200\x00\x03000\x00"
+// That byte sequence, when Unpack() and subsequential Pack() created a
+// panic: runtime error: slice bounds out of range
+// which was attributed to the fact that NSEC RR length computation was different (and smaller)
+// then when within packDataNsec.
+func TestCrashNSEC3(t *testing.T) {
```
miekg: Isn't this a dup of the NSEC test then?

msg_helpers.go
```diff
@@ -587,6 +587,29 @@ func unpackDataNsec(msg []byte, off int) ([]uint16, int, error) {
 	return nsec, off, nil
 }
 
+// bitMapDataLen is a helper function which computes the "maximum" length of
+// a the NSEC Type BitMap field.
+func bitMapDataLen(bitmap []uint16) int {
```
miekg: `typeBitMapLen` is maybe slightly better, because that's what the fields in the types is called.

fuzz_test.go
chantra: so it is a similar issue, but with a different input. I do admit that I copied the comment from the NSEC test and adapted it, but forgot to rename NSEC to NSEC3 in 1 spot.

chantra:
* updated helper name
* updated TestCrashNSEC3 comment

chantra:
address nits

Issue #975, opened 2019-05-22
Handling of Z flag in defaultMsgAcceptFunc


chantra:
Following the thread [[dns-operations] The last remaining DNS header flag](https://lists.dns-oarc.net/pipermail/dns-operations/2019-May/018724.html) it seems rather than [rejecting the Message](https://github.com/miekg/dns/blob/5c2ec9f7e458912d656335743784d0195a5a423e/acceptfunc.go#L38), the Z bits should be reset to 0 and the message processed just like if nothing was wrongly set.
While `RFC1035 4.1.1. Header section format` does not seem to say much other than:
```
Z               Reserved for future use.  Must be zero in all queries
                and responses.
```
and in this case it is non-zero in the query, it would make sense for "future use" that whatever they may be set to, the response would just set them to 0 as a signal that it does not support any specific use of (currently) Z flags.

https://ednscomp.isc.org currently return the following warning when the response echo the bits:

```
Plain DNS with last reserved header bit set (zflag)
dig +norec +noad +noedns +zflag soa zone @server
expect: SOA
expect: NOERROR
expect: Z bit to be clear in response
See RFC1035, 4.1.1. Header section format
```

A simple patch that would make this work would be:

```patch
@@ -25,6 +25,7 @@
 )
 
 func defaultMsgAcceptFunc(dh Header) MsgAcceptAction {
+	dh.Bits = dh.Bits & ^uint16(_Z)
 	if isResponse := dh.Bits&_QR != 0; isResponse {
 		return MsgIgnore
 	}
@@ -35,9 +36,6 @@
 		return MsgReject
 	}
 
-	if isZero := dh.Bits&_Z != 0; isZero {
-		return MsgReject
-	}
 	if dh.Qdcount != 1 {
 		return MsgReject
 	}
```
but this would make the flag cleared before the request is actually handled, which may not be the best.
Another approach could be to clear the bits when setting the response (which would make more sense IMHO), like in the helper functions in https://github.com/miekg/dns/blob/d051b464e9f024b65ce9674dece9d3e905b31828/defaults.go

Throughts? cc @yarikk

miekg:
[ Quoting <notifications@github.com> in "[miekg/dns] Handling of Z flag in d..." ]
[...]

PR #976, opened 2019-05-22
ignore Z flag in queries, clear Z flag in automatic replies

yarikk:
Context in issue #975.

cc @chantra

### DIFF
--- a/server.go
+++ b/server.go
@@ -571,6 +571,7 @@ func (srv *Server) serveDNS(m []byte, w *response) {
 		req.SetRcodeFormatError(req)
 		// Are we allowed to delete any OPT records here?
 		req.Ns, req.Answer, req.Extra = nil, nil, nil
+		req.Zero = false
 
 		w.WriteMsg(req)
 		fallthrough
--- a/server_test.go
+++ b/server_test.go
@@ -187,6 +187,34 @@ func TestServing(t *testing.T) {
 	}
 }
 
+// Verify that the server responds to a query with Z flag on, ignoring the flag, and does not echoes it back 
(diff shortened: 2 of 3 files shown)


yarikk:
`genreport -i zfla` is verified to return "`zflag=ok`" with this patch applied.

server.go
```diff
@@ -571,6 +571,7 @@ func (srv *Server) serveDNS(m []byte, w *response) {
 		req.SetRcodeFormatError(req)
 		// Are we allowed to delete any OPT records here?
 		req.Ns, req.Answer, req.Extra = nil, nil, nil
+		req.Zero = false
```
chantra: while this is not currently explicitly tested, it seems like a good default when rejecting the message.

server_test.go
```diff
@@ -187,6 +187,41 @@ func TestServing(t *testing.T) {
...
+	if err != nil {
+		t.Fatal("failed to exchange example.com with +zflag", err)
+	}
+	if r.Zero {
+		t.Error("the response should not have Z flag set - even for a query which does")
+	}
+	if r.Rcode != RcodeSuccess {
+		t.Errorf("expected rcode %v, got %v", RcodeSuccess, r.Rcode)
+	}
+	if len(r.Extra) == 0 {
```
miekg: seems a bit of a stretch to test all this all well, r.Zero and r.Rcode should be sufficient

yarikk: got it

yarikk:
The excessive testing code removed, leaving with the only two checks for the aspects this diff changes. @miekg

PR #979, opened 2019-05-29
Prohibit newlines before record data in the ZoneParser

### LINKED ISSUE #901: Zone file parsing ignores meaningful line boundaries
When parsing zone files, a line break immediately following record type is ignored, resulting in acceptance and misinterpretation of invalid input:
```go
$ cat main.go
package main

import (
	"fmt"
	"github.com/miekg/dns"
	"strings"
)

func main() {
	fmt.Printf("; miekg/dns version: %v\n", dns.Version)

	zone := `
$ORIGIN example.com.
example.com. 3600 IN SOA ns.example.net. hostmaster.example.com. 1 86400 60 86400 3600
example.com. 3600 NS ns.example.net.

example.com. 60 PX
1000 TXT 1K
`
	for token := range dns.ParseZone(strings.NewReader(zone), "", "") {
		if token.Error != nil {
			fmt.Printf("; ERROR: %v\n", token.Error)
		} else {
			fmt.Println(token.RR.String())
		}
	}
}

$ go run main.go
; miekg/dns version: 1.1.3
example.com.	3600	IN	SOA	ns.example.net. hostmaster.example.com. 1 86400 60 86400 3600
example.com.	3600	IN	NS	ns.example.net.
example.com.	60	IN	PX	1000 TXT.example.com. 1K.example.com.
```

tmthrgd:
This also includes several simplifications to the parsing code.

Another approach that doesn't need to add the `Peek` method is to pass the first `lex` token directly to the `parse` method like: `parse(c *zlexer, l lex, origin string) *ParseError`. That may be simpler and preferred, but it doesn't otherwise add much.

There is one `TODO` in this PR, previously `slurpRemainder` was only called for some RR types on  dynamic updates. I'm not sure if this is actually observable or not though.

This may still need some polish, but it's here for review at least.

Fixes #901

### DIFF
--- a/dns.go
+++ b/dns.go
@@ -54,7 +54,7 @@ type RR interface {
 	// parse parses an RR from zone file format.
 	//
 	// This will only be called on a new and empty RR type with only the header populated.
-	parse(c *zlexer, origin, file string) *ParseError
+	parse(c *zlexer, origin string) *ParseError
 
 	// isDuplicate returns whether the two RRs are duplicates.
 	isDuplicate(r2 RR) bool
@@ -105,7 +105,7 @@ func (h *RR_Header) unpack(msg []byte, off int) (int, error) {
 	panic("dns: internal error: unpack should never be called on RR_Header")
 }
 
-func (h *RR_Header) parse(c *zlexer, origin, file string) *ParseError {
+func (h *RR_Header) parse(c *zlexer, origin string) *ParseError {
 	panic("dns: internal error: parse should never be called on RR_Header")
 }
 
--- a/privaterr.go
+++ b/privaterr.go
@@ -68,7 +68,7 @@ func (r *PrivateRR) unpack(msg []byte, off int) (int, error) {
 	return off, err
 }
 
-func (r *PrivateRR) parse(c *zlexer, origin, file string) *ParseError {
+func (r *PrivateRR) parse(c *zlexer, origin string) *ParseError {
 	var l lex
 	text := make([]string, 0, 2) // could be 0..N elements, median is probably 1
 Fetch:
@@ -85,7 +85,7 @@ Fetch:
 
 	err := r.Data.Parse(text)
 	if err != nil {
-		return &ParseError{file, err.Error(), l}
+		return &ParseError{"", err.Error(), l}
 	}
 
 	return nil
--- a/scan.go
+++ b/scan.go
@@ -666,6 +666,8 @@ func (zp *ZoneParser) Next() (RR, bool) {
 				}
 
 				return rr, true
+			} else if l.value == zNewline {
+				return zp.setParseError("unexpected newline", l)
 			}
 
 			err := rr.parse(zp.c, zp.origin)
--- a/scan.go
+++ b/scan.go
@@ -656,6 +656,18 @@ func (zp *ZoneParser) Next() (RR, bool) {
 				rr = &RFC3597{Hdr: *h}
 			}
 
+			if allowDynamicUpdate(rr) && zp.c.Peek().token == "" {
+				// This is a dynamic update rr.
+
+				// TODO(tmthrgd): Previously slurpRemainder was only called
+				// for certain RR types, which may have been important.
+				if err := slurpRemainder(zp.c); err != nil {
+					return zp.setParseError(err.err, err.lex)
+				}
+
+				return rr, true
+			}
+
 			err := rr.parse(zp.c, zp.origin)
 			if err != nil {
 				// If err.lex is nil than we have encounter an unknown RR type
@@ -688,6 +700,15 @@ func canParseAsRR(rrtype uint16) bool {
 	}
 }
 
+func allowDynamicUpdate(rr RR) bool {
+	switch rr.(type) {
+	case *TKEY, *RFC3597, *PrivateRR:
+		return false
+	default:
+		return true
+	}
+}
+
 type zlexer struct {
 	br io.ByteReader
 
@@ -699,7 +720,8 @@ type zlexer struct {
 	comBuf  string
 	comment string
 
-	l lex
+	l       lex
+	cachedL *lex
 
 	brace  int
 	quote  bool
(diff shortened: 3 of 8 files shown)


scan.go
```diff
@@ -671,6 +689,27 @@ func (zp *ZoneParser) Next() (RR, bool) {
...
+	case TypeANY, TypeNULL, TypeOPT, TypeTSIG:
+		return false
+	default:
+		return true
+	}
+}
+
+func allowDynamicUpdate(rr RR) bool {
+	switch rr.(type) {
+	case *TKEY, *RFC3597, *PrivateRR:
```
miekg: Think we should only check for PrivateRRs

privaterr.go
```diff
@@ -85,7 +85,7 @@ Fetch:
 
 	err := r.Data.Parse(text)
 	if err != nil {
-		return &ParseError{file, err.Error(), l}
+		return &ParseError{"", err.Error(), l}
```
miekg: think `file` is/was quite useful here (and in the other `parse` functions)

dns.go
```diff
@@ -54,7 +54,7 @@ type RR interface {
 	// parse parses an RR from zone file format.
 	//
 	// This will only be called on a new and empty RR type with only the header populated.
-	parse(c *zlexer, origin, file string) *ParseError
+	parse(c *zlexer, origin string) *ParseError
```
miekg: I think we can just give this a `*ZoneParser` which holds all the info we need (origin, file and the lexer)

privaterr.go
tmthrgd: The file is added back to the error inside `ZoneParser.Next`. I agree that's not entirely clear, but I'm not sure where that could be documented. (`zp.setParseError` uses `zp.file` to fill in the `file` field).

scan.go
tmthrgd: Acknowledged.

dns.go
tmthrgd: We could do that. I'm not sure it really gives us much though.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Prohibit newlines b..." ]
>> @@ -85,7 +85,7 @@ Fetch:
>
> 	err := r.Data.Parse(text)
> 	if err != nil {
>-		return &ParseError{file, err.Error(), l}
>+		return &ParseError{"", err.Error(), l}
>
>The file is added back to the error inside `ZoneParser.Next`. I agree that's not entirely clear, but I'm not sure where that could be documented. (`zp.setParseError` uses `zp.file` to fill in the `file` field).

ok, fair enough. Maybe we can drop that (file) field at some point then?

scan.go
```diff
@@ -671,6 +693,25 @@ func (zp *ZoneParser) Next() (RR, bool) {
...
+	case TypeANY, TypeNULL, TypeOPT, TypeTSIG:
+		return false
+	default:
+		return true
+	}
+}
+
+// allowDynamicUpdate returns true if the record allows for dynamic updates
+// when parsing zone files.
+func allowDynamicUpdate(rr RR) bool {
```
miekg:
```suggestion
func isPrivateRR(rr RR) bool {
```

scan.go
```diff
@@ -671,6 +693,25 @@ func (zp *ZoneParser) Next() (RR, bool) {
...
+func canParseAsRR(rrtype uint16) bool {
+	switch rrtype {
+	case TypeANY, TypeNULL, TypeOPT, TypeTSIG:
+		return false
+	default:
+		return true
+	}
+}
+
+// allowDynamicUpdate returns true if the record allows for dynamic updates
```
miekg:
```suggestion
// isPrivateRR  returns true if rr is of type PrivateRR
```

scan.go
```diff
@@ -671,6 +693,25 @@ func (zp *ZoneParser) Next() (RR, bool) {
...
+// when parsing zone files.
```
miekg:
```suggestion
```

scan.go
```diff
@@ -671,6 +693,25 @@ func (zp *ZoneParser) Next() (RR, bool) {
...
+func allowDynamicUpdate(rr RR) bool {
+	_, isPrivate := rr.(*PrivateRR)
+	return !isPrivate
```
miekg:
```suggestion
	return isPrivate
```

scan.go
```diff
@@ -650,19 +648,43 @@ func (zp *ZoneParser) Next() (RR, bool) {
...
-					e.lex = l // Uh, dirty
+			var rr RR
+			if newFn, ok := TypeToRR[h.Rrtype]; ok && canParseAsRR(h.Rrtype) {
+				rr = newFn()
+				*rr.Header() = *h
+			} else {
+				rr = &RFC3597{Hdr: *h}
+			}
+
+			if allowDynamicUpdate(rr) && zp.c.Peek().token == "" {
```
miekg:
```suggestion
			if !isPrivateRR(rr) && zp.c.Peek().token == "" {
```

miekg:
the test can be slightly smaller and it might make sense to test with braces as well as that is valid and should parse:
~~~ go
func TestUnexpectedNewline(t *testing.T) {
	zone := `
example.com. 60 PX
1000 TXT 1K
`
	zp := NewZoneParser(strings.NewReader(zone), "example.com.", "")
	for _, ok := zp.Next(); ok; _, ok = zp.Next() {
	}

	const expect = `dns: unexpected newline: "\n" at line: 2:18`
	if err := zp.Err(); err == nil || err.Error() != expect {
		t.Errorf("expected error to contain %q, got %v", expect, err)
	}

	zone = `
example.com. 60 PX (
1000 TXT 1K )
`
	zp = NewZoneParser(strings.NewReader(zone), "example.com.", "")
	for t, ok := zp.Next(); ok; t, ok = zp.Next() {
	}

	if err := zp.Err(); err != nil {
		t.Errorf("expected an error, got none")
	}
}
~~~

miekg:
some other nits, but lgtm.

tmthrgd:
This fell off my radar, but I addressed the nits.

miekg:
awesome, thank you!

scan.go
```diff
@@ -671,6 +694,18 @@ func (zp *ZoneParser) Next() (RR, bool) {
 	return nil, false
 }
 
+// canParseAsRR returns true if the record type can be parsed as a
+// concrete RR. It blacklists certain record types that must be parsed
+// according to RFC 3597 because they lack a presentation format.
+func canParseAsRR(rrtype uint16) bool {
```
gibson042: I would invert this and rename it to something like `mustParseAsGeneric`, using vocabulary from [RFC 3597](https://tools.ietf.org/html/rfc3597#section-5).

tmthrgd: If I touch this again in a future PR, I can certainly consider doing that.

PR #980, opened 2019-05-31
Migrate to Go modules

### LINKED ISSUE #921: go mod
Recently CoreDNS moved to Go modules; this was relatively painless. 
After Go 1.13 this will be the default for Go which fits our release cadance well, as we the support the 2 latest Go release. So within the next 3 months (or is it 6?) we should move to Go modules as well.

myhro:
Hi Miek,

This PR removes the current Go dep/vendor dependencies, adding support for Go modules.

There are a couple details still pending:

- Currently, if cloned under `$GOPATH` and running, for instance, `go test`, the dependencies won't be found as the `GO111MODULE` variable still defaults to `auto`. What about adding a `Makefile`, which sets `GO111MODULE=on` when running commands like `make test`?
- The reference to `GOPATH` in the README should be removed, preferable changing the `go get` command to `git clone`, but I'm not sure about the `go build` one. As the project doesn't generate any binaries, would it make sense to change it to `go test`, or even `make test` if we go for option `1`?

This closes #921.

Regards,
Tiago.


tmthrgd:
go1.13 is changing the semantics of `GO111MODULE=auto` so it will work even under `$GOPATH`. Alternatively we could just continue adding the `vendor` directory—is that `go mod vendor`?

I would love to eliminate the vendor directory eventually though as it’s hard to keep up to date.

Aside from that quirk, I’m very in favour of this. Modules are go’s future and I’ve found them much nicer to work with than dep ever was.

miekg:
yes, this makes sense to merge, but AFAICT this doesn't work with Go 1.11 and we want to support the last two Go version, so we have to wait until Go 1.13 before this can be merged.

myhro:
Tom,

> go1.13 is changing the semantics of `GO111MODULE=auto` so it will work even under `$GOPATH`.

Yes, but as explained in the parent issue, we have to keep support for the 2 latest Go releases. So, in order to rely on a behaviour changed on `1.13`, we would have to wait until `1.14`.

> Alternatively we could just continue adding the `vendor` directory—is that `go mod vendor`?
> 
> I would love to eliminate the vendor directory eventually though as it’s hard to keep up to date.

Sorry, but I didn't properly understood it. What are the benefits of keeping the `vendor/` directory? To support Go versions `< 1.11`?

---

Miek,

> yes, this makes sense to merge, but AFAICT this doesn't work with Go 1.11 and we want to support the last two Go version, so we have to wait until Go 1.13 before this can be merged.

The Travis CI run was green on `1.11`, but I'll confirm running it on my machine to make sure it isn't broken. Travis CI indeed executes a lot of steps in preparation for a Go build environment that might have affected that.

myhro:
Outside `$GOPATH`:

```
vagrant@ubuntu-bionic:~/miekg-dns$ go version
go version go1.11.10 linux/amd64
vagrant@ubuntu-bionic:~/miekg-dns$ go test ./...
go: finding golang.org/x/sync v0.0.0-20180314180146-1d60e4601c6f
go: finding golang.org/x/sys v0.0.0-20180928133829-e4b3c5e90611
go: finding golang.org/x/crypto v0.0.0-20181001203147-e3636079e1a4
go: finding golang.org/x/net v0.0.0-20180926154720-4dfa2610cdf3
go: downloading golang.org/x/crypto v0.0.0-20181001203147-e3636079e1a4
go: downloading golang.org/x/sync v0.0.0-20180314180146-1d60e4601c6f
go: downloading golang.org/x/sys v0.0.0-20180928133829-e4b3c5e90611
go: downloading golang.org/x/net v0.0.0-20180926154720-4dfa2610cdf3
ok      github.com/miekg/dns    3.992s
ok      github.com/miekg/dns/dnsutil    0.002s
```

Inside `$GOPATH`:

```
vagrant@ubuntu-bionic:~/go/src/github.com/miekg/dns$ go version
go version go1.11.10 linux/amd64
vagrant@ubuntu-bionic:~/go/src/github.com/miekg/dns$ go test ./...
dnssec.go:23:2: cannot find package "golang.org/x/crypto/ed25519" in any of:
        /usr/local/go/src/golang.org/x/crypto/ed25519 (from $GOROOT)
        /home/vagrant/go/src/golang.org/x/crypto/ed25519 (from $GOPATH)
udp.go:8:2: cannot find package "golang.org/x/net/ipv4" in any of:
        /usr/local/go/src/golang.org/x/net/ipv4 (from $GOROOT)
        /home/vagrant/go/src/golang.org/x/net/ipv4 (from $GOPATH)
udp.go:9:2: cannot find package "golang.org/x/net/ipv6" in any of:
        /usr/local/go/src/golang.org/x/net/ipv6 (from $GOROOT)
        /home/vagrant/go/src/golang.org/x/net/ipv6 (from $GOPATH)
listen_go111.go:11:2: cannot find package "golang.org/x/sys/unix" in any of:
        /usr/local/go/src/golang.org/x/sys/unix (from $GOROOT)
        /home/vagrant/go/src/golang.org/x/sys/unix (from $GOPATH)
vagrant@ubuntu-bionic:~/go/src/github.com/miekg/dns$ GO111MODULE=on go test ./...
ok      github.com/miekg/dns    4.018s
ok      github.com/miekg/dns/dnsutil    0.010s
```

In the second run it used the dependencies installed on the first time, but I also tested purging the dependency cache between both of them. That confirms my suspicion of the same behaviour, requiring `GO111MODULE=on` when under `$GOPATH` on both `1.11` and `1.12`... But it works. :-)

tmthrgd:
Keeping the vendor directory should keep everything working pre-module right?

IIRC, `go build -mod=vendor` only uses the top-level vendor directory so ours will be ignored when running in module mode, but the vendor directory will continue to work with existing semantics for go1.11 and go1.12 under GOPATH or with modules disabled.

It’s even in the wiki: https://github.com/golang/go/wiki/Modules#providing-dependency-information-to-older-versions-of-go-and-non-module-consumers.

If we keep the vendor directory (`go mod vendor`) and switch to modules, everything should just work for everyone.

miekg:
Ok. Works for me. We can then delete the vendor dir after go 1.13 is out

myhro:
> Ok. Works for me. We can then delete the vendor dir after go 1.13 is out

But then `1.12` will be broken under `$GOPATH`. If we keep the `vendor/` directory, it will only be possible to remove it when `1.14` is released.

The question now is more about: do we want to rely exclusively on Go modules, given that both Go versions we currently support will work with it, or mainly use it to manage the `vendor/` directory?

My feeling is that we will be mixing two sets of dependencies: the ones `go.mod` and the ones in the `vendor/` directory, which might diverge at some point. Updating the later, with `go mod vendor`, will still be a manual and possibly error-prone process.

tmthrgd:
When to remove the vendor directory is something that we can figure out later. We can always add a check to the Travis build to check vendor is up to date.

Realistically the only approaches are to include the vendor directory now with modules or wait for several go versions until the latest two versions have `GO111MODULE=on`. 

If we switch to modules now without the vendor directory, most existing builds will get different versions or break. I just think that’s a non-starter.

go modules will always be the default with vendoring simply a fallback for older go versions.

myhro:
Ok, I think it makes sense to be extra careful in avoid potential breakages in this case. I've updated the `vendor/` directory with `go mod vendor`.

About the Travis CI check, do you think we need something more clever than running `go mod vendor` and checking if there are files changed in the current directory?

miekg:
I think we should just wait until go modules are out long enough to not make 
this complicated. It doesn't solve any bugs at this point and seems to raise a 
lot of question on how this is supposed to work.

Let's defer until after go 1.13 is out

tmthrgd:
Regardless of what is done now, this PR needs two changes to the `.travis.yml` file.

It needs the following to enable module mode for the go1.11 and go1.12 builds:
```yml
env:
    - GO111MODULE=on
```

And (optionally) the following to cache module downloads:
```yml
cache:
    directories:
        - $GOPATH/pkg/mod
```

---

I don't think there are any real problems with adopting this now, and it will be much nicer for consumers as they start to migrate to modules. I'm not convinced this is that complicated either.

myhro:
I've dropped the `go mod vendor` suggestion, as it's clear that the current Go development branch isn't offering full backward-compatibility in this regard. `GO111MODULE=on` was re-added to Travis CI as well.

I'm fine about keeping the PR open for 2-3 more months until Go `1.13` is released, rebasing/updating it by then.

tmthrgd:
“as it's clear that the current Go development branch isn't offering full backward-compatibility in this regard.”

I’m not sure I follow. Vendoring works *exactly* the same for non-module consumers as it always has and `go mod vendor` produces a vendor directory that should be fully compatible with previous go versions. If it isn’t, then that’s a bug that should be reported to the golang issue tracker.

miekg:
this library is the only thing holding me back to unifying all my development under one (src/) directory. I now resort to playing with GO11MODULE foo to make thing compile.

It's a bit annoying, so I rather want to move to go modules sooner than latter; go 1.13 is *almost* here, so from my point of view we can just move to go modules.

miekg:
if you can either fix the weird conflict, or send a new PR, I'll happily merge.

myhro:
Hi Miek, sorry for the delay. I've rebased the PR on top of the current `master`.

Can you please check if we are good to go?

miekg:
checks are OK. If something comes up we can always fix it later.

miekg:
released 1.1.16

PR #982, opened 2019-06-06
Reply with NOTIMPL when Opcode is not supported

chantra:
One of the test from DNS Compliance testing validates that if the opcode
is not supported, a NOTIMPL rcode is returned.

https://gitlab.isc.org/isc-projects/DNS-Compliance-Testing/blob/e0884144dde6704128e22acc14e2b65ba41a92b2/genreport.c#L293

This diff makes the default acceptfunc support this case and reply with
NOTIMPL instead of FORMERR.

### DIFF
--- a/acceptfunc.go
+++ b/acceptfunc.go
@@ -19,9 +19,10 @@ var DefaultMsgAcceptFunc MsgAcceptFunc = defaultMsgAcceptFunc
 type MsgAcceptAction int
 
 const (
-	MsgAccept MsgAcceptAction = iota // Accept the message
-	MsgReject                        // Reject the message with a RcodeFormatError
-	MsgIgnore                        // Ignore the error and send nothing back.
+	MsgAccept         MsgAcceptAction = iota // Accept the message
+	MsgReject                                // Reject the message with a RcodeFormatError
+	MsgIgnore                                // Ignore the error and send nothing back.
+	MsgRejectNotImplemented                        // Reject the message with a RcodeNotImplemented
 )
 
 func defaultMsgAcceptFunc(dh Header) MsgAcceptAction {
@@ -32,7 +33,7 @@ func defaultMsgAcceptFunc(dh Header) MsgAcceptAction {
 	// Don't allow dynamic updates, because then the sections can contain a whole bunch of RRs.
 	opcode := int(dh.Bits>>11) & 0xF
 	if opcode != OpcodeQuery && opcode != OpcodeNotify {
-		return MsgReject
+		return MsgRejectNotImplemented
 	}
 
 	if dh.Qdcount != 1 {
--- a/server.go
+++ b/server.go
@@ -560,18 +560,24 @@ func (srv *Server) serveDNS(m []byte, w *response) {
 	req := new(Msg)
 	req.setHdr(dh)
 
-	switch srv.MsgAcceptFunc(dh) {
+	switch action := srv.MsgAcceptFunc(dh); action {
 	case MsgAccept:
 		if req.unpack(dh, m, off) == nil {
 			break
 		}
 
 		fallthrough
-	case MsgReject:
+	case MsgReject, MsgRejectNotImplemented:
+		opcode := req.Opcode
 		req.SetRcodeFormatError(req)
+		req.Zero = false
+		if action == MsgRejectNotImplemented {
+			req.Opcode = opcode
+			req.Rcode = RcodeNotImplemented
+		}
+
 		// Are we allowed to delete any OPT records here?
 		req.Ns, req.Answer, req.Extra = nil, nil, nil
-		req.Zero = false
 
 		w.WriteMsg(req)
 		fallthrough
--- a/server_test.go
+++ b/server_test.go
@@ -215,6 +215,35 @@ func TestServeIgnoresZFlag(t *testing.T) {
 	}
 }
 
+// Verify that the server responds to a query with unsupported Opcode with a NotImplemented error and that Opcode is unchanged.
+func TestServeNotImplemented(t *testing.T) {
+	HandleFunc("example.com.", AnotherHelloServer)
+	opcode := 15
+
+	s, addrstr, err := RunLocalUDPServer(":0")
+	if err != nil {
+		t.Fatalf("unable to run test server: %v", err)
+	}
+	defer s.Shutdown()
+
+	c := new(Client)
+	m := new(Msg)
+
+	// Test that Opcode is like the unchanged from request Opcode and that Rcode is set to NotImplemnented
+	m.SetQuestion("example.com.", TypeTXT)
+	m.Opcode = opcode
+	r, _, err := c.Exchange(m, addrstr)
+	if err != nil {
+		t.Fatal("failed to exchange example.com with +zflag", err)
+	}
+	if r.Opcode != opcode {
+		t.Errorf("expected opcode %v, got %v", opcode, r.Opcode)
+	}
+	if r.Rcode != RcodeNotImplemented {
+		t.Errorf("expected rcode %v, got %v", RcodeNotImplemented, r.Rcode)
+	}
+}
+
 func TestServingTLS(t *testing.T) {
 	HandleFunc("miek.nl.", HelloServer)
 	HandleFunc("example.com.", AnotherHelloServer)


chantra:
There is yet a separate test:
```
```
{ "opcodeflg", FULL,  0, "",    0, 0x0000, 0, 0, 0, 0, 1, 1, 1, 1, 1, 1, 1, 15, 0,
	  "### dig +noedns +header-only +opcode=15 +tcflag +rec +raflag +cd +ad +aaflag +zflag ###"
	},
```
which on top of testing the opcode, also check that the flags ra,rd,ad,cd are cleared.

File: server.go [hunk 1]
```diff
@@ -573,6 +573,13 @@ func (srv *Server) serveDNS(m []byte, w *response) {
 		req.Ns, req.Answer, req.Extra = nil, nil, nil
 		req.Zero = false
 
+		w.WriteMsg(req)
+		fallthrough
```
tmthrgd: This double fallthrough doesn't work. This would be my suggestion:

```go
switch action := srv.MsgAcceptFunc(dh); action {
case MsgAccept:
	// ... exsting code ...
	fallthrough
case MsgReject, MsgNotImplemented:
	if accept == MsgNotImplemented {
		req.SetRcodeNotImplemented(req)
	} else {
		req.SetRcodeFormatError(req)
	}
	req.Ns, req.Answer, req.Extra = nil, nil, nil
	// ... existing code ...
	fallthrough
case MsgIgnore:
	// ... existing code ...
	return
}
```

server_test.go
```diff
@@ -215,6 +215,34 @@ func TestServeIgnoresZFlag(t *testing.T) {
...
+	if err != nil {
+		t.Fatal("failed to exchange example.com with +zflag", err)
+	}
+	if r.Opcode != opcode {
+		t.Errorf("expected opcode %v, got %v", opcode, r.Opcode)
+	}
+	if r.Rcode != RcodeNotImplemented {
+		t.Errorf("expected rcode %v, got %v", RcodeNotImplemented, r.Rcode)
+	}
+}
```
tmthrgd: Nit: There should be a newline after this. Run `go fmt`.

tmthrgd:
What if instead of extending `MsgAcceptAction` *every* time someone wants to reject with a different Rcode we instead add something like:

```go
func MsgRejectWithRcode(rcode uint16) MsgAcceptAction {
	return MsgReject | MsgAcceptAction(rcode) << 16
}
```

That we we end up with:
```go
return MsgRejectWithRcode(RcodeNotImplemented)
```

That way it's easy for people to reject with different Rcodes should they wish.

@miekg Thoughts?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Reply with NOTIMPL ..." ]
>What if instead of extending `MsgAcceptAction` *every* time someone wants to reject with a different Rcode we instead add something like:
>
>```go
>func MsgRejectWithRcode(rcode uint16) MsgAcceptAction {
>	return MsgReject | MsgAcceptAction(rcode) << 16
>}
>```
>
>That we we end up with:
>```go
>return MsgRejectWithRcode(RcodeNotImplemented)
>```
>
>That way it's easy for people to reject with different Rcodes should they wish.
>
>@miekg Thoughts?

are there that many? Manu is going over some DNS std things atm. The original 
intent of this is having a different accept func if you want to deviate. Adding 
parameters to the current accept func adds another way of configuring this.

(haven't looked at the PR yet though)

tmthrgd:
@miekg Sorry if wasn’t clear, I’m *not* suggesting we provide a configurable default accept `MsgAcceptFunc`. What I’m suggesting is basically a way to pass specific Rcodes from a `MsgAcceptFunc` function to the server. So you’d use it something like this:

```go
&Server{
       MsgAcceptFunc: func(Header) MsgAcceptAction {
              // ... logic ...
              return MsgRejectWithRcode(RcodeNotImplemented)
       },
}
```

miekg:
Ack, ok. That may be useful; but still not fully convinced...

Think the current focus on this particular function is temporary, because @chantra 's work and the increased focus from the DNS community on being more strict.

IOW: I'm fine with fiddling around with it for the time being - if things start to get out of hand we kick some tunables down a level

acceptfunc.go
```diff
@@ -19,9 +19,10 @@ var DefaultMsgAcceptFunc MsgAcceptFunc = defaultMsgAcceptFunc
 type MsgAcceptAction int
 
 const (
-	MsgAccept MsgAcceptAction = iota // Accept the message
-	MsgReject                        // Reject the message with a RcodeFormatError
-	MsgIgnore                        // Ignore the error and send nothing back.
+	MsgAccept         MsgAcceptAction = iota // Accept the message
+	MsgReject                                // Reject the message with a RcodeFormatError
+	MsgIgnore                                // Ignore the error and send nothing back.
+	MsgNotImplemented                        // Reject the message with a RcodeNotImplemented
```
miekg: Although not optimal we may need some more rejects here.

Propose:
* MsgReject
* MsgRejectNotImplemented

defaults.go
```diff
@@ -68,6 +68,16 @@ func (dns *Msg) SetRcodeFormatError(request *Msg) *Msg {
 	return dns
 }
 
+// SetRcodeNotImplemented creates a message with NotImplemented set.
+func (dns *Msg) SetRcodeNotImplemented(request *Msg) *Msg {
```
miekg: Probably not useful at this stage (usefulness of SetRcodeFormatError is also questionable)

server.go
```diff
@@ -560,18 +560,23 @@ func (srv *Server) serveDNS(m []byte, w *response) {
...
 
 		fallthrough
-	case MsgReject:
-		req.SetRcodeFormatError(req)
+	case MsgReject, MsgNotImplemented:
+		if action == MsgNotImplemented {
+			req.SetRcodeNotImplemented(req)
+		} else {
+			req.SetRcodeFormatError(req)
+			req.Zero = false
```
miekg: only zeroing on formatERR?

also just split the action into 2 cases in the switch

server_test.go
```diff
@@ -199,7 +199,7 @@ func TestServeIgnoresZFlag(t *testing.T) {
 
 	c := new(Client)
 	m := new(Msg)
-
```
miekg: just send a generic cleanup whitespace PR

server.go
tmthrgd: The way I structured this code with fallthroughs means we need to have this in one without rewriting it.

That being said, this can just be:
```go
req.SetRcodeFormatError(req)
if action == MsgNotImplemented {
    req.Rcode = RcodeNotImplemented
}
```

Then we can avoid introducing `SetRcodeNotImplemented`.

server.go
```diff
@@ -560,18 +560,24 @@ func (srv *Server) serveDNS(m []byte, w *response) {
...
 			break
 		}
 
 		fallthrough
-	case MsgReject:
+	case MsgReject, MsgRejectNotImplemented:
+		opcode := req.Opcode
 		req.SetRcodeFormatError(req)
+		req.Zero = false
+		if action == MsgRejectNotImplemented {
```
chantra: I removed the `SetRcodeNotImplemented` function. In the end, in light of the `opcodeflg` test, a NotImplemented message is basically a message with QR set, opcode set to original opcode, rcode set to NotImpl and all other flags cleared.
I think it could be valuable to have `SetRcodeNotImplemented` dealing with this.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Reply with NOTIMPL ..." ]
>
>
>
>>  		req.SetRcodeFormatError(req)
>+		req.Zero = false
>+		if action == MsgRejectNotImplemented {
>
>I removed the `SetRcodeNotImplemented` function. In the end, in light of the `opcodeflg` test, a NotImplemented message is basically a message with QR set, opcode set to original opcode, rcode set to NotImpl and all other flags cleared.
>I think it could be valuable to have `SetRcodeNotImplemented` dealing with this.

Such functions are all pretty trivial to write. I don't think we get a lot of 
benefit by extending the API by adding those.

chantra:
@miekg ok, fair. How does the PR look like now? Anything you want me to address?

acceptfunc.go
```diff
@@ -19,9 +19,10 @@ var DefaultMsgAcceptFunc MsgAcceptFunc = defaultMsgAcceptFunc
 type MsgAcceptAction int
 
 const (
-	MsgAccept MsgAcceptAction = iota // Accept the message
-	MsgReject                        // Reject the message with a RcodeFormatError
-	MsgIgnore                        // Ignore the error and send nothing back.
+	MsgAccept         MsgAcceptAction = iota // Accept the message
+	MsgReject                                // Reject the message with a RcodeFormatError
```
miekg: nit: these should be sorted (at least group the MsgReject*)

server.go
miekg: Can we just have 2 cases one for MsgReject and One for MsgRejectNotImplemented? Doing an extra if inside the case looks a bit weird (@tmthrgd or did you want it like this?)

tmthrgd: @miekg We need the if-statement unfortunately because I wrote this switch to rely on `fallthrough`s.

server.go
```diff
@@ -560,18 +560,24 @@ func (srv *Server) serveDNS(m []byte, w *response) {
...
 		if req.unpack(dh, m, off) == nil {
 			break
 		}
 
 		fallthrough
-	case MsgReject:
+	case MsgReject, MsgRejectNotImplemented:
+		opcode := req.Opcode
 		req.SetRcodeFormatError(req)
+		req.Zero = false
```
tmthrgd: I’d probably leave the `req.Zero = ...` line where it was and I think this would be a little cleaner. It’s not a big deal though.

chantra:
Thanks for merging @miekg 

there were a few comments you wanted me to address though. Want me to cut a separate PR? I will be rather OTG for the coming weeks though, nut I can still address the comments.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Reply with NOTIMPL ..." ]
>Thanks for merging @miekg
>
>there were a few comments you wanted me to address though. Want me to cut a separate PR? I will be rather OTG for the coming weeks though, nut I can still address the comments.

Ack, yes that would be appreciate. No rush though.

Thanks!

PR #989, opened 2019-06-24
Set the TC bit more aggressively in Truncate

### LINKED ISSUE #987: Truncate should set TC flag more often
On #854, @markdingo raised the following:

> On a related front, I think the logic behind setting the Truncate flag is a bit simplistic.
> 
> The code comment says "According to RFC 2181, the TC bit should only be set if not all of the answer RRs can be included in the response." and the code does precisely this. But that's not what RFC2181 says. It says TC should be set if a "required" RRSet doesn't fit, regardless of whether that required RRSet is in the answer section or not.
> 
> Admittedly that whole section in RFC2181 is pretty muddled, but there is at least one use-case where TC should be set when the additional section is truncated. That's when the additional section contains referral/glue RRs for a sub-domain. By way of example, if we imagine a query to an example.net name server for sub.example.net/NS which results in:
> 
> ;; QUESTION SECTION:
> ;sub.example.net.               IN      NS
> 
> ;; AUTHORITY SECTION:
> sub.example.net.        172800  IN      NS      c.sub.example.net.
> sub.example.net.        172800  IN      NS      a.sub.example.net.
> sub.example.net.        172800  IN      NS      b.sub.example.net.
> 
> ;; ADDITIONAL SECTION:
> c.sub.example.net.       172800  IN      A       ...
> a.sub.example.net.       172800  IN      A       ...
> b.sub.example.net.       172800  IN      A       ...
> 
> If just the additional section is truncated by msg.Truncate() then the client will fail to resolve the sub-domain as it won't see TC=1 to re-query with TCP and trying to resolve the sub NS address via the parent doesn't work since the address RRs are glue. You can try this for yourself with one of the TLDs such as .com. Imagine the additional section for google.com/NS is truncated with TC=0. You'll never be able to resolve google.com as you never see the NS address glue and can't get at it directly via the .com servers.
> 
> The problem here is that msg.Truncate() really has no idea whether non-Answer RRSets are "required" or not since it's just a generic function that is oblivious to the caller's intentions or role.
> 
> I'm not sure what the right approach is. Either the description of msg.Truncate() should accurately reflect that it's setting of TC is only a guide and the caller really has to make the final TC=1 decision. Alternatively it could follow the approach that PowerDNS takes in the next release which is to set TC on *any* truncation. This doesn't disagree with RFC2181 and it is a safer approach.
> 
> And, if you think about it, setting TC=1 only with an Answer truncation vs any truncation is mostly just a corner-case optimisation so forgoing this optimisation is unlikely to be harmful so the powerDNS approach isn't terrible.
> 
> Maybe not knowing what is "required" as per RFC2181 explains why some of the main DNS packages force the matter by wiping out *all* sections and setting TC=1 when any truncation is needed? They never give the client a chance to use a partial answer.
> 
> Anyway, there is at least one use-case where relying on msg.Truncate() to set TC correctly seems to be risky.

I’m opening a new issue since that PR has already been merged and comments aren’t always seen.

I’m fine making this change if others agree. The reasoning is sound and setting the TC flag when any record was removed was how I originally planned to implement this.

tmthrgd:
This changes the `(*Msg).Truncate` implementation to set the `TC` header flag whenever any RR is omitted from the response. Previously the `TC` flag was only set when an answer RR had to be omitted.

Fixes #987 (see this issue for a discussion on this approach).

/cc @markdingo

### DIFF
--- a/msg_truncate.go
+++ b/msg_truncate.go
@@ -8,8 +8,13 @@ package dns
 // record adding as many records as possible without exceeding the
 // requested buffer size.
 //
-// The TC bit will be set if any answer records were excluded from the
-// message. This indicates to that the client should retry over TCP.
+// The TC bit will be set if any records were excluded from the message.
+// This indicates to that the client should retry over TCP.
+//
+// According to RFC 2181, the TC bit should only be set if not all of the
+// "required" RRs can be included in the response. Unfortunately, we have
+// no way of knowing which RRs are required so we set the TC bit if any RR
+// had to be omitted from the response.
 //
 // The appropriate buffer size can be retrieved from the requests OPT
 // record, if present, and is transport specific otherwise. dns.MinMsgSize
@@ -71,10 +76,7 @@ func (dns *Msg) Truncate(size int) {
 		l, numExtra = truncateLoop(dns.Extra, size, l, compression)
 	}
 
-	// According to RFC 2181, the TC bit should only be set if not all of
-	// the "required" RRs can be included in the response. Unfortunately,
-	// we have no way of knowing which RRs are required so we set the TC
-	// bit if any RR had to be omitted from the response.
+	// See the function documentation for when we set this.
 	dns.Truncated = len(dns.Answer) > numAnswer ||
 		len(dns.Ns) > numNS || len(dns.Extra) > numExtra
 
--- a/msg_truncate.go
+++ b/msg_truncate.go
@@ -71,9 +71,12 @@ func (dns *Msg) Truncate(size int) {
 		l, numExtra = truncateLoop(dns.Extra, size, l, compression)
 	}
 
-	// According to RFC 2181, the TC bit should only be set if not all
-	// of the answer RRs can be included in the response.
-	dns.Truncated = len(dns.Answer) > numAnswer
+	// According to RFC 2181, the TC bit should only be set if not all of
+	// the "required" RRs can be included in the response. Unfortunately,
+	// we have no way of knowing which RRs are required so we set the TC
+	// bit if any RR had to be omitted from the response.
+	dns.Truncated = len(dns.Answer) > numAnswer ||
+		len(dns.Ns) > numNS || len(dns.Extra) > numExtra
 
 	dns.Answer = dns.Answer[:numAnswer]
 	dns.Ns = dns.Ns[:numNS]
--- a/msg_truncate_test.go
+++ b/msg_truncate_test.go
@@ -40,8 +40,8 @@ func TestRequestTruncateExtra(t *testing.T) {
 	if want, got := MinMsgSize, reply.Len(); want < got {
 		t.Errorf("message length should be bellow %d bytes, got %d bytes", want, got)
 	}
-	if reply.Truncated {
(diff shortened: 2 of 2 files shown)


msg_truncate.go
```diff
@@ -71,9 +71,12 @@ func (dns *Msg) Truncate(size int) {
 		l, numExtra = truncateLoop(dns.Extra, size, l, compression)
 	}
 
-	// According to RFC 2181, the TC bit should only be set if not all
-	// of the answer RRs can be included in the response.
-	dns.Truncated = len(dns.Answer) > numAnswer
+	// According to RFC 2181, the TC bit should only be set if not all of
```
miekg: This text can be elevated to the function doc. That also still says the TC bit is only set when the answer section is truncated.

markdingo:
(Sorry, I posted this comment to the discussion thread rather than this PR).

I like the change but I wonder whether it's missing the fact that the EDNS0 RR is removed at the top of the code and re-appended at the end after the length check? Note the popEdns0() call at the top of the function?

Maybe a test case with a call to SetUDPSize(4096) would catch this?

As an aside, with that change to the semantics, can the truncation of the RR slices be avoided if Truncated is not set? Probably a silly optimisation, but why not?

if dns.Truncated {
   dns.Answer = dns.Answer[:numAnswer]
   dns.Ns = dns.Ns[:numNS]
   dns.Extra = dns.Extra[:numExtra]
}

tmthrgd:
@markdingo The EDNS0 record size is handled explicitly here: 
https://github.com/miekg/dns/blob/d89f1e3d4bfcb2de7c3988c619d27bb6f5fac706/msg_truncate.go#L44-L49. As for adding the if-statement, slicing is efficient so I don't think it's worth such a micro-optimisation.

@miekg I'll address the documentation.

markdingo:
> The EDNS0 record size is handled explicitly here

Sorry, I wasn't very clear.

I'm under the impression that the popEdns0() call will have removed the EDNS0 RR thus reducing the Extra count by 1. That means that your new test will *think* that the message has been truncated when it hasn't really as the EDNS0 RR is added back in at the end of the code.

tmthrgd:
@markdingo Ah I see what you're saying, but that isn't a problem here. `popEdns0` does, as you said, remove the EDNS0 RR from `Extra`. The `truncateLoop` call [here](https://github.com/miekg/dns/blob/d89f1e3d4bfcb2de7c3988c619d27bb6f5fac706/msg_truncate.go#L71) then loops over the slice *without* the EDNS0 RR in it and determines the correct length.

The `len(dns.Extra) > numExtra` comparison is done when `dns.Extra` already has the EDNS0 RR removed.

markdingo:
> The `truncateLoop` call [here](https://github.com/miekg/dns/blob/d89f1e3d4bfcb2de7c3988c619d27bb6f5fac706/msg_truncate.go#L71) then loops over the slice _without_ the EDNS0 RR in it and determines the correct length.

Ahh yes, you're right. As you say, numExtra is calculated after the removal. I misread the code. Sorry about that.

On that basis I think the PR does the job just fine.

PR #992, opened 2019-07-03
Test out RelativeString()

miekg:
WIP to see how the code would look.

### DIFF
--- a/trim.go
+++ b/trim.go
@@ -1,0 +1,25 @@
+package dns
+
+import "strings"
+
+// trim trims origin from s if s is a subdomain of origin. If this is done Trim
+// will return a relative name. If s is not a subdomain it will be returned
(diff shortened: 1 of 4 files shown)


trim.go
```diff
@@ -0,0 +1,25 @@
...
+
+import "strings"
+
+// trim trims origin from s if s is a subdomain of origin. If this is done Trim
+// will return a relative name. If s is not a subdomain it will be returned
+// as-is.
+//
+// It will return "@" if s equals origin. Both arguments need to be fully qualified.
+// It assumes s and origin have the same encoding and quoting for things like \DDD constructs.
+func trim(s, origin string) string {
```
gibson042: `trim` is a bit generic; something like `relativeTo` might be better.

PR #1008, opened 2019-09-21
types: improve the performance of sprintName by ~30% and halve allocs

charlievieth:
This PR improves performance and memory usage by lazily allocating the `strings.Builder` and by inlining the `writeDomainNameByte()` logic.

```
benchmark                           old ns/op     new ns/op     delta
BenchmarkSprintName-12              180           120           -33.33%
BenchmarkSprintName_NoEscape-12     198           82.4          -58.38%

benchmark                           old allocs     new allocs     delta
BenchmarkSprintName-12              2              1              -50.00%
BenchmarkSprintName_NoEscape-12     1              0              -100.00%

benchmark                           old bytes     new bytes     delta
BenchmarkSprintName-12              48            32            -33.33%
BenchmarkSprintName_NoEscape-12     32            0             -100.00%
```

**Edit:** forgot to include the results of `BenchmarkSprintName_NoEscape`.

### DIFF
--- a/types.go
+++ b/types.go
@@ -438,24 +438,53 @@ func (rr *TXT) String() string { return rr.Hdr.String() + sprintTxt(rr.Txt) }
 
 func sprintName(s string) string {
 	var dst strings.Builder
-	dst.Grow(len(s))
+
 	for i := 0; i < len(s); {
 		if i+1 < len(s) && s[i] == '\\' && s[i+1] == '.' {
-			dst.WriteString(s[i : i+2])
+			if dst.Len() != 0 {
+				dst.WriteString(s[i : i+2])
+			}
 			i += 2
 			continue
 		}
 
 		b, n := nextByte(s, i)
-		switch {
-		case n == 0:
-			i++ // dangling back slash
-		case b == '.':
-			dst.WriteByte('.')
+		if n == 0 {
+			i++
+			continue
+		}
+		if b == '.' {
+			if dst.Len() != 0 {
+				dst.WriteByte('.')
+			}
+			i += n
+			continue
+		}
+		switch b {
+		case ' ', '\'', '@', ';', '(', ')', '"', '\\': // additional chars to escape
+			if dst.Len() == 0 {
+				dst.Grow(len(s) * 2)
+				dst.WriteString(s[:i])
+			}
+			dst.WriteByte('\\')
+			dst.WriteByte(b)
 		default:
-			writeDomainNameByte(&dst, b)
+			if ' ' <= b && b <= '~' {
+				if dst.Len() != 0 {
+					dst.WriteByte(b)
+				}
+			} else {
+				if dst.Len() == 0 {
+					dst.Grow(len(s) * 2)
+					dst.WriteString(s[:i])
+				}
+				dst.WriteString(escapeByte(b))
+			}
 		}
 		i += n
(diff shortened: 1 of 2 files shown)


types.go
```diff
@@ -438,25 +438,54 @@ func (rr *TXT) String() string { return rr.Hdr.String() + sprintTxt(rr.Txt) }
...
+			dst.WriteByte(b)
 		default:
-			writeDomainNameByte(&dst, b)
+			if ' ' <= b && b <= '~' {
+				if dst.Len() != 0 {
+					dst.WriteByte(b)
+				}
+			} else {
+				if dst.Len() == 0 {
+					dst.Grow(len(s) * 2)
```
miekg: some question here why the `*2`?

tmthrgd:
Can you use [benchstat](https://godoc.org/golang.org/x/perf/cmd/benchstat) to compare ~10 benchmark before and after runs rather than benchcmp please? (benchstat is more statistically sound and accounts for noise).

types.go
charlievieth: > The `* 2` is because we need more than `len(s)` bytes to store the result, but we don't know how much more space we need so I arbitrarily picked `* 2` as it provides plenty of headroom while still being cheaper than than `len(s)`, which requires at least `* 3` bytes and two allocs - see [`Builder.grow()`](https://github.com/golang/go/blob/release-branch.go1.13/src/strings/builder.go#L64-L70).
>
> A lower value (say `* 1.5`) could be used but one challenge is that escape sequences require either 2 or 4 bytes.  `* 2` is pretty aggressive when escaping with 2 bytes, but doesn't leave a lot of headroom when escaping with 4 bytes.

https://github.com/miekg/dns/pull/1009#discussion_r326872444

charlievieth:
> Can you use [benchstat](https://godoc.org/golang.org/x/perf/cmd/benchstat) to compare ~10 benchmark before and after runs rather than benchcmp please? (benchstat is more statistically sound and accounts for noise).

Interesting, never heard of `benchstat`.  I highly doubt it will make a difference here since the gain comes from the reduction in allocs, which isn't variable or subject noise, but I'll run it out of curiosity.

charlievieth:
Ran `benchstat` as predicted it didn't really show anything - cool tool nonetheless so thanks for cluing me in about it.

PR [`cev/sprintname-perf`](https://github.com/charlievieth/dns/tree/cev/sprintname-perf)
```
name                    time/op
SprintName-12            122ns ± 3%
SprintName_NoEscape-12  84.2ns ± 3%

name                    alloc/op
SprintName-12            32.0B ± 0%
SprintName_NoEscape-12   0.00B

name                    allocs/op
SprintName-12             1.00 ± 0%
SprintName_NoEscape-12    0.00
```

master@ 2acbc9eff343291e671c1a9ba7be3201bf66dfd6
```
name                    time/op
SprintName-12           185ns ± 1%
SprintName_NoEscape-12  202ns ± 1%

name                    alloc/op
SprintName-12           48.0B ± 0%
SprintName_NoEscape-12  32.0B ± 0%

name                    allocs/op
SprintName-12            2.00 ± 0%
SprintName_NoEscape-12   1.00 ± 0%
```

types.go
miekg: thank you

PR #1009, opened 2019-09-21
msg_helpers: improve perf of unpackString by 20% and halve allocs

charlievieth:
**Edit:** forgot to include results from: `BenchmarkUnpackString_NoEscape`
```
benchmark                             old ns/op     new ns/op     delta
BenchmarkUnpackString-12              79.7          64.5          -19.07%
BenchmarkUnpackString_NoEscape-12     51.3          51.8          +0.97%

benchmark                             old allocs     new allocs     delta
BenchmarkUnpackString-12              2              1              -50.00%
BenchmarkUnpackString_NoEscape-12     1              1              +0.00%

benchmark                             old bytes     new bytes     delta
BenchmarkUnpackString-12              48            32            -33.33%
BenchmarkUnpackString_NoEscape-12     32            32            +0.00%
```

### DIFF
--- a/msg_helpers.go
+++ b/msg_helpers.go
@@ -261,28 +261,42 @@ func packUint64(i uint64, msg []byte, off int) (off1 int, err error) {
 }
 
 func unpackString(msg []byte, off int) (string, int, error) {
-	if off+1 > len(msg) {
+	start := off + 1
+	if start > len(msg) {
 		return "", off, &Error{err: "overflow unpacking txt"}
 	}
 	l := int(msg[off])
-	if off+l+1 > len(msg) {
+	end := start + l
+	if end > len(msg) {
 		return "", off, &Error{err: "overflow unpacking txt"}
 	}
 	var s strings.Builder
-	s.Grow(l)
-	for _, b := range msg[off+1 : off+1+l] {
+	p := msg[start:end]
+	for i, b := range p {
 		switch {
 		case b == '"' || b == '\\':
+			if s.Len() == 0 {
+				s.Grow(l * 2)
+				s.Write(p[:i])
+			}
 			s.WriteByte('\\')
 			s.WriteByte(b)
 		case b < ' ' || b > '~': // unprintable
+			if s.Len() == 0 {
+				s.Grow(l * 2)
+				s.Write(p[:i])
+			}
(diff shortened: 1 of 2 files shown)


msg_helpers.go
```diff
@@ -261,28 +261,42 @@ func packUint64(i uint64, msg []byte, off int) (off1 int, err error) {
...
 	}
 	var s strings.Builder
-	s.Grow(l)
-	for _, b := range msg[off+1 : off+1+l] {
+	p := msg[start:end]
+	for i, b := range p {
 		switch {
 		case b == '"' || b == '\\':
+			if s.Len() == 0 {
+				s.Grow(l * 2)
```
miekg: why `* 2 here`?

tmthrgd:
I’m against this change as is. It adds a lot of complexity for relatively little gain (should only be observable when the string needs escaping).

That being said, I have an idea for how to do this cleaner that I’ll try out and send a PR for.

msg_helpers.go
charlievieth: The `* 2` is because we need more than `len(s)` bytes to store the result, but we don't know how much more space we need so I arbitrarily picked `* 2` as it provides plenty of headroom while still being cheaper than than `len(s)`, which requires at least `* 3` bytes and two allocs - see [`Builder.grow()`](https://github.com/golang/go/blob/release-branch.go1.13/src/strings/builder.go#L64-L70).

A lower value (say `* 1.5`) could be used but one challenge is that escape sequences require either 2 or 4 bytes.  `* 2` is pretty aggressive when escaping with 2 bytes, but doesn't leave a lot of headroom when escaping with 4 bytes.

charlievieth:
> I’m against this change as is. It adds a lot of complexity for relatively little gain (should only be observable when the string needs escaping).
> 
> That being said, I have an idea for how to do this cleaner that I’ll try out and send a PR for.

The only complexity here is the lazy allocation of the `strings.Builder`, which boils down to a few `if` statements.  Remember the real gain here is the reduction of allocs, which reduces GC pressure and is critical to improving the throughput of things like CoreDNS (which is how I identified this function).

tmthrgd:
Closing in favour of #1011.

PR #1011, opened 2019-09-21
Improve unpackString performance

tmthrgd:
This is an alternative approach to #1009.

I'm not convinced this is really worth doing, because it does add a decent amount of complexity no matter how you approach it. That being said, it does show a performance improvement and reduces the allocations when there is escaping.

```
name                       old time/op    new time/op    delta
UnpackString/Escaped-12      83.7ns ± 7%    78.2ns ± 3%   -6.50%  (p=0.000 n=10+9)
UnpackString/Unescaped-12    57.8ns ± 9%    50.4ns ±13%  -12.74%  (p=0.000 n=10+10)

name                       old alloc/op   new alloc/op   delta
UnpackString/Escaped-12       48.0B ± 0%     32.0B ± 0%  -33.33%  (p=0.000 n=10+10)
UnpackString/Unescaped-12     32.0B ± 0%     32.0B ± 0%     ~     (all equal)

name                       old allocs/op  new allocs/op  delta
UnpackString/Escaped-12        2.00 ± 0%      1.00 ± 0%  -50.00%  (p=0.000 n=10+10)
UnpackString/Unescaped-12      1.00 ± 0%      1.00 ± 0%     ~     (all equal)
```

/cc @charlievieth

### DIFF
--- a/msg_helpers.go
+++ b/msg_helpers.go
@@ -265,12 +265,25 @@ func unpackString(msg []byte, off int) (string, int, error) {
 		return "", off, &Error{err: "overflow unpacking txt"}
 	}
 	l := int(msg[off])
-	if off+l+1 > len(msg) {
+	off++
+	if off+l > len(msg) {
 		return "", off, &Error{err: "overflow unpacking txt"}
 	}
+	escapedLen := l
+	for _, b := range msg[off : off+l] {
+		switch {
+		case b == '"' || b == '\\':
+			escapedLen++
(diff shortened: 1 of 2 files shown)


miekg:
Escapes are relatively rare as well, so might be the wrong target to
optimize for

miekg:
~6- 10 % faster is nothing to sneeze at, /lgtm

charlievieth:
LGTM

PR #1012, opened 2019-09-23
Simplify unpackString

gibson042:
Smaller, simpler, and (imperceptibly) faster by always doing everything in a single loop and keeping a bookmark rather than a `WriteByte` for every unescaped character:
```
$ go test -benchmem -bench=UnpackString >benchmark.new.txt && benchstat *old*txt *new*txt
name                      old time/op    new time/op    delta
UnpackString/Escaped-8       100ns ± 0%      96ns ± 0%   ~     (p=1.000 n=1+1)
UnpackString/Unescaped-8    55.1ns ± 0%    53.8ns ± 0%   ~     (p=1.000 n=1+1)

name                      old alloc/op   new alloc/op   delta
UnpackString/Escaped-8       32.0B ± 0%     32.0B ± 0%   ~     (all equal)
UnpackString/Unescaped-8     32.0B ± 0%     32.0B ± 0%   ~     (all equal)

name                      old allocs/op  new allocs/op  delta
UnpackString/Escaped-8        1.00 ± 0%      1.00 ± 0%   ~     (all equal)
UnpackString/Unescaped-8      1.00 ± 0%      1.00 ± 0%   ~     (all equal)
```

### DIFF
--- a/msg_helpers.go
+++ b/msg_helpers.go
@@ -269,33 +269,32 @@ func unpackString(msg []byte, off int) (string, int, error) {
 	if off+l > len(msg) {
 		return "", off, &Error{err: "overflow unpacking txt"}
 	}
-	escapedLen := l
-	for _, b := range msg[off : off+l] {
-		switch {
-		case b == '"' || b == '\\':
-			escapedLen++
-		case b < ' ' || b > '~': // unprintable
-			escapedLen += 3 // escapeByte always returns four characters
-		}
-	}
-	if escapedLen == l { // no escaping needed
-		return string(msg[off : off+l]), off + l, nil
-	}
 	var s strings.Builder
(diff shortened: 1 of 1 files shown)


miekg:
like it, also removes some of the duplication.

tmthrgd:
This is nearly exactly the same as #1009 and is much more likely to over-allocate than #1011.

tmthrgd:
That being said, I don't actually object to this.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Simplify unpackStri..." ]
>That being said, I don't actually object to this.

merge?

tmthrgd:
@miekg Your choice. I prefer what’s already been merged and don’t think the duplication is that bad, but I have no objections.

PR #1013, opened 2019-09-25
Update deps

miekg:


Dreamacro:
It seems better if run `go mod tidy` after update deps.

PR #1016, opened 2019-10-02
[fuzz] Fix crashes when parsing GENERATE

chantra:
Running the fuzzer on NewRR, some crashes came up that could be
prevented by checking that the token after the range is a Blank.
This diff checks that and return an error when the blank is not found.

### DIFF
--- a/generate.go
+++ b/generate.go
@@ -53,7 +53,14 @@ func (zp *ZoneParser) generate(l lex) (RR, bool) {
 		return zp.setParseError("bad range in $GENERATE range", l)
 	}
 
-	zp.c.Next() // _BLANK
+	// _BLANK
+	if l, ok := zp.c.Next(); ok {
+		if l.value != zBlank {
+			return zp.setParseError("Expect blank after $GENERATE range", l)
+		}
+	} else {
+		return zp.setParseError("Failed to parse token after $GENERATE range", l)
+	}
 
 	// Create a complete new string, which we then parse again.
 	var s string
--- a/generate.go
+++ b/generate.go
@@ -54,12 +54,9 @@ func (zp *ZoneParser) generate(l lex) (RR, bool) {
 	}
 
 	// _BLANK
-	if l, ok := zp.c.Next(); ok {
-		if l.value != zBlank {
-			return zp.setParseError("Expect blank after $GENERATE range", l)
-		}
-	} else {
-		return zp.setParseError("Failed to parse token after $GENERATE range", l)
+	l, ok := zp.c.Next()
+	if !ok || l.value != zBlank {
+			return zp.setParseError("garbage after $GENERATE range", l)
 	}
 
 	// Create a complete new string, which we then parse again.
--- a/generate_test.go
+++ b/generate_test.go
@@ -212,10 +212,10 @@ func TestCrasherString(t *testing.T) {
 	in  string
 	err string
 }{
-		{"$GENERATE 0-300103\"$$GENERATE 2-2", "dns: Expect blank after $GENERATE range: \"\\\"\" at line: 1:19"},
-		{"$GENERATE 0-5414137360", "dns: Expect blank after $GENERATE range: \"\\n\" at line: 1:22"},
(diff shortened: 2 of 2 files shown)


generate_test.go
```diff
@@ -204,3 +204,29 @@ $GENERATE 32-158 dhcp-${-32,4,d} A 10.0.0.$
...
+
+func TestCrasherString(t *testing.T) {
+	tests := []struct{
+	in  string
+	err string
+}{
+		{"$GENERATE 0-300103\"$$GENERATE 2-2", "dns: Expect blank after $GENERATE range: \"\\\"\" at line: 1:19"},
+		{"$GENERATE 0-5414137360", "dns: Expect blank after $GENERATE range: \"\\n\" at line: 1:22"},
+		{"$GENERATE 0-5414137360", "dns: Expect blank after $GENERATE range: \"\\n\" at line: 1:22"},
+		{"$GENERATE       11522-3668518066406258", "dns: Expect blank after $GENERATE range: \"\\n\" at line: 1:38"},
```
chantra: not that this type of high range number would crash NewRR with
```
runtime: goroutine stack exceeds 1000000000-byte limit
fatal error: stack overflow
```

if the rest of the input was legit.

generate_test.go
```diff
@@ -204,3 +204,29 @@ $GENERATE 32-158 dhcp-${-32,4,d} A 10.0.0.$
...
 }
+
+
+
+func TestCrasherString(t *testing.T) {
+	tests := []struct{
+	in  string
+	err string
+}{
+		{"$GENERATE 0-300103\"$$GENERATE 2-2", "dns: Expect blank after $GENERATE range: \"\\\"\" at line: 1:19"},
```
chantra: Specific crash here was:
```
runtime: goroutine stack exceeds 1000000000-byte limit
fatal error: stack overflow
```

generate_test.go
```diff
@@ -204,3 +204,29 @@ $GENERATE 32-158 dhcp-${-32,4,d} A 10.0.0.$
...
+		{"$GENERATE 0-5414137360", "dns: Expect blank after $GENERATE range: \"\\n\" at line: 1:22"},
```
chantra: dup, will remove

generate_test.go
chantra: program hanged (timeout 10 seconds)

generate_test.go
```diff
@@ -204,3 +204,29 @@ $GENERATE 32-158 dhcp-${-32,4,d} A 10.0.0.$
...
+		{"$GENERATE 0-5414137360", "dns: Expect blank after $GENERATE range: \"\\n\" at line: 1:22"},
+		{"$GENERATE       11522-3668518066406258", "dns: Expect blank after $GENERATE range: \"\\n\" at line: 1:38"},
+		{"$GENERATE 0-200\"(;00000000000000\n$$GENERATE 0-0", "dns: Expect blank after $GENERATE range: \"\\\"\" at line: 1:16"},
```
chantra: panic: runtime error: index out of range [2048] with length 2048

generate.go
```diff
@@ -53,7 +53,14 @@ func (zp *ZoneParser) generate(l lex) (RR, bool) {
 		return zp.setParseError("bad range in $GENERATE range", l)
 	}
 
-	zp.c.Next() // _BLANK
+	// _BLANK
+	if l, ok := zp.c.Next(); ok {
+		if l.value != zBlank {
+			return zp.setParseError("Expect blank after $GENERATE range", l)
```
miekg: to be more in sync with the rest of the parsing errors, this should be something ala "garbage after $GENERATE range" (same for the other error), might also drop the if-else in that case (most errors aren't super descriptive)

generate_test.go
miekg: should we have a sane upper limit?

chantra: I think it would make sense or you would get:
```
runtime: goroutine stack exceeds 1000000000-byte limit
fatal error: stack overflow
```
eventually.
The fix currently prevent the error from happening, but the fuzzer will definitely find a way to format a string that triggers this eventually.

miekg: ok, i'll add something. What *is* a sane upper limit? Does bind9 have one?

PR #1018, opened 2019-10-02
[scan] fix crashers when parsing comment

chantra:
When dealing with comments the parsers was potentially incrementing comi
variable twice. During the second access to com[], comi was possibly
longer than maxTok, causing an out of bound error:
panic: runtime error: index out of range [2048] with length 2048


scan.go
```diff
@@ -968,6 +968,11 @@ func (zl *zlexer) Next() (lex, bool) {
 				// was inside braces and we delayed adding it until now.
 				com[comi] = ' ' // convert newline to space
 				comi++
+				if comi >= len(com) {
+					l.token = "comment length insufficient for parsing"
+					l.err = true
+					return *l, true
+				}
```
chantra: it may be well possible that breaking here would be enough. I have little knowledge of this parsing code to make a good judgement call. Sanity checking here (like is done at the beginning of the for loop) just work (tm)

chantra:
go fmt does not seem to agree with my IDE. forced pushed after using go fmt.

scan_test.go
```diff
@@ -296,3 +296,231 @@ func BenchmarkZoneParser(b *testing.B) {
...
+				"rx_H__ a\tJ  \a\xfde s 60" +
+				"itr s t td  M__6VU_S" +
+				"B45BVM2i69 W__z0AEfP" +
+				"_g6me \"\"2Z_qY05APO_0" +
+				"SgAvjFYC_0__0riU_9K9" +
+				"_C__6af__wj_UbSYy_ge" +
+				"29S_s_Qe259q_kGoddr\t" +
+				" \t\t^  \t\t \t ;;;;;;;;;" +
+				";;;;;;;;;;",
+			`dns: comment length insufficient for parsing: "comment length insufficient for parsing" at line: 1:2030`,
```
miekg: are these 2 inputs actually testing the same thing? As they are quite long, can we just use 1?

scan.go
miekg: should this be `comi+1` as we increment it again on line 978? think it's fine to error here, instead of breaking and letting the check on line 847 pick it up

PR #1020, opened 2019-10-03
Limit $GENERATE range to 65535 steps

### LINKED ISSUE #1019: limit $GENERATE range to 2^16?
a $GENERATE range that is too large will cause a panic, it make sense to limit this to something reasonable. I propose to limit the number of steps to 65535, i.e.

`(stop - start) / step > 65535`, and error out if we hit this.

miekg:
Fixes #1019

### DIFF
--- a/generate.go
+++ b/generate.go
@@ -49,14 +49,14 @@ func (zp *ZoneParser) generate(l lex) (RR, bool) {
 	if err != nil {
 		return zp.setParseError("bad stop in $GENERATE range", l)
 	}
-	if end < 0 || start < 0 || end < start {
+	if end < 0 || start < 0 || end < start || (end-start)/step > 65535 {
 		return zp.setParseError("bad range in $GENERATE range", l)
 	}
 
 	// _BLANK
 	l, ok := zp.c.Next()
 	if !ok || l.value != zBlank {
-			return zp.setParseError("garbage after $GENERATE range", l)
+		return zp.setParseError("garbage after $GENERATE range", l)
 	}
 
 	// Create a complete new string, which we then parse again.
--- a/generate_test.go
+++ b/generate_test.go
@@ -205,31 +205,27 @@ $GENERATE 32-158 dhcp-${-32,4,d} A 10.0.0.$
 	}
 }
 
-func TestOutofRangeRange(t *testing.T) {
-	// outside limit
-	data := `$GENERATE 0-65536 dhcp A 10.0.0.1`
-	zp := NewZoneParser(strings.NewReader(data), ".", "test")
-	for _, ok := zp.Next(); ok; _, ok = zp.Next() {
-	}
-	if zp.Err() == nil || !strings.Contains(zp.Err().Error(), "dns: bad range in $GENERATE") {
-		t.Errorf("Expected %q error, got %s", "bad range", zp.Err())
-	}
 
-	// within limit
-	data = `$GENERATE 0-65535 dhcp A 10.0.0.1`
-	zp = NewZoneParser(strings.NewReader(data), ".", "test")
-	for _, ok := zp.Next(); ok; _, ok = zp.Next() {
-	}
-	if zp.Err() != nil {
-		t.Errorf("Expected nil error, got %s", zp.Err())
-	}
 
-	// within limit, due to /2
-	data = `$GENERATE 0-65536/2 dhcp A 10.0.0.1`
-	zp = NewZoneParser(strings.NewReader(data), ".", "test")
-	for _, ok := zp.Next(); ok; _, ok = zp.Next() {
+func TestCrasherString(t *testing.T) {
+	tests := []struct{
+	in  string
+	err string
+}{
+		{"$GENERATE 0-300103\"$$GENERATE 2-2", "dns: garbage after $GENERATE range: \"\\\"\" at line: 1:19"},
+		{"$GENERATE 0-5414137360", "dns: garbage after $GENERATE range: \"\\n\" at line: 1:22"},
+		{"$GENERATE       11522-3668518066406258", "dns: garbage after $GENERATE range: \"\\n\" at line: 1:38"},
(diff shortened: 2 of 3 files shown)


generate_test.go
```diff
@@ -229,3 +227,32 @@ func TestCrasherString(t *testing.T) {
...
 	}
 }
+
+func TestOutofRangeRange(t *testing.T) {
+	// outside limit
+	data := `$GENERATE 0-65537 dhcp A 10.0.0.1`
+	zp := NewZoneParser(strings.NewReader(data), ".", "test")
+	for _, ok := zp.Next(); ok; _, ok = zp.Next() {
+	}
+	if !strings.Contains(zp.Err().Error(), "dns: bad range in $GENERATE directive:") {
```
tmthrgd: This should have `zp.Err() == nil || !strings.Contains(...)`.

generate.go
```diff
@@ -72,6 +72,10 @@ func (zp *ZoneParser) generate(l lex) (RR, bool) {
 		s += l.token
 	}
 
+	if (end-start)/step > 65536 {
+		return zp.setParseError("bad range in $GENERATE directive", l)
```
tmthrgd: Can we merge this with the `end < 0 || start < 0 || end < start` if-statement above?

miekg:
done. PTAL

generate.go
```diff
@@ -49,14 +49,14 @@ func (zp *ZoneParser) generate(l lex) (RR, bool) {
 	if err != nil {
 		return zp.setParseError("bad stop in $GENERATE range", l)
 	}
-	if end < 0 || start < 0 || end < start {
+	if end < 0 || start < 0 || end < start || (end-start)/step > 65535 {
```
chantra: See my comment in https://github.com/miekg/dns/issues/1019#issuecomment-537989190 to align this with BIND behaviour. TL;DR both start and stop must be positive integers between 0 and (2^31)-1

generate_test.go
```diff
@@ -205,27 +205,31 @@ $GENERATE 32-158 dhcp-${-32,4,d} A 10.0.0.$
...
 
-func TestCrasherString(t *testing.T) {
-	tests := []struct{
-	in  string
-	err string
-}{
-		{"$GENERATE 0-300103\"$$GENERATE 2-2", "dns: garbage after $GENERATE range: \"\\\"\" at line: 1:19"},
-		{"$GENERATE 0-5414137360", "dns: garbage after $GENERATE range: \"\\n\" at line: 1:22"},
-		{"$GENERATE       11522-3668518066406258", "dns: garbage after $GENERATE range: \"\\n\" at line: 1:38"},
-		{"$GENERATE 0-200\"(;00000000000000\n$$GENERATE 0-0", "dns: garbage after $GENERATE range: \"\\\"\" at line: 1:16"},
```
chantra: some of those test were generating different errors, like timeout (I believe the second test was a timeout). Let me find the original error they were triggering.

generate_test.go
```diff
@@ -205,27 +205,31 @@ $GENERATE 32-158 dhcp-${-32,4,d} A 10.0.0.$
...
+		t.Errorf("Expected nil error, got %s", zp.Err())
+	}
 
-func TestCrasherString(t *testing.T) {
-	tests := []struct{
-	in  string
-	err string
-}{
-		{"$GENERATE 0-300103\"$$GENERATE 2-2", "dns: garbage after $GENERATE range: \"\\\"\" at line: 1:19"},
-		{"$GENERATE 0-5414137360", "dns: garbage after $GENERATE range: \"\\n\" at line: 1:22"},
```
chantra: original error: `program hanged (timeout 10 seconds)`

generate_test.go
```diff
@@ -205,27 +205,31 @@ $GENERATE 32-158 dhcp-${-32,4,d} A 10.0.0.$
...
+	if zp.Err() != nil {
+		t.Errorf("Expected nil error, got %s", zp.Err())
+	}
 
-func TestCrasherString(t *testing.T) {
-	tests := []struct{
-	in  string
-	err string
-}{
-		{"$GENERATE 0-300103\"$$GENERATE 2-2", "dns: garbage after $GENERATE range: \"\\\"\" at line: 1:19"},
```
chantra: runtime: goroutine stack exceeds 1000000000-byte limit fatal error: stack overflow

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Limit $GENERATE ran..." ]
>
>
>
>>
>-func TestCrasherString(t *testing.T) {
>-	tests := []struct{
>-	in  string
>-	err string
>-}{
>-		{"$GENERATE 0-300103\"$$GENERATE 2-2", "dns: garbage after $GENERATE range: \"\\\"\" at line: 1:19"},
>-		{"$GENERATE 0-5414137360", "dns: garbage after $GENERATE range: \"\\n\" at line: 1:22"},
>
>original error: `program hanged (timeout 10 seconds)`

yes, with only 2^16-1 we don't hit this - I think the bind limits are a bit 
crazy

generate_test.go
```diff
@@ -205,27 +205,31 @@ $GENERATE 32-158 dhcp-${-32,4,d} A 10.0.0.$
...
-		{"$GENERATE 0-5414137360", "dns: garbage after $GENERATE range: \"\\n\" at line: 1:22"},
-		{"$GENERATE       11522-3668518066406258", "dns: garbage after $GENERATE range: \"\\n\" at line: 1:38"},
```
chantra: program hanged (timeout 10 seconds)

generate_test.go
chantra: panic: runtime error: index out of range [2048] with length 2048

miekg:
@chantra happy to merge this as it now?

chantra:
sure, I would be tempted to leave test case that catch garbage after the range/step token though :)

miekg:
ok, adding that back in

Issue #810, opened 2018-11-12
ZoneParser error with "TSIG" records


lanrat:
When using `dns.NewZoneParser()` on a zone file, I get the following error:

```
dns: unknown RR type: "TSIG" at line: 433:36
```
Line 433 of the zone file contains the following line:
```
monitor.reg.neustar.com. 0      ANY     TSIG    hmac-md5.sig-alg.reg.int. 1541638220 300 16 Cocdc1taK0N1SnSb4sohRQ== 4270 NOERROR 0
```

Is this a bug? For the purposes of my project I currently don't care about TSIG records, would it be possible to tell the parser to ignore records it does not support?

miekg:
[ Quoting <notifications@github.com> in "[miekg/dns] ZoneParser error with "..." ]
>When using `dns.NewZoneParser()` on a zone file, I get the following error:
>
>```
>dns: unknown RR type: "TSIG" at line: 433:36
>```
>Line 433 of the zone file contains the following line:
>```
>monitor.reg.neustar.com. 0      ANY     TSIG    hmac-md5.sig-alg.reg.int. 1541638220 300 16 Cocdc1taK0N1SnSb4sohRQ== 4270 NOERROR 0
>```
>
>Is this a bug? For the purposes of my project I currently don't care about TSIG records, would it be possible to tell the parser to ignore records it does not support?

TSIG records don't have a TEXT representation, so they can't be read in.
If you need this to somehow work, you need to use RFC 3597 represenation.

lanrat:
Unfortunately I'm not in control of the input zone file. It's being generated by Neustar with what looks like an AXFR from dig.

If they are not conforming to the spec is there another way to make this work? Or alternatively, would it be possible to allow the parser to continue after throwing a noncritical error, allowing the user to decide what errors are worth stopping on?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] ZoneParser error wi..." ]
>Unfortunately I'm not in control of the input zone file. It's being generated by Neustar with what looks like an AXFR from dig.

you're not getting a zonefile, your getting an axfr with a transaction signature 
attached. Those are 2 different things. TSIG has no meaning in a zone and 
doesn't have a text representation, so it errors.

>If they are not conforming to the spec is there another way to make this work? Or alternatively, would it be possible to allow the parser to continue after throwing a noncritical error, allowing the user to decide what errors are worth stopping on?

[...]

PR #1030, opened 2019-10-19
Provide zp.ClearError()

nerdlem:
Various zone files distributed by ICANN via the CZDS as well as via direct interfaces provided by the respective registry operator, contain malformed or bogus records.

Examples at the time of this writing include .us as distributed by Neustar and .tel as available through the CZDS.

High-performance processing pipelines can benefit from filtering those bogus records on the fly.

The ClearError() method allows these classes of applications to simply have the parser state machine to step through these errors without changing the parser or otherwise affecting its correctness.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -8,3 +8,4 @@ Peter van Dijk
 Omri Bahumi
 Alex Sergeyev
 James Hartig
+Luis Muñoz
--- a/scan.go
+++ b/scan.go
@@ -275,6 +275,14 @@ func NewZoneParser(r io.Reader, origin, file string) *ZoneParser {
 	}
 }
 
+// ClearError resets the error state of the parser. This is useful to forcibly
+// parse widely distributed but malformed zones such as .us, which include TSIG
+// pseudo-RRs. By clearing the error state from the parsing loop, the user is
+// able to skip the errors.
+func (zp *ZoneParser) ClearError() {
+	zp.parseErr = nil
+}
+
 // SetDefaultTTL sets the parsers default TTL to ttl.
 func (zp *ZoneParser) SetDefaultTTL(ttl uint32) {
 	zp.defttl = &ttlState{ttl, false}
--- a/scan_test.go
+++ b/scan_test.go
@@ -45,6 +45,36 @@ func TestParseZoneGenerate(t *testing.T) {
 	}
 }
 
+func TestClearError(t *testing.T) {
+	zone := "monitor.reg.neustar.com. 0\tANY\tTSIG\thmac-md5.sig-alg.reg.int. 1571283626 300 16 tCSU0OXlP3q2R9OAUe1lHQ== 32393 NOERROR 0\nfoo\tIN\tA\t127.0.0.1"
+	zp := NewZoneParser(strings.NewReader(zone), "", "")
+	i := 0
+	pass := false
+	cleared := false
+
+Again:
+	for rr, ok := zp.Next(); ok; rr, ok = zp.Next() {
+		i++
+		if rr != nil {
+			pass = rr.Header().Class == ClassINET && rr.Header().Rrtype == TypeA
+		}
+
+		if err := zp.Err(); err != nil {
+			zp.ClearError()
+			cleared = true
+			goto Again
+		}
+	}
+
+	if !pass {
+		t.Error("Unable to jump over TSIG pseudo-record")
+	}
+
+	if !cleared {
+		t.Error("Parse errors should have been detected")
+	}
+}
+
 func TestParseZoneInclude(t *testing.T) {
 
 	tmpfile, err := ioutil.TempFile("", "dns")


nerdlem:
Thanks for taking the time to read the PR @tmthrgd. Actually the parser recovers very well because the it is well written. It consumes tokens – and returns to an error state – until the end-of-line is consumed. At that point, it synchronizes again and continues with the next RR.

I agree that the, ahem, "zonefile" is malformed. However this change does not break the standard usage – i.e., if you use the package as per the documentation, the ZoneParser object will abort on first error – which I think is important to maintain.

For the special case of parsing these pseudo-zonefiles, the user can choose to forcibly reset the error state and resume the parsing loop. My sample code looks like this

```go
Again:
	for rr, ok := zp.Next(); ok; rr, ok = zp.Next() {
              // Do something with rr
	}

	if err := zp.Err(); err != nil {
		fmt.Println(err)
		zp.ClearError()
		goto Again
	}
```

Here's what it looks like (this is a line taken from the `us.zone.gz` file).

```
90558 IN NS
dns: unknown RR type: "TSIG" at line: 90764:36
dns: not a TTL: "hmac-md5.sig-alg.reg.int." at line: 90764:62
dns: expecting RR type or class, not this...: "300" at line: 90764:77
dns: expecting RR type or class, not this...: "GIIV8pwTGSVYsCjnjilGtQ==" at line: 90764:105
dns: expecting RR type or class, not this...: "NOERROR" at line: 90764:119
dns: expecting RR type or class, not this...: "\n" at line: 90764:121
90559 IN NS
90560 IN NS
```

miekg:
See https://twitter.com/miekg/status/1065310757657694208

miekg:
I this case complaining with .US might help

nerdlem:
Yes, I'm aware of these previous discussions. I agree on your stance, but I – and others – need the parser to deal with these pseudo-zones, which is why provided the code in hopes that the trivial change would be accepted, as it does not break the parser when used as per the documentation.

Perhaps bringing this up in dnsop? I'll also give the CZDS mailing lists a try. .us is not the only zone with breakage.  There's also .tel (which is also discussed in another thread related to this parser). I believe the method I implemented will help with both cases. Likely all of Neustar's DNS zones carry invalid contents.

Best regards

-lem

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Provide zp.ClearErr..." ]
>Yes, I'm aware of these previous discussions. I agree on your stance, but I – and others – need the parser to deal with these pseudo-zones, which is why provided the code in hopes that the trivial change would be accepted, as it does not break the parser when used as per the documentation.
>
>Perhaps bringing this up in dnsop? I'll also give the CZDS mailing lists a try. .us is not the only zone with breakage.  There's also .tel (which is also discussed in another thread related to this parser). I believe the method I implemented will help with both cases. Likely all of Neustar's DNS zones carry invalid contents.

dnsop sounds like a good thing to do. TSIG doesn't have a text representation, 
so bind9 (it's probably bind) is doing it wrong and Neustar look a bit clueless.

nerdlem:
> dnsop sounds like a good thing to do. TSIG doesn't have a text representation, so bind9 (it's probably bind) is doing it wrong and Neustar look a bit clueless.

It seems like they are doing the equivalent of a `dig axfr`. That's perhaps a shortcut they took – it could work with a bit more work I think.

nerdlem:
> > dnsop sounds like a good thing to do. TSIG doesn't have a text representation, so bind9 (it's probably bind) is doing it wrong and Neustar look a bit clueless.
> 
> It seems like they are doing the equivalent of a `dig axfr`. That's perhaps a shortcut they took – it could work with a bit more work I think.

I can indeed confirm that these TSIG pseudo-records appear when a zone transfer is triggered by `dig`. It would seem that simply passing the output through a `grep` would clear this issue effectively on Neustar part.

tmthrgd:
Is the issue only about fake TSIG records?

nerdlem:
@tmthrgd there are also zones with occasional malformed records. .tel was one of them.

tmthrgd:
@nerdlem In what way is .tel malformed? I'd much rather something like ignoring TSIG records.

nerdlem:
> @nerdlem In what way is .tel malformed? I'd much rather something like ignoring TSIG records.

Logs are gone, but there was a syntax error in a RR within the zone. I haven't seen this issue recur since October 23rd.

tmthrgd:
@nerdlem That being the case, would (optionally?) ignoring TSIG records solve your problems?

miekg:
If the TSIG was included as an unknown RR I think it would just have worked. Closing this PR, please open an issue to ponder any future direction this lib should take to tackle or not tackle this.

nerdlem:
I have another use case for the state reset. Current behavior fails hard and early on the first error. By allowing the caller to reset the state, additional errors can be detected on the zone file on a single pass.

Since this second use case is gaining traction for some of my applications, I'll be forking and maintaining the library, as I need this functionality.

Thank you very much for your time and best regards.

miekg:
Ok. Let's see how that goes, and as said feel free to open an issue

lanrat:
For visibility this is the same problem encountered in #810, which lists some more of the example where this parser is unable to parse some zones.

I've also been maintaining a fork very similar to this PR that adds the ability to skip over unparsable records. I agree its not ideal, but better than nothing.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Provide zp.ClearErr..." ]
>For visibility this is the same problem encountered in #810, which lists some more of the example where this parser is unable to parse some zones.
>
>I've also been maintaining a fork very similar to this PR that adds the ability to skip over unparsable records. I agree its not ideal, but better than nothing.

ok, If multiple folks have this need, then we should add something like ClearError

nerdlem:
I believe it would be better to add ClearErr() – or however it gets called in the end – and support the use case from upstream. That would allow me to kill my branch. Likewise @lanrat would be in a similar position :-)

lanrat:
This is my implementation of ClearErr(), which also moves the parser to the next line in an effort to skip the rest of the unknown record type.
```
func (zp *ZoneParser) ResetErr() error {
       if zp.parseErr != nil {
               nextLine := false
               for !nextLine {
                       l, _ := zp.c.Next()
                       if l.value == zNewline || l.value == zEOF {
                               nextLine = true
                       }
               }
               zp.parseErr = nil
       }
       return nil
}
```
While this works, I would prefer a solution where the parser could return unknown record types. Or at least have a new error type that is returned so it is easy to distinguish between bad records in the zone and other errors that may arise.

nerdlem:
What @lanrat proposes is more general than my `ClearError()`, which has a narrow application.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Provide zp.ClearErr..." ]
>This is my implementation of ClearErr(), which also moves the parser to the next line in an effort to skip the rest of the unknown record type.
>```
>func (zp *ZoneParser) ResetErr() error {
>       if zp.parseErr != nil {
>               nextLine := false
>               for !nextLine {
>                       l, _ := zp.c.Next()
>                       if l.value == zNewline || l.value == zEOF {
>                               nextLine = true
>                       }
>               }
>               zp.parseErr = nil
>       }
>       return nil
>}
>```
>While this works, I would prefer a solution where the parser could return unknown record types. Or at least have a new error type that is returned so it is easy to distinguish between bad records in the zone and other errors that may arise.

A new error type can work, but i'm not sure if you can capture what you're after.

Also making up an unknown RR is doable, but doesn't define a very narrow use-case? I.e.
you can parse just enough to make sense of an RR - this must be at least the Type of the
RR.

I rather have something general (enough) and let all these decision to the caller.

On naming: i *think* spelling Error fullout is more Go like, see text/scanner for example.

tmthrgd:
I still really don’t want to do this and it’s likely to be brittle and hard to maintain. There will always be some failures we can’t recover from and we could easily end up in a corrupted or invalid internal state.

#810 is also about TSIG records so we’re much better of dealing with that explicitly. Either parsing them or ignoring them, either by default or opt-in.

miekg:
Fair enough. We can also just define the presentation format for TSIG as
BIND seems to define a defacto one anyway.

nerdlem:
I would argue that constraining the issue to just the TSIG records is a limited view. This is akin to having a C compiler that bails out on the first error it finds, leaving further errors undiscovered. My use case is satisfied with a way to reset parser state so that all the errors can be reported on a single scan if so desired.

Honestly I don't understand the resistance to implement a feature that is being requested a few times and for which patches have been provided. I respect @tmthrgd view of this being "likely .. brittle and hard to maintain" but strongly disagree. The parser, as designed, will resynchronize at EOL upon a parse error.  In order to make it hard to maintain, you would have to actively change the way the parser is written now.

For pathological files – i.e., zone files that are already broken – the existing implementation would simply give up on the first error and my patches maintain my behavior. For those that wish to go further ahead, they need to actually make the effort of trapping the error, issuing the `ClearError()` call and continue the loop. I shown that code snippet on my comment on Oct 19.

Since that comment, that code has parsed in excess of 2.7E10 records composed of various custom-generated, CZDS sourced zone files and privately sourced TLD zones. I've observed no issues or flakiness. Indeed, this parser is faster and more robust that the alternatives I tried early on.

As a compromise, I would be happy to take responsibility for supporting my `ClearError()` code if this would help get the feature accepted in upstream.

miekg:
We can't make tsig a thing btw. This means it can be (possible) picked up
by zones and actually served by it. We need to keep erroring on it

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Provide zp.ClearErr..." ]
>As a compromise, I would be happy to take responsibility for supporting my `ClearError()` code if this would help get the feature accepted in upstream.

As stated I'm not againt this addition. I think part of the push back is:

* you have a broken zonefile
* the expection that ClearErr/ClearError will continue to work

tmthrgd:
The test case included in this PR doesn't pass and I presume you'd have to write code that looks something like this (*I noticed https://github.com/miekg/dns/pull/1030#issuecomment-544140542 after writing this comment*):
```go
func TestClearError(t *testing.T) {
	const zone = "$GENERATE 0-1 monitor.reg.neustar.com. 0\tANY\tTSIG\thmac-md5.sig-alg.reg.int. 1571283626 300 16 tCSU0OXlP3q2R9OAUe1lHQ== 32393 NOERROR 0\nfoo\tIN\tA\t127.0.0.1"
	//const zone = "$INCLUDE /tmp/broken\n"
	zp := NewZoneParser(strings.NewReader(zone), "", "")
	zp.SetIncludeAllowed(true)
	pass := false
	cleared := false

Again:
	for rr, ok := zp.Next(); ok; rr, ok = zp.Next() {
		pass = rr.Header().Class == ClassINET && rr.Header().Rrtype == TypeA
	}
	if zp.Err() != nil {
		zp.ClearError()
		cleared = true
		goto Again
	}

	if !pass {
		t.Error("Unable to jump over TSIG pseudo-record")
	}

	if !cleared {
		t.Error("Parse errors should have been detected")
	}
}
```

Even then that only works using `ClearError()` and fails using `ResetErr()` for reasons that I haven't looked into.

Both implementations in this thread run into numerous other problems though:

- If a parsing error occurs in a $GENERATE directive, it causes an infinite loop: `"$GENERATE 0-1 monitor.reg.neustar.com. 0\tANY\tTSIG\thmac-md5.sig-alg.reg.int. 1571283626 300 16 tCSU0OXlP3q2R9OAUe1lHQ== 32393 NOERROR 0\nfoo\tIN\tA\t127.0.0.1"`.
- If the reader returns an error, `NewZoneParser(errReader{}, "", "")`, it will cause an infinite loop.
- If a parsing error occurs in an $INCLUDEd file, it causes an infinite loop:
```go
zp := NewZoneParser(strings.NewReader("$INCLUDE /tmp/broken\n"), "", "")
zp.SetIncludeAllowed(true)
```
with /tmp/broken being:
```
$GENERATE 0-1 monitor.reg.neustar.com. 0	ANY	TSIG	hmac-md5.sig-alg.reg.int. 1571283626 300 16 tCSU0OXlP3q2R9OAUe1lHQ== 32393 NOERROR 0
foo	IN	A	127.0.0.1
```

If you're more careful and avoid clearing all errors, and instead clear only `*ParseError`s, like so:
```go
Again:
	for rr, ok := zp.Next(); ok; rr, ok = zp.Next() {
		pass = rr.Header().Class == ClassINET && rr.Header().Rrtype == TypeA
	}
	if _, ok := zp.Err().(*ParseError); ok {
		zp.ClearError()
		cleared = true
		goto Again
	}
```

You fix the erroring reader case, with very messy code, but still fall victim to infinite loops in the $GENERATE and $INCLUDE case. Except if the error reader returns a `*ParseError` for whatever reason, then you're back to an infinite loop anyway.

Maybe you expand the `ClearError`/`ResetErr` implementation to reach into the sub parser and try and jerry-rig that into a workable state (including likely somehow reaching into the underlying `*generateReader` and manipulating that), and then also give it a return value to indiciate whether to retry (`func (*ZoneParser) ClearError() (canRetry bool)`). That might get you past the infinite loops, but it still won't meet either the documentation or what a caller would expect the method to do.

It can't properly recover from parsing errors that occur in the zone lexer:
```go
zone := "; this is a really long comment " + strings.Repeat("x", maxTok) + "\n" +
		"monitor.reg.neustar.com. 0\tANY\tTSIG\thmac-md5.sig-alg.reg.int. 1571283626 300 16 tCSU0OXlP3q2R9OAUe1lHQ== 32393 NOERROR 0\n" +
		"foo\tIN\tA\t127.0.0.1"
```
or:
```go
zone := "monitor.reg.neustar.com. 0\tANY\tTSIG\thmac-md5.sig-alg.reg.int. 1571283626 300 16 tCSU0OXlP3q2R9OAUe1lHQ== 32393 NOERROR 0\n" +
		"; this is a really long comment " + strings.Repeat("x", maxTok) + "\n" +
		"foo\tIN\tA\t127.0.0.1"
```
or: `")\nfoo\tIN\tA\t127.0.0.1"`.

Even when it appears to work it can return suitably incorrect/corrupt data:
```go
func TestClearError(t *testing.T) {
	zone := "example.com.\tIN\tA\t127.0.0.1\n" +
		"monitor.reg.neustar.com. 0\tANY\tTSIG\thmac-md5.sig-alg.reg.int. 1571283626 300 16 tCSU0OXlP3q2R9OAUe1lHQ== 32393 NOERROR 0\n" +
		"foo\tIN\tA\t127.0.0.1"
	zp := NewZoneParser(strings.NewReader(zone), "", "")
	pass := false
	cleared := false
	ignored := 0

Again:
	for rr, ok := zp.Next(); ok; rr, ok = zp.Next() {
		t.Log("found record", rr)
		pass = rr.Header().Class == ClassINET && rr.Header().Rrtype == TypeA
	}
	if _, ok := zp.Err().(*ParseError); ok {
		if ignored < 10 {
			t.Log("ignored error:", zp.Err())
			ignored++
		}

		zp.ClearError()
		cleared = true
		goto Again
	}

	if err := zp.Err(); err != nil {
		t.Error(err)
	}

	if !pass {
		t.Error("Unable to jump over TSIG pseudo-record")
	}

	if !cleared {
		t.Error("Parse errors should have been detected")
	}
}
```

When run this results in:
```
=== RUN   TestClearError
--- PASS: TestClearError (0.00s)
    scan_test.go:63: ignored error: dns: unknown RR type: "TSIG" at line: 1:36
    scan_test.go:63: ignored error: dns: not a TTL: "hmac-md5.sig-alg.reg.int." at line: 1:62
    scan_test.go:63: ignored error: dns: expecting RR type or class, not this...: "300" at line: 1:77
    scan_test.go:63: ignored error: dns: expecting RR type or class, not this...: "tCSU0OXlP3q2R9OAUe1lHQ==" at line: 1:105
    scan_test.go:63: ignored error: dns: expecting RR type or class, not this...: "NOERROR" at line: 1:119
    scan_test.go:63: ignored error: dns: no blank before TTL: "\n" at line: 1:120
    scan_test.go:63: ignored error: dns: bad owner name: "foo" at line: 2:4
    scan_test.go:58: found record monitor.reg.neustar.com.      0       IN      A       127.0.0.1
```

The `A` record very clearly didn't have an owner name of `monitor.reg.neustar.com.`.

Running the code through a fuzzer also produces a few other cases where the method doesn't work properly and the code can't recover from errors like it promises, though I haven't taken any time to examine them nor did I seed the fuzzer with zone files.

----

My objections to this isn't simply stubbornness, nor am I trying to argue in bad faith.
- The method is impossible to use safely without being one maliciously crafted input away from an infinite loop; both proposed API's are insanely hard to use right and leave you uncomfortable close to infinite loops.
- It forces future design choices in the parser to be constantly recoverable.
- It's not a particularly Go-like API—I can't think of anything in the standard library or x/ repos that is like this.
- To get around the $INCLUDE and $GENERATE problems you have to be reaching into the sub parsers and that still doesn't get you a reliable way to restart the parser.
- It can't recover from parsing errors that are surfaced in the lexer.

I'm definitely sympathetic, but what you're asking for is something orthogonal to what this zone parser fundamentally is. Addressing specifically the problem of stray TSIG records is doable (@miekg by-default, or by opt-in, ignoring of TSIG records wouldn't risk surfacing them), but something more general like this just isn't and it shouldn't be within the scope of this project.

It's ultimately not my say, but I'm strongly against including anything along these lines. Sorry.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Provide zp.ClearErr..." ]
>I'm definitely sympathetic, but what you're asking for is something orthogonal to what this zone parser fundamentally is. Addressing specifically the problem of stray TSIG records is doable (@miekg by-default, or by opt-in, ignoring of TSIG records wouldn't risk surfacing them), but something more general like this just isn't and it shouldn't be within the scope of this project.
>
>It's ultimately not my say, but I'm strongly against including anything along these lines. Sorry.

Well, I think you made a very strong technical case for not adding ClearError to the api
:)

Ignoring TSIG sounds very icky, there is no precedent for ignoring records in zonefiles.

All in all, I think we're back to square 1; don't add TSIG records to your zone file and
filing a bug with ISC might be worth while as well; to add a comment in front of the TSIG
record.

nerdlem:
Thank you @tmthrgd for that sound and thorough explanation. It certainly covers use cases I'm not dealing with. I agree with @miekg that the issue here is that we're attempting to parse something that isn't really a zone file. I like the idea of filing a bug with ISC to get `dig` taken care of, although there's no warranty that the affected backend operators will update their toolchain to fix this. After all, you could easily fix this with good ole' `grep`.

I wonder if rather than trying to reset state in a parser that is a few floors below, it might be simpler to have a flag – initially fed to the main parser and inherited to any lower level parser started afterwards  – which would cause errors to be reported as a special _pseudo record_ and then simply consuming input until next newline.

Is this an approach you considered? Do you see an _a priori_ reason for this not to work? I would be happy to give it a shot if you think this might work.

Best regards

-lem

PR #1032, opened 2019-10-23
Use packed length instead of struct length fields

andrewtj:
This follows up #426. This time around I thought I'd try adjusting the lengths after packing the values which is probably a bit simpler. I haven't considered performance - if this approach seems okay, avoiding the `packUintX`'s might be worthwhile.

Also I'm a bit rusty so I may have missed some things.


miekg:
I'll try to review this soon(-ish)

PR #1033, opened 2019-10-23
[zone parser] disallow nested $GENERATE directive

chantra:
While the range number of GENERATE is now limited, one can pass
a line with 2 $GENERATE directive that will exponentially increase the
time spent generating RRs.
Limit to only one per line.
Fixes #1020

### DIFF
--- a/generate.go
+++ b/generate.go
@@ -85,6 +85,7 @@ func (zp *ZoneParser) generate(l lex) (RR, bool) {
 	}
 	zp.sub = NewZoneParser(r, zp.origin, zp.file)
 	zp.sub.includeDepth, zp.sub.includeAllowed = zp.includeDepth, zp.includeAllowed
+	zp.sub.generateDisallowed = true
 	zp.sub.SetDefaultTTL(defaultTtl)
 	return zp.subNext()
 }
--- a/scan.go
+++ b/scan.go
@@ -248,6 +248,7 @@ type ZoneParser struct {
 	includeDepth uint8
 
 	includeAllowed bool
+	generateDisallowed bool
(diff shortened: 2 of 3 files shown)


chantra:
fixing test.... had the line number wrong.

generate.go
```diff
@@ -85,6 +85,7 @@ func (zp *ZoneParser) generate(l lex) (RR, bool) {
 	}
 	zp.sub = NewZoneParser(r, zp.origin, zp.file)
 	zp.sub.includeDepth, zp.sub.includeAllowed = zp.includeDepth, zp.includeAllowed
+	zp.sub.SetGenerateAllowed(false)
```
tmthrgd: I don’t think we should add any exported API for this, just a direct `zp.sub.generateAllowed = false` should work fine.

scan.go
```diff
@@ -248,6 +248,7 @@ type ZoneParser struct {
 	includeDepth uint8
 
 	includeAllowed bool
+	generateAllowed bool
```
tmthrgd: I’d prefer we flipped this to `generateDisallowed`, that way we don’t have to initialise it in `NewZoneParser`. It’s not a big deal though and this could be more consistent.

miekg:
What's the behaviour change? The desc and possibly the commits are not
saying anything on this.

Code wise lgtm

tmthrgd:
@miekg Just that previously $GENERATE in $GENERATE was allowed (though I can't imagine anyone using it) and now it's disallowed.

miekg:
Thanks. I think changing that is fine. So lgmt

PR #1035, opened 2019-11-10
Add RdataString() to all types already implementing RR interface

UmanShahzad:
Needed a way to gather the rdata as a string minus the header string for a RR, and the only way to really do it for most RRs would be to chop off the header string after calling RR.String(). This PR adds RR.RdataString() so we get it directly, and RR.String() behaves the same in all cases.

### DIFF
--- a/dns.go
+++ b/dns.go
@@ -29,6 +29,8 @@ type RR interface {
 	// Header returns the header of an resource record. The header contains
 	// everything up to the rdata.
 	Header() *RR_Header
+	// RdataString returns the string format of the rdata of a resource record.
+	RdataString() string
 	// String returns the text representation of the resource record.
 	String() string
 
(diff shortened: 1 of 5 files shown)


miekg:
This might be what your looking for https://godoc.org/github.com/miekg/dns#Field (not super ideal api, but will get the job done)

UmanShahzad:
@miekg That's not a solution. Just see https://github.com/miekg/dns/pull/1035/files#diff-e871404fd69cab40521d6b2178847cf7R913-R955 or many other types in there with complex RdataString implementations.

UmanShahzad:
@miekg Can you check this out again?

tmthrgd:
@UmanShahzad "Needed a way to gather the rdata as a string minus the header string for a RR," Can you explain why you need this?

UmanShahzad:
@tmthrgd We were loading rdata strings from CZDS/Verisign zonefiles into a db and doing some analysis.

tmthrgd:
@UmanShahzad Not all the records actually have well formed ‘rdata strings’ and some are simply for presentation and subject to change. Without knowing more it sounds like something like JSON encoding the record structs would be more appropriate and stable.

PR #1039, opened 2019-11-29
Speed up NextLabel and PrevLabel

jpicht:
NextLabel and PrevLabel are heavily used in the "file" plugin in coredns. This change introduces benchmark tests and improves the performance (on my machine).

It eliminates all allocations and uses strings in a way that should make L1-Cache-Hits more likely.

The time for coredns "file"-plugin Requests is cut roughly in half.

### DIFF
--- a/labels.go
+++ b/labels.go
@@ -127,18 +127,19 @@ func Split(s string) []int {
 // Also see PrevLabel.
 func NextLabel(s string, offset int) (i int, end bool) {
 	for i = offset; i < len(s)-1; i++ {
-		if s[i] == '.' {
-			j := i - 1
-			for j >= 0 && s[j] == '\\' {
-				j--
-			}
-
-			if (j-i)%2 == 0 {
-				continue
-			}
-
-			return i + 1, false
+		if s[i] != '.' {
+			continue
 		}
+		j := i - 1
+		for j >= 0 && s[j] == '\\' {
+			j--
+		}
+
+		if (j-i)%2 == 0 {
+			continue
+		}
+
+		return i + 1, false
 	}
 	return i + 1, true
 }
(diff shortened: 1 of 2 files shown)


labels.go
```diff
@@ -126,18 +126,17 @@ func Split(s string) []int {
...
 	for i = offset; i < len(s)-1; i++ {
-		switch s[i] {
-		case '\\':
-			quote = !quote
-		default:
-			quote = false
-		case '.':
-			if quote {
-				quote = !quote
+		if s[i] == '.' {
```
tmthrgd: This should be `if s[i] != '.' { continue }` to reduce indenting.

labels.go
```diff
@@ -152,14 +151,31 @@ func PrevLabel(s string, n int) (i int, start bool) {
...
+
+	l := len(s) - 1
+	if s[l] == '.' {
+		l--
 	}
-	if n > len(lab) {
-		return 0, true
+
+	for ; l >= 0 && n > 0; l-- {
+		if s[l] == '.' {
```
tmthrgd: Same here.

labels.go
```diff
@@ -126,18 +126,17 @@ func Split(s string) []int {
 // The bool end is true when the end of the string has been reached.
 // Also see PrevLabel.
 func NextLabel(s string, offset int) (i int, end bool) {
-	quote := false
 	for i = offset; i < len(s)-1; i++ {
```
tmthrgd: What’s the performance like if you use `strings.Index(s[i:], ".")`?

labels.go
```diff
@@ -126,18 +126,17 @@ func Split(s string) []int {
...
-		switch s[i] {
-		case '\\':
-			quote = !quote
-		default:
-			quote = false
-		case '.':
-			if quote {
-				quote = !quote
+		if s[i] == '.' {
+			ii := i - 1
```
tmthrgd: `ii` isn’t really a nice variable name, `j` might be better.

labels.go
jpicht: Will try. That will probably heavily depend on if the optimizer inlines Index.

jpicht: It's about 25% slower using IndexByte. I implemented two versions. The code is [here](https://github.com/jpicht/dns/tree/IndexByteLoop) and [here](https://github.com/jpicht/dns/tree/IndexByteRecursive).

labels.go
jpicht: Done.

labels.go
jpicht: And done.

labels.go
jpicht: Done.

labels.go
```diff
@@ -152,14 +152,32 @@ func PrevLabel(s string, n int) (i int, start bool) {
 	if n == 0 {
 		return len(s), false
 	}
-	lab := Split(s)
-	if lab == nil {
-		return 0, true
+
+	l := len(s) - 1
+	if s[l] == '.' {
```
tmthrgd: This needs a check that `l >= 0` otherwise it will panic for an empty string.

labels.go
tmthrgd: That’s disappointing. Oh well.

labels.go
jpicht: fixed

miekg:
thank you both

PR #1041, opened 2019-12-04
Export EDNS0 interface

### LINKED ISSUE #857: EDNS0 interface needs to be exported
The EDNS0 interface is
~~~
type EDNS0 interface {
	// Option returns the option code for the option.
	Option() uint16
	// pack returns the bytes of the option data.
	pack() ([]byte, error)
	// unpack sets the data as found in the buffer. Is also sets
	// the length of the slice as the length of the option data.
	unpack([]byte) error
	// String returns the string representation of the option.
	String() string
}
~~~
This means I can't externally define a new option; i.e. a bogus 100 one to do some EDNS0 compliance testing.

omeranson:
Replace all the private methods in the EDNS0 with public methods.
Additionally, as suggested in issue #857, made Pack receive a
pre-allocated byte array, introduce a Len method, and have Pack
and Unpack return the number of octets written and read (respectively)
if there was no error.

Closes #857

### DIFF
--- a/edns.go
+++ b/edns.go
@@ -47,7 +47,8 @@ func (rr *OPT) String() string {
 		switch o.(type) {
 		case *EDNS0_NSID:
 			s += "\n; NSID: " + o.String()
-			h, e := o.pack()
+			h := make([]byte, o.Len())
+			_, e := o.Pack(h)
 			var r string
 			if e == nil {
 				for _, c := range h {
@@ -82,8 +83,7 @@ func (rr *OPT) len(off int, compression map[string]struct{}) int {
 	l := rr.Hdr.len(off, compression)
 	for _, o := range rr.Option {
 		l += 4 // Account for 2-byte option code and 2-byte option length.
-		lo, _ := o.pack()
-		l += len(lo)
+		l += o.Len()
 	}
 	return l
 }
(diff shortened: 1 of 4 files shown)


edns.go
```diff
@@ -225,77 +230,90 @@ type EDNS0_SUBNET struct {
 // Option implements the EDNS0 interface.
 func (e *EDNS0_SUBNET) Option() uint16 { return EDNS0SUBNET }
 
-func (e *EDNS0_SUBNET) pack() ([]byte, error) {
-	b := make([]byte, 4)
+func (e *EDNS0_SUBNET) Len() int {
+	l := 4
+	switch e.Family {
```
miekg: think this can be more easily done with an if
`if e.Family == 1 { return 4 }` and then handle the `Family == 2 ` case. Note sure what the RFC says about other families.

omeranson: I reviewed the RFC. It says

```
   The format of the address part depends on the value of FAMILY.  This
   document only defines the format for FAMILY 1 (IPv4) and FAMILY 2
   (IPv6), which are as follows:
```

Sounds to me like FAMILY only controls the format of ADDRESS, but SOURCE PREFIX-LENGTH and SCOPE PREFIX-LENGTH retain their meaning. In any case, only IPv4 and IPv6 are defined, to which this generalisation holds.

Therefore, I simplified the code to
````
return 4 + int((e.SourceNetmask + 8 - 1) / 8)
```

Issue #1037, opened 2019-11-15
Revisit ID randomness decision based on new performance data


jsha:
In #48 and #102 some benchmarking data was provided showing that `math/rand` is faster than `crypto/rand` (for use in generating query IDs). However, those tests took too narrow a view. It's important to look at ID generation in the actual context in which it's used.

[I wrote a more robust benchmark](https://github.com/miekg/dns/files/3849306/main.go.txt) that stands up a trivial DNS server, and sends a lot of queries at it. You can fiddle with values of `parallel` to try and max out local CPU. On my laptop, `-parallel 40` came close to maxing out my CPUs, and I was able to get an average qps of 147,210 for crypto/rand vs 147,350 for math/rand. This is a pretty tiny difference, and that's in optimal conditions. Across a real network, the qps would be much lower, which means the impact of using crypto/rand would be much lower.

I'd like to propose making `crypto/rand` the default, so users don't have to worry about whether they need more secure ID generation.

```
$ go run main.go 
2019/11/14 16:43:11    math/rand N = 100000, parallel = 40, qps = 149075.287112
2019/11/14 16:43:12  crypto/rand N = 100000, parallel = 40, qps = 149218.081585
2019/11/14 16:43:13    math/rand N = 100000, parallel = 40, qps = 146252.205194
2019/11/14 16:43:13  crypto/rand N = 100000, parallel = 40, qps = 147203.384613
2019/11/14 16:43:14    math/rand N = 100000, parallel = 40, qps = 147398.219853
2019/11/14 16:43:15  crypto/rand N = 100000, parallel = 40, qps = 147448.185616
2019/11/14 16:43:15    math/rand N = 100000, parallel = 40, qps = 147396.953876
2019/11/14 16:43:16  crypto/rand N = 100000, parallel = 40, qps = 145910.799292
2019/11/14 16:43:17    math/rand N = 100000, parallel = 40, qps = 146785.044813
2019/11/14 16:43:18  crypto/rand N = 100000, parallel = 40, qps = 147818.343743
2019/11/14 16:43:18    math/rand N = 100000, parallel = 40, qps = 146877.602488
2019/11/14 16:43:19  crypto/rand N = 100000, parallel = 40, qps = 147186.064584
2019/11/14 16:43:20    math/rand N = 100000, parallel = 40, qps = 148628.014722
2019/11/14 16:43:20  crypto/rand N = 100000, parallel = 40, qps = 146466.368147
[...]

PR #1044, opened 2019-12-05
Use crypto/rand for random id generation.

### LINKED ISSUE #1043: [security] Predictable TXID can lead to response forgeries
The default `Id` function uses math/rand, meaning the outputs are predictable, and an attacker might be able to use this to forge responses without being on path.

Seeding math/rand from crypto/rand is pointless, as the math/rand algorithm is invertible: given a sequence of outputs it's possible to reconstruct the `Rand` state and predict all future outputs. Exploitation might be a little slower because the outputs are just 16 bits, but it's likely to be possible.

Unless 0x20 or DNSSEC are used, response verification relies only on source port and TXID. They are both short, but the combination usually makes it hard for an off-path attacker to win the race against the real answer. Without the TXID, the attacker has a very good chance of success at a Kaminsky Attack.

A couple example scenarios:

* A DNS cache — the attacker makes a sequence of requests to get enough TXIDs to reverse the internal state of the RNG, then it causes a request to be issued by the cache to an upstream server, and knowing the TXID sends a spoofed answer to every port, poisoning the cache.
* A prober with an API, like say Let's Encrypt — the attacker causes a number of requests to its own domain to observe TXIDs, then causes a lookup for a target domain, and spoofs the response.

Since the performance cost seems negligible (#1037), I recommend doing the secure thing by default and just reading the 2 bytes from crypto/rand. If there are performance problems, just using a `bufio.Reader` should solve them, as most of the cost of crypto/rand is syscall overhead.

Filing publicly as asked by @miekg.

jsha:
Notably, if crypto/rand returns an error this will silently fall back to math/rand, since
the exported signature can't return an error.

Fixes #1043 and #1037

### DIFF
--- a/msg.go
+++ b/msg.go
@@ -15,7 +15,6 @@ import (
 	"encoding/binary"
 	"fmt"
 	"math/big"
-	mrand "math/rand"
 	"strconv"
 	"strings"
 )
@@ -72,16 +71,12 @@ var (
 	ErrTime          error = &Error{err: "bad time"}      // ErrTime indicates a timing error in TSIG authentication.
 )
 
-// Id by default, returns a 16 bits random number to be used as a
-// message id. This being a variable the function can be reassigned
-// to a custom function. For instance, to make it return a static value:
+// Id by default, returns a 16 bit random number (from crypto/rand) to
+// be used as a message id. This being a variable the function can be
+// reassigned to a custom function. For instance, to make it return a
+// static value:
 //
 //	dns.Id = func() uint16 { return 3 }
-//
-// Randomness comes from crypto/rand. In the rare case that there is an error
-// getting bytes from that source, this function will silently fall back to
-// the predictable math/rand. If your program has more stringent requirements,
-// you should override Id to do something different (e.g., terminate).
 var Id = id
 
 // id returns a 16 bits random number to be used as a
@@ -90,7 +85,7 @@ func id() uint16 {
 	var v uint16
 	err := binary.Read(crand.Reader, binary.BigEndian, &v)
 	if err != nil {
-		v = uint16(mrand.Uint32())
+		panic(fmt.Sprintf("failed to get randomness: %s", err))
 	}
 	return v
 }
--- a/msg.go
+++ b/msg.go
@@ -11,7 +11,7 @@ package dns
 //go:generate go run msg_generate.go
 
 import (
-	crand "crypto/rand"
+	"crypto/rand"
 	"encoding/binary"
 	"fmt"
 	"math/big"
@@ -71,10 +71,10 @@ var (
 	ErrTime          error = &Error{err: "bad time"}      // ErrTime indicates a timing error in TSIG authentication.
 )
 
-// Id by default, returns a 16 bit random number (from crypto/rand) to
-// be used as a message id. This being a variable the function can be
-// reassigned to a custom function. For instance, to make it return a
-// static value:
+// Id by default returns a 16-bit random number to be used as a message id. The
+// number is drawn from a cryptographically secure random number generator.
+// This being a variable the function can be reassigned to a custom function.
+// For instance, to make it return a static value for testing:
 //
 //	dns.Id = func() uint16 { return 3 }
 var Id = id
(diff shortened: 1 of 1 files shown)


FiloSottile:
crypto/rand basically never returns an error, unless someone set up a super-strict sandbox, in which case they probably have opinions that can be encoded as a custom Id function. I'd just go ahead and make it a panic, that's what I did in the version in the Google monorepo.

msg.go
```diff
@@ -73,53 +71,23 @@ var (
 	ErrTime          error = &Error{err: "bad time"}      // ErrTime indicates a timing error in TSIG authentication.
 )
 
-// Id by default, returns a 16 bits random number to be used as a
-// message id. The random provided should be good enough. This being a
-// variable the function can be reassigned to a custom function.
-// For instance, to make it return a static value:
+// Id by default, returns a 16 bit random number (from crypto/rand) to
+// be used as a message id. This being a variable the function can be
```
tmthrgd: I think I'd rather this be worded something like: "Id by default returns a 16-bit random number to be used as a message id. The number is drawn from a cryptographically secure random number generator."

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Use crypto/rand for..." ]
>tmthrgd requested changes on this pull request.
>
>I like this approach and it's a nice simplification. Just a few minor nits.
>
>> -		// seeding idRand upon the first call to id.
>-
>-		var seed int64
>-		var buf [8]byte
>-
>-		if _, err := crand.Read(buf[:]); err == nil {
>-			seed = int64(binary.LittleEndian.Uint64(buf[:]))
>-		} else {
>-			seed = rand.Int63()
>-		}
>-
>-		idRand = rand.New(rand.NewSource(seed))
>+	var v uint16
>+	err := binary.Read(crand.Reader, binary.BigEndian, &v)
>+	if err != nil {
>+		panic(fmt.Sprintf("failed to get randomness: %s", err))
>
>This panic message is off. It should be something like: `panic("dns: reading random id failed: " + err.Error())`.
>
>> -	if idRand == nil {
>-		// This (partially) works around
>-		// https://github.com/golang/go/issues/11833 by only
>-		// seeding idRand upon the first call to id.
>-
>-		var seed int64
>-		var buf [8]byte
>-
>-		if _, err := crand.Read(buf[:]); err == nil {
>-			seed = int64(binary.LittleEndian.Uint64(buf[:]))
>-		} else {
>-			seed = rand.Int63()
>-		}
>-
>-		idRand = rand.New(rand.NewSource(seed))
>+	var v uint16
>
>I think `var id uint16` would be better, but that's very minor.
>
>> @@ -15,10 +15,8 @@ import (
> 	"encoding/binary"
> 	"fmt"
> 	"math/big"
>-	"math/rand"
>
>With this import gone, we can import crypto/rand without renaming it. So `crand "crypto/rand"` -> `"crypto/rand"`.
>
>> @@ -73,53 +71,23 @@ var (
> 	ErrTime          error = &Error{err: "bad time"}      // ErrTime indicates a timing error in TSIG authentication.
> )
>
>-// Id by default, returns a 16 bits random number to be used as a
>-// message id. The random provided should be good enough. This being a
>-// variable the function can be reassigned to a custom function.
>-// For instance, to make it return a static value:
>+// Id by default, returns a 16 bit random number (from crypto/rand) to
>+// be used as a message id. This being a variable the function can be
>
>I think I'd rather this be worded something like: "Id by default returns a 16-bit random number to be used as a message id. The number is drawn from a cryptographically secure random number generator."

Can you post a follow up PR with these changes? (I can do it as well).

Thanks!

yongtang:
Looks like all review comments have been covered, @miekg @tmthrgd can you check to see if this PR is ready to go?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Use crypto/rand for..." ]
>Looks like all review comments have been covered, @miekg @tmthrgd can you check to see if this PR is ready to go?

Ah, let me fix the nits from Tom. I though I merged, but I only pressed 'lgtm',
my mistake

miekg:
merging and doing a small follow up PR; this allows me to tag a release where this at least is fixed.

PR #1048, opened 2019-12-11
Add support for freebsd/arm64

dmgk:


tmthrgd:
@dmgk Can you explain why this change is needed?

dmgk:
@tmthrgd Go 1.14 will support freebsd/arm64, golang.org/x/net@83d349e8ac1a and golang.org/x/sys@33540a1f6037 are earliest commits that bring in necessary arm64 bits. github.com/miekg/dns is an indirect dependency of lego port [1] [2]

[1] https://www.freshports.org/security/lego/
[2] https://github.com/go-acme/lego/pull/1019

tmthrgd:
If lego is already pulling in a newer version of x/net, then the version in the final build will be that later version and no change in this library is needed. Is there a reason this library needs to update its dependency?

dmgk:
Now there's none. Thanks, closing.

PR #1049, opened 2019-12-12
Update depencencies

miekg:
Do a go get -u and go mod tidy

This updates all deps to a newer version.


miekg:
more generic update than #1048 (which may fix that issue - whatever it is/was)

tmthrgd:
If my understanding is correct, I’m pretty certain this type of change is wrong when using modules. With MVS any importers requiring newer versions should update their own requirements. Updating here just forces consumers to pull newer dependencies unnecessarily. I’ll try and find a reference.

tmthrgd:
I think this is the comment I was thinking of https://github.com/golang/go/issues/35798#issuecomment-558358674.

miekg:
Well, but, we should still regularly update ours anyway...

PR #1050, opened 2019-12-16
code generation compatible with go modules

fcelda:
The programs to generate type definitions use `go/importer` which is not compatible with go modules and require `miekg/dns` to be installed globally. This PR replaces `go/importer` with the new `golang.org/x/tools/go/packages` which is module aware.

### DIFF
--- a/types_generate.go
+++ b/types_generate.go
@@ -11,12 +11,13 @@ import (
 	"bytes"
 	"fmt"
 	"go/format"
-	"go/importer"
 	"go/types"
 	"log"
 	"os"
 	"strings"
 	"text/template"
+
+	"golang.org/x/tools/go/packages"
 )
 
 var skipLen = map[string]struct{}{
@@ -81,9 +82,19 @@ func getTypeStruct(t types.Type, scope *types.Scope) (*types.Struct, bool) {
 	return nil, false
 }
 
+// loadModule retrieves package description for a given module.
+func loadModule(name string) (*types.Package, error) {
+	conf := packages.Config{Mode: packages.NeedTypes | packages.NeedTypesInfo}
+	pkgs, err := packages.Load(&conf, name)
+	if err != nil {
+		return nil, err
+	}
+	return pkgs[0].Types, nil
+}
+
 func main() {
 	// Import and type-check the package
-	pkg, err := importer.Default().Import("github.com/miekg/dns")
+	pkg, err := loadModule("github.com/miekg/dns")
 	fatalIfErr(err)
 	scope := pkg.Scope()
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ env:
   - GO111MODULE=on
 
 script:
+  - go generate ./... && test `git ls-files --modified | wc -l` = 0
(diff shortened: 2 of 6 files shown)


tmthrgd:
If this could resolve the issue of the generators not picking up local changes, that would be fantastic. @fcelda Does it pick up changes in the local directory? Also, are the resulting generated files the same? If so, this is a LGTM from me.

fcelda:
@tmthrgd Yes, it picks up the local changes. It's one of the reasons why I looked at it. :slightly_smiling_face:  And yes, the generated files are identical. The only thing I haven't tried is if it runs in environment without go modules.

types_generate.go
```diff
@@ -81,9 +82,19 @@ func getTypeStruct(t types.Type, scope *types.Scope) (*types.Struct, bool) {
 	return nil, false
 }
 
+// loadModule retrieves package description for a given module.
+func loadModule(name string) (*types.Package, error) {
+	conf := packages.Config{Mode: packages.NeedTypes | packages.NeedTypesInfo}
```
UmanShahzad: Tested this locally and without `packages.NeedDeps`, `net.IP` doesn't get recognized and it changes the output of the copy funcs to not use `copyIP`.

fcelda: I will have a look at this specifically again. I think I added `NeedTypesInfo` just for that. :thinking:

fcelda: It works for me, @UmanShahzad.

```
$ docker run --rm -ti golang:latest bash
root@52fb4647a303:/go# cd /tmp
root@52fb4647a303:/tmp# git clone -b mod-compatible-generate https://github.com/fcelda/dns.git
Cloning into 'dns'...
remote: Enumerating objects: 58, done.
remote: Counting objects: 100% (58/58), done.
remote: Compressing objects: 100% (46/46), done.
remote: Total 18490 (delta 25), reused 28 (delta 11), pack-reused 18432
Receiving objects: 100% (18490/18490), 16.17 MiB | 5.08 MiB/s, done.
Resolving deltas: 100% (9828/9828), done.
root@52fb4647a303:/tmp# cd dns
root@52fb4647a303:/tmp/dns# go mod download
go: finding golang.org/x/crypto v0.0.0-20191011191535-87dc89f01550
go: finding golang.org/x/mod v0.1.1-0.20191105210325-c90efee705ee
go: finding golang.org/x/net v0.0.0-20190923162816-aa69164e4478
go: finding golang.org/x/sync v0.0.0-20190423024810-112230192c58
go: finding golang.org/x/sys v0.0.0-20190924154521-2837fb4f24fe
go: finding golang.org/x/text v0.3.0
go: finding golang.org/x/tools v0.0.0-20191216052735-49a3e744a425
go: finding golang.org/x/xerrors v0.0.0-20191011141410-1b5146add898
root@52fb4647a303:/tmp/dns# go generate ./...
root@52fb4647a303:/tmp/dns# git diff
root@52fb4647a303:/tmp/dns# git status
On branch mod-compatible-generate
Your branch is up to date with 'origin/mod-compatible-generate'.

nothing to commit, working tree clean
```

fcelda: I've added additional check for this into Travis.

miekg:
go generate with the PR doesn't yield a diff, so +1. Merging.

PR #1051, opened 2019-12-17
Add ; before printing TSIG

miekg:
TSIG doesn't have a presentation format, when converting to a string
prefix the record with a semi-colon so it will be parsed as a comment.

### DIFF
--- a/tsig.go
+++ b/tsig.go
@@ -40,7 +40,7 @@ type TSIG struct {
 // TSIG has no official presentation format, but this will suffice.
 
 func (rr *TSIG) String() string {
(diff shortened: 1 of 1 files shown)


tsig.go
```diff
@@ -40,7 +40,7 @@ type TSIG struct {
 // TSIG has no official presentation format, but this will suffice.
 
 func (rr *TSIG) String() string {
-	s := "\n;; TSIG PSEUDOSECTION:\n"
+	s := "\n;; TSIG PSEUDOSECTION:\n;" // add another semi-colon to signify TSIG does not have a presentation format
```
tmthrgd: Can we get a space after the semi-colon so it's slightly more readable?

miekg:
>Can we get a space after the semi-colon so it's slightly more readable?

good point. done

PR #1052, opened 2019-12-17
WIP: implement packetaccept func

miekg:
Run a function before accepting the packet and spawning a goroutine. This is a very grude impl. and
it doesn't reject antyhing, just delays.

When this starts rejecting we must send back a SERVFAIL or REFUSED, but we haven't parsed the packet
yet... We could set the return buffer to the length of the header (12 bytes) and then set the error
code directly in that buffer - this does disregard the edns0 footer.

### DIFF
--- a/acceptfunc.go
+++ b/acceptfunc.go
@@ -1,5 +1,7 @@
 package dns
 
+import "time"
+
 // MsgAcceptFunc is used early in the server code to accept or reject a message with RcodeFormatError.
 // It returns a MsgAcceptAction to indicate what should happen with the message.
 type MsgAcceptFunc func(dh Header) MsgAcceptAction
@@ -51,4 +53,30 @@ func defaultMsgAcceptFunc(dh Header) MsgAcceptAction {
 		return MsgReject
 	}
 	return MsgAccept
+}
+
+type Accepter interface {
+	Accept(numgo int, last time.Time)
+}
+
+type PacketAccept struct {
+	Max int // Max is the maximum number of goroutines to allow.
+}
+
+func (p PacketAccept) Accept(numgo int, last time.Time) {
+	if numgo < p.Max/2 {
+		return
+	}
+	rate := 100.0 / time.Since(last).Seconds()
+	left := float64(p.Max-numgo) / rate
+	switch {
+	case left < 0.0:
+		// 50 is a random number.
+		time.Sleep(50 * time.Microsecond)
+		return
+	default:
+		sleep := time.Since(last).Nanoseconds() / 100
+		time.Sleep(time.Duration(sleep))
+		return
+	}
 }
--- a/acceptfunc.go
+++ b/acceptfunc.go
@@ -1,6 +1,9 @@
 package dns
 
-import "time"
+import (
+	"runtime"
+	"time"
+)
 
 // MsgAcceptFunc is used early in the server code to accept or reject a message with RcodeFormatError.
 // It returns a MsgAcceptAction to indicate what should happen with the message.
@@ -55,28 +58,42 @@ func defaultMsgAcceptFunc(dh Header) MsgAcceptAction {
 	return MsgAccept
 }
 
-type Accepter interface {
-	Accept(numgo int, last time.Time)
+type PacketAction int
+
+const (
+	Accept PacketAction = iota
+	Reject
+)
+
+type AcceptOption struct {
+	Last time.Time
 }
+
+type PacketAccepter interface {
+	Accept(*AcceptOption) PacketAction
+}
+
+// our default implementation
 
 type PacketAccept struct {
 	Max int // Max is the maximum number of goroutines to allow.
 }
 
-func (p PacketAccept) Accept(numgo int, last time.Time) {
+func (p PacketAccept) Accept(opt *AcceptOption) PacketAction {
+	numgo := runtime.NumGoroutine()
 	if numgo < p.Max/2 {
-		return
+		return Accept
 	}
-	rate := 100.0 / time.Since(last).Seconds()
+	rate := 100.0 / time.Since(opt.Last).Seconds()
 	left := float64(p.Max-numgo) / rate
 	switch {
 	case left < 0.0:
 		// 50 is a random number.
 		time.Sleep(50 * time.Microsecond)
-		return
+		return Accept
 	default:
-		sleep := time.Since(last).Nanoseconds() / 100
+		sleep := time.Since(opt.Last).Nanoseconds() / 100
 		time.Sleep(time.Duration(sleep))
-		return
+		return Accept
 	}
 }
--- a/server.go
+++ b/server.go
@@ -9,7 +9,6 @@ import (
 	"errors"
 	"io"
 	"net"
-	"runtime"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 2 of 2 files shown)


miekg:
Have a look @tmthrgd if something like this is desirable

acceptfunc.go
```diff
@@ -52,3 +57,43 @@ func defaultMsgAcceptFunc(dh Header) MsgAcceptAction {
 	}
 	return MsgAccept
 }
+
+type PacketAction int
```
tmthrgd: I think we should just reuse `MsgAcceptAction` here. The options (Accept, Reject, Ignore) seem like the exact same options we'd want here.

server.go
```diff
@@ -197,11 +197,14 @@ type Server struct {
...
 	// It is only supported on go1.11+ and when using ListenAndServe.
 	ReusePort bool
-	// AcceptMsgFunc will check the incoming message and will reject it early in the process.
-	// By default DefaultMsgAcceptFunc will be used.
+	// AcceptMsgFunc will check the incoming message and will reject it early in the process, but after
+	// spawning a new go-routine. By default DefaultMsgAcceptFunc will be used.
 	MsgAcceptFunc MsgAcceptFunc
+	// PacketFunc will either allow the incoming packet, deny it, or have some in-between action, like sleeping a
+	// bit. DefaultPacketFunc can be used or serve as an example. This function is called for every 100th packet received.
+	PacketAccepter
```
tmthrgd: Although the name is the same, we don't really want to embed this, otherwise `Server` unexpectedly gains an `Accept` method.

acceptfunc.go
```diff
@@ -52,3 +57,43 @@ func defaultMsgAcceptFunc(dh Header) MsgAcceptAction {
...
+}
+
+// our default implementation
+
+type PacketAccept struct {
+	Max int // Max is the maximum number of goroutines to allow.
+}
+
+func (p PacketAccept) Accept(opt *AcceptOption) PacketAction {
+	numgo := runtime.NumGoroutine()
```
tmthrgd: If we want people to respond to the number of goroutines, wouldn't it make more sense to internally track the number of goroutines *we*-create (which sync.WaitGroup is doing internally). Otherwise this could get confused by goroutines spawned by other code in the binary (including the runtime itself).

server.go
```diff
@@ -458,8 +461,18 @@ func (srv *Server) serveUDP(l *net.UDPConn) error {
 
 	rtimeout := srv.getReadTimeout()
 	// deadline is not used here
+	last := time.Now()
+	pkts := uint64(0)
+	opt := &AcceptOption{}
 	for srv.isStarted() {
 		m, s, err := reader.ReadUDP(l, rtimeout)
+		pkts++
+		if srv.PacketAccepter != nil && pkts%100 == 0 {
```
tmthrgd: This should all go after both the `err != nil` and `len(m) < headerSize` checks. The error one because we might not even have a packet at this point, and the `len(m)` because we just assume they're junk and ignore them unconditionally.

tmthrgd:
Just taking a quick look at this, I think we can achieve this without adding anything to the existing code by using a `DecorateReader`. For that we could even supply a default implementation like you're proposing. I'll try and throw together a quick prototype.

tmthrgd:
Here's a rough idea of what I had in mind:

```go
type PacketLimitOption struct{ apply func(*packetLimitReader) }

func WithMaxGoroutines(num int) PacketLimitOption {
	return PacketLimitOption{func(r *packetLimitReader) { r.maxGoroutines = num }}
}

type packetLimitReader struct {
	Reader

	maxGoroutines int

	pkts int
	last time.Time
}

func PacketLimitReader(opts ...PacketLimitOption) func(Reader) Reader {
	return func(r Reader) Reader {
		lr := &packetLimitReader{
			Reader: r,

			maxGoroutines: 100,

			last: time.Now(),
		}
		for _, opt := range opts {
			opt.apply(lr)
		}

		return lr
	}
}

func (r *packetLimitReader) ReadUDP(conn *net.UDPConn, timeout time.Duration) ([]byte, *SessionUDP, error) {
	m, s, err := r.Reader.ReadUDP(conn, timeout)
	if err != nil {
		return nil, nil, err
	}

	last := r.last
	r.last = time.Now()

	r.pkts++
	if r.pkts%100 != 0 {
		return m, s, nil
	}

	numgo := runtime.NumGoroutine()
	if numgo < r.maxGoroutines/2 {
		return m, s, nil
	}

	rate := 100.0 / time.Since(last).Seconds()
	left := float64(r.maxGoroutines-numgo) / rate
	switch {
	case left < 0.0:
		// 50 is a random number.
		time.Sleep(50 * time.Microsecond)
	default:
		sleep := time.Since(last).Nanoseconds() / 100
		time.Sleep(time.Duration(sleep))
	}

	if false {
		RejectPacket(conn, s, m, RcodeRefused)
		// return m, s, RejectPacketError(RcodeRefused)
	}
	if false {
		return nil, nil, ErrIgnorePacket
	}

	return m, s, nil
}

var ErrIgnorePacket = ignorePacketError{}

type ignorePacketError struct{}

func (ignorePacketError) Error() string   { return "dns: ignoring packet" }
func (ignorePacketError) Timeout() bool   { return false }
func (ignorePacketError) Temporary() bool { return true }

func RejectPacket(conn *net.UDPConn, s *SessionUDP, m []byte, rcode int) error {
	if rcode < 0 || rcode > 0xFFF {
		return ErrRcode
	}
	if rcode > 0xF {
		return ErrExtendedRcode
	}
7
	dh, _, err := unpackMsgHdr(m, 0)
	if err != nil {
		// Let client hang, they are sending crap; any reply can be used to amplify.
		return nil
	}

	var req Msg
	req.setHdr(dh)

	var msg Msg
	msg.SetRcode(&req, rcode)

	m, err = msg.Pack()
	if err != nil {
		return err
	}

	_, err = WriteToSessionUDP(conn, m, s)
	return err
}
```

Only `RejectPacket` (or some kind of `RejectPacketError` function) would need to be defined by this package, with `ErrIgnorePacket` being possible to implement outside, but good to implement inside.

@miekg Thoughts?

miekg:
>@miekg Thoughts?

I like it, nice idea! Let me ponder this a bit.

miekg:
@tmthrgd what are your thoughts on delaying packet ingestion? My hope is that you always delay and let the OS handle not accepting packets on this port.

tmthrgd:
@miekg Doesn't the OS keep a packet buffer of it's own?

chantra:
I like this. Some actions that can be useful to do when load-shedding:
- return an answer with TC bit set
- plainly drop the packet

Regardless, I believe packet ingestion needs to happen or the queue will
get full and packets will be dropped.

miekg:
ok, so we want always want to ingest, but avoid starting a go routine and doing a lot of work.

I like the suggestion of TC. Couple of follow up questions:

* We can do what we want with the raw binary and set TC and the return code (easy enough)
* EDNS0 will be harder to support (because we need to parse more) - so these shortcuts will be actual shortcuts

I  propose:
* Returning an answer that *only* contains the header and the question section
* Setting SERVAIL (and/or TC) on the reply

miekg:
See #1068, closing here.

PR #1054, opened 2019-12-17
Add missing EDNS0EXPIRE data unpack

austinoh:
Currently EDNS0 Data unpack for EXPIRE option isn't implemented yet.

### DIFF
--- a/msg_helpers.go
+++ b/msg_helpers.go
@@ -447,7 +447,6 @@ Option:
 		off += int(optlen)
 	case EDNS0EXPIRE:
 		e := new(EDNS0_EXPIRE)
-		e.Code = code
 		if err := e.unpack(msg[off : off+int(optlen)]); err != nil {
 			return nil, len(msg), err
 		}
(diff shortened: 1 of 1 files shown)


msg_helpers.go
```diff
@@ -445,6 +445,14 @@ Option:
 		}
 		edns = append(edns, e)
 		off += int(optlen)
+	case EDNS0EXPIRE:
+		e := new(EDNS0_EXPIRE)
+		e.Code = code
```
miekg: why this assignment?

austinoh: Since EDNS0_EXPIRE struct has Code field, just fill it with. However, if it's not consistent with other cases, I can remove e.Code assignment. Please let me know.

austinoh: However, i think it's good idea to fill the code with correct value. Otherwise, I am seeing zero code value in final return message.

tmthrgd: I don't think this is appropriate. Maybe it would make sense, but none of the others set it even though they have the field too. Consistency is what should be aimed for here.

austinoh: Sure, removed code assignment, to be consistent with other cases.

PR #1058, opened 2019-12-18
APL record support

fcelda:
This PR adds support for [APL record type](https://tools.ietf.org/html/rfc3123).

### DIFF
--- a/duplicate_generate.go
+++ b/duplicate_generate.go
@@ -102,7 +102,7 @@ func main() {
 
 				if st.Tag(i) == `dns:"apl"` {
 					o3(`for i := 0; i < len(r1.%s); i++ {
-						if !r1.%s[i].Equals(&r2.%s[i]) {
+						if !r1.%s[i].equals(&r2.%s[i]) {
 							return false
 						}
 					}`)
--- a/duplicate_generate.go
+++ b/duplicate_generate.go
@@ -100,6 +100,16 @@ func main() {
 					continue
 				}
 
+				if st.Tag(i) == `dns:"apl"` {
+					o3(`for i := 0; i < len(r1.%s); i++ {
+						if r1.%s[i].Equals(&r2.%s[i]) {
+							return false
+						}
+					}`)
+
+					continue
+				}
+
 				o3(`for i := 0; i < len(r1.%s); i++ {
 					if r1.%s[i] != r2.%s[i] {
 						return false
--- a/duplicate_generate.go
+++ b/duplicate_generate.go
@@ -102,7 +102,7 @@ func main() {
 
 				if st.Tag(i) == `dns:"apl"` {
 					o3(`for i := 0; i < len(r1.%s); i++ {
-						if r1.%s[i].Equals(&r2.%s[i]) {
+						if !r1.%s[i].Equals(&r2.%s[i]) {
 							return false
 						}
 					}`)
--- a/msg_helpers.go
+++ b/msg_helpers.go
@@ -681,3 +681,135 @@ func packDataDomainNames(names []string, msg []byte, off int, compression compre
 	}
 	return off, nil
 }
+
+var aplAddrLenToFamily = map[int]uint16{
+	net.IPv4len: 1,
+	net.IPv6len: 2,
+}
+
+var aplFamilyToAddrLen = map[uint16]int{
+	1: net.IPv4len,
+	2: net.IPv6len,
+}
+
+func packDataApl(data []APLPrefix, msg []byte, off int) (int, error) {
+	var err error
+	for i, _ := range data {
+		off, err = packDataAplPrefix(&data[i], msg, off)
+		if err != nil {
+			return len(msg), err
+		}
+	}
+	return off, nil
+}
+
+func packDataAplPrefix(p *APLPrefix, msg []byte, off int) (int, error) {
+	if len(p.Network.IP) != len(p.Network.Mask) {
+		return len(msg), &Error{err: "address and mask lengths don't match"}
+	}
+
+	family, ok := aplAddrLenToFamily[len(p.Network.IP)]
+	if !ok {
+		return len(msg), &Error{err: "unrecognized address family"}
+	}
+
+	prefix, _ := p.Network.Mask.Size()
+	addr := p.Network.IP.Mask(p.Network.Mask)[:(prefix+7)/8]
+
+	// Output is 4-byte header + address
+	if off+4+len(addr) > len(msg) {
+		return len(msg), &Error{err: "overflow packing APL prefix"}
+	}
+
+	var err error
+
+	// Write ADDRESSFAMILY
+	off, err = packUint16(family, msg, off)
+	if err != nil {
+		return len(msg), err
+	}
+	// Write PREFIX
+	off, err = packUint8(uint8(prefix), msg, off)
+	if err != nil {
+		return len(msg), err
+	}
+	// Write N and AFDLENGTH
+	var n uint8
+	if p.Negation {
+		n = 0x80
+	}
+	adflen := uint8(len(addr)) & 0x7f
+	off, err = packUint8(n|adflen, msg, off)
+	if err != nil {
+		return len(msg), err
+	}
+	// Write AFDPART
+	copy(msg[off:], addr)
+	off += len(addr)
+
+	return off, nil
+}
+
+func unpackDataApl(msg []byte, off int) ([]APLPrefix, int, error) {
+	var result []APLPrefix
+	for off < len(msg) {
+		prefix, end, err := unpackDataAplPrefix(msg, off)
+		if err != nil {
+			return nil, len(msg), err
+		}
+		off = end
+		result = append(result, *prefix)
+	}
+	return result, off, nil
+}
+
+func unpackDataAplPrefix(msg []byte, off int) (*APLPrefix, int, error) {
+	// Always has 4-byte header.
+	if off+4 > len(msg) {
+		return nil, len(msg), &Error{err: "overflow unpacking APL prefix"}
+	}
+
+	// Read ADDRESSFAMILY
+	var family uint16
+	family, off, _ = unpackUint16(msg, off)
+	ipLen, ok := aplFamilyToAddrLen[family]
+	if !ok {
+		return nil, len(msg), &Error{err: "unrecognized APL address family"}
+	}
+	// Read PREFIX
+	var prefix uint8
+	prefix, off, _ = unpackUint8(msg, off)
+	if int(prefix) > 8*ipLen {
+		return nil, len(msg), &Error{err: "APL prefix too long"}
+	}
+	// Read N and AFDLENGTH
+	var nlen uint8
+	nlen, off, _ = unpackUint8(msg, off)
+	neg := (nlen & 0x80) > 0
+	afdlen := int(nlen & 0x7f)
+	if (int(prefix)+7)/8 != afdlen {
+		return nil, len(msg), &Error{err: "invalid APL address length"}
+	}
+	// Read AFDPART
+	if off+afdlen > len(msg) {
+		return nil, len(msg), &Error{err: "overflow unpacking APL address"}
+	}
+	ip := make([]byte, ipLen)
+	copy(ip, msg[off:off+afdlen])
+	off += afdlen
+	if prefix%8 > 0 {
+		last := ip[afdlen-1]
+		zero := uint8(0xff) >> (prefix % 8)
+		if last&zero > 0 {
+			return nil, len(msg), &Error{err: "extra APL address bits"}
+		}
+	}
+
+	return &APLPrefix{
+		Negation: neg,
+		Network: net.IPNet{
+			IP:   ip,
+			Mask: net.CIDRMask(int(prefix), 8*ipLen),
+		},
+	}, off, nil
+}
--- a/msg_helpers.go
+++ b/msg_helpers.go
@@ -747,16 +747,16 @@ func unpackDataApl(msg []byte, off int) ([]APLPrefix, int, error) {
 			return nil, len(msg), err
 		}
 		off = end
-		result = append(result, *prefix)
+		result = append(result, prefix)
 	}
 	return result, off, nil
 }
 
-func unpackDataAplPrefix(msg []byte, off int) (*APLPrefix, int, error) {
+func unpackDataAplPrefix(msg []byte, off int) (APLPrefix, int, error) {
 	// Read ADDRESSFAMILY
 	family, off, err := unpackUint16(msg, off)
 	if err != nil {
-		return nil, len(msg), &Error{err: "overflow unpacking APL prefix"}
+		return APLPrefix{}, len(msg), &Error{err: "overflow unpacking APL prefix"}
 	}
 	var ipLen int
 	switch family {
@@ -765,29 +765,29 @@ func unpackDataAplPrefix(msg []byte, off int) (*APLPrefix, int, error) {
 	case 2:
 		ipLen = net.IPv6len
 	default:
-		return nil, len(msg), &Error{err: "unrecognized APL address family"}
+		return APLPrefix{}, len(msg), &Error{err: "unrecognized APL address family"}
 	}
 	// Read PREFIX
 	prefix, off, err := unpackUint8(msg, off)
 	if err != nil {
-		return nil, len(msg), &Error{err: "overflow unpacking APL prefix"}
+		return APLPrefix{}, len(msg), &Error{err: "overflow unpacking APL prefix"}
 	}
 	if int(prefix) > 8*ipLen {
-		return nil, len(msg), &Error{err: "APL prefix too long"}
+		return APLPrefix{}, len(msg), &Error{err: "APL prefix too long"}
 	}
 	// Read N and AFDLENGTH
 	nlen, off, err := unpackUint8(msg, off)
 	if err != nil {
-		return nil, len(msg), &Error{err: "overflow unpacking APL prefix"}
+		return APLPrefix{}, len(msg), &Error{err: "overflow unpacking APL prefix"}
 	}
 	neg := (nlen & 0x80) != 0
 	afdlen := int(nlen & 0x7f)
 	if (int(prefix)+7)/8 != afdlen {
-		return nil, len(msg), &Error{err: "invalid APL address length"}
+		return APLPrefix{}, len(msg), &Error{err: "invalid APL address length"}
 	}
 	// Read AFDPART
 	if off+afdlen > len(msg) {
-		return nil, len(msg), &Error{err: "overflow unpacking APL address"}
+		return APLPrefix{}, len(msg), &Error{err: "overflow unpacking APL address"}
 	}
 	ip := make([]byte, ipLen)
 	off += copy(ip, msg[off:off+afdlen])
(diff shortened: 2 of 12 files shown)


types.go
```diff
@@ -1402,11 +1483,24 @@ func euiToString(eui uint64, bits int) (hex string) {
 	return
 }
 
+// copyBytes returns a copy of a byte slice.
+func copyBytes(b []byte) []byte {
```
miekg: this is a too thin of a wrapper of `copy` right now. Please just keep copyIP and copy the netmask below manually

duplicate_generate.go
```diff
@@ -100,6 +100,16 @@ func main() {
 					continue
 				}
 
+				if st.Tag(i) == `dns:"apl"` {
+					o3(`for i := 0; i < len(r1.%s); i++ {
+						if r1.%s[i].Equals(&r2.%s[i]) {
```
tmthrgd: This should be not equals (`!...`).

msg_helpers.go
```diff
@@ -681,3 +681,135 @@ func packDataDomainNames(names []string, msg []byte, off int, compression compre
 	}
 	return off, nil
 }
+
+var aplAddrLenToFamily = map[int]uint16{
```
tmthrgd: Just use a switch statement for these. I'm pretty sure we don't use maps anywhere else.

msg_helpers.go
```diff
@@ -681,3 +681,135 @@ func packDataDomainNames(names []string, msg []byte, off int, compression compre
...
+}
+
+var aplFamilyToAddrLen = map[uint16]int{
+	1: net.IPv4len,
+	2: net.IPv6len,
+}
+
+func packDataApl(data []APLPrefix, msg []byte, off int) (int, error) {
+	var err error
+	for i, _ := range data {
```
tmthrgd: Drop the `_`.

msg_helpers.go
```diff
@@ -681,3 +681,135 @@ func packDataDomainNames(names []string, msg []byte, off int, compression compre
...
+	if p.Negation {
+		n = 0x80
+	}
+	adflen := uint8(len(addr)) & 0x7f
+	off, err = packUint8(n|adflen, msg, off)
+	if err != nil {
+		return len(msg), err
+	}
+	// Write AFDPART
+	copy(msg[off:], addr)
```
tmthrgd: `off += copy(...)`.

msg_helpers.go
```diff
@@ -681,3 +681,135 @@ func packDataDomainNames(names []string, msg []byte, off int, compression compre
...
+
+func unpackDataApl(msg []byte, off int) ([]APLPrefix, int, error) {
+	var result []APLPrefix
+	for off < len(msg) {
+		prefix, end, err := unpackDataAplPrefix(msg, off)
+		if err != nil {
+			return nil, len(msg), err
+		}
+		off = end
+		result = append(result, *prefix)
```
tmthrgd: I don't like that `unpackDataAplPrefix` returns a pointer but `unpackDataApl` returns a slice of values. Pick one and stick to it throughout.

msg_helpers.go
```diff
@@ -681,3 +681,135 @@ func packDataDomainNames(names []string, msg []byte, off int, compression compre
...
+
+func unpackDataAplPrefix(msg []byte, off int) (*APLPrefix, int, error) {
+	// Always has 4-byte header.
+	if off+4 > len(msg) {
+		return nil, len(msg), &Error{err: "overflow unpacking APL prefix"}
+	}
+
+	// Read ADDRESSFAMILY
+	var family uint16
+	family, off, _ = unpackUint16(msg, off)
```
tmthrgd: These can all use the short variable declaration (`:=`) instead of having a separate `var`.

tmthrgd: I don't like that we're ignoring the errors from `unpackUintX` in this function.

msg_helpers.go
```diff
@@ -681,3 +681,135 @@ func packDataDomainNames(names []string, msg []byte, off int, compression compre
...
+	// Read PREFIX
+	var prefix uint8
+	prefix, off, _ = unpackUint8(msg, off)
+	if int(prefix) > 8*ipLen {
+		return nil, len(msg), &Error{err: "APL prefix too long"}
+	}
+	// Read N and AFDLENGTH
+	var nlen uint8
+	nlen, off, _ = unpackUint8(msg, off)
+	neg := (nlen & 0x80) > 0
```
tmthrgd: `!= 0`

msg_helpers.go
```diff
@@ -681,3 +681,135 @@ func packDataDomainNames(names []string, msg []byte, off int, compression compre
...
+	afdlen := int(nlen & 0x7f)
+	if (int(prefix)+7)/8 != afdlen {
+		return nil, len(msg), &Error{err: "invalid APL address length"}
+	}
+	// Read AFDPART
+	if off+afdlen > len(msg) {
+		return nil, len(msg), &Error{err: "overflow unpacking APL address"}
+	}
+	ip := make([]byte, ipLen)
+	copy(ip, msg[off:off+afdlen])
```
tmthrgd: `off += copy(...)`

msg_helpers_test.go
```diff
@@ -153,3 +157,309 @@ func BenchmarkUnpackString(b *testing.B) {
...
+			"normalize 1:198.51.103.255/22",
+			false,
+			net.ParseIP("198.51.103.255").To4(),
+			net.CIDRMask(22, 32),
+			[]byte{0x00, 0x01, 0x16, 0x03, 198, 51, 100}, // 1:198.51.100.0/22
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ap := APLPrefix{
```
tmthrgd: Put the `&` here instead of at `packDataAplPrefix`.

msg_helpers_test.go
```diff
@@ -153,3 +157,309 @@ func BenchmarkUnpackString(b *testing.B) {
...
+		},
+		{
+			"unrecognized family",
+			[]byte{0x42},
+			[]byte{0xff},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ap := APLPrefix{Network: net.IPNet{IP: tt.ip, Mask: tt.mask}}
```
tmthrgd: `&` here as well.

msg_helpers_test.go
```diff
@@ -153,3 +157,309 @@ func BenchmarkUnpackString(b *testing.B) {
...
+			}
+			if off != len(msg) {
+				t.Fatalf("expected %d, got %d", len(msg), off)
+			}
+		})
+	}
+}
+
+func TestPackDataAplPrefix_BufferBounds(t *testing.T) {
+	ap := APLPrefix{
```
tmthrgd: `&` here as well.

types.go
```diff
@@ -1353,6 +1355,85 @@ func (rr *CSYNC) len(off int, compression map[string]struct{}) int {
...
+
+// APLPrefix is an address prefix hold by an APL record.
+type APLPrefix struct {
+	Negation bool
+	Network  net.IPNet
+}
+
+// String returns presentation form of the APL record.
+func (rr *APL) String() string {
+	sb := strings.Builder{}
```
tmthrgd: `var sb strings.Builder`

types.go
```diff
@@ -1353,6 +1355,85 @@ func (rr *CSYNC) len(off int, compression map[string]struct{}) int {
...
+	sb.WriteString(rr.Hdr.String())
+	for i, p := range rr.Prefixes {
+		if i != 0 {
```
tmthrgd: `i > 0`

types.go
```diff
@@ -1353,6 +1355,85 @@ func (rr *CSYNC) len(off int, compression map[string]struct{}) int {
...
+			sb.WriteByte(' ')
+		}
+		sb.WriteString(p.String())
+	}
+	return sb.String()
+}
+
+// String returns presentation form of the APL prefix.
+func (p *APLPrefix) String() string {
+	sb := strings.Builder{}
```
tmthrgd: `var sb strings.Builder`

types.go
```diff
@@ -1353,6 +1355,85 @@ func (rr *CSYNC) len(off int, compression map[string]struct{}) int {
...
+
+// String returns presentation form of the APL record.
+func (rr *APL) String() string {
+	sb := strings.Builder{}
+	sb.WriteString(rr.Hdr.String())
+	for i, p := range rr.Prefixes {
+		if i != 0 {
+			sb.WriteByte(' ')
+		}
+		sb.WriteString(p.String())
```
tmthrgd: Having a `String() string` method on both `APL` and `APLPrefix` both of which allocate isn't exactly great. Do we really need one on `APLPrefix`?

types.go
```diff
@@ -1353,6 +1355,85 @@ func (rr *CSYNC) len(off int, compression map[string]struct{}) int {
...
+
+	sb.WriteByte(':')
+
+	switch len(p.Network.IP) {
+	case net.IPv4len:
+		sb.WriteString(p.Network.IP.String())
+	case net.IPv6len:
+		// add prefix for IPv4-mapped IPv6
+		if v4 := p.Network.IP.To4(); v4 != nil {
+			sb.WriteString("::ffff:")
```
tmthrgd: Does `net.IP.String()` not do this already?

types.go
```diff
@@ -1353,6 +1355,85 @@ func (rr *CSYNC) len(off int, compression map[string]struct{}) int {
...
+		}
+		sb.WriteString(p.Network.IP.String())
+	default:
```
tmthrgd: Drop the `default` label.

types.go
```diff
@@ -1353,6 +1355,85 @@ func (rr *CSYNC) len(off int, compression map[string]struct{}) int {
...
+	prefix, _ := p.Network.Mask.Size()
+	sb.WriteString(strconv.Itoa(prefix))
+
+	return sb.String()
+}
+
+// Equals reports whether two APL prefixes are identical.
+func (a *APLPrefix) Equals(b *APLPrefix) bool {
+	return a.Negation == b.Negation &&
+		bytes.Equal(a.Network.IP, b.Network.IP) &&
```
tmthrgd: Use `net.IP.Equal` here.

types.go
```diff
@@ -1353,6 +1355,85 @@ func (rr *CSYNC) len(off int, compression map[string]struct{}) int {
...
+// Copy returns a copy of the APL prefix.
+func (p *APLPrefix) Copy() APLPrefix {
+	return APLPrefix{
+		Negation: p.Negation,
+		Network:  copyNet(p.Network),
+	}
+}
+
+// Len returns size of the prefix in wire format.
+func (p *APLPrefix) Len() int {
```
tmthrgd: How useful is this method? If it's not really needed externally it should be either unexported on preferably merged into  types_generate.go.

types_generate.go
```diff
@@ -249,6 +251,12 @@ func main() {
 					fields = append(fields, f)
 					continue
 				}
+				if t == "APLPrefix" {
+					fmt.Fprintf(b, "%s := make([]%s, len(rr.%s));\nfor i, _ := range rr.%s {\n %s[i] = rr.%s[i].Copy()\n}\n",
```
tmthrgd: Drop the `_` here.

miekg:
APL RFC https://tools.ietf.org/html/rfc3123 should also be added to the README

duplicate_generate.go
fcelda: Good catch!

types.go
fcelda: Unfortunately no. IPv4-mapped IPv6 is as IPv4 per [IP.String](https://godoc.org/net#IP.String).

I'm not sure what are the reasons but Go often stores IPv4 as IPv6 address and then performs conversions in the background so nobody would notice. The conversion to string is one example of that.

APL record is protocol-agnostic and we need to respect the address type and therefore I added the prefix to force the presentation format to be IPv6. I just checked the specification again whether the IPv4-mapped IPv6 is allowed and it is. The other implementation option I considered was to print the address as a pure IPv6 address but that would require different kind of trickery. I would stick to this solution.

types.go
fcelda: I cannot because IPv4 and equivalent IPv4-mapped IPv6 would be treated as the same (see [IP.Equal](https://godoc.org/net#IP.Equal)) while the APL resource record would be different.

msg_helpers.go
fcelda: This is used in four different places so there will be some duplication but I can do that.

msg_helpers.go
fcelda: This was a practical decision. Most of the parsing happens in `unpackDataAplPrefix`, there are eight return statements (currently after addressing other review comments), and returning nil pointer is more convenient than returning an empty structure. I would actually prefer leaving it this way. What do you think? Should I keep it or change it?

msg_helpers.go
fcelda: The only reason for the unpack functions to fail is buffer underflow and I was checking the header boundaries a few lines above. But I'm going to change this for consistency with the existing code. :+1:

types.go
fcelda: Maybe not. I will make it private method. :+1:

types.go
fcelda: I'm going to make all methods on APLPrefix private. I can imagine some of them may become handy but probably not much. :+1:

fcelda:
@miekg @tmthrgd Thank you for looking at my changes thoroughly. I've addressed all the comments you left in code or made my case in the comments. It should be ready for another round of reviews.

parse_test.go
```diff
@@ -1584,3 +1584,137 @@ func TestNULLRecord(t *testing.T) {
...
+
+			got := rr.String()
+			if got != tc.expect {
+				t.Errorf("expected %q, got %q", tc.expect, got)
+			}
+		})
+	}
+}
+
+func TestParseAPL_Errors(t *testing.T) {
```
miekg: No underscores please.

types.go
```diff
@@ -1353,6 +1355,91 @@ func (rr *CSYNC) len(off int, compression map[string]struct{}) int {
...
+	}
+
+	var afi int
+	switch len(p.Network.IP) {
+	case net.IPv4len:
+		afi = 1
+	case net.IPv6len:
+		afi = 2
+	}
+	sb.WriteString(strconv.Itoa(int(afi)))
```
miekg: this can be written inside the switch, no need for `afi` integer then

types.go
```diff
@@ -1353,6 +1355,91 @@ func (rr *CSYNC) len(off int, compression map[string]struct{}) int {
...
+func (p *APLPrefix) copy() APLPrefix {
+	return APLPrefix{
+		Negation: p.Negation,
+		Network:  copyNet(p.Network),
+	}
+}
+
+// len returns size of the prefix in wire format.
+func (p *APLPrefix) len() int {
+	// 4-byte header and the network address prefix (see https://tools.ietf.org/html/rfc3123#section-4)
```
miekg: reference the RFC by the number , not a link to tools.ietf.org

msg_helpers.go
tmthrgd: `APLPrefix{}` isn’t much longer than `nil` and this inconsistency isn’t great.

types.go
tmthrgd: Ah I see. Don’t love it, but I guess it’s okay. Is there anything wrong with just converting IPv4 as IPv6 address to IPv4 throughout?

fcelda:
OK, should be fixed now.

types.go
fcelda: I think there are the following options to write IPv4 APL in presentation format:

1. `1:192.0.2.0/24`
2. `2:192.0.2.0/24` (invalid)
3. `2:::ffff:192.0.2.0/120`
4. `2:::ffff:c000:0200/120`

Option 1 is plain IPv4. Option 2 is invalid according to the specification, we could be lax and allow it in this library but that would make it incompatible with the other libraries. Options 3 and 4 are equivalent on the wire.

Even though all the valid options are semantically the same (i.e. refer to the same network subnet), the APL has explicit notation of address family so there is an actual difference between 1 and 3 (or 4) in the wire format.

I think we cannot force conversion from IPv4-mapped IPv6 to IPv4 because then part of the information is lost. It would also lead to confusing behavior when converting from write to presentation and back to wire again.

Therefore we should stick to option 3 or 4. Go doesn't have API to print address in that format so we can either write a custom conversion function or use the trick above. For option 4, I also have this if it's prefered:
```go
ip := net.ParseIP("::ffff:192.0.2.1")
fmt.PrintlN(ip) // 192.0.2.1
ip[1] = 1 // hack
fmt.Println(ip.String()[1:]) // ::ffff:c000:201
```

msg_helpers.go
```diff
@@ -681,3 +681,129 @@ func packDataDomainNames(names []string, msg []byte, off int, compression compre
...
+	case net.IPv6len:
+		family = 2
+	default:
+		return len(msg), &Error{err: "unrecognized address family"}
+	}
+	prefix, _ := p.Network.Mask.Size()
+	addr := p.Network.IP.Mask(p.Network.Mask)[:(prefix+7)/8]
+
+	// Write ADDRESSFAMILY
+	off, err := packUint16(family, msg, off)
```
miekg: sorry to keep on nittig: but this can also put in the switch, no need for `family` var, just try to do the write there. The comment also doesn't *really* add useful context here.

msg_helpers.go
```diff
@@ -681,3 +681,129 @@ func packDataDomainNames(names []string, msg []byte, off int, compression compre
...
+	if err != nil {
+		return len(msg), err
+	}
+	// Write PREFIX
```
miekg: comment doesn't add anything

msg_helpers.go
```diff
@@ -681,3 +681,129 @@ func packDataDomainNames(names []string, msg []byte, off int, compression compre
...
+	off, err = packUint8(uint8(prefix), msg, off)
+	if err != nil {
+		return len(msg), err
+	}
+	// Write N and AFDLENGTH
+	var n uint8
+	if p.Negation {
+		n = 0x80
```
miekg: these constants come from the RFC?

msg_helpers.go
```diff
@@ -681,3 +681,129 @@ func packDataDomainNames(names []string, msg []byte, off int, compression compre
...
+	switch family {
+	case 1:
+		ipLen = net.IPv4len
+	case 2:
+		ipLen = net.IPv6len
+	default:
+		return APLPrefix{}, len(msg), &Error{err: "unrecognized APL address family"}
+	}
+	// Read PREFIX
+	prefix, off, err := unpackUint8(msg, off)
```
miekg: read prefix before the `switch family` and then do the `int(prefix)> *ipLen` check there, this way to you drop the `ipLen` variable.

I see `ipLen` is also used as the very last thing in this function, probably with building up a APLPrefix as you go, to not have to variables, but just initialize an APLPrefix along the way.

msg_helpers.go
fcelda: Yes, these are the names from the [RFC](https://tools.ietf.org/html/rfc3123#section-4). I can remove that. It's not super helpful.

```
      +---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+
      |                          ADDRESSFAMILY                        |
      +---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+
      |             PREFIX            | N |         AFDLENGTH         |
      +---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+
      /                            AFDPART                            /
      |                                                               |
      +---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+
```

msg_helpers.go
fcelda: I'm not sure I understood exactly what you proposed. I moved all the reads of the fixed-size header to the beginning of the function so that the address length handling logic is together. `ipLen` is gone and so is `neg`. Please, take a look.

fcelda:
OK, folks, another round. :upside_down_face:

miekg:
It's not super important and we can always fix it later. So

/lgtm

miekg:
I'm ready to merge this cc @tmthrgd

fcelda:
Thank you for the the review and patience, @miekg and @tmthrgd.

PR #1065, opened 2019-12-29
Support the zero length EDNS0 EXPIRE option

abh:


edns.go
```diff
@@ -543,6 +543,10 @@ func (e *EDNS0_EXPIRE) pack() ([]byte, error) {
 }
 
 func (e *EDNS0_EXPIRE) unpack(b []byte) error {
+	if len(b) == 0 {
+		// zero-length EXPIRE query, https://tools.ietf.org/html/rfc7314#section-2
```
miekg:
```suggestion
		// zero-length EXPIRE query, see RFC 7314 Section 2
```

Issue #10, opened 2012-01-09
Many errors related to "error"


bortzmeyer:
Even after patching for issue #9, godns does not compile with today's tip:

% make
8g  -o _go_.8 clientconfig.go client.go defaults.go dns.go dnssec.go edns.go keygen.go kscan.go msg.go nsec3.go rawmsg.go server.go tsig.go types.go update.go xfr.go zscan.go zscan_rr.go 
clientconfig.go:30: undefined: error
client.go:24: undefined: error
client.go:25: undefined: error
client.go:26: undefined: error
client.go:27: undefined: error
client.go:156: undefined: error
client.go:175: undefined: error
client.go:204: undefined: error
client.go:228: undefined: error
dns.go:168: undefined: error
client.go:228: too many errors
make: **\* [_go_.8] Error 1

Something apparently changed recently in error types.

PR #1067, opened 2019-12-30
Implement SVCB

### LINKED ISSUE #1064: add SVCB record
See https://tools.ietf.org/html/draft-ietf-dnsop-svcb-httpssvc-01. We should add this record as a private RR type for the time being. Wireformat is fairly straight forward, except for the value field, which resembles TXT but not entirely and it can be switched to something else by a previous value in the RR.

I was thinking off:

~~~
// SVCB RR. See RFC xxxx (https://tools.ietf.org/html/draft-ietf-dnsop-svcb-httpssvc-01)
type SVCB struct {
	Hdr      RR_Header
	Priority uint16
	Target   string   `dns:"domain-name"`
	Value    []string `dns:"txt"` // this is wrong, if priority == 0 this must be something else entirely
} 
~~~
The draft talks about SvcFieldPriority instead of `Priority`, which I really dislike, hence `Priority` in the above example. `Target` is a better name than "SvcDomainName" which is used in the draft - this also mimics the SRV record a bit.

DesWurstes:
Resolves #1064

Seems to work well. Can be tested with @ghedo's [python code](https://github.com/miekg/dns/pull/1067#pullrequestreview-395582299) and his Go code (Now my code uses the same numbers. I've updated his Go code, below, to work with the new edns-like code:)

```go
package main

import (
	"fmt"
	"github.com/DesWurstes/dns"
	"net"
	"sync"
)

func handleRequest(w dns.ResponseWriter, r *dns.Msg) {
	fmt.Println("GOT REQUEST", r)

	m := new(dns.Msg)
	m.SetReply(r)

	svcb := &dns.SVCB{
		Hdr: dns.RR_Header{
			Name:   r.Question[0].Name,
			Rrtype: dns.TypeSVCB,
			Class:  dns.ClassINET,
			Ttl:    3600,
		},
		Priority: 1,
		Target:   ".",
		Value: []dns.SVCBKeyValue{
			&dns.SVCBAlpn{
				Alpn: []string{"h2", "h3"},
			},
			&dns.SVCBPort{
				Port: 117,
			},
			&dns.SVCBIPv4Hint{
				Hint: []net.IP{net.IPv4(1, 1, 1, 1), net.IPv4(1, 1, 1, 2)},
			},
			&dns.SVCBECHConfig{
				ECH: []byte{1, 2},
			},
		},
	}

	m.Answer = append(m.Answer, svcb)
	err := w.WriteMsg(m)
	if err != nil {
		fmt.Println(err)
	}
}

func main() {
	var wg sync.WaitGroup
	wg.Add(1)

	dns.HandleFunc("svcb.example.", handleRequest)

	server := &dns.Server{Addr: ":53", Net: "udp"}
	go server.ListenAndServe()

	wg.Wait()
}
```


scan_rr.go
```diff
@@ -1696,3 +1696,40 @@ func (rr *TKEY) parse(c *zlexer, o string) *ParseError {
...
+	}
+	rr.Target = name
+
+	// Values (if any)
+	l, _ = c.Next()
+	var xs []string
+	for l.value != zNewline && l.value != zEOF {
+		switch l.value {
+		case zString:
+			xs = append(xs, l.token)
```
miekg: I think these values are always quoted? Not sure if the quotation is part of the value or not

types.go
```diff
@@ -1353,6 +1354,22 @@ func (rr *CSYNC) len(off int, compression map[string]struct{}) int {
 	return l
 }
 
+// SVCB RR. See RFC xxxx (https://tools.ietf.org/html/draft-ietf-dnsop-svcb-httpssvc-01)
+// Named ESNI and numbered 0xff9f = 65439 according to draft-ietf-tls-esni-05
+type SVCB struct {
+	Hdr      RR_Header
+	Priority uint16
+	Target   string   `dns:"domain-name"`
+	Value    []string `dns:"txt"` // if priority == 0 this is empty
```
miekg: this will encode to the wrong wire format

scan_rr.go
DesWurstes: They're of the form `a=b` or `a="b"`. Because the values of key=value pairs consist of

> basic-visible [which] is VCHAR minus DQUOTE, ";", and "\"

So I should add code to remove the double quotes if they exist.

types.go
DesWurstes: What I've just read:

> Values are in a format specific to the SvcParamKey.  Their definition
   should specify both their presentation format and wire encoding
   (e.g., domain names, binary data, or numeric values).

...

>    The "Service Binding (SVCB) Parameter Registry" defines the name
   space for parameters, including string representations and numeric  SvcParamKey values.  This registry is shared with other SVCB-
   compatible RR types, such as HTTPSSVC.

>   ACTION: create and include a reference to this registry.

Which means the wire format of values of key=value pairs depend on keys, and key- value-wireformat-type table was not written which means no one can implement wire format encoding/decoding before the spec is ready.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Implement SVCB (#10..." ]
>They're of the form `a=b` or `a="b"`. Because the values of key=value pairs consist of
>
>> basic-visible [which] is VCHAR minus DQUOTE, ";", and "\"
>
>So I should add code to remove the double quotes if they exist.

dunno, if the draft is unclear on that it should be more clear. Or send an email to dnsops
to ask for clarification.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Implement SVCB (#10..." ]
>Which means the wire format of values of key=value pairs depend on keys, and key- value-wireformat-type table was not written which means no one can implement wire format encoding/decoding before the spec is ready.

Missed this previously: https://tools.ietf.org/html/draft-ietf-dnsop-svcb-httpssvc-01#appendix-C.3

C.3.  Wire Format

   Advice from experts in DNS wire format best practices would be
   greatly appreciated to refine the proposed details, overall.

currently the format is:

  When SvcFieldValue is non-empty (ServiceForm), it contains a list of
   SvcParamKey=SvcParamValue pairs with length-prefixes for the
   SvcParamValues, each of which contains:

   o  a 2 octet field containing the SvcParamKey as an integer in
      network byte order.

   o  a 2 octet field containing the length of the SvcParamValue as an
      integer between 0 and 65535 in network byte order (but constrained
      by the RDATA and DNS message sizes).

   o  an octet string of the length defined by the previous field.

this is def *not* how the TXT record does it and a bit bonkers if you ask me.

DesWurstes:
Thank you, that helps, I'll add that.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Implement SVCB (#10..." ]
>Thank you, that helps, I'll add that.

also see https://github.com/miekg/dns/issues/1064#issuecomment-569664299

DesWurstes:
Hello again, I'm adding `key=value` wire encoding/decoding. In #1064 you suggested those `key=value` pairs should be `[]string` in the struct. I believe `[]byte` (or even a separated [][]byte) in the wireformat would suit better, because it'd allow the library to be used as a proxy with reserved and private key types (otherwise we can't deserialize the value, which is of an unknown format). In addition, packing and unpacking would be more expensive, which I think is used more often than string serialization/deserialization. Do you agree, should the struct keep Values as binary?

miekg:
I think we should help shape the wire format for this record using the
experience we gather from trying to implement it. IOW engage on the dnsop
ML to propose a new format

ghedo:
Hello @DesWurstes, are you still working on this by any chance? I'm interested in helping out (if any help is needed) to get this in shape (if you are not interested in working on this anymore I'm also available to take over the work and get it ready, if that's ok with you).

We (Cloudflare) are looking into shipping support for HTTSSVC, so this would help greatly.

> Implements for draft-ietf-tls-esni-05, thus if we don't declare it private people will try to use with Cloudflare's servers, which are draft-ietf-tls-esni-02 that is incompatible with the SVCB record.

I don't understand this part. Why does it need to be "private"? If people try to query SVCB or HTTPSSVC records on Cloudflare's DNS servers they will get back nothing, as we don't currently serve these records. In that case a client wouldn't use an ESNI version that Cloudflare servers don't support, and it could either fallback to an older version if it supports it, or not do ESNI at all.

DesWurstes:
Hello,

My implementation was incomplete because I wasn't sure if the specification would undergo major changes. It seems it has matured. Now I'll rebase it and start working on it again.

I'll change
```
Value    []string `dns:"txt"`
```
to
```
Value    []uint16 `dns:"svcb"`
```
so that it doesn't decode when not needed. (and define the type)

I wanted to make it private as the draft is not finalized so no server supports it, and those implementing SNI encryption could try to use the SVCB record instead of `_sni`.

DesWurstes:
@miekg should I make `Value` field closer to the wireformat than the presentation format? (Because other ones seem to be keeping data in base64 strings instead of byte arrays but SvcValueField can be containing a key whose value type we don't know because it's reserved, so we don't need to process SvcValueField unless it's required?)

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Implement SVCB (#10..." ]
>@miekg should I make Value field closer to the wireformat than the presentation
>format? (Because other ones seem to be keeping data in base64 strings instead
>of byte arrays but SvcValueField can be containing a key whose value type we
>don't know because it's reserved, so we don't need to process SvcValueField
>unless it's required?)

I haven't been following the latest with regards to SVCB, but the structure in go dns
should contain a field that is usable. I.e. the `A` records we use `net.IP`, not `[]byte`

/Miek

--
Miek Gieben

DesWurstes:
```
type SvcKeyValue struct {
	SvcParamKey   uint16
	SvcParamValue string
}
```

where `SvcParamValue` is the presentation format representation is fine? Then for defined keys we can add
```
func (rr *SVCB) getIpv4Hint() (ip net.IP, has bool) {
	...
}
```
etc, because the presentation format usually is helpful, but if they need it in data types this library uses, they can call functions similar to this. (Or functions taking `[]SvcKeyValue` as arguments, so that it's valid for both SVCB and HTTPSSVC)

Or to solve the problem similar to `edns.go`? For someone needing to use SVCB the EDNS approach could look too complex?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Implement SVCB (#10..." ]
>type SvcKeyValue struct {
>        SvcParamKey   uint16
>        SvcParamValue string
>}
>
>where SvcParamValue is the presentation format representation is fine? Then for
>defined keys we can add
>
>func (rr *SVCB) getIpv4Hint() (ip net.IP, has bool) {
>        ...
>}

Why is such a helper needed? The SvcParamKeys prolly will need a whole slew of constants
defined as well...

DesWurstes:
https://github.com/miekg/dns/pull/1067/files#diff-e871404fd69cab40521d6b2178847cf7R184-R219

The array of SvcKeyValues would be ordered, they can for-loop until `SvcParamKey == SVC_PORT` for example. Do you prefer:

```
type SVCB struct {
	Hdr      RR_Header
	Priority uint16
	Target   string        `dns:"domain-name"`
	esni       ESNI `dns:"-"`
        Ipv4hint Ipv4 `dns:"-"`
        Alpn        alpn `dns:"-"`
        Nodefprotocol bool `dns:"-"`
        // other valuews
	Value    []SvcKeyValue `dns:"svc"` // if priority == 0 this is empty
}
```

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Implement SVCB (#10..." ]
>https://github.com/miekg/dns/pull/1067/files#
>diff-e871404fd69cab40521d6b2178847cf7R184-R219
>
>The array of SvcKeyValues would be ordered, they can for-loop until SvcParamKey
>== SVC_PORT for example. Do you prefer:
>
>type SVCB struct {
>        Hdr      RR_Header
>        Priority uint16
>        Target   string        `dns:"domain-name"`
>        esni       ESNI `dns:"-"`
>        Ipv4hint Ipv4 `dns:"-"`
>        Alpn        alpn `dns:"-"`
>        Nodefprotocol bool `dns:"-"`
>        // other valuews
>        Value    []SvcKeyValue `dns:"svc"` // if priority == 0 this is empty
>}

I need to look at the current record, all those `dns:"-"` tagged values are super weird,
if they are not useful to a user they should not be in the struct

DesWurstes:
That was an example of solution, the specification didn't change. In other words, should we keep values in SvcFieldValue struct in the presentation format as string, with conversion functions to go types, or when decoding it should we fill, with values, go-typed spaces in our struct, which are empty when a relevant key=value pair is not found?

msg_helpers.go
```diff
@@ -659,6 +659,47 @@ func packDataNsec(bitmap []uint16, msg []byte, off int) (int, error) {
...
+		xs = append(xs, SvcKeyValue{SvcParamKey: code,
+			SvcParamValue: string(val)})
+	}
+
+	return xs, off, nil
+}
+
+func packDataSvc(options []SvcKeyValue, msg []byte, off int) (int, error) {
+	for _, el := range options {
+		off, err := packUint16(el.SvcParamKey, msg, off)
```
ghedo, line 684: I know that this is probably going to change anyway, but note that the `off` variable here shadows the `off` argument variable, so this function always ends up returning the same value that was passed in (effectively the svc fields are not encoded in the output message).

ghedo:
Oh, regarding the dnspython code, note that I accidentally ended up using different code-points there than what is used in this PR, so one or the other needs to be manually changed to make the test work.

DesWurstes:
@miekg I'm ditto-ing `edns.go`, but I've found fields that have no use:

https://github.com/miekg/dns/blob/67373879ce327b5fd112d9301d0a4d62bad6b904/edns.go#L217-L226

Code is only written to and never read (instead of `.Option()`). Should I not add this to SVCB (except local-experimental one)? (Are those fields only for backwards compatibility, or does that aim to remind the programmer that key is a part of the data stored too?)

svcb.go
```diff
@@ -0,0 +1,661 @@
...
+	return rr.SVCB.parse(c, o)
+}
+
+func (r1 *HTTPSSVC) isDuplicate(r2 RR) bool {
+	return (*r1).SVCB.isDuplicate(r2)
+}
+
+// SvcKeyValue defines a key=value pair for SvcFieldValue.
+// A SVCB RR can have multiple SvcKeyValues appended to it.
+type SvcKeyValue interface {
```
miekg, line 232: oh wow. I'm reading section 6 of the draft, that's pretty terrible range of things you can put in this record as key=value thingies.
Note that the EDNS0 stuff has been shown to be suboptimal and isn't the nicest interface to work with, so I rather not have it here again. Having different types implementing SvcKeyValue works for me, so this will be still an interface. However things like `Key` and `Read` look like overkill. I.e. the interface must be paired down to the absolute minimal

svcb.go
```diff
@@ -0,0 +1,661 @@
...
+//
+//	o := new(dns.HTTPSSVC)
+//	o.Hdr.Name = "."
+//	o.Hdr.Rrtype = dns.TypeHTTPSSVC
+//	e := new(dns.SVC_ALPN)
+//	e.Code = dns.SVCALPN
+//	e.Alpn = []string{"h2", "http/1.1"}
+//	o.Value = append(o.Value, e)
+type SVC_ALPN struct {
+	Code uint16   // Always SVCALPN
```
miekg, line 264: So Code isn't really need here? Maybe that can be the interface `Code() uint` ? Also the naming of the types should be move Go like and not uppercase with underscores (another mistake in the EDNS0 code). I'm not sure what would make sense here? A subpackage we're these things live? Or prefixing them with `Svc` ? Note that the final name is not chosen yet? (is this true?) So we might be looking into renaming this in the future

DesWurstes: Rather than prefixing every member of each struct I'd prefer to move everything to a SVCB folder/subpackage. What do you think? The specification may be extended by much in the future.

I believe `Key()` should stay in the main interface, so that its exposed parts are
```
type SvcKeyValue interface {
	// Key returns the key code of the pair.
	Key() uint16
	// String returns the string representation of the value.
	String() string
}
```

ghedo:
Just for the record, I was able to implement support for serving HTTPSSVC records in our internal DNS server after pulling these changes (including the latest API tweaks) :tada: but I'll need to wait for this PR to be merged before starting deployment to production.

The new public API exposed to applications is quite nice, so I don't really have anything more to say :+1:

DesWurstes:
Hi ghedo, thanks for the feedback you've given, and the dnspython SVCB code you've made, it was essential for me to debug my implementation.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Implement SVCB (#10..." ]
>@DesWurstes commented on this pull request.
>
>
>In svcb.go:
>
>> +
>+// SVC_ALPN pair is used to list supported connection protocols.
>+// Protocol ids can be found at:
>+// https://www.iana.org/assignments/tls-extensiontype-values/
>+// tls-extensiontype-values.xhtml#alpn-protocol-ids
>+// Basic use pattern for creating an alpn option:
>+//
>+//     o := new(dns.HTTPSSVC)
>+//     o.Hdr.Name = "."
>+//     o.Hdr.Rrtype = dns.TypeHTTPSSVC
>+//     e := new(dns.SVC_ALPN)
>+//     e.Code = dns.SVCALPN
>+//     e.Alpn = []string{"h2", "http/1.1"}
>+//     o.Value = append(o.Value, e)
>+type SVC_ALPN struct {
>+       Code uint16   // Always SVCALPN
>
>Rather than prefixing every member of each struct I'd prefer to move everything
>to a SVCB folder/subpackage. What do you think? The specification may be
>extended by much in the future.
>
>I believe Key() should stay in the main interface, so that its exposed parts
>are
>
>type SvcKeyValue interface {
>        // Key returns the key code of the pair.
>        Key() uint16
>        // String returns the string representation of the value.
>        String() string
>}

yep, this looks like a sane interface, modulo naming but we can fix that later.

Yes, I'm also inclined to move this to a subpackage, `dnssvcb` or something less
hidious...

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Implement SVCB (#10..." ]
>Just for the record, I was able to implement support for serving HTTPSSVC
>records in our internal DNS server after pulling these changes (including the
>latest API tweaks) 🎉 but I'll need to wait for this PR to be merged before
>starting deployment to production.
>
>The new public API exposed to applications is quite nice, so I don't really
>have anything more to say 👍

nice cross OSS/DNS work here :)

DesWurstes:
I was dreaming of keeping `SVCB` in `types.go` and moving everything related to the value field to a subpackage. For encode/decode and everything this would require everything in the KeyValuePair interface (`Key`, `String`,`pack`, `unpack`, `read`, `len`) to be declared public so that we can access it from the top package. We can't have a link between the package and the subpackage that is not exposed to the user. So I'd rather not move it to a subpackage. Do you have any solution ideas? Or we can move and then expose only 4 functions which we can mark "Internal, don't use" in the documentation?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Implement SVCB (#10..." ]
>I was dreaming of keeping SVCB in types.go and moving everything related to the
>value field to a subpackage. For encode/decode and everything this would

yes, that would be my preference too, all the KV stuff can move in a subpackage, but the
main RR stays here - just like all the other RRs.

>require everything in the KeyValuePair interface (Key, String,pack, unpack,
>read, len) to be declared public so that we can access it from the top package.
>We can't have a link between the package and the subpackage that is not exposed
>to the user. So I'd rather not move it to a subpackage. Do you have any
>solution ideas? Or we can move and then expose only 4 functions which we can
>mark "Internal, don't use" in the documentation?

What functions would that be? I was hoping to just expose the KV types from the
subpackage?

/Miek

--
Miek Gieben

DesWurstes:
It works if I expose these from the subpackage:

```go
func SvcInternalLen(s *SvcKeyValue) uint16 {
	return (*s).len()
}

func SvcInternalCopy(s *SvcKeyValue) SvcKeyValue {
	return (*s).copy()
}

func SvcInternalPack(s *SvcKeyValue) ([]byte, error) {
	return (*s).pack()
}

func SvcInternalUnpack(s *SvcKeyValue, b []byte) error {
	return (*s).unpack(b)
}

func SvcInternalRead(s *SvcKeyValue, str string) error {
	return (*s).read(str)
}
```

I didn't want them to be part of the interface as users of this library would look for functions they need in the keyvaluepair interface.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Implement SVCB (#10..." ]
>It works if I expose these from the subpackage:
>
>func SvcInternalLen(s *SvcKeyValue) uint16 {
>        return (*s).len()
>}
>
>func SvcInternalCopy(s *SvcKeyValue) SvcKeyValue {
>        return (*s).copy()
>}
>
>func SvcInternalPack(s *SvcKeyValue) ([]byte, error) {
>        return (*s).pack()
>}
>
>func SvcInternalUnpack(s *SvcKeyValue, b []byte) error {
>        return (*s).unpack(b)
>}
>
>func SvcInternalRead(s *SvcKeyValue, str string) error {
>        return (*s).read(str)
>}
>
>I didn't want them to be part of the interface as users of this library would
>look for functions they need in the keyvaluepair interface.

that's not pretty... alternatively we could give up on a subpackage and prefix the names
to fake the namespacing.

Having a subpackage for 1 RR doesn't really sit well we've me, all RRs are equal,
singling this one out because $REASONS, seems odd.

DesWurstes:
Currently they are prefixed. From the SVCB branch's godoc:

    Constants
    Variables
    func ActivateAndServe(l net.Listener, p net.PacketConn, handler Handler) error
    func CanonicalName(s string) string
    ...
    func StringToTime(s string) (uint32, error)
    func SvcKeyToString(svcKey uint16) string  ⬅
    func SvcStringToKey(str string) uint16  ⬅
    func TLSAName(name, service, network string) (string, error)
    ...
    type HIP
        func (rr *HIP) Header() *RR_Header
        func (rr *HIP) String() string
    type HTTPSSVC  ⬅
        func (rr *HTTPSSVC) Header() *RR_Header  ⬅
        func (rr *HTTPSSVC) String() string  ⬅
    type Handler
    ...
    type SSHFP
        func (rr *SSHFP) Header() *RR_Header
        func (rr *SSHFP) String() string
    type SVCB  ⬅
        func (rr *SVCB) Header() *RR_Header  ⬅
        func (rr *SVCB) String() string  ⬅
    type ServeMux
        func NewServeMux() *ServeMux
        func (mux *ServeMux) Handle(pattern string, handler Handler)
        func (mux *ServeMux) HandleFunc(pattern string, handler func(ResponseWriter, *Msg))
        func (mux *ServeMux) HandleRemove(pattern string)
        func (mux *ServeMux) ServeDNS(w ResponseWriter, req *Msg)
    ...
    type SessionUDP
        func ReadFromSessionUDP(conn *net.UDPConn, b []byte) (int, *SessionUDP, error)
        func (s *SessionUDP) RemoteAddr() net.Addr
    type SvcAlpn  ⬅
        func (s *SvcAlpn) Key() uint16  ⬅
        func (s *SvcAlpn) String() string  ⬅
    type SvcESNIConfig  ⬅
        func (s *SvcESNIConfig) Key() uint16  ⬅
        func (s *SvcESNIConfig) String() string  ⬅
    type SvcIPv4Hint  ⬅
        func (s *SvcIPv4Hint) Key() uint16  ⬅
        func (s *SvcIPv4Hint) String() string  ⬅
    type SvcIPv6Hint  ⬅
        func (s *SvcIPv6Hint) Key() uint16  ⬅
        func (s *SvcIPv6Hint) String() string  ⬅
    type SvcKeyValue  ⬅
    type SvcLocal  ⬅
        func (s *SvcLocal) Key() uint16  ⬅
        func (s *SvcLocal) String() string  ⬅
    type SvcNoDefaultAlpn  ⬅
        func (s *SvcNoDefaultAlpn) Key() uint16  ⬅
        func (s *SvcNoDefaultAlpn) String() string  ⬅
    type SvcPort  ⬅
        func (s *SvcPort) Key() uint16  ⬅
        func (s *SvcPort) String() string  ⬅
    type TA
        func (rr *TA) Header() *RR_Header
        func (rr *TA) String() string

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Implement SVCB (#10..." ]
>Currently they are prefixed. From the SVCB branch's godoc:

let's go with that then. I prolly want to uppercase the SVCB name (and use SVCB as the
previx for all of these) - following in the footsteps of TLSAName, but that's a minor nit.

Is there any progress on the final name of this record? Is it going to be SVCB ?

DesWurstes:
> Is there any progress on the final name of this record? Is it going to be SVCB ?

I don't know but renaming everything a few times before this pull request is merged is fine/easy for me.

Though I prefer prefixing SVC instead of SVCB because the SVC things are valid for both SVCB and HTTPSSVC, what do you think?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Implement SVCB (#10..." ]
>    Is there any progress on the final name of this record? Is it going to be
>    SVCB ?
>
>I don't know but renaming everything a few times before this pull request is
>merged is fine/easy for me.

agreed.

>Though I prefer prefixing SVC instead of SVCB because the SVC things are valid
>for both SVCB and HTTPSSVC, what do you think?

Hmmm. I think sticking with one will be better. SVC doesn't exist, also just looking at
the 2 names you would think that SVCB and SVCHTTPS would be a better fit.

DesWurstes:
> Hmmm. I think sticking with one will be better. SVC doesn't exist, also just looking at
the 2 names you would think that SVCB and SVCHTTPS would be a better fit.

I agree after noticing that I declared HTTPSSVC as a derived type of SVCB. To make HTTPSSVC declared as `&dns.HTTPSSVC{SVCB: dns.SVCB{...}}` so it'd make sense to make all of the common code based on SVCB.

svcb.go
```diff
@@ -0,0 +1,737 @@
...
+	// Key returns the numerical key code.
+	Key() uint16
+	// pack returns the encoded value.
+	pack() ([]byte, error)
+	// unpack sets the value.
+	unpack([]byte) error
+	// String returns the string representation of the value.
+	String() string
+	// read sets the value to the given string representation of the value.
+	read(string) error
```
miekg, line 248: this method is better named `parse`

svcb.go
```diff
@@ -0,0 +1,737 @@
...
+// Basic use pattern for creating an alpn option:
+//
+//	o := new(dns.HTTPSSVC)
+//	o.Hdr.Name = "."
+//	o.Hdr.Rrtype = dns.TypeHTTPSSVC
+//	e := new(dns.SVCBAlpn)
+//	e.Alpn = []string{"h2", "http/1.1"}
+//	o.Value = append(o.Value, e)
+type SVCBAlpn struct {
+	Alpn []string // Must not be of zero length
```
miekg, line 267: this comment doesn't make a lot of sense

msg_helpers.go
```diff
@@ -612,6 +613,76 @@ func packDataNsec(bitmap []uint16, msg []byte, off int) (int, error) {
...
+			return len(msg), &Error{err: "repeated svcb keys are not allowed"}
+		}
+		prev = e.Key()
+	}
+	for _, el := range pairs {
+		packed, err := el.pack()
+		if err != nil {
+			return len(msg), err
+		}
+		if len(packed) > 65535 {
```
miekg, line 669: why this check?

svcb.go
```diff
@@ -0,0 +1,737 @@
...
+
+// The spec requires the alpn keys including \ or , to be escaped.
+// In practice, no standard key including those exists.
+// Therefore those characters are not escaped.
+
+func (s *SVCBAlpn) pack() ([]byte, error) {
+	// Estimate
+	b := make([]byte, 0, 10*len(s.Alpn))
+	for _, e := range s.Alpn {
+		//x := []byte(strings.ReplaceAll(strings.ReplaceAll(e, "\\", "\\\\"), ",", "\\,"))
```
miekg, line 281: lingering comment

svcb.go
```diff
@@ -0,0 +1,737 @@
...
+//	o.Hdr.Rrtype = dns.TypeHTTPSSVC
+//	e := new(dns.SVCBAlpn)
+//	e.Alpn = []string{"h2", "http/1.1"}
+//	o.Value = append(o.Value, e)
+type SVCBAlpn struct {
+	Alpn []string // Must not be of zero length
+}
+
+func (s *SVCBAlpn) Key() uint16    { return SVCB_ALPN }
+func (s *SVCBAlpn) String() string { return strings.Join(s.Alpn[:], ",") }
```
miekg, line 271: this reslicing `[:]` is needed?

svcb.go
```diff
@@ -0,0 +1,737 @@
...
+			return "<nil>"
+		}
+		str.WriteByte(',')
+		str.WriteString(x.String())
+	}
+	return str.String()[1:]
+}
+
+func (s *SVCBIPv4Hint) read(b string) error {
+	if strings.ContainsRune(b, ':') {
```
miekg, line 471: why can't just ParseIP be used here?

types_generate.go
```diff
@@ -72,6 +72,9 @@ func getTypeStruct(t types.Type, scope *types.Scope) (*types.Struct, bool) {
 	if !ok {
 		return nil, false
 	}
+	if st.NumFields() == 0 {
```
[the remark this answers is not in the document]
DesWurstes: `type SVCBNoDefaultAlpn struct` is the first empty struct in this project, which resulted in null dereference in generators. So it now checks if a field exists before checking the type of the zeroth element.

msg.go
```diff
@@ -1005,6 +1005,10 @@ func domainNameLen(s string, off int, compression map[string]struct{}, compress
 	return len(s) + 1
 }
 
+func svcbFieldLen(s SVCBKeyValue) int {
```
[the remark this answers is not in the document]
DesWurstes: Removing in the next commit.

svcb.go
```diff
@@ -0,0 +1,737 @@
...
+var svcbKeyToString = map[uint16]string{
+	SVCB_ALPN:            "alpn",
+	SVCB_NO_DEFAULT_ALPN: "no-default-alpn",
+	SVCB_PORT:            "port",
+	SVCB_IPV4HINT:        "ipv4hint",
+	SVCB_ECHCONFIG:       "echconfig",
+	SVCB_IPV6HINT:        "ipv6hint",
+}
+
+var svcbStringToKey = map[string]uint16{
```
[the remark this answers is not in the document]
DesWurstes: Fixing.

svcb.go
```diff
@@ -0,0 +1,737 @@
...
+		return uint16(a)
+	}
+	return svcbStringToKey[str]
+}
+
+func (rr *SVCB) parse(c *zlexer, o string) *ParseError {
+	l, _ := c.Next()
+	i, e := strconv.ParseUint(l.token, 10, 16)
+	if e != nil || l.err {
+		return &ParseError{"", "bad svb Priority", l}
```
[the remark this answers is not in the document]
DesWurstes: fixing

msg_helpers.go
DesWurstes: Two-byte limit to encode value length? Right, removing.

svcb.go
DesWurstes: Linked #1107 `ParseIP` accepts `::ffff:192.168.1.1` as IPv4 but since #1108 was rejected we can apply the same behavior and reject IPv4-as-v6?

msg_generate.go
```diff
@@ -110,6 +113,8 @@ return off, err
 					o("off, err = packDataOpt(rr.%s, msg, off)\n")
 				case `dns:"nsec"`:
 					o("off, err = packDataNsec(rr.%s, msg, off)\n")
+				case `dns:"svcb-pairs"`:
```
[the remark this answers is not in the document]
DesWurstes: I'll use `SVCBPairs` or "pairs" in the source and `SVCB value pairs` in error messages, do you have a naming suggestion?

svcb.go
```diff
@@ -0,0 +1,737 @@
...
+	rr.Value = xs
+	if rr.Priority == 0 && len(xs) > 0 {
+		return &ParseError{"", "svcb aliasform can't have values", l}
+	}
+	return nil
+}
+
+// makeSVCBKeyValue returns an SVCBKeyValue struct with the key
+// or nil for reserved keys.
+func makeSVCBKeyValue(key uint16) SVCBKeyValue {
```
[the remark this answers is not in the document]
DesWurstes: We have another option: to allow private types between `SVCB_PRIVATE_LOWER` and `SVCB_PRIVATE_UPPER` and consider the rest invalid? So that this library can be used an  as a dns proxy without needing to adjust for private types. Otherwise, do I completely remove SVCBLocal or move them to somewhere else?

ghedo:
> Is there any progress on the final name of this record? Is it going to be SVCB ?

Just FTR, a new version of the draft has been published, which renames `HTTPSSVC` to `HTTPS`, while keeping `SVCB` as-is: https://mailarchive.ietf.org/arch/msg/dnsop/c1My5e6KbSf0aldTEEdVq3b2DyA/

Hopefully these will be the final names :)

DesWurstes:
@ghedo Thanks for the notice.

@miekg I need two design decisions:

Do I rename

    SVCBKeyValue `dns:"svcb-pairs"`

And how do I refer to it in printed errors?

Secondly, I was planning to keep `SVC_Local` for between `SVCB_PRIVATE_LOWER` and `SVCB_PRIVATE_UPPER` so that this library can be used as a transparent proxy in private networks. If you see no use of it should I rather remove `SVC_Local` and `SVCB_PRIVATE_LOWER` `SVCB_PRIVATE_UPPER`?

miekg:
Note I'm slow to merge because it's a massive amount of code (for 1 RR).
I'll review again and then likely merge with the caveat that a revert or
backwards incompatible changes are possible.

DesWurstes:
@miekg Before you review I need idea-help about reserved keys. Currently it accepts all unknown keys. Should it accept known keys plus keys only in the private use range (whose serialization is described in the specification) or known keys only?

tmthrgd:
I’ve held off reviewing or looking at this because it seemed to be in flux and, as @miekg said, it’s a lot of code. If this is bearing some final revision, I’m happy to take a look in the coming days.

ghedo:
FYI, SVCB and HTTPS now have official code-points as well:
https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#dns-parameters-4

```
SVCB 	64 	Service Binding 	[draft-ietf-dnsop-svcb-https-00] 	SVCB/svcb-completed-template 	2020-06-30
HTTPS 	65 	HTTPS Binding 	[draft-ietf-dnsop-svcb-https-00] 	HTTPS/https-completed-template 	2020-06-30
```

Also FWIW I updated the dnspython PR with the new code-points and all the various renames (https://github.com/rthalley/dnspython/pull/452). I haven't really tested against this yet, but I expect it to still work.

And finally, I updated our internal code to use the latest version of this PR, and everything seems to be still working fine.

DesWurstes:
Excellent, thanks for notifying me

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Implement SVCB (#10..." ]
>I’ve held off reviewing or looking at this because it seemed to be in flux and,
>as @miekg said, it’s a lot of code. If this is bearing some final revision, I’m
>happy to take a look in the coming days.

I want to minimize the LOC for this PR. So I'll try to find some time this weekend to
suggest things that can be removed. Most likely things that look like some kind of future
proofing that we don't need yet - if I can find them.

svcb.go
```diff
@@ -0,0 +1,802 @@
...
+// Basic use pattern for creating a mandatory option:
+//
+//	o := new(dns.SVCB)
+//	o.Hdr.Name = "."
+//	o.Hdr.Rrtype = dns.TypeSVCB
+//	e := new(dns.SVCBMandatory)
+//	e.Code = []uint16{65403}
+//	o.Value = append(o.Value, e)
+//  // Then add key-value pair for key65403
+type SVCBMandatory struct {
```
miekg, line 260: don't really like this name, maybe wait for the RFC to be numbered and `SVBRFCXXX` or something (also don't really like that).
Is SVCBMandatory something from the I-D ?

svcb.go
miekg: yes, that is the sane thing to do

svcb.go
DesWurstes: SVCBMandatory is the newest key-value pair in the specification.

DesWurstes:
`makeSVCBKeyValue` is called from `svcb.go` and `msg_helpers.go` so I can't remove it.

svcb.go
```diff
@@ -0,0 +1,795 @@
...
+	str.Grow(16 * len(s.Hint))
+	for _, e := range s.Hint {
+		x := e.To4()
+		if x == nil {
+			return "<nil>"
+		}
+		str.WriteByte(',')
+		str.WriteString(x.String())
+	}
+	return str.String()[1:]
```
tmthrgd, line 525: Why not just use `strings.Join` and an approach similar to `(*SVCBMandatory).String` here?

svcb.go
```diff
@@ -0,0 +1,795 @@
...
+		return errors.New("dns: not IPv4")
+	}
+	str := strings.Split(b, ",")
+	dst := make([]net.IP, 0, len(str))
+	for _, e := range str {
+		ip := net.ParseIP(e)
+		if ip == nil {
+			return errors.New("dns: bad IP")
+		}
+		dst = append(dst, ip.To4())
```
tmthrgd, line 539: `dst[i] = ip.To4()`

svcb.go
```diff
@@ -0,0 +1,795 @@
...
+	}
+	s.Hint = dst
+	return nil
+}
+
+func (s *SVCBIPv4Hint) copy() SVCBKeyValue {
+	return &SVCBIPv4Hint{
+		append(make([]net.IP, 0, len(s.Hint)), s.Hint...),
```
tmthrgd, line 547: Same as `SVCBMandatory`.

svcb.go
```diff
@@ -0,0 +1,795 @@
...
+	}
+}
+
+// SVCBECHConfig pair contains the ECHConfig structure
+// defined in draft-ietf-tls-esni [RFC TODO] to encrypt TODO
+// the SNI during the client handshake.
```
tmthrgd, line 553: Don't line wrap so aggressively.

svcb.go
```diff
@@ -0,0 +1,795 @@
...
+// Basic use pattern for creating an echconfig option:
+//
+//	o := new(dns.HTTPS)
+//	o.Hdr.Name = "."
+//	o.Hdr.Rrtype = dns.HTTPS
+//	e := new(dns.SVCBECHConfig)
+//	e.ECH = "/wH...="
+//	o.Value = append(o.Value, e)
+type SVCBECHConfig struct {
+	ECH string // This string needs to be base64 encoded
```
tmthrgd, line 563: `// Base64 encoded ...`

svcb.go
```diff
@@ -0,0 +1,795 @@
...
+}
+
+func (s *SVCBECHConfig) Key() uint16           { return SVCB_ECHCONFIG }
```
tmthrgd, line 566: `func (*SVCBECHConfig) Key() uint16`

svcb.go
```diff
@@ -0,0 +1,795 @@
...
+func (s *SVCBECHConfig) copy() SVCBKeyValue    { return &SVCBECHConfig{s.ECH} }
+func (s *SVCBECHConfig) pack() ([]byte, error) { return []byte(s.ECH), nil }
+func (s *SVCBECHConfig) unpack(b []byte) error { s.ECH = string(b); return nil }
```
tmthrgd, line 569: Don't use `xxx; yyy' form.

svcb.go
```diff
@@ -0,0 +1,795 @@
...
+func (s *SVCBECHConfig) String() string        { return s.ECH }
+func (s *SVCBECHConfig) parse(b string) error  { s.ECH = b; return nil }
+func (s *SVCBECHConfig) len() uint16           { return uint16(len(s.ECH)) }
+
+// SVCBIPv6Hint pair suggests an IPv6 address
+// which may be used to open connections if A and AAAA record
+// responses for SVCB's Target domain haven't been received.
+// In that case, optionally, A and AAAA requests can be made,
+// after which the connection to the hinted IP address may be
+// terminated and a new connection may be opened.
```
tmthrgd, line 579: Don't line wrap so aggressively.

svcb.go
```diff
@@ -0,0 +1,795 @@
...
+// Basic use pattern for creating an ipv6hint option:
+//
+//	o := new(dns.HTTPS)
+//	o.Hdr.Name = "."
+//	o.Hdr.Rrtype = dns.HTTPS
+//	e := new(dns.SVCBIPv6Hint)
+//	e.Hint = []net.IP{net.ParseIP("2001:db8::1")}
+//	o.Value = append(o.Value, e)
+type SVCBIPv6Hint struct {
+	Hint []net.IP // Always IPv6
```
tmthrgd, line 589: Elaborate. "Hint are IPv6 ..."

svcb.go
```diff
@@ -0,0 +1,795 @@
...
+}
+
+func (s *SVCBIPv6Hint) Key() uint16 { return SVCB_IPV6HINT }
```
tmthrgd, line 592: `func (*SVCBIPv6Hint) Key() uint16`

svcb.go
```diff
@@ -0,0 +1,795 @@
...
+func (s *SVCBIPv6Hint) len() uint16 { return uint16(16 * len(s.Hint)) }
+
+func (s *SVCBIPv6Hint) pack() ([]byte, error) {
+	b := make([]byte, 0, 16*len(s.Hint))
+	for _, e := range s.Hint {
+		if len(e) != net.IPv6len || e.To4() != nil {
+			return nil, errors.New("dns: not IPv6")
```
tmthrgd, line 599: "dns: hint is not IPv6 address" (and similar error messages elsewhere).

svcb.go
```diff
@@ -0,0 +1,795 @@
...
+	return b, nil
+}
+
+func (s *SVCBIPv6Hint) unpack(b []byte) error {
+	if len(b) == 0 || len(b)%16 != 0 {
+		return errors.New("dns: bad array of IPv6 addresses")
+	}
+	i := 0
+	x := make([]net.IP, 0, len(b)/16)
+	for i < len(b) {
```
tmthrgd, line 612: `for i := 0; i < len(b); i += 16 {`

svcb.go
```diff
@@ -0,0 +1,795 @@
...
+		x = append(x, net.IP(b[i:i+16]))
+		i += 16
+	}
+	s.Hint = x
+	return nil
+}
+
+// String returns "<nil>" if an invalid IPv6 address was encountered.
+// TODO DOC Do I need full definition for doc?
```
tmthrgd, line 621: Same as documentation for `(*SVCBIPv4Hint).String`.

svcb.go
```diff
@@ -0,0 +1,795 @@
...
+	var str strings.Builder
+	str.Grow(40 * len(s.Hint))
+	for _, e := range s.Hint {
+		if e.To4() != nil {
+			return "<nil>"
+		}
+		str.WriteByte(',')
+		str.WriteString(e.String())
+	}
+	return str.String()[1:]
```
tmthrgd, line 632: See `strings.Join` comment on `(*SVCBIPv4Hint)`.

svcb.go
```diff
@@ -0,0 +1,795 @@
...
+		return errors.New("dns: not IPv6")
+	}
+	str := strings.Split(b, ",")
+	dst := make([]net.IP, 0, len(str))
+	for _, e := range str {
+		ip := net.ParseIP(e)
+		if ip == nil {
+			return errors.New("dns: bad IP")
+		}
+		dst = append(dst, ip)
```
tmthrgd, line 646: `dst[i] = ip`

svcb.go
```diff
@@ -0,0 +1,795 @@
...
+	}
+	s.Hint = dst
+	return nil
+}
+
+func (s *SVCBIPv6Hint) copy() SVCBKeyValue {
+	return &SVCBIPv6Hint{
+		append(make([]net.IP, 0, len(s.Hint)), s.Hint...),
```
tmthrgd, line 654: Same as `SVCBMandatory`.

svcb.go
```diff
@@ -0,0 +1,795 @@
...
+	}
+}
+
+// SVCBLocal pair is intended for experimental/private use.
+// The key is recommended to be in the range
+// [SVCB_PRIVATE_LOWER, SVCB_PRIVATE_UPPER].
```
tmthrgd, line 660: Don't line wrap so aggressively.

svcb.go
```diff
@@ -0,0 +1,795 @@
...
+//	o.Hdr.Name = "."
+//	o.Hdr.Rrtype = dns.HTTPS
+//	e := new(dns.SVCBLocal)
+//	e.KeyCode = 65400
+//	e.Data = []byte("abc")
+//	o.Value = append(o.Value, e)
+type SVCBLocal struct {
+	KeyCode uint16 // Never 65535 or any assigned keys
+	Data    []byte // All byte sequences are allowed
+	// For the string representation, See draft-ietf-dnsop-svcb-https
```
tmthrgd, line 673: This comment seems misplaced.

svcb.go
```diff
@@ -0,0 +1,795 @@
...
+}
+
+func (s *SVCBLocal) Key() uint16           { return s.KeyCode }
+func (s *SVCBLocal) pack() ([]byte, error) { return s.Data, nil }
+func (s *SVCBLocal) unpack(b []byte) error { s.Data = b; return nil }
+func (s *SVCBLocal) len() uint16           { return uint16(len(s.Data)) }
+
+// String escapes whitespaces too, which is not required when
+// the result would be enclosed in double quotes. TODO Is this doc fine?
+// do i need definition
```
tmthrgd, line 687: Just drop this comment entirely, `String` is allowed to be undocumented.

svcb.go
```diff
@@ -0,0 +1,795 @@
...
+				fallthrough
+			case ';':
+				fallthrough
+			// As promised, optionally handle space
+			case ' ':
+				fallthrough
+			// Tab
+			case 0x09:
+				fallthrough
+			case '\\':
```
tmthrgd, line 704: Merge all these `case` statements into one:`case '"', ';' ...`.

svcb.go
```diff
@@ -0,0 +1,795 @@
...
+				str.WriteByte('\\')
+				fallthrough
```
tmthrgd, line 706: Drop the `fallthrough` and just use `str.WriteByte(e)`.

svcb.go
```diff
@@ -0,0 +1,795 @@
...
+func (s *SVCBLocal) len() uint16           { return uint16(len(s.Data)) }
+
+// String escapes whitespaces too, which is not required when
+// the result would be enclosed in double quotes. TODO Is this doc fine?
+// do i need definition
+func (s *SVCBLocal) String() string {
+	var str strings.Builder
+	str.Grow(4 * len(s.Data))
+	for _, e := range s.Data {
+		if (0x1f < e && e < 0x7f) || e == 0x09 {
```
tmthrgd, line 692: How does all of this differ from `escapeByte` (i.e. for what bytes is it different)?

svcb.go
```diff
@@ -0,0 +1,795 @@
...
+			i++
+		}
+	}
+	s.Data = bytes
+	return nil
+}
+
+func (s *SVCBLocal) copy() SVCBKeyValue {
+	return &SVCBLocal{s.KeyCode,
+		append(make([]byte, 0, len(s.Data)), s.Data...),
```
tmthrgd, line 752: Same as `SVCBMandatory`.

svcb.go
```diff
@@ -0,0 +1,795 @@
...
+		}
+	}
+	return str.String()
+}
+
+func (s *SVCBLocal) parse(b string) error {
+	bytes := make([]byte, 0, len(b))
+	i := 0
+	for i < len(b) {
+		if b[i] == '\\' {
```
tmthrgd, line 721: Can we switch this around for less indenting:
```go
if b[i] != '\\' {
	bytes = append(bytes, b[i])
	i++
	continue
}
```

svcb.go
```diff
@@ -0,0 +1,795 @@
...
+			}
+		} else {
+			str.WriteString(escapeByte(e))
+		}
+	}
+	return str.String()
+}
+
+func (s *SVCBLocal) parse(b string) error {
+	bytes := make([]byte, 0, len(b))
```
tmthrgd, line 718: Rename `bytes` to `data`.

svcb.go
```diff
@@ -0,0 +1,795 @@
...
+	i := 0
+	for i < len(b) {
+		if b[i] == '\\' {
+			if i+1 == len(b) {
+				return errors.New("dns: SVCB private/experimental key" +
+					" escape unterminated")
```
tmthrgd, line 724: Don't line wrap this.

svcb.go
```diff
@@ -0,0 +1,795 @@
...
+				if i+3 < len(b) && isDigit(b[i+2]) && isDigit(b[i+3]) {
+					a, err := strconv.ParseUint(b[i+1:i+4], 10, 8)
+					if err == nil {
+						i += 4
+						bytes = append(bytes, byte(a))
+						continue
+					}
+				}
+				return errors.New("dns: SVCB private/experimental key" +
+					" bad escaped octet")
```
tmthrgd, line 736: Don't line wrap this.

svcb.go
```diff
@@ -0,0 +1,795 @@
...
+	return &SVCBLocal{s.KeyCode,
+		append(make([]byte, 0, len(s.Data)), s.Data...),
+	}
+}
+
+func (rr *SVCB) String() string {
+	s := rr.Hdr.String() +
+		strconv.Itoa(int(rr.Priority)) + " " +
+		sprintName(rr.Target)
+	for _, element := range rr.Value {
```
tmthrgd, line 760: Rename `element` to `e`.

svcb.go
```diff
@@ -0,0 +1,795 @@
...
+		s += " " + svcbKeyToString(element.Key()) +
+			"=\"" + element.String() + "\""
```
tmthrgd, line 762: Put all of this in one line.

svcb.go
```diff
@@ -0,0 +1,795 @@
...
+	}
+	return s
+}
+
+// areSVCBPairArraysEqual checks if SVCBKeyValue arrays are equal
+// after sorting their copies. arrA and arrB have equal lengths,
+// otherwise zduplicate.go wouldn't call this function.
```
tmthrgd, line 769: Don't line wrap this so aggressively.

svcb.go
```diff
@@ -0,0 +1,795 @@
...
+func areSVCBPairArraysEqual(arrA []SVCBKeyValue, arrB []SVCBKeyValue) bool {
+	a := append(make([]SVCBKeyValue, 0, len(arrA)), arrA...)
+	b := append(make([]SVCBKeyValue, 0, len(arrB)), arrB...)
```
tmthrgd, line 772: `append([]SVCBKeyValue(nil), ...)`

svcb.go
```diff
@@ -0,0 +1,795 @@
...
+		s += " " + svcbKeyToString(element.Key()) +
+			"=\"" + element.String() + "\""
+	}
+	return s
+}
+
+// areSVCBPairArraysEqual checks if SVCBKeyValue arrays are equal
+// after sorting their copies. arrA and arrB have equal lengths,
+// otherwise zduplicate.go wouldn't call this function.
+func areSVCBPairArraysEqual(arrA []SVCBKeyValue, arrB []SVCBKeyValue) bool {
```
tmthrgd, line 770: Just name these `a` and `b`.

svcb.go
```diff
@@ -0,0 +1,795 @@
...
+	a := append(make([]SVCBKeyValue, 0, len(arrA)), arrA...)
+	b := append(make([]SVCBKeyValue, 0, len(arrB)), arrB...)
+	sort.Slice(a, func(i, j int) bool {
```
tmthrgd, line 773: These can probably be one line: `sort.Slice(a, func(i, j int) bool { return a[i].Key() < a[j].Key() })`.

svcb.go
```diff
@@ -0,0 +1,795 @@
...
+	for i, e := range a {
+		if e.Key() != b[i].Key() {
+			return false
+		}
+		b1, err1 := e.pack()
+		b2, err2 := b[i].pack()
+		if err1 != nil || err2 != nil || len(b1) != len(b2) {
+			return false
+		}
+		for bi, x := range b1 {
```
tmthrgd, line 788: Use `bytes.Equal`.

svcb.go
```diff
@@ -0,0 +1,795 @@
...
+	sort.Slice(b, func(i, j int) bool {
+		return b[i].Key() < b[j].Key()
+	})
+	for i, e := range a {
+		if e.Key() != b[i].Key() {
+			return false
+		}
+		b1, err1 := e.pack()
+		b2, err2 := b[i].pack()
+		if err1 != nil || err2 != nil || len(b1) != len(b2) {
```
tmthrgd, line 785: Leave the `len` check for `bytes.Equal`.

svcb_test.go
```diff
@@ -0,0 +1,100 @@
...
+		`key65000=4\ 3`:                  `4\ 3`,
+		`key65001="\" "`:                 `\"\ `,
+		`key65002=""`:                    ``,
+		`key65003==\"\"`:                 `=\"\"`,
+		`key65004=\254\032\030\000`:      `\254\ \030\000`,
+	}
+
+	for s, o := range svcbs {
+		key := ""
+		val := ""
```
tmthrgd, line 28: `var key, value string`.

svcb_test.go
```diff
@@ -0,0 +1,100 @@
...
+		idx := strings.IndexByte(s, '=')
```
tmthrgd, line 29: Is there a good reason to reimplement all of this instead of just using the parser code (eventually hitting `(*SVCB) parse`)?

svcb_test.go
```diff
@@ -0,0 +1,100 @@
...
+			continue
+		}
+		if str := keyValue.String(); str != o {
+			t.Errorf("`%s' should be equal to\n`%s', but is     `%s'", s, o, str)
+		}
+	}
+}
+
+func TestDecodeBadSVCB(t *testing.T) {
+	svcbs := map[int][][]byte{
```
tmthrgd, line 74: Replace this with `[]struct { key uint16; data []byte }`.

types_generate.go
```diff
@@ -272,6 +277,12 @@ func main() {
 					fields = append(fields, f)
 					continue
 				}
+				if t == "SVCBKeyValue" {
```
tmthrgd, line 280: This should probably check for ``` `dns:"pairs"` ``` rather than `"SVCBKeyValue"`.

types_generate.go
```diff
@@ -272,6 +277,12 @@ func main() {
...
+					fmt.Fprintf(b, "%s := make([]%s, len(rr.%s));\nfor i,e := range rr.%s {\n %s[i] = e.copy()\n}\n",
```
tmthrgd, line 281: You can drop the `\n`s from here, it's formatted anyway.

svcb.go
tmthrgd, line 770: Make this: `func (a []SVCBKeyValue) isEqual(b []SVCBKeyValue) bool`.

svcb.go
```diff
@@ -0,0 +1,795 @@
...
+//	e.Code = []uint16{65403}
+//	o.Value = append(o.Value, e)
+//  // Then add key-value pair for key65403
+type SVCBMandatory struct {
+	Code []uint16 // Must not include mandatory
+}
+
+func (s *SVCBMandatory) Key() uint16 { return SVCB_MANDATORY }
+func (s *SVCBMandatory) String() string {
+	str := make([]string, 0, len(s.Code))
```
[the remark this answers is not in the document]
DesWurstes: I can't, there is `(s *SVCBMandatory)`.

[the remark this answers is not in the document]
tmthrgd: How did I miss that. No worries, this can stay then.

svcb.go
```diff
@@ -0,0 +1,795 @@
...
+	return nil
+}
+
+func (s *SVCBAlpn) parse(b string) error {
+	s.Alpn = strings.Split(b, ",")
+	return nil
+}
+
+func (s *SVCBAlpn) len() uint16 {
+	l := len(s.Alpn)
```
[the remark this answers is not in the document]
DesWurstes: `l` is equal to array length plus the size of each element.

svcb.go
DesWurstes: In this case we know the maximum length of each string which is also a good approximation for length while the strings approach uses a lot of pointers and much less efficient. Anyways, no one will benchmark these mini-functions so readability is more important.

svcb.go
DesWurstes: I can't drop the `fallthrough` but now it looks like this:
```go
			switch e {
			case '"', ';', ' ', '\\', 0x09:
				str.WriteByte('\\')
				fallthrough
			default:
				str.WriteByte(e)
			}
```

svcb.go
DesWurstes: In `(a []SVCBKeyValue)`, there can't be an array. `invalid receiver type []SVCBKeyValue ([]SVCBKeyValue is not a defined type)`. For that we need to define a new type as in https://stackoverflow.com/a/26526822 like `type SVCBarray []SVCBKeyValue`

svcb.go
DesWurstes: The exact inverse plus `0x09`.

types_generate.go
DesWurstes: It needs to be consistent with the previous 8 lines of code?

types_generate.go
DesWurstes: It needs to be consistent with the previous 8 lines of code?

svcb.go
DesWurstes: `isDuplicate` shouldn't modify its inputs.

DesWurstes:
@tmthrgd Thank you for thorough review, I've addressed them all.

types_generate.go
tmthrgd: I'd rather this check for `dns:"pairs"`, the others probably should be changed at some later date to use tags as well instead. It might be better to leave this as is for now and do something in a later PR. I'll leave it up to you.

types_generate.go
tmthrgd: That's fine then.

svcb.go
tmthrgd: Of course it can't be :facepalm:.

DesWurstes: Then how should I name `a` and `b` below?

svcb.go
tmthrgd: What I'm saying is replace this with:
```go
case ...:
	str.WriteByte('\\')
	str.WriteByte(e)
default:
	str.WriteByte(e)
```

svcb.go
```diff
@@ -0,0 +1,764 @@
...
+	return nil
+}
+
+func (s *SVCBLocal) String() string {
+	var str strings.Builder
+	str.Grow(4 * len(s.Data))
+	for _, e := range s.Data {
+		if (0x1f < e && e < 0x7f) || e == 0x09 {
+			switch e {
+			case '"', ';', ' ', '\\', 0x09:
```
tmthrgd, line 682: This can't be right for 0x09 (\t), because that would be `\	` (slash then tab).

svcb.go
tmthrgd: Can we switch this around so it's like:
```go
if e < ' ' || e > '~' {
	str.WriteString(escapeByte(e))
} else {
	...
}
```

svcb.go
tmthrgd: They can share the same name: `a = append(...)`.

svcb.go
tmthrgd: My suggestion here is just to put the existing code onto one line.

DesWurstes:
@tmthrgd 

> This can't be right for 0x09 (\t), because that would be \ (slash then tab).

Spaces and tabs are treated the same. These are allowed:

```go
1) dnskey="a b"
2) dnskey=a\ b
3) dnskey="a\ b"
```

The current code uses the third form for experimental key=value pairs. It's easy to change it to the first, would you prefer the first?

> They can share the same name: a = append(...).

If:

```go
func areSVCBPairArraysEqual(a []SVCBKeyValue, b []SVCBKeyValue) bool {
	a := append([]SVCBKeyValue(nil), a...)
	b := append([]SVCBKeyValue(nil), b...)
```

Then:
```go
./svcb.go:744:4: no new variables on left side of :=
./svcb.go:745:4: no new variables on left side of :=
```

ghedo:
Looks like this has gone silent for a while, any update? It seems most review comments have been addressed, but I guess @miekg hasn't had a chance to do his "minimization" pass yet?

svcb.go
```diff
@@ -0,0 +1,759 @@
...
+		if x == nil {
+			return "<nil>"
+		}
+		str[i] = x.String()
+	}
+	return strings.Join(str, ",")
+}
+
+func (s *SVCBIPv4Hint) parse(b string) error {
+	if strings.Contains(b, ":") {
```
tmthrgd, line 523: I don’t love this check. Is there any harm in letting people use IPv4-in-IPv6 addresses?

svcb.go
```diff
@@ -0,0 +1,759 @@
...
+	return b, nil
+}
+
+func (s *SVCBIPv6Hint) unpack(b []byte) error {
+	if len(b) == 0 || len(b)%16 != 0 {
+		return errors.New("dns: svcbipv6hint: ipv6 address byte array length not a multiple of 16")
+	}
+	x := make([]net.IP, 0, len(b)/16)
+	for i := 0; i < len(b); i += 16 {
+		x = append(x, net.IP(b[i:i+16]))
```
tmthrgd, line 609: This doesn’t check for IPv4 addresses, should it?

svcb.go
```diff
@@ -0,0 +1,759 @@
...
+		if x := e.To4(); x != nil {
+			return "<nil>"
+		}
+		str[i] = e.String()
+	}
+	return strings.Join(str, ",")
+}
+
+func (s *SVCBIPv6Hint) parse(b string) error {
+	if strings.Contains(b, ".") {
```
tmthrgd, line 628: I don’t like this check, just check `To4() != nil` inside the for-loop.

svcb.go
tmthrgd: I’d still like this to be clearer.

svcb.go
```diff
@@ -0,0 +1,759 @@
...
+			str.WriteString(escapeByte(e))
+		}
+	}
+	return str.String()
+}
+
+func (s *SVCBLocal) parse(b string) error {
+	data := make([]byte, 0, len(b))
+	i := 0
+	for i < len(b) {
```
tmthrgd, line 697: `for I := 0; ...` as `i` isn’t used outside of the loop.

tmthrgd (changes requested):
A handful of relatively minor changes, and then it should be ready to merge. Thanks for the patience.

svcb.go
DesWurstes: It's impossible to check for the address type, this is just a byte array.

DesWurstes:
> Just to confirm the wire format is also base64 encoded?

Thank you, this used to be a broken implementation. I've now verified this with ghedo's Python tool.

> I don’t love this check. Is there any harm in letting people use IPv4-in-IPv6 addresses?

This was to be consistent with #1107, unless it changed in the last few months?

svcb.go
tmthrgd: A `To4() != nil` check was what I was referring to. Not sure if it’s a good idea, but it would seem consistent with other methods on this type.

tmthrgd:
@DesWurstes
> “This was to be consistent with #1107, unless something changed in the last few months?”

My memory apparently, months is a life time ago. :sweat_smile:

If it’s consistent with elsewhere in this library, it’s fine to stay.

svcb.go
DesWurstes: Well, if it's okay for you I would prefer not adding conditions that would always evaluate to one branch, `To4()` is always `nil` for 16-byte IP arrays.

tmthrgd: That seems inconsistent with other methods on this type that all have a `To4() != nil` check on them. It’s definitely not going to be a hotspot.

DesWurstes: OK, changed it.

tmthrgd (approved):
I'm happy with this now, no need to iterate ad infinitum over minor nits. Should be fine to merge. Thanks for your patience.

Issue #997, opened 2019-07-15
too many goroutines


alpc40:
Hi,
i'm use this in windows10.  When I use 200 threads to nslookup and send UDP package to dns server, it finds that there are too many goroutines and occupying large memory of heaps.

So I review the code, and find that there is no limits when proccess UDP packages in server.go. I don't sure if it is all right.  May anybody help me?

my code:
```
func serveDNS(laddr string) error {
	serveMux := dns.NewServeMux()
	serveMux.HandleFunc(".", handleDnsRequest)
	glog.Errorf("serveDNS Begin...")
	e := make(chan error)
	for _, _net := range [...]string{"udp", "tcp"} {
		srv := &dns.Server{Addr: laddr, Net: _net, Handler: serveMux}
		go func(srv *dns.Server) {
			e <- srv.ListenAndServe()
		}(srv)
	}
	return <-e
}
```
dns code:
```
github.com/miekg/dns/server.go

for srv.isStarted() {
		m, s, err := reader.ReadUDP(l, rtimeout)
		if err != nil {
			if !srv.isStarted() {
				return nil
			}
			if netErr, ok := err.(net.Error); ok && netErr.Temporary() {
				continue
			}
			return err
		}
		if len(m) < headerSize {
			if cap(m) == srv.UDPSize {
				srv.udpPool.Put(m[:srv.UDPSize])
			}
			continue
		}
		wg.Add(1)
		go srv.serveUDPPacket(&wg, m, l, s)
	}
```

tmthrgd:
@miekg I think this is something we should fix. I'm thinking a simple opt-in/opt-out semaphore w/ optional timeout. We can just use a `chan struct{}` as our semaphore. This might help with #916 and the related CoreDNS issues.

Something like this perhaps:
```go
type Server struct {
	// Controls the maximum number of UDP queries to process concurrently. If the value is
	// negative/non-zero, no limit is applied. Once the limit is reached, the server will block or
	// reject the query depending on UDPQueryTimeout.
	MaxConcurrentUDPQueries int
	// If this timeout is positive, the server will reject the query if it can't handle the request
	// within the timeout.
	UDPQueryTimeout time.Duration
}
```

This approach isn't perfect and has problems around choosing the correct value, but I think we'll keep seeing these problems otherwise.

I'm not sure whether it makes sense to try and respond with an error or just drop the query on the floor.

@miekg Thoughts?

miekg:
[...]

PR #1068, opened 2020-01-08
Add LimitReader to drop excess packets

miekg:
This includes code from #1052 and add a tests

See for further discussion: #997

### DIFF
--- a/limit_reader.go
+++ b/limit_reader.go
@@ -1,0 +1,115 @@
+package dns
+
+import (
+	"net"
+	"runtime"
+	"time"
+)
+
+// LimitReader can be used to limit the intake of new packets. Currently it works by checking the number of
+// active goroutines. If we have too many it will refuse any new messages. Note that this check is performed every 100th packet.
+//
+// LimitReader can be used as a DecorateReader in a server as:
+//
+//	server := &Server{/* various options */}
+//	server.DecorateReader = func(r Reader) Reader { return &LimitReader{Reader: r, MaxGoroutines: 10000} }
+//
+type LimitReader struct {
+	Reader
+
+	// MaxGoroutines is the maxium number of goroutines we're willing to tolerate.
+	MaxGoroutines int
+
+	upkts int
+	tpkts int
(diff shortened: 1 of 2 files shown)


miekg:
there 1 problem though, the LimitReader can't be both used in the udp and tcp path without locking (or maybe different counters. That should be OK?)

miekg:
We could also simplify the algo: `goroutines > max => refuse` instead of trying to be smart (I think this has my preference)

miekg:
failing with
~~~
o: downloading golang.org/x/tools v0.0.0-20191216052735-49a3e744a425

go: extracting golang.org/x/tools v0.0.0-20191216052735-49a3e744a425

panic: runtime error: index out of range

goroutine 1 [running]:

go/types.(*Struct).Field(...)

	/home/travis/.gimme/versions/go1.12.14.linux.amd64/src/go/types/type.go:148

main.getTypeStruct(0x704b20, 0xc00080b7a0, 0xc000af13b0, 0xf, 0xc00096a368)

	/home/travis/gopath/src/github.com/miekg/dns/duplicate_generate.go:33 +0x1ab

main.main()

	/home/travis/gopath/src/github.com/miekg/dns/duplicate_generate.go:67 +0x1516

exit status 2

~~~
not related to this PR...?

chantra:
Wouldn't this only refuse 1 out of 100 packet when we are getting saturated?
Would it be better to return REFUSED until the number of goroutines get back under the threshold?

There is pros and cons of each approaches, but I am not sure how REFUSING 1 out of 100 will help.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Add LimitReader to ..." ]
>Wouldn't this only refuse 1 out of 100 packet when we are getting saturated?

uhm, yes ... changed :-)

>Would it be better to return REFUSED until the number of goroutines get back under the threshold?

/Miek

--
Miek Gieben

miekg:
leaving this as a way to do it, not merging

PR #1073, opened 2020-02-13
CanonicalName function to return domain name in canonical form

fcelda:
This PR adds `Canonical` function which returns given domain in canonical form.

I realise the implementation is rather trivial. There are a few places the library where a name in canonical form is required (e.g. map with TSIG keys) but canonical names are also useful to normalize names in other context. Essentially all projects and little tools that I've ever worked on and that use this library have a copy of this little function. Therefore I think it might be worth including in the library even though the implementation is trivial.

Please, let me know what you think.

Also, let me know if `Canonical` is name or if it should be something else like `CanonicalName`.

### DIFF
--- a/serve_mux.go
+++ b/serve_mux.go
@@ -35,7 +35,7 @@ func (mux *ServeMux) match(q string, t uint16) Handler {
 		return nil
 	}
 
-	q = Canonical(q)
+	q = CanonicalName(q)
 
 	var handler Handler
 	for off, end := 0, false; !end; off, end = NextLabel(q, off) {
--- a/serve_mux.go
+++ b/serve_mux.go
@@ -65,7 +65,7 @@ func (mux *ServeMux) Handle(pattern string, handler Handler) {
 	if mux.z == nil {
 		mux.z = make(map[string]Handler)
 	}
-	mux.z[Fqdn(pattern)] = handler
+	mux.z[CanonicalName(pattern)] = handler
 	mux.m.Unlock()
 }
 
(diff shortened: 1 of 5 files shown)


miekg:
I think this can make sense as an internal, non exported, function, but should not be part of the API.
Two reasons:
1) as you said, it's rather trivial
2) discoverability; I see (lot of) Go code adding a dot to a domain name and not using `Fqdn()`, because it's hard to find. So we carry `Fqdn` for the rest of time and only a minority of users actually use it

fcelda:
I can't say I disagree. :slightly_smiling_face: I really don't have strong opinion on this.

> discoverability; I see (lot of) Go code adding a dot to a domain name and not using Fqdn(), because it's hard to find. So we carry Fqdn for the rest of time and only a minority of users actually use it

This is sad because turning a name into a fully qualified one can be surprising given possible character escaping. Having a tested library function is beneficial.

tmthrgd:
I prefer `CanonicalName` to just `Canonical`. I do think there's value in adding this depending on how often it actually gets used. That being said, it is definitely trivial so it's not the most compelling addition.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Canonical function ..." ]
>I prefer `CanonicalName` to just `Canonical`. I do think there's value in adding this depending on how often it actually gets used. That being said, it is definitely trivial so it's not the most compelling addition.

Fair enough, that makes 2 people think this is useful. Please make that name change,
otherwise /lgtm

fcelda:
OK, I will rename the function. Give me a few days please.

serve_mux.go
```diff
@@ -36,7 +35,7 @@ func (mux *ServeMux) match(q string, t uint16) Handler {
 		return nil
 	}
 
-	q = strings.ToLower(q)
+	q = CanonicalName(q)
```
tmthrgd: Could you also change `Fqdn(pattern)` to `CanonicalName(pattern)` in `Handle` and `HandleRemove`.

fcelda: Updated. :+1:

fcelda:
Thank you. :100:

Issue #640, opened 2018-02-13
TSIG verification via nsupdate failing "dns: bad signature"


ennetech:
Hello!
I have a simple DNS server inplace and i'm using this piece of code to implement tsig validation:
```
secret := conf.SystemConfig.TsigKey
logger.Info("USING THIS KEY "+secret)
pack, _ := r.Pack()
err = dns.TsigVerify(pack, secret, "", false)
if err != nil {
	logger.Error("TSIG VERIFICATION FAILED " + err.Error())
} else {
	logger.Info("TSIG VERIFICATION SUCCEDEED")
}
```
(https://github.com/ennetech/consul-dns/blob/master/pkg/dns/dns.go Line 70)

When sending this update (nsupdate -v update):
```
debug yes
key hmac-md5:mydomain.com. {{ conf.SystemConfig.TsigKey }}
server 192.168.1.101 53535 // Local server
zone mydomain.com.
update add snd.mydomain.com. 300 A 127.0.0.1
send
```
I receive an error "dns: bad signature", if i substitute "snd.mydomain.com." with "snd" the TSIG verification passes but the update request is wrong

I am generating the keys this way:
```
	// hmac-md5
	hmacmd5 := hmac.New(md5.New, []byte(generatingSecret))
	res += "hmac-md5: " + base64.StdEncoding.EncodeToString(hmacmd5.Sum(nil))
	// hmac-sha1
	hmacsha1 := hmac.New(sha1.New, []byte(generatingSecret))
	res += "hmac-sha1: " + base64.StdEncoding.EncodeToString(hmacsha1.Sum(nil))
	// hmac-sha256
	hmacsha256 := hmac.New(sha256.New, []byte(generatingSecret))
	res += "hmac-sha256: " + base64.StdEncoding.EncodeToString(hmacsha256.Sum(nil))
	// hmac-sha512
	hmacsha512 := hmac.New(sha512.New, []byte(generatingSecret))
	res += "hmac-sha512: " + base64.StdEncoding.EncodeToString(hmacsha512.Sum(nil))
```

I am experiencing the same error with letsencrypt certbot but not with hashicorp terraform
RFC2136+RFC2845

andrewtj:
> I receive an error "dns: bad signature", if i substitute "snd.mydomain.com." with "snd" the TSIG verification passes but the update request is wrong

[...]

PR #1077, opened 2020-02-21
Tsig secret provider

0x5a17ed:
These changes replace the secrets map present in both the server and the client with a new resolver interface that will be inquired to resolve a secret by its canonical name into the corresponding secret bytes. The new approach is more flexible and allows server implementations to be free in their secrets managing strategy.

So far these changes are just a draft and haven't been battle tested yet.

### DIFF
--- a/client.go
+++ b/client.go
@@ -20,9 +20,9 @@ const (
 
 // A Conn represents a connection to a DNS server.
 type Conn struct {
-	net.Conn                         // a net.Conn holding the connection
-	UDPSize        uint16            // minimum receive buffer for UDP messages
-	TsigSecret     map[string]string // secret(s) for Tsig map[<zonename>]<base64 secret>, zonename must be in canonical form (lowercase, fqdn, see RFC 4034 Section 6.2)
+	net.Conn                          // a net.Conn holding the connection
+	UDPSize        uint16             // minimum receive buffer for UDP messages
+	TsigSecrets    TsigSecretResolver // TsigSecrets will be used to resolve the given secret name in canonical form (lowercase, fqdn, see RFC 4034 Section 6.2) into the corresponding secret bytes.
 	tsigRequestMAC string
 }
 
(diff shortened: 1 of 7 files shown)


miekg:
[ Quoting <notifications@github.com> in "[miekg/dns] Tsig secret provider (#..." ]
>These changes replace the secrets map present in both the server and the client with a new resolver interface that will be inquired to resolve a secret by its canonical name into the corresponding secret bytes. The new approach is more flexible and allows server implementations to be free in their secrets managing strategy.
>
>So far these changes are just a draft and haven&#39;t been battle tested yet.
>You can view, comment on, or merge this pull request online at:
>
>  https://github.com/miekg/dns/pull/1077
>
>-- Commit Summary --
>
>  * Ensure tsig state is verified in TestServerRoundtripTsig
>  * Use a separate resolver to resolve TSig secrets instead of a map

This seems like a nice idea. You probably want to add to `Conn` and `Client` instead of
replacing to remain backwards compatible.

Is this new interface being use/planned to be used?

0x5a17ed:
> This seems like a nice idea. You probably want to add to `Conn` and `Client` instead of
replacing to remain backwards compatible.

Thanks for the quick response and the nice feedback! :+1: Regarding the backwards compatibility, I will see what I can do, but I have to coordinate with my supervisor about it first. It would help if you could tell me whenever this is a hard requirement to get this feature merged. It shouldn't be hard though.

> Is this new interface being use/planned to be used?

It is planned to be used in a software project. I haven't gotten around to implement it yet but that's on the plate. So the implementation might or might not change depending on how I manage to get this implemented into that project. That's why I submitted the PR as a draft. My supervisor asked me to upload it to Github before testing it to get some feedback but I'm not exactly sure if this is the final state of the implementation so it might be subject to change and reviewing it a waste of time. Proceed at your own risk. :stuck_out_tongue_winking_eye:

corny:
@miekg In case we keep backwards compatibility - when do you want to drop it to remove the technical debt?

Migrating to the new interface is quite easy. Before:
```golang
conn.TsigSecret = map[string]string{"test.": "so6ZGir4GPAqINNh9U5c3A=="}
```

After:
```golang
conn.TsigSecrets = &TsigSecretMap{}
conn.TsigSecrets.SetBase64Secret("test.", "so6ZGir4GPAqINNh9U5c3A==")
```

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Tsig secret provide..." ]
>@miekg In case we keep backwards compatibility - when do you want to drop it to remove the technical debt?

probably never

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Tsig secret provide..." ]
>> This seems like a nice idea. You probably want to add to `Conn` and `Client` instead of
>replacing to remain backwards compatible.
>
>Thanks for the quick response and the nice feedback! :+1: Regarding the backwards compatibility, I will see what I can do, but I have to coordinate with my supervisor about it first. It would help if you could tell me whenever this is a hard requirement to get this feature merged. It shouldn't be hard though.
>
>> Is this new interface being use/planned to be used?
>
>It is planned to be used in a software project. I haven't gotten around to implement it yet but that's on the plate. So the implementation might or might not change depending on how I manage to get this implemented into that project. That's why I submitted the PR as a draft. My supervisor asked me to upload it to Github before testing it to get some feedback but I'm not exactly sure if this is the final state of the implementation so it might be subject to change and reviewing it a waste of time. Proceed at your own risk. :stuck_out_tongue_winking_eye:

Wait, so there is only one future user of this? Then I probably hold off and merging this
into master and wait for other folks to chime in that this solves a specific problem they
have with the current impl.

0x5a17ed:
> Wait, so there is only one future user of this? Then I probably hold off and merging this
into master and wait for other folks to chime in that this solves a specific problem they
have with the current impl.

The project, that I'm currently maintaining, fetches secrets for the TSIG verification from a database in a dynamic way. The previous map interface to the TSIG handling in the DNS server provided by this module was kind of "lame" for this use case because it required to update the map every time a key was added, deleted or modified. It is otherwise not clear how to handle TSIG verification in a DNS server that uses this module in a way that allows the keyset to be modified. The previous implementation in said project tackled this issue by deferring the TSIG verification until the `ServeDNS` method in the `Handler` interface. This brings its own set of issues though and makes it impossible to verify requests properly, if the original message cannot be reproduced by serializing the parsed message back into its raw form. It seems that this issue here was about the same problem: <https://github.com/miekg/dns/issues/640>. So there might be more users of this DNS module hidden in the woods with the same problem, they just worked around it one way or the other and their use cases were simple enough for now to not have the workaround cause them any problems.

0x5a17ed:
It seems these changes are relatively stable so I remove the "Draft` marker. It's ready to be reviewed I guess.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Tsig secret provide..." ]
>It seems these changes are relatively stable so I remove the "Draft` marker. It's ready to be reviewed I guess.

See earlier message, if this is just nice to have with no users, it will not be merged.
And I haven't seen any noise that this is indeed a must-have feature

corny:
Do you see another way to solve the retrieval of secrets from an external database? The current approach involves re-implementing the serve handler.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Tsig secret provide..." ]
>Do you see another way to solve the retrieval of secrets from an external database? The current approach involves re-implementing the serve handler.

it's a map, it's exported; you can write to up. Granted there is no locking, so that could
be changed, but that's a couple of lines.

The more I think about this PR, the less I think this is a really good addition

corny:
The approach of just having an interface to implement gives the freedom to adopt the library to the specific needs of any project. The map is just one implementation. I see no elegant way of keeping the map up to date when an external database is being updated.

PR #1080, opened 2020-02-24
update noRdata to limit acceptable empty RDATA cases

jinmeiib:
The current implementation of `noRdata` in msg_helpers.go just checks whether the RDLENGTH is 0, and its callers skip any further validation.  But it allows invalid empty RDATA for most types of RRs (A, AAAA...) in wire format to be successfully "unpacked".  According to the code comment, the intent seems to allow empty RDATA in a dynamic update request.  If so, we can limit the case to some specific RR classes since it's only possible for "delete" variants of update requests.

So I propose including this check in `noRdata`.  The actual diff also covers some special types, mainly for allowing empty OPT RR (which is indeed valid, but its `unpack` rejects).

I personally think it would be even better if `UnpackRR` is more aware of the unpack context and also checks OPCODE so it won't accidentally accept broken wire form data from the net, but that will be a much bigger change, so I chose to suggest this small adjustment first.

I'd also note that the `unpack` implementation of some RR types will have to be updated to reject empty RDATA when it's not allowed.  Just from a quick look, at least TXT.unpack should be updated (empty TXT RDATA isn't possible per RFC1035).  There are probably some more.  See the comment in revised `TestNoRdataPack`.

### DIFF
--- a/dns_test.go
+++ b/dns_test.go
@@ -169,22 +169,39 @@ func TestNoRdataPack(t *testing.T) {
 
 func TestNoRdataUnpack(t *testing.T) {
 	data := make([]byte, 1024)
-	for typ, fn := range TypeToRR {
-		if typ == TypeSOA || typ == TypeTSIG || typ == TypeTKEY {
+	classes := []uint16{ClassINET, ClassNONE, ClassANY}
+	for typ, _ := range TypeToRR {
+		switch typ {
+		case TypeSOA, TypeTSIG, TypeTKEY:
 			// SOA, TSIG will not be seen (like this) in dyn. updates?
 			// TKEY requires length fields to be present for the Key and OtherData fields
 			continue
-		}
-		r := fn()
-		*r.Header() = RR_Header{Name: "miek.nl.", Rrtype: typ, Class: ClassINET, Ttl: 16}
-		off, err := PackRR(r, data, 0, nil, false)
-		if err != nil {
-			// Should always works, TestNoDataPack should have caught this
-			t.Errorf("failed to pack RR: %v", err)
+		case TypeDHCID, TypeSPF, TypeAPL, TypeAVC, TypeTXT, TypeOPENPGPKEY, TypeEID, TypeNIMLOC, TypeNINFO:
+			// UnpackRR leniently allows empty RDATA for these types.
+			// At least some of them (e.g. TXT) should obviously be rejected, so this list may have to be updated in future.
 			continue
 		}
-		if _, _, err := UnpackRR(data[:off], 0); err != nil {
-			t.Errorf("failed to unpack RR with zero rdata: %s: %v", TypeToString[typ], err)
+		for _, class := range classes {
+			// Build wire-format empty RDATA.  We use RFC3597{} as packRR of concrete struct types could append some data.
+			// Root owner name is chosen to make offset calculation easy.
+			r := &RFC3597{}
+			*r.Header() = RR_Header{Name: ".", Rrtype: typ, Class: class, Ttl: 16}
+			off, err := PackRR(r, data, 0, nil, false)
+			if off != 11 { // 1 byte owner name followed by fixed 10-byte RR header
+				t.Errorf("unexpected pack offset for %s: %d", TypeToString[typ], off)
+			}
+			if err != nil {
+				// Should always works, TestNoDataPack should have caught this
+				t.Errorf("failed to pack RR: %v", err)
+				continue
+			}
+			rr, _, err := UnpackRR(data[:off], 0)
+			if err != nil && (class != ClassINET || !canParseAsRR(typ)) {
+				t.Errorf("failed to unpack RR with zero rdata: %s: %v", TypeToString[typ], err)
+			}
+			if err == nil && class == ClassINET && canParseAsRR(typ) {
+				t.Errorf("unexpectedly succeeded to unpack RR with zero rdata: %v", rr)
+			}
 		}
 	}
 }
--- a/msg_helpers.go
+++ b/msg_helpers.go
@@ -171,8 +171,10 @@ func fromBase64(s []byte) (buf []byte, err error) {
 
 func toBase64(b []byte) string { return base64.StdEncoding.EncodeToString(b) }
 
-// dynamicUpdate returns true if the Rdlength is zero.
-func noRdata(h RR_Header) bool { return h.Rdlength == 0 }
+// noRdata returns true if the Rdlength can be and is indeed zero.
+// It's allowed for dynamic update delete records.
+// This function also returns true for types that lacks a presentation format.
+func noRdata(h RR_Header) bool { return h.Rdlength == 0 && (h.Class == ClassANY || h.Class == ClassNONE || !canParseAsRR(h.Rrtype))}
 
 func unpackUint8(msg []byte, off int) (i uint8, off1 int, err error) {
 	if off+1 > len(msg) {
--- a/update_test.go
+++ b/update_test.go
@@ -34,7 +34,7 @@ func TestDynamicUpdateUnpack(t *testing.T) {
 
 func TestDynamicUpdateZeroRdataUnpack(t *testing.T) {
 	m := new(Msg)
-	rr := &RR_Header{Name: ".", Rrtype: 0, Class: 1, Ttl: ^uint32(0), Rdlength: 0}
+	rr := &RR_Header{Name: ".", Rrtype: 0, Class: ClassNONE, Ttl: ^uint32(0), Rdlength: 0}
 	m.Answer = []RR{rr, rr, rr, rr, rr}
 	m.Ns = m.Answer
 	for n, s := range TypeToString {


miekg:
this stuff is super under specified and the opcode is not tied to the RR, so I don't think you can just claim that this is invalid binary data. It can exist on the wire so we should be able to parse it.

jinmeiib:
> It can exist on the wire

In which case the does following RR validly (i.e. confirming the protocol standard) exist on the wire?
Type = TXT
Class = IN
RDLEN = 0
RDATA: empty

As far as I understand the RFC, this is not possible either per RFC1035 or RFC2136.  So I can only imagine a broken DNS message if we see it on the wire.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] update noRdata to l..." ]
>> It can exist on the wire
>
>In which case the does following RR validly (i.e. confirming the protocol standard) exist on the wire?
>Type = TXT
>Class = IN
>RDLEN = 0
>RDATA: empty
>
>As far as I understand the RFC, this is not possible either per RFC1035 or RFC2136.  So I can only imagine a broken DNS message if we see it on the wire.

Why not? So you can't use this in dynamic updates? That would be odd and would go against
the intent of dyn updates.

If you're right you should bring it up on dnsops, i see no reason to decide here that this
cannot be done, because clearly a rdlen=0 is saying that there will not be rdata

jinmeiib:
> Why not? So you can't use this in dynamic updates? 

In the case of CLASS=IN, it means an addition of RR to a class IN zone.  From Section 3.4.2.6 of RFC2136, for example:
```
   CLASS    TYPE     RDATA    Meaning
   ---------------------------------------------------------
   ANY      ANY      empty    Delete all RRsets from a name
   ANY      rrset    empty    Delete an RRset
   NONE     rrset    rr       Delete an RR from an RRset
   zone     rrset    rr       Add to an RRset
```
(In this case 'zone' = IN).  So the RDATA is 'rr', and must be able to be added to a zone, just like loadbale to a zone from a zone file.  It may be able to be empty type-specific basis (although the only type I know of that allows it is NULL), and will have to be validated per type basis, e.g., in `AAAA.unpack`.

But the current code allows, for example, the following in an update request:
```
   CLASS    TYPE     RDATA    Meaning
   zone=IN  AAAA     empty    (what to add to the zone?)
```

This PR just tries to prevent that, deferring the validation to the type specific unpack code.  Do you still think this is a protocol specification issue and I should bring it to dnsop?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] update noRdata to l..." ]
>> Why not? So you can't use this in dynamic updates?
>
>This PR just tries to prevent that, deferring the validation to the type specific unpack code.  Do you still think this is a protocol specification issue and I should bring it to dnsop?

This would mean that parsing rdata is class dependent? That's slightly less worse than
opcode dependent, but still sounds like a new requirement to me. Can an RR have different
RDATA semantics based on class? is essentially the question.

jinmeiib:
> Can an RR have different RDATA semantics based on class? is essentially the question.

No, except for some very class-specific types such as AAAA or A.  And that's actually why we can't simply allow empty RDATA in `UnpackRRWithHeader` like this:

```
	if noRdata(h) {
		return rr, off, nil
	}
```

and we should rather call corresponding `unpack` for the RR to see if the empty RDATA is allowed for that type:
```
	off, err = rr.unpack(msg, off)
	if err != nil {
		return nil, end, err
	}
```

The dynamic update delete cases are very special exceptions to these.  RFC2136 allows the RDATA can be empty when (and only when) the RRCLASS is ANY so that the record can represent the semantics of "delete all RRs of this name (and type)".  As far as I know, no other RFCs allow an outright acceptance of an empty RDATA.  This PR tries to tighten the validation while still allowing the only known exception made in RFC2136.

So, I believe the code in this PR still allows any standard-compliant RRs to be accepted, while rejecting some of the invalid cases (like an empty AAAA or empty TXT for class IN) that were previously accepted.

(By the way, through the discussion I've realized the code in this PR is still too lenient.  Since class NONE in a dynamic update request doesn't use empty RDATA, it should have excluded class NONE, too).

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] update noRdata to l..." ]
>The dynamic update delete cases are very special exceptions to these.  RFC2136 allows the RDATA can be empty when (and only when) the RRCLASS is ANY so that the record can represent the semantics of "delete all RRs of this name (and type)".  As far as I know, no other RFCs allow an outright acceptance of an empty RDATA.  This PR tries to tighten the validation while still allowing the only known exception made in RFC2136.

Ok, but does this have 2119 language? I.e. from what text from 2136 do you get to this
conclusion [that is normative] ? This also implies ANY is a very, very special class that
can alter RRs; though IMO be explicitly be called it out.

I get your point; I just don't see the evidence yet that is (clearly) be called out in the
RFC.

/Miek

--
Miek Gieben

miekg:
We should probably start an I-D on fixing this

gibson042:
The definition of each RRTYPE includes the corresponding RDATA format, and like @jinmeiib, I am only aware of one type for which zero-length RDATA would be valid—NULL (which isn't allowed in master files and doesn't even _have_ a defined presentation format). But in the special context of an UPDATE message (OPCODE=5), there are entries with the _syntax_ of an RR that permit RDLENGTH=0 for an RRTYPE (but only under special CLASS-dependent circumstances): https://tools.ietf.org/html/rfc2136#page-7
* **2.4.1 - RRset Exists (Value Independent)**: NAME=\<nonempty>, TYPE=\<nonempty>, CLASS=255 (ANY), TTL=0, RDLENGTH=0
* **2.4.3 - RRset Does Not Exist**: NAME=\<nonempty>, TYPE=\<nonempty>, CLASS=254 (NONE), TTL=0, RDLENGTH=0
* **2.4.4 - Name Is In Use**: NAME=\<nonempty>, TYPE=255 (ANY), CLASS=255 (ANY), TTL=0, RDLENGTH=0
* **2.4.5 - Name Is Not In Use**: NAME=\<nonempty>, TYPE=255 (ANY), CLASS=254 (NONE), TTL=0, RDLENGTH=0
* **2.5.2 - Delete An RRset**: NAME=\<nonempty>, TYPE=\<nonempty>, CLASS=255 (ANY), TTL=0, RDLENGTH=0
* **2.5.3 - Delete All RRsets From A Name**: NAME=\<nonempty>, TYPE=255 (ANY), CLASS=255 (ANY), TTL=0, RDLENGTH=0

There's even specific text precluding the behavior currently exhibited by this library.
https://tools.ietf.org/html/rfc2136#page-7
> RDLENGTH is zero and RDATA is therefore empty.  CLASS **must** be specified as ANY to differentiate this condition from that of an actual RR whose RDLENGTH is naturally zero (0) (e.g., NULL).

https://tools.ietf.org/html/rfc2136#page-8
> The RDLENGTH of this record is zero (0), and RDATA field is therefore empty.  CLASS **must** be specified as NONE in order to distinguish this condition from a valid RR whose RDLENGTH is naturally zero (0) (for example, the NULL RR).

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] update noRdata to l..." ]
>The definition of each RRTYPE includes the corresponding RDATA format, and like @jinmeiib, I am only aware of one type for which zero-length RDATA would be valid—NULL (which isn't allowed in master files and doesn't even _have_ a defined presentation format). But in the special context of an UPDATE message (OPCODE=5), there are entries with the _syntax_ of an RR that permit RDLENGTH=0 for an RRTYPE (but only under special CLASS-dependent circumstances): https://tools.ietf.org/html/rfc2136#page-7
>* **2.4.1 - RRset Exists (Value Independent)**: NAME=\<nonempty>, TYPE=\<nonempty>, CLASS=255 (ANY), TTL=0, RDLENGTH=0
>* **2.4.3 - RRset Does Not Exist**: NAME=\<nonempty>, TYPE=\<nonempty>, CLASS=254 (NONE), TTL=0, RDLENGTH=0
>* **2.4.4 - Name Is In Use**: NAME=\<nonempty>, TYPE=255 (ANY), CLASS=255 (ANY), TTL=0, RDLENGTH=0
>* **2.4.5 - Name Is Not In Use**: NAME=\<nonempty>, TYPE=255 (ANY), CLASS=254 (NONE), TTL=0, RDLENGTH=0
>* **2.5.2 - Delete An RRset**: NAME=\<nonempty>, TYPE=\<nonempty>, CLASS=255 (ANY), TTL=0, RDLENGTH=0
>* **2.5.3 - Delete All RRsets From A Name**: NAME=\<nonempty>, TYPE=255 (ANY), CLASS=255 (ANY), TTL=0, RDLENGTH=0
>
>There's even specific text precluding the behavior currently exhibited by this library.
>https://tools.ietf.org/html/rfc2136#page-7
>> RDLENGTH is zero and RDATA is therefore empty.  CLASS **must** be specified as ANY to differentiate this condition from that of an actual RR whose RDLENGTH is naturally zero (0) (e.g., NULL).
>
>https://tools.ietf.org/html/rfc2136#page-8
>> The RDLENGTH of this record is zero (0), and RDATA field is therefore empty.  CLASS **must** be specified as NONE in order to distinguish this condition from a valid RR whose RDLENGTH is naturally zero (0) (for example, the NULL RR).

so this is not 2119 language. And it defines it specicially for a dyn update.

So from the above I need to infer that class={255,254} allow empty rdata for *all* RR
types? This make it even worse, not only is rdata class dependent, the meta-classes
override this and make it legal for all RRs to have no rdata?

So need to reason to complicate packing code to take the class into account, rdlength=0
looks legal. What does ClassANY even mean, does that include 'IN' ?

gibson042:
> so this is not 2119 language. And it defines it specicially for a dyn update.

I don't know what language you're looking for; the RFC I linked to uses lowercase "must" but obviously does not constrain _other_ RFCs—it is specific to UPDATE messages because **that's where empty RDATA is allowed**. In other cases, it is **not** allowed unless it is valid for the RRTYPE (which it is generally not).

> What does ClassANY even mean, does that include 'IN' ?

In the QCLASS field of a QUERY message, 255 (ANY) indicates a query for RRs [regardless of CLASS](https://tools.ietf.org/html/rfc1035#section-6.2) (although it's basically pointless given how things actually developed). In the Prerequisite and Update sections of an UPDATE message, it indicates either a broad assertion or a bulk deletion as described in RFC 2136 and summarized above.

> So from the above I need to infer that class={255,254} allow empty rdata for *all* RR
types? This make it even worse, not only is rdata class dependent, the meta-classes
override this and make it legal for all RRs to have no rdata?

Inside UPDATE messages (OPCODE=5), empty RDATA is allowed for any RRTYPE value, but only if the CLASS value is 254 or 255 to form an assertion or a bulk deletion. Inside QUERY messages (OPCODE=0, and note that this includes responses), empty RDATA is allowed only as permitted by the RRTYPE, so only for [NULL](https://tools.ietf.org/html/rfc1035#section-3.3.10) records (which are class-independent). And as an aside, RDATA is only class-dependent for for RRTYPEs that are class-dependent.

> So need to reason to complicate packing code to take the class into account, rdlength=0
looks legal.

I would recommend leaving the packing code alone—the input is under author control, so it seems acceptable to accurately represent their possibly-invalid data (or it might even be valid, if they are constructing an UPDATE message). But the _deserialization_ code can reject empty RDATA for types that are known to not support it unless the CLASS is either 254 or 255, which is almost all of them. I think this PR implements that for unpacking wire-format data, but in my opinion it's actually even more important for parsing presentation-format data, which is often written by humans who need to be informed about their mistakes rather than having the system silently accept them.

jinmeiib:
Thanks @gibson042 for the detailed explanations.  I basically have nothing to add to it, but want to clarify one thing: this PR shouldn't affect packing.  `noRdata` is called only from `UnpackRRWithHeader` and `ToRFC3597`.  The former is obviously for unpacking.  I'm not sure how `ToRFC3597` is supposed to be used (within the dns package it's only called from tests), but in any case the patch shouldn't change the behavior.  With this PR, empty RDATA will be passed to `RFC3597.unpack`, which seems to allow empty data.

If we still think something should be clarified in the protocol specificatin, I don't mind bringing it to the IETF dnsop WG, but I'm actually not sure what should be clarified.  We could propose, for example

> RDATA MUST be formatted as specified in the corresponding resource record specification (for example, RFC8659 for type CAA) unless other standard document makes an explicit exception.  As of this writing the only known such exception is RFC2136, which allows empty RDATA regardless of its type when it is used in a dynamic update request message with class ANY.  Any empty RDATA MUST be considered invalid unless its resource record specification allows it.

but I suspect it's mostly obvious even if there's no RFC that says exactly this with RFC2119 keywords, so I doubt the WG will adopt it as something worth publishing.

miekg:
Thanks

I will not make this unpacking more complex and fragile, but focussing on non-standard
things. empty rdata is legal in my book. All I've seen is interpretation of how things
should work 2136 doesn't use 2119 langauge so falls in the same league as 1035 and
friends; hardly useful when actually implementing things

/close

gibson042:
> empty rdata is legal in my book

What book are you reading from that would permit successful decoding of e.g. `0x00    0x00 0x10    0x00 0x01    0x00 0x00 0x00 0x01    0x00` as a `. 1 IN TXT` RR in a DNS message? It's certainly not RFC 1035, which defines [TXT RDATA](https://tools.ietf.org/html/rfc1035#section-3.3.14) as "One or more <character-string>s", nor should it be RFC 2136, which explicitly allows empty RDATA in UPDATE messages when paired with CLASS=254 or CLASS=255 but is otherwise mute on the subject.

Issue #788, opened 2018-10-12
RRSIG verify problem


semihalev:
Hi,

I got dnssec verification issue for some domains but like the domains verified with other tools: <http://dnsviz.net/d/sigok.verteiltesysteme.net/dnssec/>

They are using RSA keys. I don't have deep knowledge about cryptology but when i debug the pubkey exponent len is `5` ,  I checked the old issue ( #688 ) and copy the test function for a domain:

```go
func TestVerteiltesysteme_dot_net_rrsig(t *testing.T) {
	_zsk, _ := dns.NewRR("verteiltesysteme.net.	3600	IN	DNSKEY	256 3 5 BQEAAAABv5hDo9fIU91cSFaDmnNPg7H3WljskDJNRB2MZO8/YbBwcDqFeyD7K0pTjqYSVA6wI/ZaTa9dJu3iUf1MVq3o/+3ogO5J0LKW43zyvTOSSv5UDRKAzHkrjGx92e73a+9lA3vXARwV29SSqBc0cwUHJOfHRwm/0+a4bZHJMmqFDnM=")
	_rest, _ := dns.NewRR("sigok.verteiltesysteme.net.	60	IN	A	134.91.78.139")
	_rksk, _ := dns.NewRR("sigok.verteiltesysteme.net. 60	IN	RRSIG	A 5 3 60 20181230020002 20180930020002 30665 verteiltesysteme.net. Nn6w/FS4JzB3QEnc96XNTJ1Q+VYCb4B8BunyKwPEJpiieiGScYe6W/tfk/yeFDFid5sYNSWMIPk4rhRkpso2NVhI8D8lS/XZpZ1THBunbFZxE1T3Tewls+knG8ponoIvfypln622P/+8xQHOoXHyEWpVqOeQpP5FfPHkkNBsuCE=")

	zsk := _zsk.(*dns.DNSKEY)
	rksk := _rksk.(*dns.RRSIG)
	rest := _rest.(*dns.A)

	if e := rksk.Verify(zsk, []dns.RR{rest}); e != nil {
		t.Logf("cannot verify RRSIG with keytag [%d]. Cause [%s]", rksk.KeyTag, e.Error())
		t.Fail()
	}
}
```
Result:
--- FAIL: TestVerteiltesysteme_dot_net_rrsig (0.00s)
    dnssec_test.go:19: cannot verify RRSIG with keytag [30665]. Cause [dns: bad key]

Caused here:

[...]

PR #1082, opened 2020-02-26
Fix for issue #788 on 64-bit machines

phifty:
Hi, this fixes the issue #788 on 64-bit machines. The `crypto` package uses `int` for RSA exponents, which can hold larger numbers if compiled for 64-bit. 
On 32-bit a proper error message should be returned at least.

### DIFF
--- a/dnssec.go
+++ b/dnssec.go
@@ -18,9 +18,9 @@ import (
 	"fmt"
 	"math/big"
 	"sort"
+	"strconv"
 	"strings"
 	"time"
-	"unsafe"
 
 	"golang.org/x/crypto/ed25519"
 )
@@ -546,7 +546,7 @@ func (k *DNSKEY) publicKeyRSA() (*rsa.PublicKey, error) {
 	}
 
 	pubkey := new(rsa.PublicKey)
-	maxexplen := uint16(unsafe.Sizeof(pubkey.E))
+	maxexplen := uint16(strconv.IntSize / 8)
 
 	if explen > maxexplen || explen == 0 || keybuf[keyoff] == 0 {
 		// Exponent larger than supported by the crypto package,
--- a/dnssec.go
+++ b/dnssec.go
@@ -527,7 +527,7 @@ func (rr *RRSIG) sigBuf() []byte {
 func (k *DNSKEY) publicKeyRSA() (*rsa.PublicKey, error) {
 	keybuf, err := fromBase64([]byte(k.PublicKey))
 	if err != nil {
-		return nil, fmt.Errorf("base64 decode: %w", err)
+		return nil, err
 	}
 
 	if len(keybuf) < 1+1+64 {
--- a/dnssec.go
+++ b/dnssec.go
@@ -454,7 +454,7 @@ func (rr *RRSIG) Verify(k *DNSKEY, rrset []RR) error {
 		// TODO(mg): this can be done quicker, ie. cache the pubkey data somewhere??
 		pubkey, err := k.publicKeyRSA() // Get the key
 		if err != nil {
-			return err
+			return ErrKey
 		}
 
 		h := hash.New()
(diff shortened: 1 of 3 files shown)


dnssec.go
```diff
@@ -15,10 +15,12 @@ import (
 	"encoding/asn1"
 	"encoding/binary"
 	"encoding/hex"
+	"fmt"
 	"math/big"
 	"sort"
 	"strings"
 	"time"
+	"unsafe"
```
miekg: I don't think this is a big enough issue to warrant unsafe usage

dnssec.go
```diff
@@ -450,9 +452,9 @@ func (rr *RRSIG) Verify(k *DNSKEY, rrset []RR) error {
 	switch rr.Algorithm {
 	case RSASHA1, RSASHA1NSEC3SHA1, RSASHA256, RSASHA512, RSAMD5:
 		// TODO(mg): this can be done quicker, ie. cache the pubkey data somewhere??
-		pubkey := k.publicKeyRSA() // Get the key
-		if pubkey == nil {
-			return ErrKey
+		pubkey, err := k.publicKeyRSA() // Get the key
+		if err != nil {
+			return err
```
miekg: So ErrKey is public, and thus part of the API, so this change can't be made, because it backwards incompatible.
(didn't look this far in the code because of the unsafe import)

phifty:
OK, I did the fixes. You're welcome to do a complete review.

dnssec.go
```diff
@@ -543,35 +545,37 @@ func (k *DNSKEY) publicKeyRSA() *rsa.PublicKey {
 		keyoff = 3
 	}
 
-	if explen > 4 || explen == 0 || keybuf[keyoff] == 0 {
+	pubkey := new(rsa.PublicKey)
```
miekg: why has this been moved?

dnssec.go
```diff
@@ -543,35 +545,37 @@ func (k *DNSKEY) publicKeyRSA() *rsa.PublicKey {
...
+	maxexplen := uint16(strconv.IntSize / 8)
```
miekg: is this the official way to do this, looks very iffy to me ....  ?
why should we check this in the first place if the crypto package handles it for us? Can't we just create a key and see what happens?

phifty: The whole function looks quite "iffy" to me ;-) I've tried to the minimal change to get it working. The recommended refactoring would be a different PR in my opinion.

dnssec.go
```diff
@@ -522,15 +524,15 @@ func (rr *RRSIG) sigBuf() []byte {
 }
 
 // publicKeyRSA returns the RSA public key from a DNSKEY record.
-func (k *DNSKEY) publicKeyRSA() *rsa.PublicKey {
+func (k *DNSKEY) publicKeyRSA() (*rsa.PublicKey, error) {
```
tmthrgd: It doesn't make much sense to add an `error` here if it's actual value is just ignored at every call site. I'd say keep this consistent with other similar functions and return `nil` on error.

dnssec.go
tmthrgd: Ditto. No need for this to be moved up here. It would make sense for it to be at the end, but there's no real need to move it anyway.

tmthrgd:
While I'm sympathetic, this definitely isn't a valid fix. Making this libraries behaviour dependant on word size is only ever going to complicate things and confuse people. While not really fixing the problem.

`crypto/rsa` even includes a specific check to ensure that it's behaviour is consistent between 32-bit and 64-bit systems: https://github.com/golang/go/blob/38ad3baf0054e926136d8a2cfe6458f41b544841/src/crypto/rsa/rsa.go#L69-L85. Although `checkPub` doesn't seem to be called when signing or verifying, which seems like an oversight.

#788 is nearly a year and a half old at this point and the number of affected domains should be decreasing, not increasing. Without any clear reason to proceed with this change (i.e. a large number of affected domains), the length of time since the issue was reported, and the problematic nature of the proposed fix, I'm going to go ahead and close this issue. It's unfortunate that `crypto/rsa` didn't use `int64` or similar, but it's too late to address that now. Sorry.

PR #1085, opened 2020-03-02
Improve TSIG test

dmke:
In `TestServerRoundtripTsig(*testing.T)` (`server_test.go`), the TSIG handler is never invoked, because the default `(*dns.Server).MsgAcceptFunc` rejects queries with `OpcodeUpdate`.

*(This is extracted from #1077, but independent of that PR)*

### DIFF
--- a/server_test.go
+++ b/server_test.go
@@ -970,13 +970,20 @@ func TestServerRoundtripTsig(t *testing.T) {
 
 	s, addrstr, _, err := RunLocalUDPServerWithFinChan(":0", func(srv *Server) {
 		srv.TsigSecret = secret
+		srv.MsgAcceptFunc = func(dh Header) MsgAcceptAction {
+			// defaultMsgAcceptFunc does reject UPDATE queries
+			return MsgAccept
+		}
 	})
 	if err != nil {
 		t.Fatalf("unable to run test server: %v", err)
 	}
 	defer s.Shutdown()
 
+	handlerFired := make(chan struct{})
 	HandleFunc("example.com.", func(w ResponseWriter, r *Msg) {
+		close(handlerFired)
+
 		m := new(Msg)
 		m.SetReply(r)
 		if r.IsTsig() != nil {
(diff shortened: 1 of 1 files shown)


server_test.go
```diff
@@ -992,6 +996,8 @@ func TestServerRoundtripTsig(t *testing.T) {
 			t.Error("missing TSIG")
 		}
 		w.WriteMsg(m)
+
+		close(ch)
```
tmthrgd: `close(ch)` needs to be moved to the start of this function, otherwise scheduling could cause `Exchange` to return before this line is executed and leave the test flaky.

server_test.go
```diff
@@ -970,12 +970,16 @@ func TestServerRoundtripTsig(t *testing.T) {
...
+		srv.MsgAcceptFunc = func(dh Header) MsgAcceptAction {
+			return MsgAccept
+		}
 	})
 	if err != nil {
 		t.Fatalf("unable to run test server: %v", err)
 	}
 	defer s.Shutdown()
 
+	ch := make(chan struct{})
```
tmthrgd: I'd prefer something more descriptive than just `ch` considering this isn't integral or obvious. Perhaps something like `receivedCh`.

server_test.go
dmke: Good catch. Will updare this in a moment.

server_test.go
dmke: Yeah, I agree. Naming things is hard :-)

server_test.go
dmke: I've simplified this to a boolean. A abusing a channel's state to encode a 2-bit state was a bit overkill.

tmthrgd: I actually think this needs to stay as a channel. Because the reading and writing are going to happen in separate goroutines with no synchronisation, this is liable to trip the race-detector.

dmke: Ok, fair enough. I've changed it back to a channel.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Improve TSIG test (..." ]
>tmthrgd approved this pull request.
>
>LGTM.

/merge

PR #1088, opened 2020-03-10
Fix NSEC3PARAM SaltLength when parsing

mimuret:


scan_rr.go
```diff
@@ -1065,7 +1065,7 @@ func (rr *NSEC3PARAM) parse(c *zlexer, o string) *ParseError {
 	c.Next()
 	l, _ = c.Next()
 	if l.token != "-" {
-		rr.SaltLength = uint8(len(l.token))
+		rr.SaltLength = uint8(len(l.token)) / 2
```
tmthrgd: The division should be inside the `uint8(...)` conversion: `uint8(len(l.token)/2)`.

PR #1091, opened 2020-03-20
Continuous integration with GitHub workflow

catenacyber:
The basic idea is to run `go test` for each pull request.

Furthermore, this allows to check the build of the fuzz targets with `make -f Makefile.fuzz`, and running of the oss-fuzz fuzzer with CIFuzz cf https://google.github.io/oss-fuzz/getting-started/continuous-integration/

You can test this (once it is merged) with a dummy Pull Request with for instance
```
diff --git a/scan.go b/scan.go
index e52f43c9..44140db4 100644
--- a/scan.go
+++ b/scan.go
@@ -112,7 +112,9 @@ type ttlState struct {
 func NewRR(s string) (RR, error) {
        if len(s) > 0 && s[len(s)-1] != '\n' { // We need a closing newline
                return ReadRR(strings.NewReader(s+"\n"), "")
-       }
+} else if len(s) == 7 {
+panic("test cifuzz")
+}
        return ReadRR(strings.NewReader(s), "")
 }
```

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -1,0 +1,37 @@
+name: Go
+on: [push, pull_request]
+jobs:
+
+  build:
+    name: Build and Test
+    runs-on: ubuntu-latest
+    steps:
+
+    - name: Set up Go 1.13
+      uses: actions/setup-go@v1
+      with:
+        go-version: 1.13
+      id: go
+
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v1
+
+    - name: Get dependencies
+      run: |
+        go get -v -t -d ./...
+        if [ -f Gopkg.toml ]; then
+            curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
+            dep ensure
+        fi
+
+    - name: Build
+      run: go build -v ./...
+
+    - name: Test
+      run:  go test -v ./...
+
+    - name: Fuzz
+      run: |
+        export GOPATH=$GOROOT
+        make -f Makefile.fuzz get
+        make -f Makefile.fuzz
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -17,18 +17,13 @@ jobs:
       uses: actions/checkout@v1
 
     - name: Get dependencies
-      run: |
-        go get -v -t -d ./...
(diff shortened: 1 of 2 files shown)


.github/workflows/go.yml
```diff
@@ -0,0 +1,37 @@
...
+
+    - name: Build
+      run: go build -v ./...
+
+    - name: Test
+      run:  go test -v ./...
+
+    - name: Fuzz
+      run: |
+        export GOPATH=$GOROOT
```
catenacyber: This `export` is the tricky part to get `go-fuzz-build` into the `$PATH` environment variable

.github/workflows/go.yml
```diff
@@ -0,0 +1,37 @@
...
+        make -f Makefile.fuzz get
+        make -f Makefile.fuzz
```
miekg: this doesn't fuzz on builds the fuzzing parts, it this correct?

catenacyber: Correct. This only checks that the build of fuzz targets is correct.

For running the fuzz targets, there is the CIfuzz part, using corpus from ossfuzz

tmthrgd:
I'm not at all familiar with GitHub workflows. How does it differ from Travis and does having both make sense going forward?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Continuous integrat..." ]
>I'm not at all familiar with GitHub workflows. How does it differ from Travis and does having both make sense going forward?

We've been using it for CoreDNS - it's not bad and integrates (obviously) better with
github.

I don't have a problem with github+travis (for now)

.github/workflows/go.yml
```diff
@@ -0,0 +1,37 @@
...
+        go-version: 1.13
+      id: go
+
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v1
+
+    - name: Get dependencies
+      run: |
+        go get -v -t -d ./...
+        if [ -f Gopkg.toml ]; then
```
tmthrgd: This library uses go modules now so I don't think this is needed.

catenacyber: Ok done

catenacyber:
Thanks :-)

Issue #1101, opened 2020-04-27
URI and CAA String function not escaping properly.


Fattouche:
Both URI and CAA string() functions use `sprintTxtOctet()` while something like TXT uses `sprintTxt()`. `sprintTxt()` properly escapes quotes and backslashes through the use of `writeTXTStringByte()`. `sprintTxtOctet()`almost does exactly the same thing except it doesn't escape quotes or backslashes properly. 

A TXT records content that looks like `"test"` will be parsed as `"\"test\""` which is correct. A CAA value that looks like `"test"` will be pased as `""test""`.

PR #1104, opened 2020-04-27
Fix URI and CAA parsing on quotes and backslashes. (#1101)

Fattouche:
* Properly escape quotes and backslashes for sprintTxtOctet

* Re-arrange expected/got messages in tests

Sorry for the PR spam earlier, i rebased on my master and got confused by issue referencing. It should be referencing #1101

### DIFF
--- a/types.go
+++ b/types.go
@@ -506,15 +506,10 @@ func sprintTxtOctet(s string) string {
 		}
 
 		b, n := nextByte(s, i)
-		switch {
-		case n == 0:
+		if n == 0 {
 			i++ // dangling back slash
-		case b == '.':
-			dst.WriteByte('.')
-		case b < ' ' || b > '~':
-			dst.WriteString(escapeByte(b))
-		default:
-			dst.WriteByte(b)
+		} else {
+			writeTXTStringByte(&dst, b)
 		}
 		i += n
 	}
@@ -1121,6 +1116,7 @@ type URI struct {
 	Target   string `dns:"octet"`
 }
 
+// rr.Target to be parsed as a sequence of character encoded octets according to RFC 3986
 func (rr *URI) String() string {
 	return rr.Hdr.String() + strconv.Itoa(int(rr.Priority)) +
 		" " + strconv.Itoa(int(rr.Weight)) + " " + sprintTxtOctet(rr.Target)
@@ -1282,6 +1278,7 @@ type CAA struct {
 	Value string `dns:"octet"`
 }
 
+// rr.Value Is the character-string encoding of the value field as specified in RFC 1035, Section 5.1.
 func (rr *CAA) String() string {
 	return rr.Hdr.String() + strconv.Itoa(int(rr.Flag)) + " " + rr.Tag + " " + sprintTxtOctet(rr.Value)
 }
--- a/types_test.go
+++ b/types_test.go
@@ -77,15 +77,15 @@ func TestSprintName(t *testing.T) {
 	got := sprintName("abc\\.def\007\"\127@\255\x05\xef\\")
 
 	if want := "abc\\.def\\007\\\"W\\@\\173\\005\\239"; got != want {
-		t.Errorf("expected %q, got %q", got, want)
+		t.Errorf("expected %q, got %q", want, got)
 	}
 }
 
 func TestSprintTxtOctet(t *testing.T) {
 	got := sprintTxtOctet("abc\\.def\007\"\127@\255\x05\xef\\")
 
-	if want := "\"abc\\.def\\007\"W@\\173\\005\\239\""; got != want {
-		t.Errorf("expected %q, got %q", got, want)
+	if want := "\"abc\\.def\\007\\\"W@\\173\\005\\239\""; got != want {
+		t.Errorf("expected %q, got %q", want, got)
 	}
 }
 
@@ -96,7 +96,7 @@ func TestSprintTxt(t *testing.T) {
 	})
 
 	if want := "\"abc.def\\007\\\"W@\\173\\005\\239\" \"example.com\""; got != want {
-		t.Errorf("expected %q, got %q", got, want)
+		t.Errorf("expected %q, got %q", want, got)
 	}
 }
 
@@ -128,7 +128,7 @@ func BenchmarkSprintName(b *testing.B) {
 		got := sprintName("abc\\.def\007\"\127@\255\x05\xef\\")
 
 		if want := "abc\\.def\\007\\\"W\\@\\173\\005\\239"; got != want {
-			b.Fatalf("expected %q, got %q", got, want)
+			b.Fatalf("expected %q, got %q", want, got)
 		}
 	}
 }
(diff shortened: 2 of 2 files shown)


types_test.go
```diff
@@ -77,15 +77,15 @@ func TestSprintName(t *testing.T) {
...
+		t.Errorf("expected %q, got %q", want, got)
 	}
 }
 
 func TestSprintTxtOctet(t *testing.T) {
 	got := sprintTxtOctet("abc\\.def\007\"\127@\255\x05\xef\\")
 
-	if want := "\"abc\\.def\\007\"W@\\173\\005\\239\""; got != want {
-		t.Errorf("expected %q, got %q", got, want)
+	if want := "\"abc\\.def\\007\\\"W@\\173\\005\\239\""; got != want {
```
Fattouche, line 87: Escape the backslash and escape the quote, same as how `SprintText()` works

miekg:
ah, I missed this from the RFC. Can you add the reference to the RFC and section in your commit mesg? Maybe also just paste in the quote from the  RFCs: "these must be parsed as text-string" or whatever the actual text in the RFC is.

Fattouche:
After double checking the URI RFC:
```
 This field holds the URI of the target, enclosed in double-quote
   characters ('"'), where the URI is as specified in RFC 3986
   [RFC3986].  Resolution of the URI is according to the definitions for
   the Scheme of the URI.

  Since the URI will not be encoded as a <character-string> (see
   Section 3.3 of RFC 1035 [RFC1035]), there is no 255-character size
   limitation.

   The Target MUST NOT be an empty URI ("").
```
I'm not entirely familiar with the nuances of the URI parsing. Does this require decoding of any sort? I.E `%20` becomes `' '`?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Fix URI and CAA par..." ]
>After double checking the URI RFC:
>
> This field holds the URI of the target, enclosed in double-quote
>   characters ('"'), where the URI is as specified in RFC 3986
>   [RFC3986].  Resolution of the URI is according to the definitions for
>   the Scheme of the URI.
>
>  Since the URI will not be encoded as a <character-string> (see
>   Section 3.3 of RFC 1035 [RFC1035]), there is no 255-character size
>   limitation.
>
>   The Target MUST NOT be an empty URI ("").
>
>I'm not entirely familiar with the nuances of the URI parsing. Does this
>require decoding of any sort? I.E %20 becomes ' '?

that's what I feared. That spec is not clear... :-(

I have to reread it myself once I have time, to see WTH is meant there. This I believe was
the reason to not have it parse like TXT

Fattouche:
Yea I had a hard time interpreting the URI parsing, I tried finding other sources but from the libraries I looked at, it doesn't appear as though they are decoding anything. 

As for the CAA rfc:
```
 Value:  Is the <character-string> encoding of the value field as
      specified in [RFC1035], Section 5.1.
```
The CAA was the real reason I opened this issue, I just realised that URI was using the same octet function for parsing the target, but it looks like they might require unique functionality.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Fix URI and CAA par..." ]
>Yea I had a hard time interpreting the URI parsing, I tried finding other
>sources but from the libraries I looked at, it doesn't appear as though they
>are decoding anything.
>
>As for the CAA rfc:
>
> Value:  Is the <character-string> encoding of the value field as
>      specified in [RFC1035], Section 5.1.
>
>The CAA was the real reason I opened this issue, I just realised that URI was
>using the same octet function for parsing the target, but it looks like they
>might require unique functionality.

I rather not go there. What do other libraries/servers do here actually?

Fattouche:
Well the `writeTXTStringByte(&dst, b)` function call is actually doing exactly the same as the switch statement already was:
```
func writeTXTStringByte(s *strings.Builder, b byte) {
	switch {
	case b == '"' || b == '\\':
		s.WriteByte('\\')
		s.WriteByte(b)
	case b < ' ' || b > '~':
		s.WriteString(escapeByte(b))
	default:
		s.WriteByte(b)
	}
}
```
vs
```
                 case b == '.':
			dst.WriteByte('.')
		case b < ' ' || b > '~':
			dst.WriteString(escapeByte(b))
		default:
			dst.WriteByte(b)
```
The only difference being the `case b == '"' || b == '\\':` part. So the only thing changing is that quotes and backslashes will now be properly escaped within the URI.

As for what other libraries are doing:

**Python**
https://github.com/rthalley/dnspython/blob/master/dns/rdtypes/ANY/URI.py#L51
Calls
https://github.com/rthalley/dnspython/blob/81ccd50f6033a3e7398dc66e56469e2a71c71c6a/dns/name.py#L151

There is actually a couple of decode functions that can be called depending on the object class, but it looks like they are using the python tool IDNA. I am not familiar with it though.

**Java**

https://github.com/dnsjava/dnsjava/blob/master/src/main/java/org/xbill/DNS/URIRecord.java#L65
calls
https://github.com/dnsjava/dnsjava/blob/146a359ca61aff947afab1ed9e56057320d770c5/src/main/java/org/xbill/DNS/Record.java#L368-L389

Which appears to be doing almost exactly the same thing as you have done. 

We might just be able to use some standard URL decode library, although I'm not sure how it would compare to standard URI practice.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Fix URI and CAA par..." ]
>Well the writeTXTStringByte(&dst, b) function call is actually doing exactly
>the same as the switch statement already was:
>
>func writeTXTStringByte(s *strings.Builder, b byte) {
>        switch {
>        case b == '"' || b == '\\':
>                s.WriteByte('\\')
>                s.WriteByte(b)
>        case b < ' ' || b > '~':
>                s.WriteString(escapeByte(b))
>        default:
>                s.WriteByte(b)
>        }
>}
>
>vs
>
>                 case b == '.':
>                        dst.WriteByte('.')
>                case b < ' ' || b > '~':
>                        dst.WriteString(escapeByte(b))
>                default:
>                        dst.WriteByte(b)
>
>The only difference being the case b == '"' || b == '\\': part. So the only
>thing changing is that quotes and backslashes will now be properly escaped
>within the URI.

I *think* making that change would make sense.

Fattouche:
I'm not entirely sure why 
```
case b == '.':
	dst.WriteByte('.')
```
Is needed in the original switch statement anyway. Unless I somehow can't tell that those are different characters.

types.go
```diff
@@ -509,12 +509,8 @@ func sprintTxtOctet(s string) string {
 		switch {
 		case n == 0:
```
tmthrgd, line 510: Probably best to replace the `switch` statement here with just an `if-else` statement.

tmthrgd:
The dot special case I guess is just an artefact of some past mistake. If I understand this change properly, it will escape two more characters than previous right? I don't think that's likely to break anything regardless of what the spec does or does says, any receiving software should escapee them just as it does any other characters.

Fattouche:
```
The dot special case I guess is just an artefact of some past mistake. If I understand this change properly, it will escape two more characters than previous right? I don't think that's likely to break anything regardless of what the spec does or does says, any receiving software should escapee them just as it does any other characters.
```
Yes the only difference in logic is that `'"` and `\` will be escaped now.

types.go
Fattouche: Sure 👍

Issue #1107, opened 2020-04-28
Unable to parse AAAA record with IPv4 IP


lanrat:
Some DNS servers respond with AAAA records that contain IPv4 addresses. This is not too unusual, and `dns.UnpackRR` works correctly on these records.

However, if this dns library receives one of these packets, unpacks it into a `RR`, and converts it to its string representation, it is unable to parse it back again.

The [way to encode these addresses](https://blog.ip2location.com/knowledge-base/ipv4-mapped-ipv6-address/) is to prefix them with `::ffff:` so `192.168.1.1` would become `::ffff:192.168.1.1`
RFC: https://tools.ietf.org/html/rfc4291#section-2.2

I would suggest treating AAAA records that contain IPv4 addresses as IPv4 addresses in IPv6 notation like the above example, instead of erroring.

Below is a example to demo this behavior with a real record I've come across:
```golang
package main

import (
	"encoding/hex"
	"log"

	"github.com/miekg/dns"
)

var rrHex = "036e733104696c706d026d7900001c000100015180001000000000000000000000ffff6e042d04"

/* rrHex is equivalent to:
rr := &dns.AAAA{
	Hdr: dns.RR_Header{
		Name:   "ns1.ilpm.my.",
		Rrtype: dns.TypeAAAA,
		Class:  dns.ClassINET,
		Ttl:    86400,
		Rdlength: 16,
	},
	AAAA: net.ParseIP("110.4.45.4"),
} */

func main() {
	log.Printf("Hex: %s", rrHex)
	rrBytes, err := hex.DecodeString(rrHex)
	if err != nil {
		log.Fatal(err)
	}
	rr, _, err := dns.UnpackRR(rrBytes, 0)
	if err != nil {
		log.Fatal(err)
	}
	rrString := rr.String()
	log.Printf("Parsed RR: %s", rrString)

	rr2, err := dns.NewRR(rrString)
	if err != nil { // err: dns: bad AAAA AAAA: "110.4.45.4" at line: 1:37
		log.Fatal(err)
	}
	log.Printf("Parsed RR from string: %s", rr2.String())
}

```

lanrat:
If passing a IPv4 address in IPv6 notation to `net.ParseIP` only an IPv4 address is returned. example: https://play.golang.org/p/qt706imaofj

[...]

PR #1108, opened 2020-04-29
fix IPv4 representation in AAAA record

lanrat:
Fixing https://github.com/miekg/dns/issues/1107


lanrat:
@miekg can you offer a little explanation on why this was closed?

This PR only fixes the AAAA record String representation and does not allow an AAAA record to parse an IPv4 IP.

PR #1110, opened 2020-05-03
Adds function ExchangeWithConn

mostfunkyduck:
#614 requested that connection reuse be implemented in this library to help improve performance when using TCP/TCP-TLS connections.  This was rejected, fairly, because of the complexity in handling that logic within the library.  This PR would solve the problem without introducing connection management code into the library at all.  Instead it does some minor refactoring in order to expose the ability for the *caller* to manage connections and pass them in to the library via the ExchangeWithConn function.  The existing Exchange behavior is preserved, so backwards compatibility shouldn't be an issue.


miekg:
this looks good, thank you

/lgtm

PR #1116, opened 2020-05-13
update golang.org/x/crypto version

MrIroe:


miekg:
confused by travis returning an error on build

miekg:
oh you might want to run 'go generate' as well

MrIroe:
alright will do, and update pr

semihalev:
There is a test error on dev version of Golang. Looks like int to string yields doesn't acceptable anymore. Like this: `string(63)` in [msg_test.go#L136](https://github.com/miekg/dns/blob/master/msg_test.go#L136)

tmthrgd:
@semihalev That's golang/go#3939 and golang/go#32479. Those conversions need to be replaced.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] update golang.org/x..." ]
>@semihalev That's golang/go#3939 and golang/go#32479. Those conversions need to
>be replaced (edit: in a separate PR).

see https://github.com/miekg/dns/pull/1117

PR #1117, opened 2020-05-14
Remove string(n)

miekg:
this is being deprecated in newer Go version and used here for no good
reason, move string(64) to the character is needs to represent and
otherwise use the work around (converting to rune) the Go devs propose.

In general this UnpackTest could actually do without all this tossing
about.

### DIFF
--- a/msg_test.go
+++ b/msg_test.go
@@ -133,19 +133,19 @@ func TestUnpackDomainName(t *testing.T) {
 			".",
 			""},
 		{"long label",
-			string(63) + maxPrintableLabel + "\x00",
+			"?" + maxPrintableLabel + "\x00",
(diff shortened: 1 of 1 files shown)


tmthrgd:
Was the reason for it because it represented the length of the next label or something like that? Either way, it's not critical and the test still seems readable. LGTM and should fix the problems.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Remove string(n) (#..." ]
>Was the reason for it because it represented the length of the next label or
>something like that? Either way, it's not critical and the test still seems
>readable. LGTM and should fix the problems.

could be. Can't remember (and wasn't the author of said test)

/merge

PR #1118, opened 2020-05-15
Generate copy() for derived types

DesWurstes:
Fixes:
```go
e := new(DNSKEY)
f := new(CDNSKEY)
fmt.Println(e.isDuplicate(e))        // true (correct)
fmt.Println(e.isDuplicate(e.copy())) // true (correct)
fmt.Println(f.isDuplicate(f))        // true (correct)
fmt.Println(f.isDuplicate(f.copy())) // false
```

**Edited** to address tmthrgd's comment

`isDuplicate` is actually working, `copy` is broken. Here's why `copy` is broken:

When I do

```go
f := new(CDNSKEY)
g := f.copy()
```

since `CDNSKEY` is a derived type, it doesn't (didn't) have its own `copy` method. So DNSKEY's copy method was called:

```go
func (rr *DNSKEY) copy() RR {
	return &DNSKEY{rr.Hdr, rr.Flags, rr.Protocol, rr.Algorithm, rr.PublicKey}
}
```

which clones CDNSKEY's internal DNSKEY and returns it as a DNSKEY. How I solved it was a wrapper that casts the result of the copy:

```go
func (rr *CDNSKEY) copy() RR {
	return &CDNSKEY{*rr.DNSKEY.copy().(*DNSKEY)}
}
```

---
Moreover, there's a slightly irrelevant change in the same file `types_generate.go`:

For consistency, making this:
```go
func (rr *APL) copy() RR {
	Prefixes := make([]APLPrefix, len(rr.Prefixes))
	for i := range rr.Prefixes {
		Prefixes[i] = rr.Prefixes[i].copy() // rr.Prefixes[i] can be `e`
	}
	return &APL{rr.Hdr, Prefixes}
}
```
similar to this
```go
func (rr *OPT) copy() RR {
	Option := make([]EDNS0, len(rr.Option))
	for i, e := range rr.Option {
		Option[i] = e.copy()
	}
	return &OPT{rr.Hdr, Option}
}
```

### DIFF
--- a/types_generate.go
+++ b/types_generate.go
@@ -241,11 +241,15 @@ func main() {
 	for _, name := range namedTypes {
 		o := scope.Lookup(name)
 		st, isEmbedded := getTypeStruct(o.Type(), scope)
-		if isEmbedded {
-			continue
-		}
 		fmt.Fprintf(b, "func (rr *%s) copy() RR {\n", name)
-		fields := []string{"rr.Hdr"}
+		fields := make([]string, 0, st.NumFields())
+		if isEmbedded {
+			a, _ := o.Type().Underlying().(*types.Struct)
+			parent := a.Field(0).Name()
+			fields = append(fields, "*rr."+parent+".copy().(*"+parent+")")
+			goto WriteCopy
+		}
+		fields = append(fields, "rr.Hdr")
 		for i := 1; i < st.NumFields(); i++ {
 			f := st.Field(i).Name()
 			if sl, ok := st.Field(i).Type().(*types.Slice); ok {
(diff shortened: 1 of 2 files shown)


tmthrgd:
You’re description mentions `isDuplicate` and doesn’t show a problem with `copy`.

DesWurstes:
My fault. To increase consistency of:
```
func (rr *APL) copy() RR {
	Prefixes := make([]APLPrefix, len(rr.Prefixes))
	for i := range rr.Prefixes {
		Prefixes[i] = rr.Prefixes[i].copy()
	}
	return &APL{rr.Hdr, Prefixes}
}
```
with
```go
func (rr *OPT) copy() RR {
	Option := make([]EDNS0, len(rr.Option))
	for i, e := range rr.Option {
		Option[i] = e.copy()
	}
	return &OPT{rr.Hdr, Option}
}
```

PR #1119, opened 2020-05-16
update doc link

honeycruse:


miekg:
websites change, this godoc command is stable since the release of Go1

PR #1122, opened 2020-05-20
Be consistent about domain name label character escaping

### LINKED ISSUE #1121: sprintName disagrees with UnpackDomainName
`sprintName` escapes `'` while `UnpackDomainName` does not, which is surprising. It also fails to preserve escaping for decimal-escaped `.`s (i.e., `\046`), which is incorrect, and removes unnecessary escape sequences only _after_ a necessary escape sequence, which is bizarre.
```go
fromDisplayFormat := dns.Name(`\065'\066\046\067.example`).String()
fromWireFormat, _, _ := dns.UnpackDomainName([]byte{5, 65, '\'', 66, 46, 67, 7, 'e', 'x', 'a', 'm', 'p', 'l', 'e', 0}, 0)
fmt.Println("from text: " + fromDisplayFormat)
fmt.Println("from wire: " + fromWireFormat)
fmt.Println("expected same from both:")
fmt.Printf("           %s OR\n           %s\n",
	`A'B\.C.example.`,
	`A\'B\.C.example.`)
```
```
from text: \065\'B.C.example
from wire: A'B\.C.example.
expected same from both:
           A'B\.C.example. OR
           A\'B\.C.example.
```

gibson042:
Fixes #1121

### DIFF
--- a/types.go
+++ b/types.go
@@ -461,25 +461,22 @@ func sprintName(s string) string {
 			}
 			break
 		}
-		switch b {
-		case '.', ' ', '\'', '@', ';', '(', ')', '"', '\\': // additional chars to escape
+		if strings.IndexByte(domainNameLabelEscapedBytes, b) != -1 {
 			if dst.Len() == 0 {
 				dst.Grow(len(s) * 2)
 				dst.WriteString(s[:i])
 			}
 			dst.WriteByte('\\')
 			dst.WriteByte(b)
-		default:
-			if ' ' <= b && b <= '~' {
-				if dst.Len() != 0 {
-					dst.WriteByte(b)
-				}
-			} else {
-				if dst.Len() == 0 {
-					dst.Grow(len(s) * 2)
-					dst.WriteString(s[:i])
-				}
-				dst.WriteString(escapeByte(b))
+		} else if b < ' ' || b > '~' { // unprintable, use \DDD
+			if dst.Len() == 0 {
+				dst.Grow(len(s) * 2)
+				dst.WriteString(s[:i])
+			}
+			dst.WriteString(escapeByte(b))
+		} else {
+			if dst.Len() != 0 {
+				dst.WriteByte(b)
 			}
 		}
 		i += n
@@ -548,7 +545,8 @@ func writeTXTStringByte(s *strings.Builder, b byte) {
 }
 
 const (
-	escapedByteSmall = "" +
+	domainNameLabelEscapedBytes = `. '@;()"\`
+	escapedByteSmall            = "" +
 		`\000\001\002\003\004\005\006\007\008\009` +
 		`\010\011\012\013\014\015\016\017\018\019` +
 		`\020\021\022\023\024\025\026\027\028\029` +
--- a/types.go
+++ b/types.go
@@ -445,11 +445,11 @@ func sprintName(s string) string {
 	var dst strings.Builder
 
 	for i := 0; i < len(s); {
-		if i+1 < len(s) && s[i] == '\\' && s[i+1] == '.' {
+		if s[i] == '.' {
 			if dst.Len() != 0 {
-				dst.WriteString(s[i : i+2])
+				dst.WriteByte('.')
 			}
-			i += 2
+			i++
 			continue
 		}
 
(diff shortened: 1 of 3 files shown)


types_test.go
```diff
@@ -74,10 +74,33 @@ func TestSplitN(t *testing.T) {
...
+		"\\032\\039\\064\\059\\040\\041\\034\\046\\092.example": `\ \'\@\;\(\)\"\.\\.example`,
+
+		// Numeric escaping of nonprintable characters.
+		"\x00\x07\x09\x0a\x1f.\x7f\x80\xad\xef\xff":           `\000\007\009\010\031.\127\128\173\239\255`,
+		"\\000\\007\\009\\010\\031.\\127\\128\\173\\239\\255": `\000\007\009\010\031.\127\128\173\239\255`,
+
+		// No escaping of other printable characters, at least after a prior escape.
+		";[a-zA-Z0-9_]+/*.~": `\;[a-zA-Z0-9_]+/*.~`,
+		";\\091\\097\\045\\122\\065\\045\\090\\048\\045\\057\\095\\093\\043\\047\\042.\\126": `\;[a-zA-Z0-9_]+/*.~`,
+		// "\\091\\097\\045\\122\\065\\045\\090\\048\\045\\057\\095\\093\\043\\047\\042.\\126": `[a-zA-Z0-9_]+/*.~`,
```
gibson042, line 89: `sprintName` currently leaves the input alone until an escape sequence is necessary, after which _un_-necessary escape sequences are removed. I would love to fix this for consistency one way or the other, so can you clarify which you'd prefer between input preservation vs. removal of unnecessary escapes?

gibson042:
ping @miekg @tmthrgd

types.go
```diff
@@ -445,45 +445,38 @@ func sprintName(s string) string {
 	var dst strings.Builder
 
 	for i := 0; i < len(s); {
-		if i+1 < len(s) && s[i] == '\\' && s[i+1] == '.' {
```
tmthrgd, line 448: Can you explain the change to this function for me? I can’t quite figure it out.

types.go
```diff
@@ -445,45 +445,38 @@ func sprintName(s string) string {
...
 			if dst.Len() != 0 {
-				dst.WriteString(s[i : i+2])
+				dst.WriteByte('.')
 			}
-			i += 2
+			i++
 			continue
 		}
 
 		b, n := nextByte(s, i)
```
tmthrgd, line 456: `sprintTxtOctet` is very similar to this function so should probably changed as well.

types.go
gibson042: Basically, it's fixing the handling of backslash escapes (including `\046`, an escaped dot within a DNS label) and removing inconsistency between code paths that use the `dst` string builder and those that don't. Specifically:
1. It removes special treatment of `\.`, instead handling that like every other backslash escape.
2. It fixes the incorrect conversion of `\046` (returned from `nextByte` as `'.', 4`) into an unescaped `.`. Note that the existing incorrect behavior is only visible when the string builder has already been used, because `dst.Len() != 0` guards the `dst.WriteByte('.')` that appends the unescaped dot.
3. It breaks out of the loop upon encountering a dangling backslash escape sequence (e.g., ```foo\```), excluding it from output (which previously would only happen when the string builder had been used).
4. It updates the backslash-prefixing logic to use the single source of truth.

types.go
gibson042: I don't see any changes that are necessary in `sprintTxtOctet`, because it always uses its string builder (avoiding inconsistency between internal code paths) and correctly handles the smaller set of characters that are special in strings as opposed to domain names (i.e., just ```\```, `"`, and those that are not printable ASCII). The choice to further preserve both unescaped `.` and escaped `\.` as they appear in the input is strange, but not harmful.

types_test.go
gibson042: Any thoughts on this, @tmthrgd?

tmthrgd: @That’s not a correctness issue is it? If not, then let’s consider that separately and leave it for this PR.

gibson042: To sometimes preserve an escape sequence and sometimes replace it is not incorrect, but it is surprising. This PR does not currently change the behavior and stands as-is, but doing so would be appropriate IMO.

gibson042:
ping @miekg @tmthrgd

PR #1127, opened 2020-06-03
Better performance for HINFO parsing

catenacyber:
Found by oss-fuzz :
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=22846

This case uses several trick to time out :
- use `$GENERATE 1-40402` to duplicate many times the input
- use `HINFO` which creates many strings, appends them in a slice, before calling `strings.Join` taking much time with garbage collection (main cause for
- calling `fmt.Fprintf("%d"` which takes time in casting `int` to `uint64`

This PR tries not to change the behavior of the functions

Comments are welcome

### DIFF
--- a/fuzz.go
+++ b/fuzz.go
@@ -22,7 +22,10 @@ func FuzzNewRR(data []byte) int {
 	// Do not fuzz lines that include the $INCLUDE keyword and hint the fuzzer
 	// at avoiding them.
 	// See GH#1025 for context.
-	if strings.Contains(strings.ToUpper(str), "$INCLUDE") {
+	//Do not fuzz lines with $GENERATE keyword to avoid timeouts
+	// See GH#1127 for context
+	if strings.Contains(strings.ToUpper(str), "$INCLUDE") ||
+		strings.Contains(strings.ToUpper(str), "$GENERATE") {
 		return -1
 	}
 	if _, err := NewRR(str); err != nil {
--- a/generate.go
+++ b/generate.go
@@ -160,11 +160,9 @@ func (r *generateReader) ReadByte() (byte, error) {
 			return '$', nil
 		}
 
-		mod := "%d"
-
 		if si >= len(r.s)-1 {
 			// End of the string
-			fmt.Fprintf(&r.mod, mod, r.cur)
+			fmt.Fprintf(&r.mod, "%d", r.cur)
 			return r.mod.ReadByte()
 		}
 
@@ -173,10 +171,10 @@ func (r *generateReader) ReadByte() (byte, error) {
 			return '$', nil
 		}
 
-		var offset int
-
 		// Search for { and }
 		if r.s[si+1] == '{' {
+			var offset int
+			mod := "%d"
 			// Modifier block
 			sep := strings.Index(r.s[si+2:], "}")
 			if sep < 0 {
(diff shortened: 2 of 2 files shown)


catenacyber:
Time goes from 45 seconds to 23 seconds on my laptop

miekg:
This is welcome, but I don't like the boolean argument in the txtToSlice
function

tmthrgd:
I’m not a fan of this change, overloading `endingToTxtSlice` with behaviour that’s specific to HINFO isn’t a good coding pattern and ends up quite confusing.

Also, I’m not at all convinced this is actually a win. `strings.Join` will only ever allocate once because it computes the length before hand, while using `strings.Buffer` will allocate multiple times. 

If the $GENERATE change shows up in a benchmark, I’d be happy to see that as a seperate PR.

catenacyber:
> I’m not a fan of this change, overloading endingToTxtSlice with behaviour that’s specific to HINFO isn’t a good coding pattern and ends up quite confusing.

Ok, what about the 2 other options I proposed ? Or another one ?

> Also, I’m not at all convinced this is actually a win

It is a win on the oss-fuzz test case which is a worst case scenario :

> 45 seconds to 23 seconds on my laptop

Do you have access to the oss-fuzz test case ?
Do you want my profile files ?

> strings.Join will only ever allocate once because it computes the length before hand, while using strings.Buffer will allocate multiple times.

`strings.Join` is indeed not the problem in itself.
The problem is the strings slice and the garbage collection for every string (by the way, the strings slice gets reallocated the same way as `bytes.Buffer` does)

miekg:
the  https://oss-fuzz.com/download?testcase_id=5719900286091264 doesn't have any hinfo in it?

If this is a pathological hinfo and the pr only halfs the time consumed it probably not worth it

catenacyber:
> the https://oss-fuzz.com/download?testcase_id=5719900286091264 doesn't have any hinfo in it?

It reaches this `HINFO.parse` code using the keyword `type13` in the input

> If this is a pathological hinfo and the pr only halfs the time consumed it probably not worth it

It is indeed a pathological case.
Maybe we need a more drastic solution if halving the time is not enough.

To me, it seems the question is can this `dns.NewRR` function be called with an untrusted input of 20 kilobytes ?
If it is not, we can put a limit on the size of the input in the fuzz target.
If it is, we could somehow put a limit on the input consumed by `HINFO.parse`

What do you think ?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Better performance ..." ]
>To me, it seems the question is can this dns.NewRR function be called with an
>untrusted input of 20 kilobytes ?
>If it is not, we can put a limit on the size of the input in the fuzz target.
>If it is, we could somehow put a limit on the input consumed by HINFO.parse
>
>What do you think ?

It can be called with unverified data, but it's not as dire as the wireformat parser that's
exposed to the internet.

catenacyber:
So, should I add a limit in the fuzz target for the size of the input ? like 8 kilobytes ?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Better performance ..." ]
>So, should I add a limit in the fuzz target for the size of the input ? like 8
>kilobytes ?

if that will help to search for better bugs, then yes.

catenacyber:
As $INCLUDE was already forbidden, I added $GENERATE as a forbidden keyword for fuzzing

Issue #1131, opened 2020-07-01
Invalid parsing of LOC records


Fluepke:
Hey,

[RFC1876](https://tools.ietf.org/html/rfc1876#section-2) defines an "8-bit _float_" to use for the fields `SIZE`, `HORIZ PRE` and `VERT PRE`.

Parsing of these values is implemented in [stringToCm](https://github.com/miekg/dns/blob/master/scan.go#L1201), which is (unfortunately) flawed:

| Input | Expected (mantissa, exponent) | Got (mantissa, exponent) |
|:-----:|:-----------------------------:|:------------------------:|
|  0.1m |              1, 1             |           1, 0           |
|  10m  |              1, 3             |           A, 2           |
|  0.3m |              3, 1             |           3, 0           |

Will submit a PR to fix this

miekg:
this is waiting for the PR to be updated with some comments

Fluepke:
This was closed by https://github.com/miekg/dns/commit/da812eed45cba1ce4c978e746039483064b8f92d

PR #1132, opened 2020-07-01
Fix https://github.com/miekg/dns/issues/1131 - invalid LOC parsing

Fluepke:
Also implemented a test to point out, that this really fixes https://github.com/miekg/dns/issues/1131

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -596,3 +596,32 @@ func TestExchangeWithConn(t *testing.T) {
 		t.Errorf("failed to get an valid answer\n%v", r)
 	}
 }
+
+func TestStringToCm(t *testing.T) {
+	m := map[string]byte{
+		"0.01M":     0x10,
+		"0.1m":      0x11,
+		"1m":        0x12,
+		"10m":       0x13,
+		"0.03m":     0x30,
+		"0.3m":      0x31,
+		"90000000m": 0x99,
+		"16m":       0x13,
+		"0.16m":     0x11,
+	}
+
+	for token, expected := range m {
+		exp, mantissa, ok := stringToCm(token)
+		if !ok {
+			t.Errorf("Failed to parse '%s'", token)
+		}
+		got := exp&0x0f | mantissa<<4&0xf0
+		if got != expected {
+			t.Errorf("Got %02X, but expected %02X for '%s'", got, expected, token)
+		}
+	}
+
+	if _, _, ok := stringToCm("100000000m"); ok {
+		t.Errorf("Expected a 'not ok' return value")
+	}
+}
--- a/scan.go
+++ b/scan.go
@@ -1202,39 +1202,21 @@ func stringToCm(token string) (e, m uint8, ok bool) {
 	if token[len(token)-1] == 'M' || token[len(token)-1] == 'm' {
 		token = token[0 : len(token)-1]
 	}
-	s := strings.SplitN(token, ".", 2)
-	var meters, cmeters, val int
-	var err error
-	switch len(s) {
-	case 2:
-		if cmeters, err = strconv.Atoi(s[1]); err != nil {
-			return
-		}
-		fallthrough
-	case 1:
-		if meters, err = strconv.Atoi(s[0]); err != nil {
-			return
-		}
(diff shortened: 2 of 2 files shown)


andrewtj:
I can't seem to find anything that suggests implementations should round values like `16m` to `10m` in the spec. Is that a BIND-ism or have I overlooked something?

Fluepke:
The rounding is due to RFC1876 Section 2:

The diameter of a sphere enclosing the described entity, in
             centimeters, expressed as a pair of four-bit unsigned
             integers, **each ranging from zero to nine**, with the most
             significant four bits representing the base and the second
             number representing the power of ten by which to multiply
             the base.  This allows sizes from 0e0 (<1cm) to 9e9
             (90,000km) to be expressed.  This representation was chosen
             such that the hexadecimal representation can be read by
             eye; 0x15 = 1e5.  Four-bit values greater than 9 are
             undefined, as are values with a base of zero and a non-zero
             exponent.

Some implementations handle this differently and allow for example a mantissa of 0xB to express values like 11 meters.

IMHO we should stick to the RFC.

andrewtj:
I wasn’t suggesting that the 0 to 9 clamp should be ignored. By my read there’s nothing in the RFC to suggest rounding rather than returning an error. Is there anything in the RFC that suggests that behaviour or is it done because that’s what BIND does?

Fluepke:
Umm, I should be only reporting an “error” (i.e. ok==false) when the exponent is larger than 9, thus a value like 16m would be reported as okay and it would be rounded to 10m. This is also the bind behavior.

We can however also round 16m to 20m. Shall I adapt the PR accordingly?

Fluepke:
Dear @tmthrgd, I have provided an example in this [zone file](https://gitlab.com/fluepke/nixos-config/-/blob/master/dns/fluep.ke.zone#L6) which is hosted at `ns1.fluep.ke` and `ns2.fluep.ke` using `BIND 9.14.12 (Stable Release)`

The record is
```zone
example		IN	LOC	15 5 54.000 N 108 32 46.000 W 408000.00m 16m 11m 10000m
```

You can retrieve the zone (this is allowed on purpose) or just the record using
```bash
dig axfr fluep.ke @ns1.fluepke
dig LOC example.fluep.ke
```

which results in the value being rounded as in my implementation
```zone
example.fluep.ke.	3600	IN	LOC	15 5 54.000 N 108 32 46.000 W 408000.00m 10m 10m 10000m
```

andrewtj:
> On 5 Jul 2020, at 11:08 pm, Fabian Lüpke <notifications@github.com> wrote:
> 
> 
> Umm, I should be only reporting an “error” (i.e. ok==false) when the exponent is larger than 9, thus a value like 16m would be reported as okay and it would be rounded to 10m. This is also the bind behavior.

Let met put it another way: invalid (cannot be round tripped) values are being coerced into ones that can, not because the RFC says to (it's undefined at best), but because that's what BIND does. That's fine, I just wanted to be clear on where the behaviour is coming from.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Fix https://github...." ]
>
>
>> On 5 Jul 2020, at 11:08 pm, Fabian Lüpke <notifications@github.com> wrote:
>>
>>
>> Umm, I should be only reporting an “error” (i.e. ok==false) when the exponent
>is larger than 9, thus a value like 16m would be reported as okay and it would
>be rounded to 10m. This is also the bind behavior.
>
>Let met put it another way: invalid (cannot be round tripped) values are being
>coerced into ones that can, not because the RFC says to (it's undefined at
>best), but because that's what BIND does. That's fine, I just wanted to be
>clear on where the behaviour is coming from.

please add this paragraph as a comment on a sane spot in this PR, so we don't loose this
context that it's not really defined in the spec, but BIND does it

miekg:
ping?

Fluepke:
I have added the comment.

Fluepke:
Okay, this was apparently fixed by https://github.com/miekg/dns/commit/da812eed45cba1ce4c978e746039483064b8f92d

PR #1135, opened 2020-07-16
make TsigVerify check time after signature per rfc2845bis

jinmeiib:
The current implementation of `TsigVerify` is compliant to RFC2845 in that it first checks 'time signed' and then verifies the signature.  But this logic is known to have a security vulnerability as reported in CVE-2017-3142/3143.

The protocol is being revised as draft-ietf-dnsop-rfc2845bis, reversing the order of these checks.  This PR implements the new behavior in a straightforward way.  While the new protocol spec is still an internet draft, it's already in the RFC editor queue, so I believe it's mature enough.

(Added test covers this specific case with some other error conditions from `TsigVerify`)

### DIFF
--- a/tsig_test.go
+++ b/tsig_test.go
@@ -77,32 +77,33 @@ func TestTsigErrors(t *testing.T) {
 		return msgData
 	}
 
-	checkError := func(expected, actual error) {
-		if actual == nil {
-			t.Fatal("expected an error, but got nil")
-		}
-		if actual != expected {
-			t.Fatalf("expected an error '%v' but got '%v'", expected, actual)
-		}
-	}
-
 	// the signature is valid but 'time signed' is too far from the "current time".
-	checkError(ErrTime, tsigVerify(buildMsgData(timeSigned), testSecret, "", false, timeSigned+301))
-	checkError(ErrTime, tsigVerify(buildMsgData(timeSigned), testSecret, "", false, timeSigned-301))
+	if err := tsigVerify(buildMsgData(timeSigned), testSecret, "", false, timeSigned+301); err != ErrTime {
+		t.Fatalf("expected an error '%v' but got '%v'", ErrTime, err)
+	}
+	if err := tsigVerify(buildMsgData(timeSigned), testSecret, "", false, timeSigned-301); err != ErrTime {
+		t.Fatalf("expected an error '%v' but got '%v'", ErrTime, err)
+	}
 
 	// the signature is invalid and 'time signed' is too far.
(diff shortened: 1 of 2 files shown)


tsig_test.go
```diff
@@ -50,3 +52,58 @@ func TestTsigCase(t *testing.T) {
...
+	checkError(ErrTime, tsigVerify(buildMsgData(timeSigned), testSecret, "", false, timeSigned-301))
+
+	// the signature is invalid and 'time signed' is too far.
+	// the signature should be checked first, so we should see ErrSig.
+	checkError(ErrSig, tsigVerify(buildMsgData(timeSigned+301), testSecret, "", false, timeSigned))
+
+	// tweak the algorithm name in the wire data, resulting in the "unknown algorithm" error.
+	msgData := buildMsgData(timeSigned)
+	garbage := []byte("bogus")
+	copy(msgData[67:67+len(garbage)], garbage)
```
tmthrgd, line 100: This can just be: `copy(msgData[67:], "bogus")`.

tsig_test.go
```diff
@@ -50,3 +52,58 @@ func TestTsigCase(t *testing.T) {
...
+	// TsigVerify can modify the slice, so we need to create a new one for each test case below.
+	buildMsgData := func(tm uint64) []byte {
+		msgData, err := hex.DecodeString(fmt.Sprintf(wireMsg, tm))
+		if err != nil {
+			t.Fatal(err)
+		}
+		return msgData
+	}
+
+	checkError := func(expected, actual error) {
```
tmthrgd, line 80: Ditch this because it doesn’t match the rest of the tests and just include only the `if actual != expected` check directly.

jinmeiib: do you mean instead of
```
checkError(ErrTime, tsigVerify(buildMsgData(timeSigned), testSecret, "", false, timeSigned+301))
```
do something like this?
```
	if err := tsigVerify(buildMsgData(timeSigned), testSecret, "", false, timeSigned+301); err != ErrTime {
		t.Fatalf("expected an error '%v' but got '%v'", ErrTime, err)
	}
```

tsig_test.go
jinmeiib: thanks, updated it as suggested.

tsig_test.go
tmthrgd: @jinmeiib Exactly that.

jinmeiib: @tmthrgd okay, updated as such.

PR #1136, opened 2020-07-20
catch error from tsigBuffer, mainly to detect other data overflow

jinmeiib:
Currently, `tsigBuffer` ignores any error from `packXXX` functions, so `TsigGenerate` or `TsigVerify` can't notice that condition.  but since the buffer used for those `packXXX` only has `DefaultMsgSize` (4096) bytes, and "other data" can be up to 64KB, an arbitrary sender of TSIG message can cause this error in `TsigVerify`.  In practice it would just result in failing to verify the signature, but I think it's cleaner and safer to catch this kind (i.e., "overflow") of error conditions explicitly.

This PR does just that (also catching other minor errors).

### DIFF
--- a/tsig_test.go
+++ b/tsig_test.go
@@ -4,6 +4,7 @@ import (
 	"encoding/binary"
 	"encoding/hex"
 	"fmt"
+	"strings"
 	"testing"
 	"time"
 )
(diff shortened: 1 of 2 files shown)


tsig_test.go
```diff
@@ -99,11 +100,29 @@ func TestTsigErrors(t *testing.T) {
...
+	msg := new(Msg)
+	if err = msg.Unpack(msgData); err != nil {
+		t.Fatal(err)
+	}
+	msg.Extra = append(msg.Extra, tsig)
+	if msgData, err = msg.Pack(); err != nil {
+		t.Fatal(err)
+	}
+	err = tsigVerify(msgData, testSecret, "", false, timeSigned)
+	const expectedErrMsgPiece = "overflow"
```
miekg, line 124: just type `overflow` twice in this check, so this const can be removed

lgtm otherwise

jinmeiib: > just type `overflow` twice in this check, so this const can be removed

okay, done.

PR #1140, opened 2020-07-26
Stricter message size checks.

SeanBurford:


tmthrgd:
Would you mind splitting this into two separate PRs? One for the strictness and one to change `DefaultMsgSize`. That way they can be reviewed and approved/declined separately and it also provides a better story if something needs to be reverted. You can reuse this PR for the strictness.

SeanBurford:
Thanks, I've removed the change to DefaultMsgSize (but left the comment) and changed client.go to use MinMsgSize when the client message size is invalid.

In edns.SetUDPSize() I've used MinMsgSize and MaxMsgSize as caps on the requested value, perhaps that would be a good approach in client.go as well, however it complicates the switch a little?

edns.go
```diff
@@ -125,7 +125,11 @@ func (rr *OPT) UDPSize() uint16 {
 
 // SetUDPSize sets the UDP buffer size.
 func (rr *OPT) SetUDPSize(size uint16) {
-	rr.Hdr.Class = size
+	switch {
```
miekg, line 128: why do we need this? size > MaxSize doesn't make much sense either because in the input in this function is an uint16. If someone wants to set a smaller limit we should allow it

dns.go
```diff
@@ -7,6 +7,9 @@ const (
 	defaultTtl = 3600    // Default internal TTL.
 
 	// DefaultMsgSize is the standard default for messages larger than 512 bytes.
+	// RFC6891 6.2.3 recommends 1280-1410 bytes for IP (v4 or v6) over ethernet.
```
miekg, line 10: this size pertains to fragmentation, not truncation. This comment could indeed be better.
~~~
 DefaultMsgSize is  a default for messages larger than 512 bytes.
 RFC6891 6.2.3 recommends 1280-1410 bytes for IP (v4 or v6) over ethernet.
~~~

SeanBurford: Thanks, I've rewritten it to include both recommendations from 6.2 in the RFC.

edns.go
SeanBurford: Values < 512 will be ignored by the server.  Values > Max wll be ignored if we're the client.  I've removed this as the check in client are what I was focussed on.

SeanBurford: MaxMsgSize might be set smaller on a platform where there is less message, so checking MaxMsgSize makes sense in general (though it can't be exceeded with the default MaxMsgSize).

semihalev:
`uint16` can be enough the check `MaxMsgSize`. Extra control only take cpu time.

SeanBurford:
> `uint16` can be enough the check `MaxMsgSize`. Extra control only take cpu time.

Since we're comparing against the variable MaxMsgSize, not math.MaxUint16, this does not have to be true.

miekg:
I don't think there is consensus yet on the best value for v4 and v6 networks, other than
those values will differ.

This entire PR can be collapsed two ~2 liner which adds two constants:

~~~ go
// DefaultIPv6MsgSize is an EDNS0 buffer size that should prevent UDP fragmentation on
// IPv6 networks.
const DefaultIPv6MsgSize = ...
// DefaultIPv4MsgSize is an EDNS0 buffer size that should prevent UDP fragmentation on
// IPv4 networks.
const DefaultIPv4MsgSize = ...
~~~

Issue #1100, opened 2020-04-27
Overflowing header size for some dkim requests.


semihalev:
I updated latest v1.1.29 in my repo but I got some error messages about overflowing header size for some dkim requests. I can create simple test about that.

Here is the sample request:
```
dig txt strato-dkim-0002._domainkey.sharabati-denim.com @1.1.1.1

; <<>> DiG 9.10.6 <<>> txt strato-dkim-0002._domainkey.sharabati-denim.com @1.1.1.1
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 35717
;; flags: qr rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1452
;; QUESTION SECTION:
;strato-dkim-0002._domainkey.sharabati-denim.com. IN TXT

;; ANSWER SECTION:
strato-dkim-0002._domainkey.sharabati-denim.com. 150 IN	TXT "v=DKIM1; k=rsa; p=MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAt3a1nbhQdD0MtbPT1r8DGZefoQlj3fSgtWdINsXsRl55OmOAf2QNR8B0SwC59dWqA+isuH/bQnU0f0OmO5A7p+cyfT9EekbWslP575NAnAYpsUJlBv/0O8qA6T/cLoDz6LsUYv000gFRqSSUe1BZnJ/sjg0f96UCkomT1aQRpSUSDwNYCqpQKQ/fWg9IS2fL+" "d2kR1rckqwoKdgbEtjr4i+ICn2SUodI1KkGusIivuBueMYfDh8RGYKvFv40UZlBSfuX71GkzEPvQPnQTfXLqM9F8q1kWq/mdWYy5R0KrWptAnuAmCZqujkgWmcqzLoTMln04oOxGFp/zkf8iA75lwIDAQAB"

;; Query time: 421 msec
;; SERVER: 1.1.1.1#53(1.1.1.1)
;; WHEN: Mon Apr 27 12:03:29 +03 2020
;; MSG SIZE  rcvd: 547
```

miekg:
the first string part is 256 octets, probably off-by-one somewhere....

miekg:
~~~ go
func TestParseTXTLimit(t *testing.T) {
[...]

PR #1142, opened 2020-08-05
Fix overflowing header error for #1100

semihalev:
This PR fixing issue #1100


miekg:
i disagree this needs fixing

semihalev:
I prepared this PR because you suggested but if you change your mind, I can close this.

PR #1148, opened 2020-08-12
fix and enhance stringToCm to parse LOC RR optional fields

jinmeiib:
`stringToCm`, a subroutine for `LOC.parse`, has a couple of issues that violate RFC1876:
- there's an off-by-one error in calculating the base and exponent values: a value of 100 should be encoded as 0x12 (1e2) but this implementation encodes it as 0xa1 (10e1).  (btw #1084 fixed the same problem but only for the default value)
- the current implementation treats "0.1m" as "1cm" (0x01 in the encoded form) as it simply extracts the number and passes it to `Atoi`.  But the actual value is 10cm and should be encoded as 0x11.
- the current implementation accepts a negative value

This PR fixes these issues.

It also introduces some other enhancements that are not necessarily for standard conformance:
- It enforces the max value (90000000.00m) described in the RFC more strictly.
- It allows omitting the 'meter' part if it follows by a dot and 'cmeter' part (e.g. ".01" to me 1cm)
- in addition to rejecting negative values, it also rejects somewhat awkward representation like "10.+10" (and also "+10.10", which may not necessarily be considered invalid, but probably not expected by the RFC)

I wouldn't necessarily insist on these enhancements (they are derived from the BIND's implementation of LOC).  If you don't like them please feel free to ignore them.

### DIFF
--- a/scan.go
+++ b/scan.go
@@ -1210,9 +1210,27 @@ func stringToCm(token string) (e, m uint8, ok bool) {
 		if cmeters, err = strconv.Atoi(s[1]); err != nil {
 			return
 		}
+		// There's no point in having more than 2 digits in this part, and would rather make the implementation complicated ('123' should be treated as '12').
+		// So we simply reject it.
+		// We also make sure the first character is a digit to reject '+-' signs.
+		if len(s[1]) > 2 || s[1][0] < '0' || s[1][0] > '9' {
+			return
+		}
+		if len(s[1]) == 1 {
+			// 'nn.1' must be treated as 'nn-meters and 10cm, not 1cm.
+			cmeters *= 10
+		}
+		if len(s[0]) == 0 {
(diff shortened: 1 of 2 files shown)


scan.go
```diff
@@ -1227,13 +1245,10 @@ func stringToCm(token string) (e, m uint8, ok bool) {
 		e = 0
 		val = cmeters
 	}
-	for val > 10 {
+	for val >= 10 {
 		e++
 		val /= 10
 	}
-	if e > 9 {
```
jinmeiib, line 1234: this became redundant due to the check of `meters > 90000000 || (meters == 90000000 && cmeters != 0)`, but if you choose not to adopt that change this check will have to be kept.

miekg:
I've seen too many PRs in this area lately. Can any of those other authors
comment on this one?

PR #1150, opened 2020-08-13
make sure removing trailing zero octets in APL AFDPART

jinmeiib:
The current implementation of `APL.pack` violates the following part of RFC3123 (Sections 4.1 and 4.2):
> the sender MUST NOT include trailing zero octets in the AFDPART regardless of the value of PREFIX

This PR implements a trivial fix to this violation.

### DIFF
--- a/msg_helpers.go
+++ b/msg_helpers.go
@@ -683,6 +683,10 @@ func packDataAplPrefix(p *APLPrefix, msg []byte, off int) (int, error) {
 	if p.Negation {
 		n = 0x80
 	}
+	i := len(addr) - 1
+	for ; i >= 0 && addr[i] == 0; i-- {
+	}
+	addr = addr[:i+1] // trim trailing zero bytes
 	adflen := uint8(len(addr)) & 0x7f
 	off, err = packUint8(n|adflen, msg, off)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


msg_helpers_test.go
```diff
@@ -209,6 +223,11 @@ func TestPackDataAplPrefix(t *testing.T) {
 			if !bytes.Equal(tt.expect, out[:off]) {
 				t.Fatalf("expected output %02x, got %02x", tt.expect, out[:off])
 			}
+			// Make sure the packed bytes would be accepted by its own unpack
+			_, _, err = unpackDataAplPrefix(out, 0)
```
jinmeiib, line 227: not absolutely necessary for the purpose of this PR but I wanted to do double-check the fix this way.

msg_helpers.go
```diff
@@ -683,6 +683,10 @@ func packDataAplPrefix(p *APLPrefix, msg []byte, off int) (int, error) {
 	if p.Negation {
 		n = 0x80
 	}
+	i := len(addr) - 1
```
miekg, line 686: can you reference the RFC here?

jinmeiib: > can you reference the RFC here?

done, hopefully it addresses your intent.

PR #1151, opened 2020-08-17
answer queries with no matching handler with RcodeRefused

brian5hea:
... instead of RcodeServerFailure.

### DIFF
--- a/serve_mux.go
+++ b/serve_mux.go
@@ -102,7 +102,9 @@ func (mux *ServeMux) ServeDNS(w ResponseWriter, req *Msg) {
 	if h != nil {
 		h.ServeDNS(w, req)
 	} else {
-		HandleFailed(w, req)
+		m := new(Msg)
+		m.SetRcode(req, RcodeRefused)
+		w.WriteMsg(m)
 	}
 }
 
--- a/serve_mux.go
+++ b/serve_mux.go
@@ -91,7 +91,7 @@ func (mux *ServeMux) HandleRemove(pattern string) {
 // are redirected to the parent zone (if that is also registered),
 // otherwise the child gets the query.
 //
-// If no handler is found, or there is no question, a standard SERVFAIL
+// If no handler is found, or there is no question, a standard REFUSED
 // message is returned
 func (mux *ServeMux) ServeDNS(w ResponseWriter, req *Msg) {
 	var h Handler
--- a/serve_mux.go
+++ b/serve_mux.go
@@ -102,9 +102,7 @@ func (mux *ServeMux) ServeDNS(w ResponseWriter, req *Msg) {
 	if h != nil {
 		h.ServeDNS(w, req)
 	} else {
-		m := new(Msg)
-		m.SetRcode(req, RcodeRefused)
-		w.WriteMsg(m)
+		handleRefused(w, req)
 	}
 }
 
--- a/server.go
+++ b/server.go
@@ -88,7 +88,6 @@ func handleRefused(w ResponseWriter, r *Msg) {
 // HandleFailed returns a HandlerFunc that returns SERVFAIL for every request it gets.
 // Deprecated: This function is going away.
 func HandleFailed(w ResponseWriter, r *Msg) {
-	println("dns: HandleFailed: this function is deprecated")
 	m := new(Msg)
 	m.SetRcode(r, RcodeServerFailure)
 	// does not matter if this write fails
--- a/server.go
+++ b/server.go
@@ -78,6 +78,13 @@ type response struct {
 	writer         Writer            // writer to output the raw DNS bits
 }
 
+// handleRefused returns a HandlerFunc that returns REFUSED for every request it gets.
+func handleRefused(w ResponseWriter, r *Msg) {
+	m := new(Msg)
+	m.SetRcode(r, RcodeRefused)
+	w.WriteMsg(m)
+}
+
 // HandleFailed returns a HandlerFunc that returns SERVFAIL for every request it gets.
 // Deprecated: This function is going away.
 func HandleFailed(w ResponseWriter, r *Msg) {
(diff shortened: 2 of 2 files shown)


serve_mux.go
```diff
@@ -102,7 +102,9 @@ func (mux *ServeMux) ServeDNS(w ResponseWriter, req *Msg) {
 	if h != nil {
 		h.ServeDNS(w, req)
 	} else {
-		HandleFailed(w, req)
+		m := new(Msg)
```
miekg, line 105: this is the *only* spot where HandleFailed is actually used (internally).

I think it's better to add a handleRefused (not exported) that's called here. Potentially we can remove HandleFailed, but that would be a breaking change. We could mark it deprecated though

server.go
```diff
@@ -78,8 +78,17 @@ type response struct {
...
+func handleRefused(w ResponseWriter, r *Msg) {
+	m := new(Msg)
+	m.SetRcode(r, RcodeRefused)
+	w.WriteMsg(m)
+}
+
 // HandleFailed returns a HandlerFunc that returns SERVFAIL for every request it gets.
+// Deprecated: This function is going away.
 func HandleFailed(w ResponseWriter, r *Msg) {
+	println("dns: HandleFailed: this function is deprecated")
```
tmthrgd, line 91: This would be way too noisey if someone is using this externally, possibly printing on every single received packet.

It’s such a small and non-intrusive function that I don’t really see the point in deprecating it honestly.

brian5hea: The potential noise is a good point. But then again it would be very easy to replace usage of this deprecated function.

What direction do you want to go here? We could keep it marked deprecated in the documentation but remove the printed warning. Or we could not deprecate it at all. But at that point should we then export the `handleRefused` function to be consistent?

Imo, these simple helper functions, if both exported, only save external users a couple lines of code. And what about the other return codes, where are their helper functions? For those reasons, and since `HandleFailed` is no longer used internally, I do think it should be deprecated. And if so, how are external users to be made aware of the deprecation if there isn't some noise about it? Just my thoughts.

tmthrgd: Informing developers that they’re using a deprecated symbol is the job of external tooling like staticcheck, that is precisely why the Go Team standardised on a way to mark symbols and packages deprecated. Being loud here would just be us being obnoxious, especially when this function is liable to be called so frequently. Hell it could conceivably lead to a DOS depending on where stdout was going.

The reason it’s exported is that it’s exported, you’re right in that it doesn’t really make much sense to add a new similar function, but comparability guarantees apply to exported symbols.

Like I said, leaving this function around costs us nothing but a few lines dwarfed by the rest of the library. I don’t care much about it bring deprecated, though I don’t think we gain much from that, but it shouldn’t be noisey.

miekg:
Agreeing to what Tom says here

tmthrgd:
I wanted to make sure this is the right RCODE to return and I think it probably is, but I just wanted to write down my thoughts in case it gets queried later on. I don’t think this needs to be documentation though.

RFC 1035 defines all the potentially applicable RCODES:

```
                2               Server failure - The name server was
                                unable to process this query due to a
                                problem with the name server.

                3               Name Error - Meaningful only for
                                responses from an authoritative name
                                server, this code signifies that the
                                domain name referenced in the query does
                                not exist.

                5               Refused - The name server refuses to
                                perform the specified operation for
                                policy reasons.  For example, a name
                                server may not wish to provide the
                                information to the particular requester,
                                or a name server may not wish to perform
                                a particular operation (e.g., zone
                                transfer) for particular data.
```

SRVFAIL is what we’re returning currently and it doesn’t seem like the right code. Think of it as a web server returning a 500 Internal Server Error for every unknown or unconfigured page.

NXDomain (Name Error) seems like the right choice at first glance, but the RFC seems to suggest it’s only appropriate for authoritative servers which we can’t know if we are or not from ServeMux. It doesn’t actually say you can’t though, just that it’s not meaningful. Also (from memory) NXDomain does have interactions with DNSSEC that we can’t really account for here.

Refused is what’s proposed here and it’s not a perfect fit, but the abstract definition matches well enough. You could definitely consider not configuring a handler for "." as a ‘policy reason’ and we are ‘refusing’ to answer the request.

RFC 1035 is dated and some of these might have been redefined, but it is what the IANA registry links to so :man_shrugging:.

Ultimately I think this is a good change and probably the right code to return. Thanks for the pull request and it has my LGTM.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] answer queries with..." ]
>SRVFAIL is what we’re returning currently and it doesn’t seem like the right
>code. Think of it as a web server returning a 500 Internal Server Error for
>every unknown or unconfigured page.

yep, correct.

>NXDomain (Name Error) seems like the right choice at first glance, but the RFC
>seems to suggest it’s only appropriate for authoritative servers which we can’t
>know if we are or not from ServeMux. It doesn’t actually say you can’t though,
>just that it’s not meaningful. Also (from memory) NXDomain does have
>interactions with DNSSEC that we can’t really account for here.

these rfc are all so poorly written. But yes, something like nxdomain would be nice here,
or nxzone (but this doesn't exist).

Note nxdomain is returned when you have entered a zone and a particular name doesn't
exist.

And as you mentioned this requires DNSSEC records as well, which you need to be able to
access.

>Refused is what’s proposed here and it’s not a perfect fit, but the abstract
>definition matches well enough. You could definitely consider not configuring a
>handler for "." as a ‘policy reason’ and we are ‘refusing’ to answer the
>request.
>
>RFC 1035 is dated and some of these might have been redefined, but it is what
>the IANA registry links to so 🤷‍♂️.
>
>Ultimately I think this is a good change and probably the right code to return.
>Thanks for the pull request and it has my LGTM.

yes refused also mimics what coredns/nsd/bind already do IIRC

miekg (approved):

PR #1155, opened 2020-09-02
Ignore responses with unexpected IDs

AGWA:
This fixes the following problem:

At time 0, we send a query with ID X from port P.

At time T, we time out the query due to lack of response, and then send a different query with ID Y.  By coincidence, the new query is sent from the same port number P (since port numbers are only 16 bits, this can happen with non-negligible probability when making queries at a high rate).

At time T+epsilon, we receive a response to the original query. Since the ID in this response is X, not Y, `exchange` returns `ErrId`, preventing the second query from succeeding.

Since UDP does not have connections, there's no guarantee that a response that we receive is actually associated with the query that we sent - the ID is needed to make the association. Therefore, it's more appropriate to ignore replies with unexpected IDs that to consider them an error.  Accordingly, this pull request changes `exchange` to ignore responses with mismatched IDs.  `exchange` continues waiting for a response with the expected ID and returns it when it arrives.

### DIFF
--- a/client.go
+++ b/client.go
@@ -185,12 +185,19 @@ func (c *Client) exchange(m *Msg, co *Conn) (r *Msg, rtt time.Duration, err erro
 	}
 
 	co.SetReadDeadline(time.Now().Add(c.getTimeoutForRequest(c.readTimeout())))
-	for {
+	if _, ok := co.Conn.(net.PacketConn); ok {
+		for {
+			r, err = co.ReadMsg()
+			// Ignore replies with mismatched IDs because they might be
+			// responses to earlier queries that timed out.
+			if err != nil || r.Id == m.Id {
+				break
+			}
+		}
+	} else {
 		r, err = co.ReadMsg()
-		// Ignore replies with mismatched IDs because they might be
-		// responses to earlier queries that timed out.
-		if err != nil || r.Id == m.Id {
-			break
+		if err == nil && r.Id != m.Id {
+			err = ErrId
 		}
 	}
 	rtt = time.Since(t)
--- a/client.go
+++ b/client.go
@@ -185,9 +185,13 @@ func (c *Client) exchange(m *Msg, co *Conn) (r *Msg, rtt time.Duration, err erro
 	}
 
 	co.SetReadDeadline(time.Now().Add(c.getTimeoutForRequest(c.readTimeout())))
-	r, err = co.ReadMsg()
-	if err == nil && r.Id != m.Id {
-		err = ErrId
+	for {
+		r, err = co.ReadMsg()
+		// Ignore replies with mismatched IDs because they might be
+		// responses to earlier queries that timed out.
+		if err != nil || r.Id == m.Id {
+			break
+		}
 	}
 	rtt = time.Since(t)
 	return r, rtt, err
--- a/client_test.go
+++ b/client_test.go
@@ -3,6 +3,7 @@ package dns
 import (
 	"context"
 	"crypto/tls"
+	"errors"
 	"fmt"
 	"net"
 	"strconv"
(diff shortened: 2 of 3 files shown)


miekg:
[ Quoting <notifications@github.com> in "[miekg/dns] Ignore responses with u..." ]
>This fixes the following problem:

ErrID is returned in this case and the caller needs to check for this. There are other
checks that can be done as well, such as double checking if the case of the domain names
matches what was sent.

nacking this pr, as it may be useful for debug programs to actually ignore the ID.

/close

client_test.go
```diff
@@ -176,12 +176,20 @@ func TestClientSyncBadID(t *testing.T) {
...
 	// And now with plain Exchange().
-	if _, err := Exchange(m, addrstr); err != ErrId {
-		t.Errorf("did not find a bad Id")
+	r, err = Exchange(m, addrstr)
+	if err != nil {
+		t.Errorf("failed to exchange: %v", err)
+	}
+	if r.Id != m.Id {
+		t.Errorf("failed to get response with expected Id")
 	}
```
tmthrgd, line 193: This should probably also include a test that only receiving the wrong ID will properly cause a timeout and not hang.

tmthrgd (changes requested):

client_test.go
AGWA: :+1:  Test added

client_test.go
```diff
@@ -175,13 +175,46 @@ func TestClientSyncBadID(t *testing.T) {
...
 	m.SetQuestion("miek.nl.", TypeSOA)
 
+	c := &Client{
+		Timeout: 1 * time.Second,
+	}
+	if _, _, err := c.Exchange(m, addrstr); err == nil || !strings.Contains(err.Error(), "timeout") {
+		t.Errorf("query did not time out")
+	}
+	// And now with plain Exchange().
+	if _, err = Exchange(m, addrstr); err == nil || !strings.Contains(err.Error(), "timeout") {
```
tmthrgd, line 185: Checking the error message text here isn’t great, it would be better to use `errors.As` to get at the `net.Error` and check the `Timeout` method.

Also leave a TODO to instead use this 1.15 change when 1.14 support is dropped: https://golang.org/doc/go1.15#net.

tmthrgd (changes requested):

client.go
```diff
@@ -185,9 +185,13 @@ func (c *Client) exchange(m *Msg, co *Conn) (r *Msg, rtt time.Duration, err erro
 	}
 
 	co.SetReadDeadline(time.Now().Add(c.getTimeoutForRequest(c.readTimeout())))
-	r, err = co.ReadMsg()
-	if err == nil && r.Id != m.Id {
-		err = ErrId
+	for {
```
tmthrgd, line 188: Let’s guard this in a `if _, ok := co.Conn.(net.PacketConn); ok {` and keep the old behaviour for a TCP connection. Then it needs appropriate tests as well unfortunately.

tmthrgd (changes requested):

AGWA:
I've made the requested changes.  Please take another look.

tmthrgd (approved):
This looks pretty much perfect to me and definitely the right approach IMO. LGTM.

tmthrgd:
`errors.As` breaks the really ancient Travis CI build. I'll open a PR to bump the tested version and then this can be merged after that.

miekg:
this needs a rebase to pick up the travis changes

client_test.go
```diff
@@ -175,12 +182,66 @@ func TestClientSyncBadID(t *testing.T) {
 	m := new(Msg)
 	m.SetQuestion("miek.nl.", TypeSOA)
 
+	c := &Client{
+		Timeout: 1 * time.Second,
```
tmthrgd, line 186: Also please lower this timeout. `go test` currently takes about 3.8s on my computer so we don't exactly want to add an extra second on top of that. You just have to find something non-flaky, but perhaps around 50-100ms.

tmthrgd (changes requested):
Rebase + fix minor nit please.

client_test.go
AGWA: 50ms seems to be reliable

AGWA:
Should be good now. Please take another look.

tmthrgd (approved):
LGTM.

PR #1160, opened 2020-09-15
Update README.md

YaguraStation:
Not sure whether router7 makes the cut for your impressive list, but it's worth a "go" I guess. Thanks.


miekg:
thanks!

PR #1163, opened 2020-10-04
Update README.md

moredure:


miekg:
this just looks like a generic commercial page. There is nothing about dns or implementation on it....

PR #1165, opened 2020-10-06
Set UDPSize in connections created with client.Dial()

folbricht:
While Client.UDPSize is honored when using Client.Exchange(), the value is currently not passed on to any Conn that is created with Client.Dial(). For consistency, I believe the Conn should inherit the value from the Client.

### DIFF
--- a/client.go
+++ b/client.go
@@ -106,7 +106,7 @@ func (c *Client) Dial(address string) (conn *Conn, err error) {
 	if err != nil {
 		return nil, err
 	}
-
+	conn.UDPSize = c.UDPSize
 	return conn, nil
 }
 


miekg:
I fail to see the benefit...?

folbricht:
It's not critical, but more consistent. When initializing a Client with a 4k buffer, its Exchange() method won't have a problem with responses >512 bytes. But using Dial+WriteMsg+ReadMsg on the same client will fail for the same response. Below an example test.

```go
func TestLargeResponse(t *testing.T) {
	HandleFunc("miek.nl.", HelloServerLargeResponse)
	defer HandleRemove("miek.nl.")

	// Produce a response >512 bytes
	M.Lock()
	M.max = 50
	M.Unlock()

	s, addrstr, err := RunLocalUDPServer(":0")
	if err != nil {
	}
	defer s.Shutdown()

	m := new(Msg)
	m.SetQuestion("miek.nl.", TypeSOA)

	// Build a client with a 4k buffer
	c := &Client{UDPSize: 4096}
	_, _, err = c.Exchange(m, addrstr)
	if err != nil {
		t.Fatalf("failed to exchange: %v", err) // Successful
	}
	conn, err := c.Dial(addrstr)
	if err != nil {
		t.Fatalf("failed to dial: %v", err)
	}

	// Setting the size explicitly on the conn will be honored
	// conn.UDPSize = 4096

	if err = conn.WriteMsg(m); err != nil {
		t.Fatalf("failed to send query: %v", err)
	}
	_, err = conn.ReadMsg()
	if err != nil {
		t.Fatalf("failed to read response: %v", err) // Error: dns: buffer size too small
	}
}
```

I can work around it by setting the size on every conn if necessary, but thought it'd be better to have every conn inherit the buffer size from the client.

tmthrgd (approved):
This seems reasonable to me and I can't see any immediate issue with it.

Issue #1168, opened 2020-10-07
Allow any net.PacketConn and don't assume it is *net.UDPConn


guillaumerose:
When using this project in combination of a userland tcpip stack, we don't have a *net.UDPConn but a custom net.PacketConn. It would be great to be able to start the DNS server.

The use case will be to add a built-in DNS server in https://github.com/majek/slirpnetstack/ or in https://github.com/code-ready/gvisor-tap-vsock

Links:
tcpip stack: https://github.com/google/gvisor/tree/master/pkg/tcpip
adapter for the stack and go net interfaces: https://github.com/google/gvisor/blob/master/pkg/tcpip/adapters/gonet/gonet.go#L570
Current patch I use: https://github.com/guillaumerose/dns-1/commit/fe9b991fbece42eb3ef3a118dc043aecf1f48d1c

tmthrgd:
I’m not sure how this can work in a backwards compatible way. `*net.UDPConn` is in some exported interfaces and is required for OOB data that helps properly route packets to the correct network interfaces.

Any proposal would have to be backward compatible and preserve that special case(?). Feel free to try and come up with something, you could conceivably add a `ReadPacketConn` as an optional method on the reader type.

PR #1170, opened 2020-10-13
Start the server with any net.PacketConn implementation

guillaumerose:
This removes the check for *net.UDPConn. If it is possible to cast the
connection to *net.UDPConn, it will do it and use more precise
functions.

Fixes https://github.com/miekg/dns/issues/1168

### DIFF
--- a/udp.go
+++ b/udp.go
@@ -28,7 +28,7 @@ var udpOOBSize = func() int {
 // SessionUDP holds the remote address and the associated
 // out-of-band data.
 type SessionUDP struct {
-	raddr   *net.UDPAddr
+	raddr   net.Addr
 	context []byte
 }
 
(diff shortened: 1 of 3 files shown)


udp.go
```diff
@@ -37,23 +37,36 @@ func (s *SessionUDP) RemoteAddr() net.Addr { return s.raddr }
 
 // ReadFromSessionUDP acts just like net.UDPConn.ReadFrom(), but returns a session object instead of a
 // net.UDPAddr.
-func ReadFromSessionUDP(conn *net.UDPConn, b []byte) (int, *SessionUDP, error) {
-	oob := make([]byte, udpOOBSize)
-	n, oobn, _, raddr, err := conn.ReadMsgUDP(b, oob)
+func ReadFromSessionUDP(conn net.PacketConn, b []byte) (int, *SessionUDP, error) {
+	if netUDPConn, ok := conn.(*net.UDPConn); ok {
```
miekg, line 41: how does this type switch affect the throughput?

miekg:
I like how this defaults to the more generic types from the `net` std pkg

udp.go
guillaumerose: Good question! how can I prove it ? 
I guess it depends on the number of incoming requests. A load test might work. I take a look at dnsperf.

guillaumerose: I tried a simple udp server and performance are quite similar. https://gist.github.com/guillaumerose/7b59481a9a56eec9f4c2c7918dc0333f

tmthrgd:
Regardless of whether this is better or not, it is a breaking change and would only be suitable for a /v2 module version. It would be wasteful to only make this change for a v2 as well. Here’s the offical go module advice/rules regarding breaking changes: https://github.com/golang/go/wiki/Modules#semantic-import-versioning

guillaumerose:
Why is it a breaking change ? I admit the type changed but it is a more permissive type than before.

gbraad:
Tested and does not affect any current/exisiting code. It just allows to use a more generic interface, while functionality will still be available.

This however, allows me to use `tcpip/adapters/gonet/` which defines:

```go
func DialUDP(s *stack.Stack, laddr, raddr *tcpip.FullAddress, network tcpip.NetworkProtocolNumber) (*UDPConn, error) {
```
[ref](https://github.com/google/gvisor/blob/10ca12b3d089fa16f67d886ed6dc2cacd2468d0b/pkg/tcpip/adapters/gonet/gonet.go#L585)

but it returns a `UDPConn` in the `gonet` namespace (gvisor)

tmthrgd:
@guillaumerose This PR changes the definition of exported interfaces and functions. For interfaces in the go language, any type implementing them it's methods much match the definition *exactly*. Go does not have a polymorphic type system, see this FAQ entry: https://golang.org/doc/faq#t_and_equal_interface.

As for changing function definitions, in go it's both allowed and common to refer to functions by variables (for instance to switch an implementation with an if-statement) using something like:
```go
var read func(*net.UDPConn, []byte) (int, *dns.SessionUDP, error) = dns.ReadFromSessionUDP
// ...
n, s, err := read(conn, b)
```
which would be broken.

As a rule, any change to an exported interface, function or type definition is not backwards compatible. This would be a breaking change only for callers who use the `Reader` interface or `ReadFromSessionUDP` and `WriteToSessionUDP`, but regardless of how limited that may or may not be, go now has solid guidelines for breaking changes with modules at v1 or above.

Anyway, I believe this is (relatively) easily achievable with a slightly different approach. Please check out #1174 and let me know if that will suit your needs.

PR #1171, opened 2020-10-13
Retain TC bit if original message was already marked truncated

chrisohaver:
Retain TC bit if original message was already marked truncated.

Fixes coredns/coredns#4183

This could alternately be fixed in coredns/coredns, although I can't think how reverting the TC bit would be desirable in any scenario, hence suggesting the fix upstream here.

Signed-off-by: Chris O'Haver <cohaver@infoblox.com>


tmthrgd (approved):
This seems reasonable and likely to solve real world bugs. LGTM.

PR #1173, opened 2020-10-14
Update doc for Truncate

miekg:
Follow up after #1171, update the docs about TC and use the MinMsgSize
constant for 512 in the code.


tmthrgd (approved):
LGTM.

PR #1174, opened 2020-10-14
Support generic net.PacketConn's for the Server

### LINKED ISSUE #1168: Allow any net.PacketConn and don't assume it is *net.UDPConn

tmthrgd:
This commit adds support for listening on generic net.PacketConn's for UDP DNS requests, previously *net.UDPConn was the only supported type.

In the event of a future v2 of this module, this should be streamlined.

This is a backward compatible alternative to #1170 and supersedes it.

Fixes #1168
Closes #1170

### DIFF
--- a/server.go
+++ b/server.go
@@ -72,9 +72,10 @@ type response struct {
 	tsigStatus     error
 	tsigRequestMAC string
 	tsigSecret     map[string]string // the tsig secrets
-	udp            *net.UDPConn      // i/o connection if UDP was used
+	udp            net.PacketConn    // i/o connection if UDP was used
 	tcp            net.Conn          // i/o connection if TCP was used
 	udpSession     *SessionUDP       // oob data to get egress interface right
+	pcSession      net.Addr          // address to use when writing to a generic net.PacketConn
 	writer         Writer            // writer to output the raw DNS bits
 }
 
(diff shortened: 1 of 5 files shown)


tmthrgd:
/cc @guillaumerose @gbraad

Could you both test to see if this PR would meet your needs?

guillaumerose:
I just tried it and it works perfectly with my custom net.PacketConn. Thanks.

gbraad (approved):

gbraad:
Looks good. So what would the next steps be, as you mentioned a lack of test cases.

tmthrgd:
@guillaumerose Great to hear!

@gbraad Just the test cases, and then it should be good to merge. They should be pretty similar to the existing UDP test cases, a lot of copying and pasting. Might be able to smash that out tonight or in a few days time.

tmthrgd:
I have absolutely no clue what on earth is wrong with the tests. The two 'failing' lines don't remotely match up (one is blank!). When I run the tests locally both on my branch and using `git fetch origin pull/1174/head:pr1174 && git checkout pr1174`, the tests pass fine. I'm perplexed, both actions/checkout and Travis seem to have somehow ended up with a broken checkout.

server.go
```diff
@@ -72,9 +72,10 @@ type response struct {
 	tsigStatus     error
 	tsigRequestMAC string
 	tsigSecret     map[string]string // the tsig secrets
-	udp            *net.UDPConn      // i/o connection if UDP was used
+	udp            net.PacketConn    // i/o connection if UDP was used
 	tcp            net.Conn          // i/o connection if TCP was used
 	udpSession     *SessionUDP       // oob data to get egress interface right
+	pcSession      net.Addr          // address to use when writing to a generic net.PacketConn
```
miekg, line 78: pcSession needs to be carried separately from the PacketConn?

server.go
```diff
@@ -692,7 +743,10 @@ func (w *response) Write(m []byte) (int, error) {
 
 	switch {
 	case w.udp != nil:
-		return WriteToSessionUDP(w.udp, m, w.udpSession)
+		if u, ok := w.udp.(*net.UDPConn); ok {
+			return WriteToSessionUDP(u, m, w.udpSession)
+		}
+		return w.udp.WriteTo(m, w.pcSession)
```
miekg, line 749: IIRC with just `Write` this should work as well? Or does it go to the default interface or some such?

server.go
tmthrgd: This gives us the peers remote address, we can only get our local address from the connection because it’s an unconnected listen socket.

server.go
tmthrgd: The connection is listening so we need to send back to the specific remote address. The session UDP stuff handles the specific interface, but we can’t use that for generic PacketConn’s.

miekg:
>The connection is listening so we need to send back to the specific remote
>address. The session UDP stuff handles the specific interface, but we can’t use
>that for generic PacketConn’s.

fair enough.

I think this looks good

/lgmt

(the bot helping us has been upgrade - github deprecated some fluff from their API)

guillaumerose:
Thanks a lot!

PR #1175, opened 2020-10-15
tests: reduce timeouts and iterations

miekg:
This reduces the time it takes to run the test. Shorter timeouts on
clients to avoid awaiting for the detault timeouts.

It's also reduces the iterations in some test functions, this doesn't
seem to impact the tests indicating those numbers where random to begin
with.

Stop using Google Public DNS in tests as well: it's faster, keeps things
local and avoids spilling info to Google.

This brings the test duration down from ~8s to ~3s on my machine.

~~~
PASS
ok  	github.com/miekg/dns	2.751s

Switched to branch 'master'
Your branch is up-to-date with 'origin/master'.
PASS
ok  	github.com/miekg/dns	7.915s
~~~


server_test.go
```diff
@@ -698,7 +699,7 @@ func checkInProgressQueriesAtShutdownServer(t *testing.T, srv *Server, addr stri
 	})
 	defer HandleRemove("example.com.")
 
-	client.Timeout = 10 * time.Second
+	client.Timeout = 100 * time.Millisecond
```
tmthrgd, line 702: I’d switch this to 1s, this timeout should never be hit unless the test fails (unless I’m very wrong). It’s a bit more reliable than lowering it to 100ms.

server_test.go
```diff
@@ -668,7 +669,7 @@ func init() {
 }
 
 func checkInProgressQueriesAtShutdownServer(t *testing.T, srv *Server, addr string, client *Client) {
-	const requests = 100
+	const requests = 5
```
tmthrgd, line 672: 100 is probably overkill in most cases, but 5 might be a bit too few. What about 15 or 25?

length_test.go
```diff
@@ -399,7 +399,7 @@ func TestMsgCompressLengthLargeRecordsAllValues(t *testing.T) {
 	msg := new(Msg)
 	msg.Compress = true
 	msg.SetQuestion("redis.service.consul.", TypeSRV)
-	for i := 0; i < 900; i++ {
+	for i := 0; i < 90; i++ { // this was 900, for unknown reasons, brought down to 90 - questioning the value of the test in general. (for 900 this takes 2.41s)
```
tmthrgd, line 402: I think this was 900 for a good reason, git blame should point out what the PR was. I think compression was failing at specific byte boundaries and the high number was about catching that. It might be possible to raise the minimum instead. Put in the 900*2 records and then only compression for higher values.

tmthrgd: Perhaps leave this test for a separate PR.

miekg: this is the top time taker (2.9s on my machine) - but yeah, prolly easier to discuss in another pr

server_test.go
miekg: done

server_test.go
miekg: made it 15 + comment

miekg:
move the contentious bit out of this PR, all comments addressed, so I'm merging.

PR #1176, opened 2020-10-15
Make the 900 number lower in this test

miekg:
Original PR doesn't make clear why this is 900-now assuming we want to
cross the 14 bit length boundary.

Up to 900 creates a super large (>2^16) message. Not sure why that needs
testing.

Also remove the packs at lower sizes.


tmthrgd (approved):
As long as this is still hitting the problematic boundary, LGTM.

PR #1177, opened 2020-10-15
Add odohconfig parameter, defined in draft-pauly-dprive-oblivious-doh-02

chris-wood:
[draft-pauly-oblivious-doh-02](https://tools.ietf.org/html/draft-pauly-dprive-oblivious-doh-02) defines an "odohconfig" Service Binding parameter "odohconfig", which contains the public configuration information for Oblivious DoH. This structure is nearly identical to the "echconfig" parameter. The value of this parameter is a list of `ObliviousDoHConfig` structures.

### DIFF
--- a/svcb.go
+++ b/svcb.go
@@ -12,7 +12,8 @@ import (
 
 type SVCBKey uint16
 
-// Keys defined in draft-ietf-dnsop-svcb-https-02 Section 11.1.2.
+// Keys defined in draft-ietf-dnsop-svcb-httpssvc-03, Section 12.1
+// Keys in the range 32768-65280 are the First Come First Served range
 const (
 	SVCB_MANDATORY       SVCBKey = 0
 	SVCB_ALPN            SVCBKey = 1
@@ -21,6 +22,7 @@ const (
 	SVCB_IPV4HINT        SVCBKey = 4
 	SVCB_ECHCONFIG       SVCBKey = 5
 	SVCB_IPV6HINT        SVCBKey = 6
+	SVCB_ODOHCONFIG      SVCBKey = 32769 // draft-pauly-dprive-oblivious-doh-02
 	svcb_RESERVED        SVCBKey = 65535
 )
 
@@ -31,6 +33,7 @@ var svcbKeyToStringMap = map[SVCBKey]string{
 	SVCB_PORT:            "port",
 	SVCB_IPV4HINT:        "ipv4hint",
 	SVCB_ECHCONFIG:       "echconfig",
+	SVCB_ODOHCONFIG:      "odohconfig",
 	SVCB_IPV6HINT:        "ipv6hint",
 }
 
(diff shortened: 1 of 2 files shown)


miekg:
a far are we from wg last call for this draft?

chris-wood:
> a far are we from wg last call for this draft?

Quite far! It's not even an adopted draft, yet. Nevertheless, I wanted to kick this off using an experimental codepoint.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Add odohconfig para..." ]
>    a far are we from wg last call for this draft?
>
>Quite far! It's not even an adopted draft, yet. Nevertheless, I wanted to kick
>this off using an experimental codepoint.

let's wait for adoption (at least). Btw if experimental code points need code changes in
_this_ library we need to fix that.
>
/Miek

--
Miek Gieben

chris-wood:
> Btw if experimental code points need code changes in _this_ library we need to fix that.

Hmm, how would you recommend we fix that?

miekg:
I don't know how that would look. Likely it requires changing the current interface as well (so the current mandatory options use the same interface).

Lekensteyn:
External users cannot implement the `SVCBKeyValue` interface because some member functions are private. However odohconfig support can still be implemented through the `SVCBLocal` type which is designated for such experimentation.

Support for it would look like this:
```
	odohconfigs := []byte("an ObliviousDoHConfigs structure")
	h := new(dns.HTTPS)
	h.Hdr = dns.RR_Header{Name: ".", Rrtype: dns.TypeHTTPS, Class: dns.ClassINET}
	e := &dns.SVCBLocal{KeyCode: 32769, Data: odohconfigs}
	h.Value = append(h.Value, e)
	fmt.Println(h)
```
The serialized result looks like:
```
.	0	IN	HTTPS	0  key32769="an\ ObliviousDoHConfigs\ structure"
```

It does not look as nice though, ideally it would look like:
```
.	0	IN	HTTPS	0  echconfig="YW5cIE9ibGl2aW91c0RvSENvbmZpZ3NcIHN0cnVjdHVyZQ=="
```

but that would require publishing more details of the interface.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Add odohconfig para..." ]
>External users cannot implement the SVCBKeyValue interface because some member
>functions are private. However odohconfig support can still be implemented
>through the SVCBLocal type which is designated for such experimentation.
>
>Support for it would look like this:
>
>        odohconfigs := []byte("an ObliviousDoHConfigs structure")
>        h := new(dns.HTTPS)
>        h.Hdr = dns.RR_Header{Name: ".", Rrtype: dns.TypeHTTPS, Class: dns.ClassINET}
>        e := &dns.SVCBLocal{KeyCode: 32769, Data: odohconfigs}
>        h.Value = append(h.Value, e)
>        fmt.Println(h)

this would be good to add to the documentation.

>The serialized result looks like:
>
>.       0       IN      HTTPS   0  key32769="an\ ObliviousDoHConfigs\ structure"
>
>It does not look as nice though, ideally it would look like:
>
>.       0       IN      HTTPS   0  echconfig="YW5cIE9ibGl2aW91c0RvSENvbmZpZ3NcIHN0cnVjdHVyZQ=="
>
>but that would require publishing more details of the interface.

It could also just require some Registration functions. Happy to discuss what's needed
here and move the current impl. over to use the same functions.


/Miek

--
Miek Gieben

PR #1178, opened 2020-10-16
svcb: doc updates

miekg:
The SVBC record didn't have a class, so add that and use struct literal
to put it all on 1 one. Use `s` for SVCB records, and `h` for HTTPS to
be more consistent.


tmthrgd (approved):
Fine by me.

PR #1182, opened 2020-10-21
svcb: documentation fixes

Lekensteyn:
`&dns.HTTPS{Hdr: ...}` fails due to `cannot use promoted field SVCB.Hdr
in struct literal of type dns.HTTPS`. Fix this, less readable
alternatives include using `dns.SVCB` directly, or `&dns.HTTPS{SVCB:
dns.SVCB{Hdr: ...}}`.

Fix the draft reference, draft-02 has not been published yet. Fix ECH,
its type has changed from base64 to a byte array.

### DIFF
--- a/svcb.go
+++ b/svcb.go
@@ -12,7 +12,7 @@ import (
 
 type SVCBKey uint16
 
-// Keys defined in draft-ietf-dnsop-svcb-https-02 Section 11.1.2.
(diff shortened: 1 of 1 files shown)


miekg (approved):

miekg:
we need a test that extracts all Go documenation and tries to compile it, so it will stay
current *and* we know it compiles.

CoreDNS does this for it's config: https://github.com/coredns/coredns/blob/master/test/readme_test.go

Issue #1186, opened 2020-10-23
CI fuzzing is slow


tmthrgd:
The CI fuzzing action (added in #1091) is currently set to run for [10 minutes](https://github.com/miekg/dns/blob/6d41f4302200a0133146d4e7f09f3e8d6d709ac9/.github/workflows/cifuzz.yml#L16) which is quite a long time for a CI to run for on every pull request. I do know that fuzzing obviously works better the longer it's run for, but 10 minutes seems excessively long for pull requests (most of which never even touch the fuzzed code).

Should we consider lowering it and if so to how long?

/cc @catenacyber as the author of #1091

tmthrgd:
Also the fuzzing step isn't automatically cancelled if I push a new commit to a PR, see #1188 for an example. The Travis CI build for the old commit will be cancelled but the fuzzing will just power on for the full 10 minutes. I'm not sure if GitHub Actions (or whatever it's called) have a way to do this or not.

catenacyber:
Thanks for this feedback.
I reported it to https://github.com/google/oss-fuzz/issues/4580

> 10 minutes seems excessively long for pull requests

That really depends on your project/workflow.
From the fuzzing point of view, 10 minutes seem like a good time for finding some obvious bug.
You may lower this time if it seems too long for your CI.
Or we should try to run it only for relevant PRs

catenacyber:
It looks like we can restrict CIFuzz to certain PRs by configuring paths of interesting files :
https://google.github.io/oss-fuzz/getting-started/continuous-integration/#branches-and-paths

tmthrgd:
That could be slightly brittle without knowing how different go files contribute to the build, but it might be worth a shot. Do you want to open a PR trying that?

miekg:
think we should move this to a github workflow cronjob and complete from the PRs checks.

miekg:
revisit the fuzzing once we standardize on go 1.18

PR #1192, opened 2020-11-03
Runs CIFuzz only when go files are modified

catenacyber:
and run it for 5 minutes instead of 10

Cf #1186 cc @tmthrgd


catenacyber:
Looks like it is working since CIFuzz is not running on this PR, right ?

PR #1193, opened 2020-11-06
Remove travis

miekg:
This also removed the codecov that was still done. We could potentially
re-add as an action, but I don't really miss it.


tmthrgd:
I wouldn’t miss the code coverage thing, I’ve never quite understood how to interpret it. I would miss Travis though, the GitHub action is still on go1.14 with no sign that it will be updated to support 1.15 anytime soon, it’s been several months with no movement from the owner. 1.16 will be out soon and if it doesn’t get updated by then, we’d be testing only an unsupported version. Travis also gets us the latest two not just one which I think is valuable as both are still supported.

miekg:
for coredns I used action@v2 and said I wanted to use 1.15, that just seemed to have
worked... https://github.com/coredns/coredns/blob/master/.github/workflows/go.test.yml
(which started as a copy from the workflow here in this repo, so thanks! :) )

I also removed travis oauth from my account, so it's broken for all my repos since a few
minutes ago

miekg:
@tmthrgd added 1.15 and seems to work - things are now so fast we can re-enable testing 1.14 *and* 1.15 again.

miekg:
seems to do what we need.

tmthrgd:
> I also removed travis oauth from my account, so it's broken for all my repos since a few minutes ago

Well that would do it. 😆

As long as we can continue to test the latest two versions and won’t have to wait around too long once new versions are released, this should be fine.

I know there’s an open issue about go1.15 in that action repo, maybe it only affects v1? 🤷‍♂️

PR #1195, opened 2020-11-19
Circleci project setup

bigcook12345:
With the ability, to be able to "Merge", set up the workflow/respitories, with the right configuration. I am actually able to see exactly what it is, that I have been missing to begin with? I'm just hoping to be able to get it right this time? There should be a bit less "t's" crossed and "i's" dotted. Going in "Blinder" then I thought wishing myself "Luck" on getting it right. Worth a shot...

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,21 @@
+version: 2.1
+
+orbs:
+  python: circleci/python@0.2.1
+
+jobs:
+  build-and-test:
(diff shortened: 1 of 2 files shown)


bigcook12345 (commented):
Learning how to process the work, instead of continuing the action of "Hard-Forking"...

miekg:
I'm unsure what this is, but we use github actions to build and test. thanks

PR #1200, opened 2020-12-09
When using TCP connections send only one packet

### LINKED ISSUE #1199: TLS connections send packet length & packet data seperately
I think for some reason the tls library is not honouring the `net.Buffers` code here https://github.com/miekg/dns/blob/428cef31872d7aee10c55a9450929bc9b645b292/client.go#L336 ?

When dumping + decrypting with wireshark, a normal TCP connection is doing single-packet based queries, however TLS is always sending 2 packets per query (& for response when upstream is running this library too). Here is a screenshot from wireshark:

image

mzealey:
net.Buffers relies on writeBuffer to send all the pieces in one go. The
tls library does not support this so the length and content get split
into 2 packets. This patch removes net.Buffers and replaces with a new
single memory allocation to ensure the length+content get sent in one
packet. This fixes #1199

### DIFF
--- a/client.go
+++ b/client.go
@@ -330,11 +330,11 @@ func (co *Conn) Write(p []byte) (int, error) {
 		return co.Conn.Write(p)
 	}
 
-	l := make([]byte, 2)
-	binary.BigEndian.PutUint16(l, uint16(len(p)))
+	msg_with_length := make([]byte, 2+len(p))
+	binary.BigEndian.PutUint16(msg_with_length, uint16(len(p)))
+	copy(msg_with_length[2:], p[:])
 
-	n, err := (&net.Buffers{l, p}).WriteTo(co.Conn)
-	return int(n), err
+	return co.Conn.Write(msg_with_length)
 }
 
 // Return the appropriate timeout for a specific request
--- a/server.go
+++ b/server.go
@@ -752,11 +752,11 @@ func (w *response) Write(m []byte) (int, error) {
 			return 0, &Error{err: "message too large"}
 		}
 
-		l := make([]byte, 2)
-		binary.BigEndian.PutUint16(l, uint16(len(m)))
+		msg_with_length := make([]byte, 2+len(m))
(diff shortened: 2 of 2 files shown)


miekg:
I think we had something like this way before. Also to prevent further flip flopping on this code there should be at least a test to detect this.
@tmthrgd from the top of my head, I think you are the one that moved to net.Bufffer? (not related to this issue, but close enough)

tmthrgd:
> I think we had something like this way before. Also to prevent further flip flopping on this code there should be at least a test to detect this.
> @tmthrgd from the top of my head, I think you are the one that moved to net.Bufffer? (not related to this issue, but close enough)

Indeed I did. `net.Buffer` is a net win for TCP so I don’t think we should do this as is. crypto/tls has an internal buffer anyway so perhaps a better fix would be for Go to add `net.Buffer` support for `tls.Conn`. If that doesn’t happen, or we’re not prepared to wait, then perhaps gating this on `*net.TCPConn` might be a solution.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] When using TCP conn..." ]
>    I think we had something like this way before. Also to prevent further flip
>    flopping on this code there should be at least a test to detect this.
>    @tmthrgd from the top of my head, I think you are the one that moved to
>    net.Bufffer? (not related to this issue, but close enough)
>
>Indeed I did. net.Buffer is a net win for TCP so I don’t think we should do
>this as is. crypto/tls has an internal buffer anyway so perhaps a better fix
>would be for Go to add net.Buffer support for tls.Conn. If that doesn’t happen,
>or we’re not prepared to wait, then perhaps gating this on *net.TCPConn might
>be a solution.

Is there a golang issue we can link to?

tmthrgd:
> [ Quoting <notifications@github.com> in "Re: [miekg/dns] When using TCP conn..." ]
> Is there a golang issue we can link to?

Not to my knowledge. It would be worth someone opening one and hearing what the Go team think.

mzealey:
Superceded by #1219

PR #1201, opened 2020-12-16
Add GSS-TSIG support

bodgit:
Here's another go at adding GSS-TSIG support.

Inspired by how crypto/ssh handles this I've defined a simple interface for an external GSSAPI implementation and added this to the `Client` and `Conn` structs. I've added two new internal `tsigGenerateGSSAPI()` and `tsigVerifyGSSAPI()` functions that are passed this interface and then refactored the original exported `TsigGenerate()` and `TsigVerify()` functions to use these, passing `nil` for the GSSAPI interface, so existing code using those will behave the same as before.

I've left the new functions unexported and I haven't touched the `Server` or `Transfer` structs yet. I've not added any tests either but that should be easy enough if you approve.

### DIFF
--- a/client.go
+++ b/client.go
@@ -23,7 +23,7 @@ type Conn struct {
 	net.Conn                         // a net.Conn holding the connection
 	UDPSize        uint16            // minimum receive buffer for UDP messages
 	TsigSecret     map[string]string // secret(s) for Tsig map[<zonename>]<base64 secret>, zonename must be in canonical form (lowercase, fqdn, see RFC 4034 Section 6.2)
-	TsigGSS                          // An implementation of the TsigGSS interface.
+	TsigProvider                     // An implementation of the TsigProvider interface.
 	tsigRequestMAC string
 }
 
@@ -41,7 +41,7 @@ type Client struct {
 	ReadTimeout    time.Duration     // net.Conn.SetReadTimeout value for connections, defaults to 2 seconds - overridden by Timeout when that value is non-zero
 	WriteTimeout   time.Duration     // net.Conn.SetWriteTimeout value for connections, defaults to 2 seconds - overridden by Timeout when that value is non-zero
 	TsigSecret     map[string]string // secret(s) for Tsig map[<zonename>]<base64 secret>, zonename must be in canonical form (lowercase, fqdn, see RFC 4034 Section 6.2)
-	TsigGSS                          // An implementation of the TsigGSS interface.
+	TsigProvider                     // An implementation of the TsigProvider interface.
 	SingleInflight bool              // if true suppress multiple outstanding queries for the same Qname, Qtype and Qclass
 	group          singleflight
 }
@@ -177,7 +177,7 @@ func (c *Client) exchange(m *Msg, co *Conn) (r *Msg, rtt time.Duration, err erro
 		co.UDPSize = c.UDPSize
 	}
 
-	co.TsigSecret, co.TsigGSS = c.TsigSecret, c.TsigGSS
+	co.TsigSecret, co.TsigProvider = c.TsigSecret, c.TsigProvider
 	t := time.Now()
 	// write with the appropriate write timeout
 	co.SetWriteDeadline(t.Add(c.getTimeoutForRequest(c.writeTimeout())))
@@ -224,11 +224,11 @@ func (co *Conn) ReadMsg() (*Msg, error) {
 		return m, err
 	}
 	if t := m.IsTsig(); t != nil {
-		if _, ok := co.TsigSecret[t.Hdr.Name]; !ok {
+		if _, ok := co.TsigSecret[t.Hdr.Name]; !ok && co.TsigProvider == nil {
 			return m, ErrSecret
 		}
 		// Need to work on the original message p, as that was used to calculate the tsig.
-		err = tsigVerifyGSS(p, co.TsigSecret[t.Hdr.Name], co.tsigRequestMAC, false, co.TsigGSS)
+		err = tsigVerifyProvider(p, co.TsigSecret[t.Hdr.Name], co.tsigRequestMAC, false, co.TsigProvider)
 	}
 	return m, err
 }
@@ -306,10 +306,10 @@ func (co *Conn) WriteMsg(m *Msg) (err error) {
 	var out []byte
 	if t := m.IsTsig(); t != nil {
 		mac := ""
-		if _, ok := co.TsigSecret[t.Hdr.Name]; !ok {
+		if _, ok := co.TsigSecret[t.Hdr.Name]; !ok && co.TsigProvider == nil {
 			return ErrSecret
 		}
-		out, mac, err = tsigGenerateGSS(m, co.TsigSecret[t.Hdr.Name], co.tsigRequestMAC, false, co.TsigGSS)
+		out, mac, err = tsigGenerateProvider(m, co.TsigSecret[t.Hdr.Name], co.tsigRequestMAC, false, co.TsigProvider)
 		// Set for the next read, although only used in zone transfers
 		co.tsigRequestMAC = mac
 	} else {
--- a/client.go
+++ b/client.go
@@ -23,7 +23,7 @@ type Conn struct {
 	net.Conn                         // a net.Conn holding the connection
 	UDPSize        uint16            // minimum receive buffer for UDP messages
 	TsigSecret     map[string]string // secret(s) for Tsig map[<zonename>]<base64 secret>, zonename must be in canonical form (lowercase, fqdn, see RFC 4034 Section 6.2)
-	GSSAPI         GSSAPI            // an implementation of the GSSAPI interface
+	TsigGSS                          // An implementation of the TsigGSS interface.
 	tsigRequestMAC string
 }
 
@@ -41,7 +41,7 @@ type Client struct {
 	ReadTimeout    time.Duration     // net.Conn.SetReadTimeout value for connections, defaults to 2 seconds - overridden by Timeout when that value is non-zero
 	WriteTimeout   time.Duration     // net.Conn.SetWriteTimeout value for connections, defaults to 2 seconds - overridden by Timeout when that value is non-zero
 	TsigSecret     map[string]string // secret(s) for Tsig map[<zonename>]<base64 secret>, zonename must be in canonical form (lowercase, fqdn, see RFC 4034 Section 6.2)
-	GSSAPI         GSSAPI            // an implementation of the GSSAPI interface
+	TsigGSS                          // An implementation of the TsigGSS interface.
 	SingleInflight bool              // if true suppress multiple outstanding queries for the same Qname, Qtype and Qclass
 	group          singleflight
 }
@@ -177,7 +177,7 @@ func (c *Client) exchange(m *Msg, co *Conn) (r *Msg, rtt time.Duration, err erro
 		co.UDPSize = c.UDPSize
 	}
 
-	co.TsigSecret, co.GSSAPI = c.TsigSecret, c.GSSAPI
+	co.TsigSecret, co.TsigGSS = c.TsigSecret, c.TsigGSS
 	t := time.Now()
 	// write with the appropriate write timeout
 	co.SetWriteDeadline(t.Add(c.getTimeoutForRequest(c.writeTimeout())))
@@ -228,7 +228,7 @@ func (co *Conn) ReadMsg() (*Msg, error) {
 			return m, ErrSecret
 		}
 		// Need to work on the original message p, as that was used to calculate the tsig.
-		err = tsigVerifyGSSAPI(p, co.TsigSecret[t.Hdr.Name], co.tsigRequestMAC, false, co.GSSAPI)
+		err = tsigVerifyGSS(p, co.TsigSecret[t.Hdr.Name], co.tsigRequestMAC, false, co.TsigGSS)
 	}
 	return m, err
 }
@@ -309,7 +309,7 @@ func (co *Conn) WriteMsg(m *Msg) (err error) {
 		if _, ok := co.TsigSecret[t.Hdr.Name]; !ok {
 			return ErrSecret
 		}
-		out, mac, err = tsigGenerateGSSAPI(m, co.TsigSecret[t.Hdr.Name], co.tsigRequestMAC, false, co.GSSAPI)
+		out, mac, err = tsigGenerateGSS(m, co.TsigSecret[t.Hdr.Name], co.tsigRequestMAC, false, co.TsigGSS)
 		// Set for the next read, although only used in zone transfers
 		co.tsigRequestMAC = mac
 	} else {
--- a/client.go
+++ b/client.go
@@ -225,7 +225,7 @@ func (co *Conn) ReadMsg() (*Msg, error) {
 	}
 	if t := m.IsTsig(); t != nil {
 		if co.TsigProvider != nil {
-			err = tsigVerifyProvider(p, co.tsigRequestMAC, false, co.TsigProvider)
+			err = tsigVerifyProvider(p, co.TsigProvider, co.tsigRequestMAC, false)
 		} else {
 			if _, ok := co.TsigSecret[t.Hdr.Name]; !ok {
 				return m, ErrSecret
@@ -311,7 +311,7 @@ func (co *Conn) WriteMsg(m *Msg) (err error) {
 	if t := m.IsTsig(); t != nil {
 		mac := ""
 		if co.TsigProvider != nil {
-			out, mac, err = tsigGenerateProvider(m, co.tsigRequestMAC, false, co.TsigProvider)
+			out, mac, err = tsigGenerateProvider(m, co.TsigProvider, co.tsigRequestMAC, false)
 		} else {
 			if _, ok := co.TsigSecret[t.Hdr.Name]; !ok {
 				return ErrSecret
--- a/client.go
+++ b/client.go
@@ -224,11 +224,15 @@ func (co *Conn) ReadMsg() (*Msg, error) {
 		return m, err
 	}
 	if t := m.IsTsig(); t != nil {
-		if _, ok := co.TsigSecret[t.Hdr.Name]; !ok && co.TsigProvider == nil {
-			return m, ErrSecret
+		if co.TsigProvider != nil {
+			err = tsigVerifyProvider(p, co.tsigRequestMAC, false, co.TsigProvider)
+		} else {
+			if _, ok := co.TsigSecret[t.Hdr.Name]; !ok {
+				return m, ErrSecret
+			}
+			// Need to work on the original message p, as that was used to calculate the tsig.
+			err = TsigVerify(p, co.TsigSecret[t.Hdr.Name], co.tsigRequestMAC, false)
 		}
-		// Need to work on the original message p, as that was used to calculate the tsig.
-		err = tsigVerifyProvider(p, co.TsigSecret[t.Hdr.Name], co.tsigRequestMAC, false, co.TsigProvider)
 	}
 	return m, err
 }
@@ -306,10 +310,14 @@ func (co *Conn) WriteMsg(m *Msg) (err error) {
 	var out []byte
 	if t := m.IsTsig(); t != nil {
 		mac := ""
-		if _, ok := co.TsigSecret[t.Hdr.Name]; !ok && co.TsigProvider == nil {
-			return ErrSecret
+		if co.TsigProvider != nil {
+			out, mac, err = tsigGenerateProvider(m, co.tsigRequestMAC, false, co.TsigProvider)
+		} else {
+			if _, ok := co.TsigSecret[t.Hdr.Name]; !ok {
+				return ErrSecret
+			}
+			out, mac, err = TsigGenerate(m, co.TsigSecret[t.Hdr.Name], co.tsigRequestMAC, false)
 		}
-		out, mac, err = tsigGenerateProvider(m, co.TsigSecret[t.Hdr.Name], co.tsigRequestMAC, false, co.TsigProvider)
 		// Set for the next read, although only used in zone transfers
 		co.tsigRequestMAC = mac
 	} else {
--- a/client.go
+++ b/client.go
@@ -23,6 +23,7 @@ type Conn struct {
 	net.Conn                         // a net.Conn holding the connection
 	UDPSize        uint16            // minimum receive buffer for UDP messages
 	TsigSecret     map[string]string // secret(s) for Tsig map[<zonename>]<base64 secret>, zonename must be in canonical form (lowercase, fqdn, see RFC 4034 Section 6.2)
+	GSSAPI         GSSAPI            // an implementation of the GSSAPI interface
 	tsigRequestMAC string
 }
 
@@ -40,6 +41,7 @@ type Client struct {
 	ReadTimeout    time.Duration     // net.Conn.SetReadTimeout value for connections, defaults to 2 seconds - overridden by Timeout when that value is non-zero
 	WriteTimeout   time.Duration     // net.Conn.SetWriteTimeout value for connections, defaults to 2 seconds - overridden by Timeout when that value is non-zero
 	TsigSecret     map[string]string // secret(s) for Tsig map[<zonename>]<base64 secret>, zonename must be in canonical form (lowercase, fqdn, see RFC 4034 Section 6.2)
+	GSSAPI         GSSAPI            // an implementation of the GSSAPI interface
 	SingleInflight bool              // if true suppress multiple outstanding queries for the same Qname, Qtype and Qclass
 	group          singleflight
 }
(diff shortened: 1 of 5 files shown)


tmthrgd:
This idea looks reasonable enough (though I won’t review the actual implementation just yet), it appears entirely backwards compatible. Can you show an example of how the interface might be implemented?

bodgit:
To begin with you need to establish a GSSAPI context with the target server.

You're basically doing what is documented in [RFC 2743](https://tools.ietf.org/html/rfc2743#page-42) with `GSS_Init_sec_context()`. You first call it passing `GSS_C_NO_CONTEXT` for the input_context_handle with a NULL input_token. The result of this is an output_context_handle and an output_token the latter of which needs to be sent to the target server.

To do this you send a TKEY request with the target server, the TKEY name is generated with something like this:
```golang
dns.Fqdn(fmt.Sprintf("%d.sig-%s", rng.Int31(), host)) // this is what nsupdate uses
```
and the request is created something like this:
```golang
msg := &dns.Msg{
	MsgHdr: dns.MsgHdr{
		RecursionDesired: false,
	},
	Question: make([]dns.Question, 1),
	Extra:    make([]dns.RR, 1),
}

msg.Question[0] = dns.Question{
	Name:   keyname,
	Qtype:  dns.TypeTKEY,
	Qclass: dns.ClassANY,
}

msg.Id = dns.Id()

msg.Extra[0] = &dns.TKEY{
	Hdr: dns.RR_Header{
		Name:   keyname,
		Rrtype: dns.TypeTKEY,
		Class:  dns.ClassANY,
		Ttl:    0,
	},
	Algorithm:  "gss-tsig.",
	Mode:       3,
	Inception:  inception,
	Expiration: expiration,
	KeySize:    uint16(len(output_token)),
	Key:        hex.EncodeToString(output_token),
}
```
That is sent to the server which responds (hopefully) with a TKEY answer. The key payload is now the input_token which is then used to call `GSS_Init_sec_context()` again and hopefully you get `GSS_S_COMPLETE`. If you get `GSS_S_CONTINUE_NEEDED` you repeat the exchange.

At this point you have a negotiated context which you should store using the TKEY name you created. That TKEY name is the TSIG secret name however there's no secret to speak of, the context is the secret here.

So an implementation of the interface would look something like this:
```
type context struct {
	client *client.Client
	key    types.EncryptionKey
}

type Example struct {
	ctx map[string]context 
}

...

func (e *Example) Generate(msg []byte, tsig *dns.TSIG) ([]byte, error) {
	if dns.CanonicalName(tsig.Algorithm) != dns.GSS {
		return nil, dns.ErrKeyAlg
	}

	ctx, ok := e.ctx[tsig.Hdr.Name] // this is where you retrieve your negotiated context
	if !ok {
		return nil, dns.ErrSecret
	}

	token := gssapi.MICToken{
		Flags:     gssapi.MICTokenFlagAcceptorSubkey,
		SndSeqNum: 0,
		Payload:   msg,
	}

	if err := token.SetChecksum(ctx.key, keyusage.GSSAPI_INITIATOR_SIGN); err != nil {
		return nil, err
	}

	b, err := token.Marshal()
	if err != nil {
		return nil, err
	}

	return b, nil
}

func (e *Example) Verify(stripped []byte, tsig *dns.TSIG) error {
	if dns.CanonicalName(tsig.Algorithm) != dns.GSS {
		return dns.ErrKeyAlg
	}

	ctx, ok := e.ctx[tsig.Hdr.Name] // retrieve the negotiated context
	if !ok {
		return dns.ErrSecret
	}

	mac, err := hex.DecodeString(tsig.MAC)
	if err != nil {
		return err
	}

	var token gssapi.MICToken
	err = token.Unmarshal(mac, true)
	if err != nil {
		return err
	}
	token.Payload = stripped

	// This is the actual verification bit
	_, err = token.Verify(ctx.key, keyusage.GSSAPI_ACCEPTOR_SIGN)
	if err != nil {
		return err
	}

	return nil
}
```
That example is based on using github.com/jcmturner/gokrb5 for the GSSAPI implementation, on Windows you would use github.com/alexbrainman/sspi instead which has a similar API but uses SSPI.

You would then use it something like:
```golang
host := "ns.example.com"

e := new(Example)
keyname := e.NegotiateTKEYRecord(host) // this does the TKEY exchange as documented above

client := new(dns.Client)
client.Net = "tcp"
client.GSSAPI = e
client.TsigSecret = map[string]string{keyname: ""} // doesn't matter what the secret is set to

// Use the DNS client as normal
msg := new(dns.Msg)
msg.SetUpdate(dns.Fqdn("example.com"))

insert, err := dns.NewRR("test.example.com. 300 A 192.0.2.1")
if err != nil {
        panic(err)
}
msg.Insert([]dns.RR{insert})

msg.SetTsig(keyname, dns.GSS, 300, time.Now().Unix())

rr, _, err := client.Exchange(msg, net.JoinHostPort(host, "53"))
if err != nil {
        panic(err)
}

if rr.Rcode != dns.RcodeSuccess {
        fmt.Printf("DNS error: %s (%d)\n", dns.RcodeToString[rr.Rcode], rr.Rcode)
}
```
There's probably a bit more housekeeping but that's essentially it.

bodgit:
The server side would be similar however you would have to instead respond to any TKEY requests by calling `GSS_Accept_sec_context()` as per the RFC but once you've done that the TSIG generate and verify code would be similar, perhaps just swapping the key usage so you would generate the response signatures using `keyusage.GSSAPI_ACCEPTOR_SIGN` and verify using `keyusage.GSSAPI_INITIATOR_SIGN` continuing the gokrb5 example.

tmthrgd:
That doesn’t seem too unreasonable and likely useable in practice. Is GSS-TSIG the only TSIG-extension-type-thing out there or are there others?

bodgit:
I've not seen any other mechanisms other than GSS-TSIG, apart from the standard HMAC methods. Of course GSS itself stands for Generic Security Service so it is potentially a transport for any number of authentication methods although realistically it is just Kerberos.

miekg:
this looks good IMO. Maybe a example on the code/documentation makes sense? (mustn't be too big though)

client.go
```diff
@@ -23,6 +23,7 @@ type Conn struct {
 	net.Conn                         // a net.Conn holding the connection
 	UDPSize        uint16            // minimum receive buffer for UDP messages
 	TsigSecret     map[string]string // secret(s) for Tsig map[<zonename>]<base64 secret>, zonename must be in canonical form (lowercase, fqdn, see RFC 4034 Section 6.2)
+	GSSAPI         GSSAPI            // an implementation of the GSSAPI interface
```
miekg, line 26:
```suggestion
	GSSAPI                     // An implementation of the GSSAPI interface.
```
I think we can remove the second `GSSAPI` here. While typing this I wonder if there is a better name for the interface. TsigGSS or something that points to tsig?

client.go
```diff
@@ -40,6 +41,7 @@ type Client struct {
 	ReadTimeout    time.Duration     // net.Conn.SetReadTimeout value for connections, defaults to 2 seconds - overridden by Timeout when that value is non-zero
 	WriteTimeout   time.Duration     // net.Conn.SetWriteTimeout value for connections, defaults to 2 seconds - overridden by Timeout when that value is non-zero
 	TsigSecret     map[string]string // secret(s) for Tsig map[<zonename>]<base64 secret>, zonename must be in canonical form (lowercase, fqdn, see RFC 4034 Section 6.2)
+	GSSAPI         GSSAPI            // an implementation of the GSSAPI interface
```
miekg, line 44: see previous comment

client.go
```diff
@@ -226,7 +228,7 @@ func (co *Conn) ReadMsg() (*Msg, error) {
 			return m, ErrSecret
 		}
 		// Need to work on the original message p, as that was used to calculate the tsig.
-		err = TsigVerify(p, co.TsigSecret[t.Hdr.Name], co.tsigRequestMAC, false)
+		err = tsigVerifyGSSAPI(p, co.TsigSecret[t.Hdr.Name], co.tsigRequestMAC, false, co.GSSAPI)
```
miekg, line 231: I would keep the old names and just add the new parameter

tsig.go
```diff
@@ -39,6 +40,14 @@ type TSIG struct {
 	OtherData  string `dns:"size-hex:OtherLen"`
 }
 
+// GSSAPI provides the API to plug-in RFC 3645 GSS-TSIG signed messages
```
miekg, line 43:
```suggestion
// GSSAPI provides the API to plug-in RFC 3645 GSS-TSIG signed messages.
```

client.go
bodgit: As it's an exported/public function so I didn't want to change the signature and break backwards compatibility. If that's acceptable I can update the original functions.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Add GSS-TSIG suppor..." ]
>@bodgit commented on this pull request.
>
>
>In client.go:
>
>> @@ -226,7 +228,7 @@ func (co *Conn) ReadMsg() (*Msg, error) {
>                        return m, ErrSecret
>                }
>                // Need to work on the original message p, as that was used to calculate the tsig.
>-               err = TsigVerify(p, co.TsigSecret[t.Hdr.Name], co.tsigRequestMAC, false)
>+               err = tsigVerifyGSSAPI(p, co.TsigSecret[t.Hdr.Name], co.tsigRequestMAC, false, co.GSSAPI)
>
>As it's an exported/public function so I didn't want to change the signature
>and break backwards compatibility. If that's acceptable I can update the
>original functions.

ah, right, missed the T -> t

yep, proceed :)

client.go
bodgit: I renamed the interface to `TsigGSS`.

client.go
bodgit: Based on renaming the interface to `TsigGSS` I shortened the new functions to `tsigGenerateGSS` and `tsigVerifyGSS` to match.

bodgit:
I think I've addressed all of the comments. I added an RFC mention to the list in the README, a skeletal example to doc.go and some basic testing of the interface with a dummy implementation.

tmthrgd:
I do seem to recall someone previously wanted to dynamically update or handle the `TsigSecret` field in the past, so perhaps making this a little bit more generic and allowing this to handle any algorithm including the HMAC ones if provided? Wdyt @miekg?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Add GSS-TSIG suppor..." ]
>I do seem to recall someone previously wanted to dynamically update or handle
>the TsigSecret field, so perhaps making this a little bit more generic and
>allowing this to handle any algorithm including the HMAC ones if provided? Wdyt
>@miekg?

good plan!

tmthrgd:
@bodgit Would you be willing to update this PR to make it a more general mechanism as described in https://github.com/miekg/dns/pull/1201#issuecomment-751941255? We can always come back to how it is now if that doesn’t work out.

bodgit:
The interface itself can be made more general. All that is missing is to just include the key secret, (i.e. the value for the relevant `TsigSecret` key), in the two method signatures. The GSS code doesn't use it but the HMAC code would need that.

@tmthrgd Is that what you mean?

tmthrgd:
@bodgit Not quite, I mean for it to be an optional replacement for the current system. It might still be worthwhile passing in a secret if it’s configured, but I’m not sure. Does that make sense?

bodgit:
Not entirely, if I'm honest.

I think you'd still want to keep the `TsigSecret` map, but with an accompanying `map[string]TsigGSS` (with a rename of that interface) for mapping each algorithm to its implementation. So you'd end up with something like this in the `Client`:
```golang
map[string]TsigGSS{
        HmacSHA1:   tsigHmacSHA1,
        HmacSHA224: tsigHmacSHA224,
        ...
        GSS:        ...,
}
```
Although all of the HMAC algorithms can actually share the same implementation, you just `switch` based on the algorithm like the current code does.

Is that what you mean? Perhaps it's easier for me to grasp if you have an example

tmthrgd:
@bodgit No worries, I'll try to clarify. Basically I'm suggesting something like:

```go
type Client struct {
	// ...
	TsigSecret map[string]string
	TsigProvider interface { ... }
	// ...
}
```

If `TsigProvider` (bikeshed away) is `nil`, everything operates as it does now, but if `TsigProvider` is non-`nil`, then the provided interface handles all of it. It can switch on the algorithm and select secrets however it pleases. The interface would probably look pretty much the same as the `TsigGSS` you suggested.

bodgit:
Oh I see, so this one interface gets called for any number of algorithms? In this case should the `TsigSecret` map be relied on at all or should the `TsigProvider` be expected to have that information itself?

bodgit:
Ok, I think I've done what you wanted. I've basically run `s/GSS/Provider/g` over the code and removed the various GSS mentions as there's no need for them. The `TsigProvider` interface is expected to have its own secret store (where applicable) which I think is what your intention is.

client.go
```diff
@@ -23,6 +23,7 @@ type Conn struct {
 	net.Conn                         // a net.Conn holding the connection
 	UDPSize        uint16            // minimum receive buffer for UDP messages
 	TsigSecret     map[string]string // secret(s) for Tsig map[<zonename>]<base64 secret>, zonename must be in canonical form (lowercase, fqdn, see RFC 4034 Section 6.2)
+	TsigProvider                     // An implementation of the TsigProvider interface.
```
tmthrgd, line 26: This should be `TsigProvider TsigProvider` so as not to embed it's methods in `Conn`.

client.go
```diff
@@ -40,6 +41,7 @@ type Client struct {
 	ReadTimeout    time.Duration     // net.Conn.SetReadTimeout value for connections, defaults to 2 seconds - overridden by Timeout when that value is non-zero
 	WriteTimeout   time.Duration     // net.Conn.SetWriteTimeout value for connections, defaults to 2 seconds - overridden by Timeout when that value is non-zero
 	TsigSecret     map[string]string // secret(s) for Tsig map[<zonename>]<base64 secret>, zonename must be in canonical form (lowercase, fqdn, see RFC 4034 Section 6.2)
+	TsigProvider                     // An implementation of the TsigProvider interface.
```
tmthrgd, line 44: Ditto here.

client.go
tmthrgd: Also probably the comment should be something a little more detailed and specific.

tsig.go
```diff
@@ -39,6 +39,14 @@ type TSIG struct {
 	OtherData  string `dns:"size-hex:OtherLen"`
 }
 
+// TsigProvider provides the API to plug-in a custom TSIG implementation.
+type TsigProvider interface {
+	// Generate is passed the DNS message to be signed and partial TSIG RR and returns the MAC bytes and any error.
```
tmthrgd, line 44: nit: "MAC bytes" -> "signature".

tsig.go
```diff
@@ -39,6 +39,14 @@ type TSIG struct {
...
+	Generate([]byte, *TSIG) ([]byte, error)
+	// Verify is passed the DNS message to be verified and TSIG RR and returns any error.
```
tmthrgd, line 46: This probably needs to be expanded slightly. Something explaining it should only return `nil` if the signature is valid and an error otherwise.

tsig.go
```diff
@@ -119,25 +131,33 @@ func TsigGenerate(m *Msg, secret, requestMAC string, timersOnly bool) ([]byte, s
...
-	case HmacSHA512:
-		h = hmac.New(sha512.New, rawsecret)
-	default:
-		return nil, "", ErrKeyAlg
-	}
-	h.Write(buf)
 	// Copy all TSIG fields except MAC and its size, which are filled using the computed digest.
 	*t = *rr
-	t.MAC = hex.EncodeToString(h.Sum(nil))
+	if provider == nil {
```
tmthrgd, line 136: Instead of having to special case `provider == nil` what if we provide a really simple (unexported) HMAC provider and use that throughout?

Basically this:
```go
type tsigHMACProvider string // The secret key to verify with.

func (key tsigHMACProvider) Generate(...) ([]byte, error) {
	secret, err := fromBase64([]byte(key))
	// HMAC stuff here
}

func (key tsigHMACProvider) Verify(...) { ... }

...
```

Then `TsigGenerate` and `TsigVerify` just create it unconditionally and `ReadMsg` and `WriteMsg` use it if `TsigProvider` is `nil`. I think that should make this much cleaner.

tmthrgd (commented):
This looks really good so far. It's exactly what I meant.

tmthrgd (changes requested):
Clicked the wrong option before. Some changes needed.

client.go
bodgit: Updated.

client.go
bodgit: Updated.

tsig.go
bodgit: Done.

tsig.go
bodgit: Done.

bodgit:
I realised that if the `TsigProvider` is expected to provide its own secrets, we don't need to pass the secret in the `tsigGenerateProvider` and `tsigVerifyProvider` functions so the signatures are simplified slightly. The `tsigHMACProvider` code is pretty simple as `Verify` is literally just `Generate` plus an equality check.

tsig.go
bodgit: I think that's done now.

tsig.go
```diff
@@ -98,14 +148,13 @@ type timerWireFmt struct {
 // timersOnly is false.
 // If something goes wrong an error is returned, otherwise it is nil.
 func TsigGenerate(m *Msg, secret, requestMAC string, timersOnly bool) ([]byte, string, error) {
+	return tsigGenerateProvider(m, requestMAC, timersOnly, tsigHMACProvider(secret))
+}
+
+func tsigGenerateProvider(m *Msg, requestMAC string, timersOnly bool, provider TsigProvider) ([]byte, string, error) {
```
tmthrgd, line 154: I think switch the order of parameters here, move `provider TsigProvider` between `m *Msg` and `requestMAC string`. That way the secret-related part is in the same place as it is for `TsigGenerate`.

tsig.go
```diff
@@ -156,49 +193,28 @@ func TsigGenerate(m *Msg, secret, requestMAC string, timersOnly bool) ([]byte, s
 // If the signature does not validate err contains the
 // error, otherwise it is nil.
 func TsigVerify(msg []byte, secret, requestMAC string, timersOnly bool) error {
-	return tsigVerify(msg, secret, requestMAC, timersOnly, uint64(time.Now().Unix()))
+	return tsigVerify(msg, requestMAC, timersOnly, uint64(time.Now().Unix()), tsigHMACProvider(secret))
+}
+
+func tsigVerifyProvider(msg []byte, requestMAC string, timersOnly bool, provider TsigProvider) error {
```
tmthrgd, line 199: Ditto here.

tsig.go
```diff
@@ -156,49 +193,28 @@ func TsigGenerate(m *Msg, secret, requestMAC string, timersOnly bool) ([]byte, s
...
+	return tsigVerify(msg, requestMAC, timersOnly, uint64(time.Now().Unix()), provider)
 }
 
 // actual implementation of TsigVerify, taking the current time ('now') as a parameter for the convenience of tests.
-func tsigVerify(msg []byte, secret, requestMAC string, timersOnly bool, now uint64) error {
-	rawsecret, err := fromBase64([]byte(secret))
-	if err != nil {
-		return err
-	}
+func tsigVerify(msg []byte, requestMAC string, timersOnly bool, now uint64, provider TsigProvider) error {
```
tmthrgd, line 204: And ditto here.

tsig.go
```diff
@@ -89,6 +97,48 @@ type timerWireFmt struct {
 	Fudge      uint16
 }
 
+type tsigHMACProvider string
```
tmthrgd, line 100: I think move this and it's associated methods so it comes right after `type TsigProvider interface { ... }`.

tsig.go
```diff
@@ -39,6 +39,14 @@ type TSIG struct {
 	OtherData  string `dns:"size-hex:OtherLen"`
 }
 
+// TsigProvider provides the API to plug-in a custom TSIG implementation.
+type TsigProvider interface {
```
tmthrgd, line 43: This should be moved above `type TSIG struct { ... }` so it doesn't sit before the type's methods.

tsig.go
```diff
@@ -39,6 +39,14 @@ type TSIG struct {
...
+	// Generate is passed the DNS message to be signed and the partial TSIG RR. It returns the signature and nil, otherwise an error.
+	Generate([]byte, *TSIG) ([]byte, error)
```
tmthrgd, line 45: These parameters should be named, so something like: `(msg []byte, rr *TSIG)` or whatever is appropriate. That's a little clearer for people. Same with `Verify` below.

tsig_test.go
```diff
@@ -231,9 +232,128 @@ func TestTSIGHMAC224And384(t *testing.T) {
...
-			if err = tsigVerify(msgData, tc.secret, "", false, timeSigned); err != nil {
+			if err = tsigVerify(msgData, "", false, timeSigned, tsigHMACProvider(tc.secret)); err != nil {
 				t.Error(err)
 			}
 		})
 	}
 }
+
+const (
+	testGoodKeyName = "goodkey."
```
tmthrgd, line 243: Just use the `const testGoodKeyName = "goodkey."` form here.

tsig_test.go
```diff
@@ -231,9 +232,128 @@ func TestTSIGHMAC224And384(t *testing.T) {
...
+	testGoodMAC   = []byte{0, 1, 2, 3}
+)
+
+// testProvider always generates the same MAC and only accepts the one signature
+type testProvider struct {
+	GenerateAllKeys bool
+}
+
+func (provider *testProvider) Generate(_ []byte, t *TSIG) ([]byte, error) {
+	switch {
```
tmthrgd, line 257: Use an `if` not a `switch` here.

tsig_test.go
```diff
@@ -231,9 +232,128 @@ func TestTSIGHMAC224And384(t *testing.T) {
...
+	case t.Hdr.Name == testGoodKeyName || provider.GenerateAllKeys:
+		return testGoodMAC, nil
+	default:
+		return nil, testErrBadKey
+	}
+}
+
+func (*testProvider) Verify(_ []byte, t *TSIG) error {
+	switch t.Hdr.Name {
```
tmthrgd, line 266: A `switch` would probably be fine here, but an `if` will probably be a little cleaner.

tsig_test.go
```diff
@@ -231,9 +232,128 @@ func TestTSIGHMAC224And384(t *testing.T) {
...
+	case testGoodKeyName:
+		return nil
+	default:
+		return testErrBadKey
+	}
+}
+
+func TestTsigGenerateProvider(t *testing.T) {
+	tables := map[string]struct {
```
tmthrgd, line 275: Use a slice instead of a map as the randomised map order can be annoying. Also just using `keyname` as the test name would probably be clear enough and fine.

tsig_test.go
```diff
@@ -231,9 +232,128 @@ func TestTSIGHMAC224And384(t *testing.T) {
...
+			expectedMAC := hex.EncodeToString(table.mac)
+			if mac != expectedMAC {
+				t.Fatalf("MAC doesn't match: expected '%s' but got '%s'", table.mac, expectedMAC)
+			}
+		})
+	}
+}
+
+func TestTsigVerifyProvider(t *testing.T) {
+	tables := map[string]struct {
```
tmthrgd, line 320: Ditto here.

tmthrgd (changes requested):
Just a handful of small nits to address, and then I think this looks good to me.

tsig.go
bodgit: Done. Originally I kept the order so it was `TsigGenerate()` plus a provider but with the original secret parameter changed this makes more sense.

tsig.go
bodgit: I did have it there when I first wrote it and then thought it maybe should be closer to the code that used it. I was right first time 😄

tsig.go
bodgit: Good point, sorted.

tsig.go
bodgit: Done.

tsig_test.go
bodgit: I think I might have had multiple constants at one point and `go fmt` doesn't collapse back to the single form. Sorted.

tsig_test.go
bodgit: Fixed. I think I originally had multiple conditions.

tsig_test.go
bodgit: Fixed.

tsig_test.go
bodgit: Fixed.

tmthrgd (approved):
Thank you for addressing all the comments. This LGTM and I think it's a worthwhile addition.

tmthrgd:
I have no idea "CIFuzz / Fuzzing (pull_request)" is failing and the failure message isn't remotely helpful. I don't think it's related to this PR though. I did try rerunning it, thinking it might just be ephemeral, but that didn't seem to help. Not sure if we should try and diagnose that or presume it's unrelated and merge anyway.

bodgit:
I noticed it was failing on some of the other PR's so I don't believe it was something I broke. I couldn't figure out what the immediate breakage was, looks like Python trying to reference a nonexistent dictionary key.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Add GSS-TSIG suppor..." ]
>@tmthrgd approved this pull request.
>
>Thank you for addressing all the comments. This LGTM and I think it's a
>worthwhile addition.

thanks all, I'll just merge then :)

/merge

PR #1202, opened 2020-12-17
Fix comment about SVCB values in AliasMode

janik-cloudflare:
When the priority is zero, the record is in AliasMode and so the value must be empty. This changes the comment accordingly.


tmthrgd (approved):
LGTM.

PR #1203, opened 2020-12-18
Export MakeSVCBKeyValue

janik-cloudflare:
This would be very helpful for deserializing `SVCBKeyValue`.

Because `SVCBKeyValue` is an interface, it's currently not possible to deserialize structs like `SVCBAlpn` if we don't already know their concrete type. This change would allow us to serialize the `SVCBKey` and use it to instantiate the appropriate type when deserializing.

Currently, `Key()` is exported which allows us to get the `SVCBKey` from an `SVCBKeyValue`, but it's not possible to create a new `SVCBKeyValue` from a `SVCBKey`.

Thank you!

### DIFF
--- a/msg_helpers.go
+++ b/msg_helpers.go
@@ -627,7 +627,7 @@ func unpackDataSVCB(msg []byte, off int) ([]SVCBKeyValue, int, error) {
 		if err != nil || off+int(length) > len(msg) {
 			return nil, len(msg), &Error{err: "overflow unpacking SVCB"}
 		}
-		e := makeSVCBKeyValue(SVCBKey(code))
+		e := MakeSVCBKeyValue(SVCBKey(code))
 		if e == nil {
 			return nil, len(msg), &Error{err: "bad SVCB key"}
 		}
--- a/svcb.go
+++ b/svcb.go
@@ -149,7 +149,7 @@ func (rr *SVCB) parse(c *zlexer, o string) *ParseError {
 					}
 				}
 			}
-			kv := makeSVCBKeyValue(svcbStringToKey(key))
+			kv := MakeSVCBKeyValue(svcbStringToKey(key))
 			if kv == nil {
 				return &ParseError{l.token, "bad SVCB key", l}
 			}
@@ -173,8 +173,8 @@ func (rr *SVCB) parse(c *zlexer, o string) *ParseError {
 	return nil
 }
 
-// makeSVCBKeyValue returns an SVCBKeyValue struct with the key or nil for reserved keys.
-func makeSVCBKeyValue(key SVCBKey) SVCBKeyValue {
+// MakeSVCBKeyValue returns an SVCBKeyValue struct with the key or nil for reserved keys.
+func MakeSVCBKeyValue(key SVCBKey) SVCBKeyValue {
 	switch key {
 	case SVCB_MANDATORY:
 		return new(SVCBMandatory)
--- a/svcb_test.go
+++ b/svcb_test.go
@@ -27,7 +27,7 @@ func TestSVCB(t *testing.T) {
 
 	for _, o := range svcbs {
 		keyCode := svcbStringToKey(o.key)
-		kv := makeSVCBKeyValue(keyCode)
+		kv := MakeSVCBKeyValue(keyCode)
 		if kv == nil {
 			t.Error("failed to parse svc key: ", o.key)
 			continue
@@ -87,7 +87,7 @@ func TestDecodeBadSVCB(t *testing.T) {
 		},
 	}
 	for _, o := range svcbs {
-		err := makeSVCBKeyValue(SVCBKey(o.key)).unpack(o.data)
+		err := MakeSVCBKeyValue(SVCBKey(o.key)).unpack(o.data)
 		if err == nil {
 			t.Error("accepted invalid svc value with key ", SVCBKey(o.key).String())
 		}


tmthrgd:
Can you show an example of where this would be useful?

miekg:
As with any public API expansion, this will require a strong reason.

janik-cloudflare:
Yes, definitely. We serialize and deserialize RRs, which is currently possible for all types except SVCB/HTTPS.

Because SVCB/HTTPS records contain an array of `SVCBKeyValue` (which is an interface), we need a way to store the concrete type (e.g. `SVCBAlpn`) of each `SVCBKeyValue`. This can be done with the `.Key()` method. However, when deserializing, there's currently no public API that I'm aware of that would convert the key back to, say, an `SVCBAlpn` variable that we can deserialize.

Please consider the following example with `msgpack` (of course, this also applies to `json` etc.).

```
package main

import "github.com/miekg/dns"
import "github.com/vmihailenco/msgpack"
import "reflect"
import "fmt"

func main() {
	init, err := dns.NewRR("example.net. 3600 IN SVCB 1 example.com. alpn=\"h3,h2\" ipv6hint=\"::1\"")
	if err != nil {
		fmt.Println("init failed")
		return
	}

	encoded, err := msgpack.Marshal(init)
	if err != nil {
		fmt.Println("marshal failed")
		return
	}

	var decoded *dns.SVCB
	err = msgpack.Unmarshal(encoded, &decoded)
	if err != nil {
		fmt.Println("unmarshal failed")
		return
	}

	fmt.Printf("old: %v\nnew: %v\n", init, decoded)
}

func init() {
	msgpack.Register([]dns.SVCBKeyValue{}, encodeArrayOfSVCBKeyValue, decodeArrayOfSVCBKeyValue)
}

func encodeArrayOfSVCBKeyValue(enc *msgpack.Encoder, v reflect.Value) error {
	ary := v.Interface().([]dns.SVCBKeyValue)
	num := len(ary)

	if num > 0xFF {
		return fmt.Errorf("too many elements in []SVCBKeyValue: %d", num)
	}

	// Encode array length
	if err := enc.Encode(uint8(num)); err != nil {
		return err
	}

	// Encode each element
	for _, kv := range ary {
		// Encode the key
		if err := enc.Encode(kv.Key()); err != nil {
			return err
		}

		// Encode the value
		if err := enc.Encode(kv); err != nil {
			return err
		}
	}

	return nil
}

func decodeArrayOfSVCBKeyValue(dec *msgpack.Decoder, v reflect.Value) error {
	var num uint8

	// Decode array length
	if err := dec.Decode(&num); err != nil {
		return err
	}

	ary := make([]dns.SVCBKeyValue, num)

	// Decode each element
	for i := uint8(0); i < num; i++ {
		// Decode the key
		var key dns.SVCBKey
		if err := dec.Decode(&key); err != nil {
			return err
		}

		// Decode the value
		// XXX: Requires exported MakeSVCBKeyValue
		val := dns.MakeSVCBKeyValue(key)
		if err := dec.Decode(val); err != nil {
			return err
		}

		ary[i] = val
	}

	v.Set(reflect.ValueOf(ary))
	return nil
}
```

This has a custom encoder/decoder for `[]SVCBKeyValue` that (de)serializes the type of each `SVCBKeyValue`.The decoder (`decodeArrayOfSVCBKeyValue`) calls the currently unexported `MakeSVCBKeyValue`.

Without this custom encoder/decoder implementation, the `SVCBAlpn` would be serialized as something like `{"alpn":["h3","h2"]}` and the deserializer would not know how to convert that back to a `SVCBKeyValue` (because it cannot know that the correct type to deserialize into is `SVCBAlpn`). In the case of `msgpack`, attempting to run the code above without the custom encoder/decoder results in this error:

```
panic: reflect.Set: value of type map[string]interface {} is not assignable to type dns.SVCBKeyValue
```

I don't think it's possible to safely deserialize interfaces like this without making the deserializer aware of the possible concrete types, and I couldn't find a way to do this with the current public interface.

Thank you for considering this!

DesWurstes:
I couldn't find another way. That's also what this library does with `TypeToRR`, the SVCB parsing code and `makeDataOpt` internally.

> We serialize and deserialize RRs, which is currently possible for **all types** except SVCB/HTTPS.

EDNS0 needs a similar solution too, or how do you currently solve this [very same problem](https://github.com/miekg/dns/blob/0972db68349b0635ca91da9ee82cf0d795dacb30/msg_helpers.go#L441)?

janik-cloudflare:
@DesWurstes Nice find! We don't currently support EDNS0 records in the code that requires (de)serialization. I think you're correct, we would need that function too if we wanted to support EDNS0 records.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Export MakeSVCBKeyV..." ]
>@DesWurstes Nice find! We don't currently support EDNS0 records in the code
>that requires (de)serialization. I think you're correct, we would need that
>function too if we wanted to support EDNS0 records.

Making the EDNS0 functionality public is a long standing issue. We merged a PR once did
that, but that led to mysterious failures, so we reverted. As said then, the EDNS0
interface isn't pretty and should have a cleaner public API. So for SCVB/HTTPS we should
find the right way to export functionality.

tmthrgd:
While I’m very sympathetic to your/Cloudflare’s goal here, I don’t think this really carries its weight. As far as I can tell this would only be useful when deserialising values with something like encoding/json, which I don’t believe will be all that common.

It also seems like it could cause problems deserialising if a concrete type is added for some future option that previously serialised as SVCBLocal.

I think for now just copy pasting this code is the right solution. That’s definitely something that can be reconsidered in the future though.

janik-cloudflare:
Yes, I think that makes sense. Let's keep this unexported for now. This function is only a small part of the serialization logic anyway, and like @tmthrgd said I think it would make sense to handle `SVCBLocal` conversions properly if we're going to officially support this here.

Thank you for looking at this!

Issue #1204, opened 2020-12-26
Data function for RR


Kloenk:
Currently the RR interface only has a String() function which also prints the Header. IMHO there should also be a function which prints only the Data as string, without the header informations.

miekg:
Every RR has public struct members that allow you to extract this information with the correct types. If you only care about the rdata a strings you can just convert to string and index that at the correct position.

Kloenk:
I found the type, but then I have to create a switch case for every possible type.
Npt sure what you mean with index at the correct position, how would I get this index?

aaranmcguire:
I do like the idea of this PR as it would clean up some of my logic, but if we don't want to do this, the easy way of just getting the data is:

```
strings.Replace(rr.String(), rr.Header().String(), "", 1)
```

Works for all types, so no need to make a switch case.

miekg:
Yes you either need to type switch or do what Aaran says. There is no need
to add this code to the lib, easy to do outside of it.

/close

PR #1205, opened 2020-12-26
Data function on RR to get content

Kloenk:
Create a Data function on the RR interface to get the content of the RR without the header.

Fixes: #1204


Kloenk:
I don't understand the output of the CIFuzz. Maybe someone can help me with that?

PR #1208, opened 2021-01-22
Implementation of zone digest (ZONEMD)

madestro:
https://tools.ietf.org/html/draft-ietf-dnsop-dns-zone-digest-14


miekg:
ah nice. Can you clean up this PR and remove any non go files? Also revert the go.mod updates

miekg:
also needs a bit of tests

tmthrgd (changes requested):
Unrelated changes need to be removed.

generate.go
```diff
@@ -188,7 +188,7 @@ func (r *generateReader) ReadByte() (byte, error) {
 			if errMsg != "" {
 				return 0, r.parseError(errMsg, si+3+sep)
 			}
-			if r.start+offset < 0 || int64(r.end) + int64(offset) > 1<<31-1 {
+			if r.start+offset < 0 || int64(r.end)+int64(offset) > 1<<31-1 {
```
tmthrgd, line 191: This change is unrelated.

go.mod
```diff
@@ -3,9 +3,9 @@ module github.com/miekg/dns
 go 1.12
 
 require (
-	golang.org/x/crypto v0.0.0-20191011191535-87dc89f01550
-	golang.org/x/net v0.0.0-20190923162816-aa69164e4478
-	golang.org/x/sync v0.0.0-20190423024810-112230192c58
-	golang.org/x/sys v0.0.0-20190924154521-2837fb4f24fe
-	golang.org/x/tools v0.0.0-20191216052735-49a3e744a425 // indirect
+	golang.org/x/crypto v0.0.0-20200622213623-75b288015ac9
```
tmthrgd, line 6: The changes to go.mod and go.sum are unrelated.

scan_rr.go
```diff
@@ -846,6 +846,42 @@ func (rr *CSYNC) parse(c *zlexer, o string) *ParseError {
...
+	if e1 != nil || l.err {
+		return &ParseError{"", "bad ZONEMD Scheme", l}
+	}
+	rr.Scheme = uint8(i)
+
+	c.Next() // zBlank
+	l, _ = c.Next()
+	if i, err := strconv.ParseUint(l.token, 10, 8); err != nil {
+		tokenUpper := strings.ToUpper(l.token)
+		i, ok := StringToAlgorithm[tokenUpper]
```
tmthrgd, line 869: This is wrong. The draft specifies only two supported algorithms SHA384 with value 1 and SHA512 with value 2; that doesn't match StringToAlgorithm. The "ZONEMD Hash Algorithms" will be a separate registry to the one used by DNSSEC.

Also, the draft doesn't specific the use of the mnemonic's here at all, instead only "unsigned decimal integer" are allowed. This should just be a parse error.

types.go
```diff
@@ -1361,6 +1361,23 @@ func (rr *CSYNC) len(off int, compression map[string]struct{}) int {
 	return l
 }
 
+// ZONEMD RR, from draft-ietf-dnsop-dns-zone-digest
+type ZONEMD struct {
+	Hdr    RR_Header
+	Serial uint32
+	Scheme uint8
```
tmthrgd, line 1368: There should probably be constants for this and Hash.

tmthrgd (changes requested):

scan_rr.go
madestro: you are right, we are removing that part

types.go
madestro: done in #edcbb0b

go.mod
eriverosr: it should be fixed in #e62bf83

tmthrgd: I'm still seeing the go.mod and go.sum changes in the PR diff.

scan_rr.go
```diff
@@ -846,6 +846,70 @@ func (rr *CSYNC) parse(c *zlexer, o string) *ParseError {
 	return nil
 }
 
+func (rr *ZONEMD) parse(c *zlexer, o string) *ParseError {
+	l, _ := c.Next()
+	if (l.token == "\\#") {
```
tmthrgd, line 851: This is definitely not the correct place to do this nor the correct pull request to do this in, please back this out.

types.go
```diff
@@ -148,6 +149,22 @@ const (
...
 	OpcodeNotify = 4
 	OpcodeUpdate = 5
+
+)
+
+// Used in ZONEMD https://tools.ietf.org/html/draft-ietf-dnsop-dns-zone-digest-14
+
+const (
+	// ZoneMD Accepted Schemes
+	ZoneMDSchemeReserved = 0
```
tmthrgd, line 159: I don't think we need the reserved constants, they're really just noise.

tmthrgd (changes requested):

scan_rr.go
madestro: where to, then?

types.go
madestro: disagree

scan_rr.go
tmthrgd: It looks like we only support RFC3597 in `(*ZoneParser).Next` if we don't recognise the RR type, that does seem like a bug to me. That bug needs to be fixed in the ZoneParser to work for *all* record types using the existing wire format parsing functions not here in an ad-hoc way for this specific one.

types.go
tmthrgd: The usable constants were, the reserved ones serve no purpose and have no practical use whatsoever. It's not merely that they take up extra lines, they also appear in the documentation, require long term compatibility and impose a burden on developers understanding the code or documentation.

scan_rr.go
tmthrgd: I've opened #1211 to add support for this.

miekg (approved):

miekg:
think all requested changes have been implemented.

miekg:
wait, where are the tests?

eriverosr:
I've just added a test that checks if the library is able to parse the example ZONEMD RRs from [RFC 8976](https://tools.ietf.org/html/rfc8976). Tell us if you can think of any other type of tests to add.

parse_test.go
```diff
@@ -1921,3 +1921,75 @@ func TestUnpackRRWithHeaderInvalidLengths(t *testing.T) {
...
+		rr, err := NewRR(i)
+		if err != nil {
+			t.Error("failed to parse RR: ", err)
+			continue
+		}
+		if rr.String() != o {
+			t.Errorf("`%s' should be equal to\n`%s', but is     `%s'", i, o, rr.String())
+		}
+	}
+
```
miekg, line 1994:
```suggestion
```

miekg (approved):

PR #1214, opened 2021-01-31
Fix Rdlength related parsing bug in RFC3597 records

tmthrgd:
`(*RFC3597).fromRFC3597` failed to set `Rdlength` on the RR header. This is used (since yesterday) when parsing records from a zone file in the RFC 3597 "Unknown DNS Resource Record Type" format.

Also while we're here, validate the length doesn't exceed the maximum RDATA length when parsing in that format and harden `UnpackRRWithHeader`.

Updates #1211 (my bad)

/cc @catenacyber

### DIFF
--- a/dns.go
+++ b/dns.go
@@ -136,7 +136,10 @@ func (rr *RFC3597) ToRFC3597(r RR) error {
 func (rr *RFC3597) fromRFC3597(r RR) error {
 	hdr := r.Header()
 	*hdr = rr.Hdr
-	hdr.Rdlength = uint16(len(rr.Rdata) / 2)
+
+	// Can't overflow uint16 as the length of Rdata is validated in (*RFC3597).parse.
+	// We can only get here when rr was constructed with that method.
+	hdr.Rdlength = uint16(hex.DecodedLen(len(rr.Rdata)))
 
 	if noRdata(*hdr) {
 		// Dynamic update.
--- a/dns.go
+++ b/dns.go
@@ -134,9 +134,11 @@ func (rr *RFC3597) ToRFC3597(r RR) error {
 
 // fromRFC3597 converts an unknown RR representation from RFC 3597 to the known RR type.
 func (rr *RFC3597) fromRFC3597(r RR) error {
-	*r.Header() = rr.Hdr
+	hdr := r.Header()
+	*hdr = rr.Hdr
+	hdr.Rdlength = uint16(len(rr.Rdata) / 2)
 
-	if len(rr.Rdata) == 0 {
+	if noRdata(*hdr) {
 		// Dynamic update.
 		return nil
 	}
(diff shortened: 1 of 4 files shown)


tmthrgd:
It's late here so if people could give it a good looking over, that would be appreciated.

msg.go
```diff
@@ -624,18 +624,25 @@ func UnpackRRWithHeader(h RR_Header, msg []byte, off int) (rr RR, off1 int, err
...
+		return rr, off, nil
+	}
 
 	off, err = rr.unpack(msg, off)
 	if err != nil {
 		return nil, end, err
 	}
 	if off != end {
-		return &h, end, &Error{err: "bad rdlength"}
+		return nil, end, &Error{err: "bad rdlength"}
```
miekg, line 645: I think the idea was to leave the header returned so the caller could still do something with it. This is also probable part of the assumed API of this lib nowdays

dns.go
```diff
@@ -134,9 +134,11 @@ func (rr *RFC3597) ToRFC3597(r RR) error {
 
 // fromRFC3597 converts an unknown RR representation from RFC 3597 to the known RR type.
 func (rr *RFC3597) fromRFC3597(r RR) error {
-	*r.Header() = rr.Hdr
+	hdr := r.Header()
+	*hdr = rr.Hdr
+	hdr.Rdlength = uint16(len(rr.Rdata) / 2)
```
miekg, line 139: this is the crucial line. The other changes might be helping, but backing them out and focusing on the minimal fix it seems to boil down to:
~~~
*r.Header() = rr.Hdr
r.Header().Rdlength = uint16(len(rr.Rdata) / 2)
~~~
which probably also requires a comment about the /2 along the lines of 2 hex digits for every octet (or some such)

msg.go
```diff
@@ -624,18 +624,25 @@ func UnpackRRWithHeader(h RR_Header, msg []byte, off int) (rr RR, off1 int, err
...
-	if noRdata(h) {
-		return rr, off, nil
+	if off < 0 {
+		return nil, off, &Error{err: "off must be positive"}
 	}
 
 	end := off + int(h.Rdlength)
+	if end < off || end > len(msg) {
+		return nil, end, &Error{err: "bad rdlength"}
+	}
```
miekg, line 634: this might still be a good addition, but do we have something that hits this?

miekg:
Hi @tmthrgd thanks for working on this. I think the most minimal fix is just 1 line, left couple of other comments as well.

The tests are def. /lgtm

msg.go
tmthrgd: I've reverted the changes to `UnpackRRWithHeader` and will propose them separately.

msg.go
tmthrgd: It's definitely a code path that can be triggered so it is needed, but I'll propose it separately with proper tests.

dns.go
tmthrgd: The changes to `(*RFC3597).parse` are also important, otherwise it can overflow a `uint16`. I've updated it to be a bit clearer. It should be minimal enough now.

miekg (approved):

PR #1215, opened 2021-02-01
Validate Rdlength and off in UnpackRRWithHeader

tmthrgd:
This was split out of #1214.

I don't think this can be hit from `UnpackRR` as `unpackHeader` does validate it doesn't exceed `msg`, but it can definitely be hit by calling `UnpackRRWithHeader` directly.


tmthrgd:
The `off` check doesn't seem to be done anywhere else `off` is taken so I'm not sure whether it should be introduced here or not, happy to back that out.

miekg (approved):

PR #1217, opened 2021-02-08
Update Truncate doc with compress behaviour

### LINKED ISSUE #1216: CompressAndTruncate
Currently the [Truncate function](https://github.com/miekg/dns/blob/35023fab5c1d57c9ac56da33a897af8794eb9567/msg_truncate.go#L24) disables compression, unless the does not fit in `size` bytes.
E.g after calling `msg.Truncate(1234)`, `msg.Compress` will be false when returning if its uncompressed length is < 1234. Even though the caller may have set msg.Compress to true explicitly before.

As much as I do understand the reasoning, I am not sure why this is forced on the user of the library. 

What would be the right way to support both use cases? Would simply removing https://github.com/miekg/dns/blob/35023fab5c1d57c9ac56da33a897af8794eb9567/msg_truncate.go#L41-L42 be OK? 
But this would change the current behaviour though, on the other hand, if someone had set msg.Compress before hand, they are probably not replying what they expect to reply.

Should another function be created to allow to Truncate without side effect on the dns.Compress attribute?
As much as I understand the attempt to compress before truncating, I can also see people that would rather truncate and not compress...

What would you see as the best middle ground here?

chantra:
This is a documentation update to highlight the behaviour of Truncate, which will reset dns.Compress to false when the message fits in the requested size without truncation, and make it the caller responsibility to set it back to true if they wish to compress, regardless of fitting, uncompressed, in the requested message size in the first place or not.
Fixes #1216

### DIFF
--- a/msg_truncate.go
+++ b/msg_truncate.go
@@ -8,6 +8,11 @@ package dns
 // record adding as many records as possible without exceeding the
 // requested buffer size.
 //
+// Note that if the message fits within the requested size without compression,
+// Truncate will reset the message's `Compress` attribute to `false`. It is
+// the caller responsibility to set it back to `true` if they wish to compress
+// the payload regardless.
+//
 // The TC bit will be set if any records were excluded from the message.
 // If the TC bit is already set on the message it will be retained.
 // TC indicates that the client should retry over TCP.
--- a/msg_truncate.go
+++ b/msg_truncate.go
@@ -8,10 +8,10 @@ package dns
 // record adding as many records as possible without exceeding the
 // requested buffer size.
 //
-// Note that if the message fits within the requested size without compression,
-// Truncate will reset the message's `Compress` attribute to `false`. It is
-// the caller responsibility to set it back to `true` if they wish to compress
-// the payload regardless.
+// If the message fits within the requested size without compression,
+// Truncate will set the message's Compress attribute to false. It is
+// the caller's responsibility to set it back to true if they wish to
+// compress the payload regardless.
 //
 // The TC bit will be set if any records were excluded from the message.
 // If the TC bit is already set on the message it will be retained.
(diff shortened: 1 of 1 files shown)


msg_truncate.go
```diff
@@ -8,6 +8,11 @@ package dns
 // record adding as many records as possible without exceeding the
 // requested buffer size.
 //
+// Note that if the message fits within the requested size without compression,
```
tmthrgd, line 11: I don't think this needs to say 'Note', just 'If the message...'

msg_truncate.go
```diff
@@ -8,6 +8,11 @@ package dns
...
+// Truncate will reset the message's `Compress` attribute to `false`. It is
```
tmthrgd, line 12: s/reset/set/

msg_truncate.go
```diff
@@ -8,6 +8,11 @@ package dns
...
+// the caller responsibility to set it back to `true` if they wish to compress
```
tmthrgd, line 13: s/caller/caller's/

msg_truncate.go
tmthrgd, line 12: Also drop the backquotes, they don't appear widely in the documentation so far.

tmthrgd (changes requested):
Minor nits, otherwise seems fine.

msg_truncate.go
```diff
@@ -8,6 +8,11 @@ package dns
 // record adding as many records as possible without exceeding the
 // requested buffer size.
 //
+// If the message fits within the requested size without compression,
+// Truncate will set the message's Compress attribute to false. It is
+// the caller's responsibility to set it back to true if they wish to
+// compress the payload regardless.
```
tmthrgd, line 14: s/regardless/regardless of size/

tmthrgd (changes requested):
One very minor nit, then LGTM.

tmthrgd (approved):
Thanks for your patience. LGTM.

PR #1218, opened 2021-02-09
TCP Query Pipelining and TCP buffers pool

ameshkov:
RFC 7766 recommends implementing query pipelining, ie. process all incoming
queries concurrently and write responses out-of-order.

This pull request changes TCP server behavior to match this recommendation.
Aside from that, there are two more changes:
1. In order to do fewer allocations, I added a sync.Pool for TCP buffers.
2. Default value for MaxTCPQueries has been changed to -1 (unlimited). Our tests with AdGuard DNS show that limiting the number of queries per connection does not save the server resources. On the contrary, it leads to increased load since clients have to re-connect, and the total number of active connections stays pretty much the same.

### DIFF
--- a/server.go
+++ b/server.go
@@ -15,7 +15,8 @@ import (
 )
 
 // Default maximum number of TCP queries before we close the socket.
-const maxTCPQueries = 128
+// By default, it is unlimited.
+const maxTCPQueries = -1
 
 // aLongTimeAgo is a non-zero time, far in the past, used for
 // immediate cancelation of network operations.
@@ -205,6 +206,9 @@ type Server struct {
 	// Default buffer size to use to read incoming UDP messages. If not set
 	// it defaults to MinMsgSize (512 B).
 	UDPSize int
+	// Default buffer size to use to read incoming TCP messages. If not set
+	// it defaults to MinMsgSize (512 B).
+	TCPSize int
 	// The net.Conn.SetReadTimeout value for new connections, defaults to 2 * time.Second.
 	ReadTimeout time.Duration
 	// The net.Conn.SetWriteTimeout value for new connections, defaults to 2 * time.Second.
@@ -236,6 +240,9 @@ type Server struct {
 
 	// A pool for UDP message buffers.
 	udpPool sync.Pool
+
+	// A pool for TCP message buffers.
+	tcpPool sync.Pool
 }
 
 func (srv *Server) isStarted() bool {
@@ -245,7 +252,7 @@ func (srv *Server) isStarted() bool {
 	return started
 }
 
-func makeUDPBuffer(size int) func() interface{} {
+func makePacketBuffer(size int) func() interface{} {
 	return func() interface{} {
 		return make([]byte, size)
 	}
@@ -258,6 +265,9 @@ func (srv *Server) init() {
 	if srv.UDPSize == 0 {
 		srv.UDPSize = MinMsgSize
 	}
+	if srv.TCPSize == 0 {
+		srv.TCPSize = MinMsgSize
+	}
 	if srv.MsgAcceptFunc == nil {
 		srv.MsgAcceptFunc = DefaultMsgAcceptFunc
 	}
@@ -265,7 +275,8 @@ func (srv *Server) init() {
 		srv.Handler = DefaultServeMux
 	}
 
-	srv.udpPool.New = makeUDPBuffer(srv.UDPSize)
+	srv.udpPool.New = makePacketBuffer(srv.UDPSize)
+	srv.tcpPool.New = makePacketBuffer(srv.TCPSize)
 }
 
 func unlockOnce(l sync.Locker) func() {
@@ -549,13 +560,18 @@ func (srv *Server) serveTCPConn(wg *sync.WaitGroup, rw net.Conn) {
 		limit = maxTCPQueries
 	}
 
+	var wgCo sync.WaitGroup
+
 	for q := 0; (q < limit || limit == -1) && srv.isStarted(); q++ {
 		m, err := reader.ReadTCP(w.tcp, timeout)
+
 		if err != nil {
 			// TODO(tmthrgd): handle error
 			break
 		}
-		srv.serveDNS(m, w)
+
+		wgCo.Add(1)
+		go srv.serveTCPPacket(&wgCo, m, w)
 		if w.closed {
 			break // Close() was called
 		}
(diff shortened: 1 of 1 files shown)


server.go
```diff
@@ -666,8 +692,15 @@ func (srv *Server) readTCP(conn net.Conn, timeout time.Duration) ([]byte, error)
 		return nil, err
 	}
 
-	m := make([]byte, length)
+	m := srv.tcpPool.Get().([]byte)
+	if int(length) > len(m) {
+		srv.tcpPool.Put(m)
+		return nil, ErrBuf
```
rdrozhdzh, line 698: wouldn't it be better to allocate new buffer here instead of dropping the query and eventually closing the connection?

miekg:
disconnecting clients is to open up ports on the server side, now a couple of rogue clients can easily fill up all ports. Another side effect is that clients learn to reconnect.

How is this different from `go srv.serveTCPConn(&wg, rw)`? It's doesn't take the entire `conn` with it?

Would be nice to test this is as well.

It also breaks the symmetry with UDP, which I don't like, so that probably warrants some fixes

rdrozhdzh:
This won't work without a change like #1200

ameshkov:
> This probably won't work without a change like #1200

Yeah, #1200 is definitely needed for this to work properly with TLS.

ameshkov:
> How is this different from go srv.serveTCPConn(&wg, rw)? It's doesn't take the entire conn with it?

It processes just one DNS query and continues to read new DNS queries from the connection.
The current implementation waits for the DNS query to be resolved and only then it gets to the next query.

> disconnecting clients is to open up ports on the server side, now a couple of rogue clients can easily fill up all ports. Another side effect is that clients learn to reconnect.

Tbh, I don't see what's the difference with the current implementation in regards to this PR.

> Would be nice to test this is as well.

Do you mean to add a unit-test for that?

server.go
ameshkov: Tbh, I am not sure about that. On one hand, you're right, we may want to process ALL queries. On the other hand, we would need to allocate a 64kb buffer that will end up in the `sync.Pool` which is an undesirable outcome.

Also, despite that DNS queries aren't limited by 512 bytes by the RFC, I hardly can imagine a situation when a DNS query is bigger than that.

ameshkov:
@rdrozhdzh 

Regarding #1200, what about a solution like that:

```
	if _, ok := co.Conn.(*net.TCPConn); ok {
		l := make([]byte, 2)
		binary.BigEndian.PutUint16(l, uint16(len(p)))
		n, err := (&net.Buffers{l, p}).WriteTo(co.Conn)
		return int(n), err
	}

	msg := make([]byte, 2+len(p))
	binary.BigEndian.PutUint16(msg, uint16(len(p)))
	copy(msg[2:], p[:])
	return co.Conn.Write(msg)
```

This way we would avoid making extra allocations when they're not needed, but on the other hand, we'll improve the situation with DOT a little bit. If you're okay with this approach, I can change the code accordingly and come up with a unit test to check this.

tmthrgd:
There are multiple separate changes here which is awkward to review. Could you separate them into individual PRs please?

ameshkov:
@tmthrgd np, I'll try to separate it.

Just one quick question before that - I guess I should start with improving #1200, my PR won't work properly without it in the case of TLS. Does the [proposed approach](https://github.com/miekg/dns/pull/1218#issuecomment-775919830) make sense to you?

server.go
rdrozhdzh: > we would need to allocate a 64kb buffer

you can allocate the buffer of the size exactly needed for reading this particular message, since you already know the message size.

> that will end up in the sync.Pool which is an undesirable outcome.

you don't need to put the big buffers to the pool, i.e. check the buffer capacity before putting back.

rdrozhdzh:
> 
> 
> @rdrozhdzh
> 
> Regarding #1200, what about a solution like that:
> 
> ```
> 	if _, ok := co.Conn.(*net.TCPConn); ok {
> 		l := make([]byte, 2)
> 		binary.BigEndian.PutUint16(l, uint16(len(p)))
> 		n, err := (&net.Buffers{l, p}).WriteTo(co.Conn)
> 		return int(n), err
> 	}
> 
> 	msg := make([]byte, 2+len(p))
> 	binary.BigEndian.PutUint16(msg, uint16(len(p)))
> 	copy(msg[2:], p[:])
> 	return co.Conn.Write(msg)
> ```
> 
> This way we would avoid making extra allocations when they're not needed, but on the other hand, we'll improve the situation with DOT a little bit. If you're okay with this approach, I can change the code accordingly and come up with a unit test to check this.

I would suggest to add the func `PackWithLength()` like below, and use it for TCP/TLS connections
```go
func (dns *Msg) PackWithLength() ([]byte, error) {
  packLen := msgLenWithCompressionMap(dns, nil) + 1
  buf = make([]byte, packLen +2)
  msg, err := dns.PackBuffer(buf[2:])
  msgLen := len(msg)
  binary.BigEndian.PutUint16(buf, uint16(msgLen))
  return buf[:msgLen+2], err
}
```
This logic is based on the code from https://github.com/miekg/dns/blob/35023fab5c1d57c9ac56da33a897af8794eb9567/msg.go#L720-L825

> If you're okay with this approach

My opinion is not decisive here, you need an approval from @miekg or @tmthrgd

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] TCP Query Pipelinin..." ]
>There are multiple separate changes here which is awkward to review. Could you
>separate them into individual PRs please?

splitting this sgtm. Let's close this and discuss separately.

mzealey:
I implemented something similar and because of the double-packet issue that #1219 addressed (& I previously did an attempt at in #1200) I had to wrap the connection in a mutex like:


```
+type MutexWriter struct {
+       dns.Writer
+       mu       sync.RWMutex
+}
+func NewMutexWriter(w dns.Writer) dns.Writer {
+       return &MutexWriter{Writer: w}
+}
+
+func (w *MutexWriter) Write(m []byte) (int, error) {
+       w.mu.Lock()
+       defer w.mu.Unlock()
+       return w.Writer.Write(m)
+}
...
DecorateWriter: NewMutexWriter
```

I'm not totally sure if this would still be required or not iirc in load-testing with a large number of cpus, even with #1200 I had to have this mutex lock in place on the conn.

miekg:
Note https://github.com/dnsfilter/dns-multiplex is a repo that needed something similar and that's working code. I haven't looked at it yet, but it's good to safe the link to it and maybe we can import bits and pieces.

mzealey:
So I can see that the TCP pool aspect of this PR went into (as yet unmerged) #1221, but I cannot see any PR for the multiplex side of things at https://github.com/miekg/dns/pull/1218/files#diff-366e46a40f6f60b4f7614eb0976bb51820364bf5ca6ccc4787eb49d7bdbef3e6R561-R608 ?

ameshkov:
Well, I planned to make the next PR after #1221 gets some attention.

PR #1219, opened 2021-02-09
Send DNS query in one packet when using TCP/TLS

### LINKED ISSUE #1199: TLS connections send packet length & packet data seperately

ameshkov:
Attempt to improve #1200.
This fixes #1199 and this is also crucial for #1218.

This way we'll continue to use `net.Buffers` for TCP connections and only switch to allocating a new buffer for TLS.
Regarding additional unit-tests, it seems that both code branches are already covered by existing unit-tests.

### DIFF
--- a/client.go
+++ b/client.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"runtime"
 	"strings"
 	"time"
 )
@@ -340,8 +341,8 @@ func (co *Conn) Write(p []byte) (int, error) {
 		return co.Conn.Write(p)
 	}
 
-	if _, ok := co.Conn.(*net.TCPConn); ok {
-		// Limit usage of net.Buffers to net.TCPConn only
+	if _, ok := co.Conn.(*net.TCPConn); ok && runtime.GOOS == "linux" {
+		// Limit usage of net.Buffers to net.TCPConn on Linux only
 		// In the case of TLS using it results in splitting it
 		// into two packets
 
@@ -354,7 +355,7 @@ func (co *Conn) Write(p []byte) (int, error) {
 
 	msg := make([]byte, 2+len(p))
 	binary.BigEndian.PutUint16(msg, uint16(len(p)))
-	copy(msg[2:], p[:])
+	copy(msg[2:], p)
 	return co.Conn.Write(msg)
 }
 
--- a/client.go
+++ b/client.go
@@ -9,7 +9,6 @@ import (
 	"fmt"
 	"io"
 	"net"
-	"runtime"
 	"strings"
 	"time"
 )
@@ -339,18 +338,6 @@ func (co *Conn) Write(p []byte) (int, error) {
 
 	if _, ok := co.Conn.(net.PacketConn); ok {
 		return co.Conn.Write(p)
-	}
-
-	if _, ok := co.Conn.(*net.TCPConn); ok && runtime.GOOS == "linux" {
-		// Limit usage of net.Buffers to net.TCPConn on Linux only
-		// In the case of TLS using it results in splitting it
-		// into two packets
-
-		l := make([]byte, 2)
-		binary.BigEndian.PutUint16(l, uint16(len(p)))
-
-		n, err := (&net.Buffers{l, p}).WriteTo(co.Conn)
-		return int(n), err
 	}
 
 	msg := make([]byte, 2+len(p))
--- a/client.go
+++ b/client.go
@@ -340,11 +340,22 @@ func (co *Conn) Write(p []byte) (int, error) {
 		return co.Conn.Write(p)
 	}
 
-	l := make([]byte, 2)
-	binary.BigEndian.PutUint16(l, uint16(len(p)))
+	if _, ok := co.Conn.(*net.TCPConn); ok {
+		// Limit usage of net.Buffers to net.TCPConn only
+		// In the case of TLS using it results in splitting it
+		// into two packets
 
-	n, err := (&net.Buffers{l, p}).WriteTo(co.Conn)
-	return int(n), err
+		l := make([]byte, 2)
+		binary.BigEndian.PutUint16(l, uint16(len(p)))
+
+		n, err := (&net.Buffers{l, p}).WriteTo(co.Conn)
+		return int(n), err
+	}
+
+	msg := make([]byte, 2+len(p))
+	binary.BigEndian.PutUint16(msg, uint16(len(p)))
+	copy(msg[2:], p[:])
+	return co.Conn.Write(msg)
 }
 
 // Return the appropriate timeout for a specific request
--- a/server.go
+++ b/server.go
@@ -752,11 +752,22 @@ func (w *response) Write(m []byte) (int, error) {
 			return 0, &Error{err: "message too large"}
 		}
 
-		l := make([]byte, 2)
-		binary.BigEndian.PutUint16(l, uint16(len(m)))
+		if _, ok := w.tcp.(*net.TCPConn); ok {
+			// Limit usage of net.Buffers to net.TCPConn only
+			// In the case of TLS using it results in splitting it
+			// into two packets
 
-		n, err := (&net.Buffers{l, m}).WriteTo(w.tcp)
-		return int(n), err
+			l := make([]byte, 2)
+			binary.BigEndian.PutUint16(l, uint16(len(m)))
+
+			n, err := (&net.Buffers{l, m}).WriteTo(w.tcp)
+			return int(n), err
+		}
+
+		msg := make([]byte, 2+len(m))
+		binary.BigEndian.PutUint16(msg, uint16(len(m)))
+		copy(msg[2:], m[:])
+		return w.tcp.Write(msg)
 	default:
 		panic("dns: internal error: udp and tcp both nil")
 	}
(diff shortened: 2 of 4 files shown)


client.go
```diff
@@ -340,11 +340,22 @@ func (co *Conn) Write(p []byte) (int, error) {
...
+		l := make([]byte, 2)
+		binary.BigEndian.PutUint16(l, uint16(len(p)))
+
+		n, err := (&net.Buffers{l, p}).WriteTo(co.Conn)
+		return int(n), err
+	}
+
+	msg := make([]byte, 2+len(p))
+	binary.BigEndian.PutUint16(msg, uint16(len(p)))
+	copy(msg[2:], p[:])
```
tmthrgd, line 357: `s/p[:]/p/`

server.go
```diff
@@ -752,11 +752,22 @@ func (w *response) Write(m []byte) (int, error) {
...
+			l := make([]byte, 2)
+			binary.BigEndian.PutUint16(l, uint16(len(m)))
+
+			n, err := (&net.Buffers{l, m}).WriteTo(w.tcp)
+			return int(n), err
+		}
+
+		msg := make([]byte, 2+len(m))
+		binary.BigEndian.PutUint16(msg, uint16(len(m)))
+		copy(msg[2:], m[:])
```
tmthrgd, line 769: `s/m[:]/m/`

tmthrgd (changes requested):

tmthrgd:
I think we should back out #934 (the net.Buffers part) and not use net.Buffers at all here. There are far too many conditionals to pass before it batch writes:

```go
// Buffers contains zero or more runs of bytes to write.
//
// On certain machines, for certain types of connections, this is
// optimized into an OS-specific batch write operation (such as
// "writev").
type Buffers [][]byte
```

Currently that happens on darwin dragonfly freebsd illumos linux netbsd openbsd and windows, but not any other Go-supported platforms. It would have been preferable for us if Go had just exposed a writev wrapper directly and allowed us to handle fallbacks as needed.

Ultimately #934 was more about cleaning up the writing code. I switched it to net.Buffers on the assumption it would be faster, but I didn't have any numbers to back that up. I suspect it still probably is, but I don't think it's worth it. It's also clearly much slower for TLS.

ameshkov:
I am bothered a little by the sheer number of allocations the DNS library makes. If it's possible to avoid extra allocations here at least on Linux, it'd be for the good. Let me please do some tests to check if using `net.Buffers` does any good there.

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Send DNS query in o..." ]
>I think we should back out #934 (the net.Buffers part) and not use net.Buffers
>at all here. There are far too many conditionals to pass before it batch
>writes:

I have no particular attachment to net.Buffers. Only the general sense of that if we can
outsource to the std lib for something non-trivial we should.

This is part of the reason that I'm looking forward to DoH/DoQ, i.e. swap out all the
'let's do networking ourselves, it will be fun' for a vetted, shared, QUIC implemenation.

tmthrgd:
> I am bothered a little by the sheer number of allocations the DNS library makes. If it's possible to avoid extra allocations here at least on Linux, it'd be for the good. Let me please do some tests to check if using `net.Buffers` does any good there.

By all means benchmark away. It actually shouldn't introduce any new allocations as both buffers escape to the heap as it is. What would change is that instead of a 2-byte and n-byte allocation, you'd get an n-byte and 2+n-byte allocation. The biggest overhead in that case is probably actually the copy, which I don't think would be that significant.

It is unfortunate that allocations are as common as they are, but that's somewhat unavoidable. A dns/v2 could potentially eliminate this in all cases by always packing messages with a length prefix that is sliced off for UDP and kept for TCP. (I feel like I remember one of the proposed x/net/dns packages doing exactly that). That wouldn't be possible with backward compatibility requirements though.

ameshkov:
Here it is:
https://github.com/ameshkov/dns/blob/bench/issue_test.go#L48

Msg size = 1000

MacOS:
```
BenchmarkNoNetBuffers-16    	    6178	    251207 ns/op	    4453 B/op	      44 allocs/op
BenchmarkNetBuffers-16      	    4820	    259940 ns/op	    3593 B/op	      49 allocs/op
```

Linux:
```
BenchmarkNoNetBuffers-4   	    7790	    161764 ns/op	    4553 B/op	      48 allocs/op
BenchmarkNetBuffers-4     	    8644	    172493 ns/op	    3712 B/op	      53 allocs/op
```

Msg size = 200

MacOS:
```
BenchmarkNoNetBuffers-16    	    6984	    243791 ns/op	    2821 B/op	      44 allocs/op
BenchmarkNetBuffers-16      	    4585	    264033 ns/op	    2780 B/op	      49 allocs/op
```

Linux:
```
BenchmarkNoNetBuffers-4   	    7945	    172557 ns/op	    2964 B/op	      48 allocs/op
BenchmarkNetBuffers-4     	    6979	    185948 ns/op	    2931 B/op	      54 allocs/op
```

rdrozhdzh:
> A dns/v2 could potentially eliminate this in all cases by always packing messages with a length prefix that is sliced off for UDP and kept for TCP. (I feel like I remember one of the proposed x/net/dns packages doing exactly that). That wouldn't be possible with backward compatibility requirements though.

@tmthrgd , what do you think about my suggestion in https://github.com/miekg/dns/pull/1218#issuecomment-776020259? This is close to what you're saying here, and it looks backward compatible to me, since we'll add another function for providing the data prefixed with length.

tmthrgd:
> > A dns/v2 could potentially eliminate this in all cases by always packing messages with a length prefix that is sliced off for UDP and kept for TCP. (I feel like I remember one of the proposed x/net/dns packages doing exactly that). That wouldn't be possible with backward compatibility requirements though.
> 
> @tmthrgd , what do you think about my suggestion in [#1218 (comment)](https://github.com/miekg/dns/pull/1218#issuecomment-776020259)? This is close to what you're saying here, and it looks backward compatible to me, since we'll add another function for providing the data prefixed with length.

For some hypothetical dns/v2 I’d say something along those lines would be good. I don’t think it’s really feasible here though, we have exported methods that take the raw message as a []byte (that’s where this change is) and I don’t think duplicating and specialising in higher up methods is ideal, and we have to deal with TSIG (on the server side) that works with []byte. Definitely something to consider though, it might be possible to do as some kind of larger refactoring though.

tmthrgd:
@ameshkov You might want to try running a benchmark like I did here: https://gist.github.com/tmthrgd/cfcc4b245267cd8d2571eaeeda0f83ad

ameshkov:
@tmthrgd performance is pretty much the same: https://gist.github.com/ameshkov/66a69888a31328efbf78b45500cda634

ameshkov:
So far I don't see much difference between using `net.Buffers` and allocating a new array. Updated the PR accordingly.

tmthrgd:
> @tmthrgd performance is pretty much the same: https://gist.github.com/ameshkov/66a69888a31328efbf78b45500cda634

Thanks for that. It definitely seems like simplicity is the best option here.

tmthrgd (approved):
LGTM.

miekg:
this needs a test, unsure how though, but it needs a tests. We keep touching this code and without a test there is no in-tree memory of how things should work/look

ameshkov:
Well, testing the desired behavior without mocking net.Conn would be rather problematic. Also, it'd require making almost identical tests for both client and server part.

What if I create ioutil.go and move this code there? This way we'd be able to write a simple test checking that a packet is written in one go and avoid duplication.

miekg:
Would mocking a net.Conn be a lot of code? We would only need to check for the double write, so I'm assuming this could be done in "not a lot of lines of code"?

ameshkov:
Well, I guess if we just want to check the TLS case, using `tls.Conn` would be enough.

miekg:
Let's just try and see how it looks? Just detecting the non-double (i.e. single) write should be good enough, so technically this should be doable on the client side only

ameshkov:
Okay, let's try it. Just added a test for the TLS server, it's rather trivial.

Check it out plz, if this approach works for you, I'll use the same for the dns.Client

tmthrgd:
If you want a test like that, which is reasonable, it’s very easy to just write a:

```go
type countingConn struct { net.Conn; writes int } 
Func (* countingConn) Write...
```

tmthrgd:
We definitely don’t want to be half-parsing TLS record headers.

ameshkov:
This is exactly what I suggested in [the first place](https://github.com/miekg/dns/pull/1219#issuecomment-778071376), but I don't see where to inject this connection into the server, and that's why I suggested moving the code that writes to a separate function and test just it alone.

ameshkov:
Okay, nvm, I should've read the code more carefully. It's enough to create a `response` instance.

ameshkov:
Updated the PR, added unit-tests, thanks for the hints!

tmthrgd (approved):
LGTM.

PR #1221, opened 2021-02-13
Added tcpPool to avoid extra allocations with TCP

ameshkov:
I find it rather strange that there's already a udpPool, but nothing for TCP so here it is.

Also, added a bench for the TCP server. A quick test shows that there's no major difference with or without this change (unless the query size becomes large enough), but at least it makes TCP and UDP more "symmetrical".

### DIFF
--- a/server.go
+++ b/server.go
@@ -205,6 +205,10 @@ type Server struct {
 	// Default buffer size to use to read incoming UDP messages. If not set
 	// it defaults to MinMsgSize (512 B).
 	UDPSize int
+	// Default buffer size to use to read incoming TCP messages. If not set
+	// it defaults to MinMsgSize (512 B). DNS messages that don't fit the TCPSize
+	// will be read, but will cause extra allocations.
+	TCPSize int
 	// The net.Conn.SetReadTimeout value for new connections, defaults to 2 * time.Second.
 	ReadTimeout time.Duration
 	// The net.Conn.SetWriteTimeout value for new connections, defaults to 2 * time.Second.
@@ -236,6 +240,8 @@ type Server struct {
 
 	// A pool for UDP message buffers.
 	udpPool sync.Pool
+	// A pool for TCP message buffers.
+	tcpPool sync.Pool
 }
 
 func (srv *Server) isStarted() bool {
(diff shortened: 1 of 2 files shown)


ameshkov:
Test result:
https://gist.github.com/ameshkov/1b62932cf8c66c40893d05942e55dc74

Performance is pretty much the same. Slightly higher RSS without tcpPool (as expected).

tmthrgd:
What did you have `TCPSize` set to in that benchmark and do you know what the message size from the benchmark was?

ameshkov:
> What did you have TCPSize set to in that benchmark and do you know what the message size from the benchmark was?

Kept it default (512B)

tmthrgd:
> > What did you have TCPSize set to in that benchmark and do you know what the message size from the benchmark was?
> 
> Kept it default (512B).
> The benchmark message always fits it.

Is it possible to try with larger values? (It's been a while since I used the benchmarking tool).

ameshkov:
> Is it possible to try with larger values?

Any idea on a different benchmark tool?

With `dnstrace` I can't make DNS queries larger than the default TCPSize since the domain name can't be larger than 255B and that's the only thing that can be changed there.

Results with the largest possible with this tool:
https://gist.github.com/ameshkov/f2d120691f53e961ce4be7ee93c78964

Tbh, the difference is not huge and might be just a fluctuation. I assume it would be noticeable for larger packets, but it won't reflect a real-life situation where DNS queries are pretty small. Anyways, I think that it's a good idea to gradually get rid of most of them, why not this one first:)

tmthrgd:
> Any idea on a different benchmark tool?

No that's the only one I know unfortunately.

> Tbh, the difference is not huge and might be just a fluctuation. I assume it would be noticeable for larger packets, but it won't reflect a real-life situation where DNS queries are pretty small. Anyways, I think that it's a good idea to gradually get rid of most of them, why not this one first:)

If there's little to know measurable performance improvement, then I don't think we should make this change. It's more complexity and it exposes more knobs. Without a compelling reason then I think we should leave it be. Feel free to expand on why you might disagree.

ameshkov:
The problem is that I don't disagree, this one PR in isolation does not make much difference.

Is there any plans for v2 with better memory handling?

tmthrgd:
> The problem is that I don't disagree, this one PR in isolation does not make much difference.

Fair enough. Then I suggest we close this until/unless there's a compelling use case.

> Is there any plan for v2 with better memory handling?

Not to my knowledge. There are a number of changes I've had in mind for a v2, but certainly nothing justifying the effort and upgrades.

Issue #1226, opened 2021-02-24
The UDPConn should be closed when `setUDPSocketOptions` returned an error


nirui:
Hi all :D

I'm just reading the source code, and discovered a small potential problem here: https://github.com/miekg/dns/blob/67bd57debd4a81b6473b6b12f76e42f863279305/server.go#L323

The function `setUDPSocketOptions` above operates on an already-opened UDPConn, so if an error caused an abort action there, the UDPConn must be closed before doing `return e`.

I think the code should be:

    u := l.(*net.UDPConn)
    if e := setUDPSocketOptions(u); e != nil {
        u.Close() // Maybe handle this error too? No?
        return e
    }

tmthrgd:
Yep good catch, feel free to send a PR. There’s no need to handle the error from Close though.

PR #1227, opened 2021-02-25
Close the UDPConn before returning if setUDPSocketOptions fails

nirui:
Detail: https://github.com/miekg/dns/issues/1226


tmthrgd (approved):
LGTM. CI failure is unrelated.

Issue #1223, opened 2021-02-18
oss-fuzz build broken


tmthrgd:
In go1.16 module mode is on by default (https://golang.org/doc/go1.16#go-command) which has caused the fuzzer to fail today.

Here is the relevant part of the error log:
```
Step #4: + compile_go_fuzzer github.com/miekg/dns FuzzNewRR fuzz_newrr fuzz
Step #4: Running go-fuzz -tags fuzz -func FuzzNewRR -o fuzz_newrr.a github.com/miekg/dns
Step #4: -: no required module provides package github.com/miekg/dns: working directory is not part of a module
```

I don’t know if this is a problem is this repo, the oss-fuzz repo or the oss-fuzz project, but it’s a problem somewhere. Maybe a newer version of go-fuzz needs to be used or module mode forced off. I’m not familiar enough with it to know.

miekg:
ah. Well maybe go 1.16 will finally be the extra push to make all these tools
'module' aware.

We'll see what happens

mvdan:
That's happening in https://github.com/google/oss-fuzz/issues/5213, FYI. It's a bit tricky because their design is very GOPATH-centric.

miekg:
revisit the fuzzing once we standardize on go 1.18

PR #1237, opened 2021-03-04
Bump go versions in .github/workflows/go.yml

tmthrgd:
These are very out of date and it's causing persistent failures in the fuzz CI step. I hadn't paid much attention as I thought they were related to #1223, but I'm pretty sure it's because the go version is so out-of-date.

**Edit**: An example of the errors for posterity:

```
Error: /opt/hostedtoolcache/go/1.14.15/x64/pkg/mod/github.com/dvyukov/go-fuzz@v0.0.0-20210103155950-6a8e9d1f2415/go-fuzz-build/main.go:267:16: undefined: packages.NeedName
Error: /opt/hostedtoolcache/go/1.14.15/x64/pkg/mod/github.com/dvyukov/go-fuzz@v0.0.0-20210103155950-6a8e9d1f2415/go-fuzz-build/main.go:293:16: cannot use func literal (type func(*token.FileSet, string, []byte) (*ast.File, error)) as type func(*token.FileSet, string) (*ast.File, error) in assignment
Error: /opt/hostedtoolcache/go/1.14.15/x64/pkg/mod/github.com/dvyukov/go-fuzz@v0.0.0-20210103155950-6a8e9d1f2415/go-fuzz-build/main.go:419:13: undefined: packages.NeedName
```

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -39,6 +39,5 @@ jobs:
 
     - name: Fuzz
       run: |
-        export GOPATH=$GOROOT
         make -f Makefile.fuzz get
         make -f Makefile.fuzz
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go: [ '1.14', '1.15' ]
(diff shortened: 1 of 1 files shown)


tmthrgd:
Yeah that fixed it.

PR #1243, opened 2021-03-16
Use ed25519 from Go standard lib

cesarkuroiwa:
ed25519 was introduced into the standard lib with [Go 1.13](https://golang.org/doc/go1.13#crypto/ed25519), so I believe it would be desired to use it, thus removing some external deps


cesarkuroiwa:
Not sure what this means or if I can do anything to fix it

go.mod
```diff
@@ -3,7 +3,6 @@ module github.com/miekg/dns
 go 1.12
```
tmthrgd, line 3: You should bump this to the version that introduced crypto/ed25519.

tmthrgd:
I have no objections to this. I think go1.12 is well outside what we intend to support anyway. As for the CI failure, I think that’s unrelated, but I’ll have to see if I can figure it out.

go.mod
cesarkuroiwa: My bad, sorry about that. Thanks

miekg (approved):

PR #1244, opened 2021-03-17
Reduce memory allocations of SVCBIPv4Hint , SVCBIPv6Hint and SVCBMandatory parse functions

moredure:


tmthrgd:
Could you clarify in the title and pull request body what the purpose of this change is? thanks.

Issue #948, opened 2019-03-27
ZoneParser doesn't validate hex/base64 fields


tmthrgd:
The `ZoneParser` code doesn't validate that hex fields are actually valid hex.

The following test case fails:
```go
func TestParsingInvalidHex(t *testing.T) {
	rr, err := NewRR(`miek.nl. 3600 IN TYPE1234 \# 3 zz-top`)
	if err != nil {
		return
	}
	t.Error("expected error, got <nil>") //line scan_test.go:200

	if rr == nil {
		return
	}
	t.Errorf("expected nil record, got %v", rr) //line scan_test.go:205

	msg := make([]byte, Len(rr))
	if _, err = PackRR(rr, msg, 0, nil, false); err != nil {
		t.Errorf("unexpected error packing record: %v", err) //line scan_test.go:209
	}
}
```

with this output:
```
--- FAIL: TestParsingInvalidHex (0.00s)
    scan_test.go:200: expected error, got <nil>
    scan_test.go:205: expected nil record, got miek.nl. 3600    CLASS1  TYPE1234        \# 3 zz-top
    scan_test.go:209: unexpected error packing record: encoding/hex: invalid byte: U+007A 'z'
```

The correct behaviour would have been for `NewRR` to return a parsing error.

I found this with some brief fuzzing.

/cc @miekg

miekg:
Does bind check this? Under the motto garbage in; garbage out?
(probably does; but would be nice to double check)

tmthrgd:
The problem is that it then produces an invalid `RR` that `PackRR` chokes on. IMHO anything `ZoneParser` produces should be able to be packed.

miekg:
Fair enough.

tmthrgd:
I'm also seeing this from non-validated base64 fields.

```
	" TYPE49 000000000000" +
	"00000000000000000000" +
	"00000000000000000000" +
	"000000000000000000\x1f"
```

```
panic: illegal base64 data at input byte 70

goroutine 1 [running]:
github.com/tmthrgd/dns.FuzzNewRR(0x7fa3a1c24000, 0x4f, 0x200000, 0xc00005b950)
	/tmp/go-fuzz-build332085670/gopath/src/github.com/tmthrgd/dns/fuzz.go:27 +0x1db
go-fuzz-dep.Main(0x5b84b0)
	/tmp/go-fuzz-build332085670/goroot/src/go-fuzz-dep/main.go:54 +0xb6
main.main()
[...]

PR #1245, opened 2021-03-21
zoneparser: document not checking syntax

miekg:
Add a couple of lines that we're not syntax checking.

Closes: #948

### DIFF
--- a/scan.go
+++ b/scan.go
@@ -150,6 +150,10 @@ func ReadRR(r io.Reader, file string) (RR, error) {
 // The text "; this is comment" is returned from Comment. Comments inside
 // the RR are returned concatenated along with the RR. Comments on a line
 // by themselves are discarded.
+//
+// The parser will not syntax check anything, e.g. illegal base64 in RRSIGs will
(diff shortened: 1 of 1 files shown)


scan.go
```diff
@@ -150,6 +150,10 @@ func ReadRR(r io.Reader, file string) (RR, error) {
 // The text "; this is comment" is returned from Comment. Comments inside
 // the RR are returned concatenated along with the RR. Comments on a line
 // by themselves are discarded.
+//
+// The parser will not syntax check anything, e.g. illegal base64 in RRSIGs will
```
tmthrgd, line 154: I think we do check somethings sometimes, so perhaps this would be better worded something like ‘Callers should not assume the output is...’.

miekg:
[ Quoting ***@***.***> in "Re: [miekg/dns] zoneparser: documen..." ]
>I think we do check somethings sometimes, so perhaps this would be better
>worded something like ‘Callers should not assume the output is...’.

updated

tmthrgd (approved):

Issue #1169, opened 2020-10-10
Add `Z` method to OPT RR


miekg:
The MBZ - must - be -zero in the OPT RR bit me today and I discovered there isn't a handy method that prints it value.

PR #1246, opened 2021-03-22
OPT: add Z and SetZ methods

miekg:
Fixes: #1169

### DIFF
--- a/edns.go
+++ b/edns.go
@@ -150,12 +150,12 @@ func (rr *OPT) SetDo(do ...bool) {
 
 // Z returns the Z part of the OPT RR as a uint16 with only the 15 least significant bits used.
 func (rr *OPT) Z() uint16 {
-	return uint16(rr.Hdr.Ttl & 0x007FFF)
+	return uint16(rr.Hdr.Ttl & 0x7FFF)
 }
 
 // SetZ sets the Z part of the OPT RR, note only the 15 least significant bits of z are used.
 func (rr *OPT) SetZ(z uint16) {
-	rr.Hdr.Ttl = rr.Hdr.Ttl&^0x7FFF | uint32(z)
+	rr.Hdr.Ttl = rr.Hdr.Ttl&^0x7FFF | uint32(z&0x7FFF)
 }
 
 // EDNS0 defines an EDNS0 Option. An OPT RR can have multiple options appended to it.
--- a/edns.go
+++ b/edns.go
@@ -148,6 +148,16 @@ func (rr *OPT) SetDo(do ...bool) {
 	}
 }
 
+// Z returns the Z part of the OPT RR as a uint16 with only the 15 least significant bits used.
+func (rr *OPT) Z() uint16 {
+	return uint16(rr.Hdr.Ttl & 0x007FFF)
+}
+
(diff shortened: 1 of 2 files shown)


miekg:
please double check my bit mask, see https://tools.ietf.org/html/rfc6891#section-6.1.3

edns.go
```diff
@@ -148,6 +148,16 @@ func (rr *OPT) SetDo(do ...bool) {
...
 }
 
+// Z returns the Z part of the OPT RR as a uint16 with only the 15 least significant bits used.
+func (rr *OPT) Z() uint16 {
+	return uint16(rr.Hdr.Ttl & 0x007FFF)
+}
+
+// SetZ sets the Z part of the OPT RR, note only the 15 least significant bits of z are used.
+func (rr *OPT) SetZ(z uint16) {
+	rr.Hdr.Ttl = rr.Hdr.Ttl&0xFFFF7FFF | uint32(z)
```
tmthrgd, line 158: Shouldn’t this just be `&^ 0x7fff`? Like the intent is to mask off then or right?

miekg:
[ Quoting ***@***.***> in "Re: [miekg/dns] OPT: add Z and SetZ..." ]
>+// SetZ sets the Z part of the OPT RR, note only the 15 least significant bits of z are used.
>+func (rr *OPT) SetZ(z uint16) {
>+       rr.Hdr.Ttl = rr.Hdr.Ttl&0xFFFF7FFF | uint32(z)
>
>Shouldn’t this just be &^ 0x7fff? Like the intent is to mask off then or right?

so:

0x7fff is "0111 1111  1111 1111", with &^ this becomes "1000 0000  0000 0000", so that
leaves the DO bit, before we OR our value back in.

So yes.

/Miek

--
Miek Gieben

miekg:
PTAL

edns.go
```diff
@@ -148,6 +148,16 @@ func (rr *OPT) SetDo(do ...bool) {
...
 }
 
+// Z returns the Z part of the OPT RR as a uint16 with only the 15 least significant bits used.
+func (rr *OPT) Z() uint16 {
+	return uint16(rr.Hdr.Ttl & 0x007FFF)
+}
+
+// SetZ sets the Z part of the OPT RR, note only the 15 least significant bits of z are used.
+func (rr *OPT) SetZ(z uint16) {
+	rr.Hdr.Ttl = rr.Hdr.Ttl&^0x7FFF | uint32(z)
```
tmthrgd, line 158: The `z` needs a mask as well to match the doc.

edns.go
```diff
@@ -148,6 +148,16 @@ func (rr *OPT) SetDo(do ...bool) {
 	}
 }
 
+// Z returns the Z part of the OPT RR as a uint16 with only the 15 least significant bits used.
+func (rr *OPT) Z() uint16 {
+	return uint16(rr.Hdr.Ttl & 0x007FFF)
```
tmthrgd, line 153: Nit: write the mask value the same way in both functions. Either leading zeroes or no leading series.

tmthrgd (changes requested):
Small changes.

edns.go
miekg: and done

tmthrgd (approved):
LGTM.

PR #1252, opened 2021-03-29
Allow custom Dialer implementation

coffeemakr:
Currently only a `net.Dialer` can be passed to the client. This prevents the client from using a custom implementation, because `net.Dialer` is not a interface. E.g. it is not possible to use a proxy.

If the Dialer field type is replaced with an interface, we can still use `net.Dialer`, but also specify a custom Dialer. Because `tls.DialWithDialer` also uses `net.Dialer` this function must be replaced with the call to the provided dialer, settings the `ServerName` in the `TlsConfig` and the handshaking.

I have tested the changes with the existing tests and the following script, that uses a running Tor instance as a proxy and connect to the Cloudflare onion address.

<details>
<summary>test.go</summary>

```go
package main

import (
	"crypto/tls"
	"fmt"
	"github.com/miekg/dns"
	"golang.org/x/net/proxy"
	"log"
	"net"
	"net/url"
)

func main() {
	var server = "dns4torpnlfs2ifuz2s2yf3fc7rdmsbhm6rw75euj35pac6ap25zgqad.onion:853"
	var proxyDialer dns.Dialer
	proxyUrl, err := url.Parse("socks5h://localhost:9050")
	if err != nil {
		log.Fatalf("invalid proxy url: %s\n", err)
	}
	proxyDialer, err = proxy.FromURL(proxyUrl, &net.Dialer{})
	if err != nil {
		log.Fatalf("invalid proxy: %s\n", err)
	}
	client := new(dns.Client)
	client.Dialer = proxyDialer
	client.TLSConfig = &tls.Config{
		ServerName: "one.one.one.one", // because we use the .onion address
	}
	client.Net = "tcp-tls"
	m := new(dns.Msg)
	m.SetQuestion(dns.Fqdn("example.com"), dns.TypeA)
	m.RecursionDesired = true
	r, _, err := client.Exchange(m, server)
	if err != nil {
		log.Fatalf("*** error: %s\n", err.Error())
	}
	if r.Rcode != dns.RcodeSuccess {
		log.Fatalf(" *** invalid answer: %s$n", dns.RcodeToString[r.Rcode])
	}
	// Stuff must be in the answer section
	for _, a := range r.Answer {
		fmt.Printf("%v\n", a)
	}
}
```

</details>

### DIFF
--- a/client.go
+++ b/client.go
@@ -384,7 +384,7 @@ func (c *Client) getTimeoutForRequest(timeout time.Duration) time.Duration {
 	}
 	// net.Dialer.Timeout has priority if smaller than the timeouts computed so
 	// far
-	if netDialer, ok := c.Dialer.(*net.Dialer); ok {
+	if netDialer, ok := c.Dialer.(*net.Dialer); ok && netDialer.Timeout != 0 {
 		if netDialer.Timeout < requestTimeout {
 			requestTimeout = netDialer.Timeout
 		}
--- a/client.go
+++ b/client.go
@@ -27,12 +27,17 @@ type Conn struct {
 	tsigRequestMAC string
 }
 
+// A Dialer is a means to establish a connection.
+type Dialer interface {
+	Dial(network, address string) (net.Conn, error)
+}
+
 // A Client defines parameters for a DNS client.
 type Client struct {
 	Net       string      // if "tcp" or "tcp-tls" (DNS over TLS) a TCP query will be initiated, otherwise an UDP one (default is "" for UDP)
 	UDPSize   uint16      // minimum receive buffer for UDP messages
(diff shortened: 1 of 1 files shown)


tmthrgd:
I don’t think this is needed, you can create your own `Conn` with whatever `net.Conn` You have and pass that to `ExchangeWithConn`. That’s all that `Exchange` does anyway.

miekg:
[ Quoting ***@***.***> in "Re: [miekg/dns] Allow custom Dialer..." ]
>I don’t think this is needed, you can create your own Conn with whatever
>net.Conn You have and pass that to ExchangeWithConn. That’s all that Exchange
>does anyway.

agreed

/close

PR #1253, opened 2021-04-07
Add support for extended DNS errors (RFC 8914).

tangrs:


edns.go
```diff
@@ -673,3 +676,94 @@ func (e *EDNS0_PADDING) copy() EDNS0 {
 	copy(b, e.Padding)
 	return &EDNS0_PADDING{b}
 }
+
+const (
+	EDECodeOther uint16 = iota
```
miekg, line 681: I would name these all `ExtendedRcodeOther` to be inline of the naming of the normal RcodeXXX

edns.go
```diff
@@ -673,3 +676,94 @@ func (e *EDNS0_PADDING) copy() EDNS0 {
 	copy(b, e.Padding)
 	return &EDNS0_PADDING{b}
 }
+
+const (
```
miekg, line 680: needs a comment and point to the RFC

miekg: README can also be extended by adding this RFC

edns.go
```diff
@@ -673,3 +676,94 @@ func (e *EDNS0_PADDING) copy() EDNS0 {
...
+	EDECodeProhibited
+	EDECodeStaleNXDOMAINAnswer
+	EDECodeNotAuthoritative
+	EDECodeNotSupported
+	EDECodeNoReachableAuthority
+	EDECodeNetworkError
+	EDECodeInvalidData
+)
+
+var EDECodeToString = map[uint16]string{
```
miekg, line 708: might need the reverse as well, just like normal rcodes

edns.go
tangrs: On my reading of RFC8914, these are "info codes" and is specific to that RFC, and doesn't really seem to be an Rcode in the common DNS definition. I'm worried this naming scheme may cause confusion with the "extended Rcode" as defined in RFC6891.

Perhaps ExtendedErrorCode as a prefix might be more suitable here (or even ExtendedEcode)?

tangrs: On further thought, perhaps ExtendedErrorInfo?

miekg: hmm good point, ExtendedRcode already exist, hmm `ErrorCode` maybe?

tangrs: I wonder if ErrorCode is a little too generic, and might be confused with something else? Ideally, I think the word "extended" should appear somewhere.

Other than that, I'm happy to go ahead with the `ErrorCode` prefix, if you think it's appropriate.

miekg:
[ Quoting ***@***.***> in "Re: [miekg/dns] Add support for ext..." ]
***@***.*** commented on this pull request.
>
>
>In edns.go:
>
>> @@ -673,3 +676,94 @@ func (e *EDNS0_PADDING) copy() EDNS0 {
>        copy(b, e.Padding)
>        return &EDNS0_PADDING{b}
> }
>+
>+const (
>+       EDECodeOther uint16 = iota
>
>I wonder if ErrorCode is a little too generic, and might be confused with
>something else? Ideally, I think the word "extended" should appear somewhere.
>
>Other than that, I'm happy to go ahead with the ErrorCode prefix, if you think
>it's appropriate.

ExtendedErrorCode seems fine as well.

tangrs:
Sounds good, I think I'll go with `ExtendedErrorCode` in that case. It's a bit long, but seems to be the least ambiguous.

edns.go
tangrs: Seeing as we're not dealing with Rcodes here, is there a good use-case for the reverse lookup?

miekg:
[ Quoting ***@***.***> in "Re: [miekg/dns] Add support for ext..." ]
***@***.*** commented on this pull request.
>
>
>In edns.go:
>
>> +     EDECodeNSECMissing
>+       EDECodeCachedError
>+       EDECodeNotReady
>+       EDECodeBlocked
>+       EDECodeCensored
>+       EDECodeFiltered
>+       EDECodeProhibited
>+       EDECodeStaleNXDOMAINAnswer
>+       EDECodeNotAuthoritative
>+       EDECodeNotSupported
>+       EDECodeNoReachableAuthority
>+       EDECodeNetworkError
>+       EDECodeInvalidData
>+)
>+
>+var EDECodeToString = map[uint16]string{
>
>Seeing as we're not dealing with Rcodes here, is there a good use-case for the
>reverse lookup?

dunno, just for symmetry with the rest I would add it.

edns.go
```diff
@@ -673,3 +676,101 @@ func (e *EDNS0_PADDING) copy() EDNS0 {
 	copy(b, e.Padding)
 	return &EDNS0_PADDING{b}
 }
+
+// Extended DNS Error Info Codes (RFC 8914).
```
miekg, line 680:
```suggestion
// Extended DNS Error Codes (RFC 8914).
```

PR #1256, opened 2021-04-22
Fix copy() in SVCBIPv4Hint and SVCBIPv6Hint

ameshkov:
The problem with the current implementation is that it is not a real deep copy,
it points to the same base arrays behind the slices. This was causing some
issues in real-life application.

### DIFF
--- a/svcb.go
+++ b/svcb.go
@@ -511,9 +511,9 @@ func (s *SVCBIPv4Hint) parse(b string) error {
 }
 
 func (s *SVCBIPv4Hint) copy() SVCBKeyValue {
-	hint := []net.IP(nil)
-	for _, ip := range s.Hint {
-		hint = append(hint, copyIP(ip))
+	hint := make([]net.IP, len(s.Hint))
+	for i, ip := range s.Hint {
+		hint[i] = copyIP(ip)
 	}
 
 	return &SVCBIPv4Hint{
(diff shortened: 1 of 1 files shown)


svcb.go
```diff
@@ -511,8 +511,13 @@ func (s *SVCBIPv4Hint) parse(b string) error {
 }
 
 func (s *SVCBIPv4Hint) copy() SVCBKeyValue {
+	hint := []net.IP(nil)
```
tmthrgd, line 514: I’d prefer:

```go
hint := make([]net.IP, len(s.Hint))
for i, ip := range s.Hint {
hint[i] = copyIP(ip)
}

return &SVCBIPv4Hint{hint}
```

svcb.go
```diff
@@ -629,8 +634,13 @@ func (s *SVCBIPv6Hint) parse(b string) error {
 }
 
 func (s *SVCBIPv6Hint) copy() SVCBKeyValue {
+	hint := []net.IP(nil)
```
tmthrgd, line 637: Ditto here.

tmthrgd (changes requested):
Small change, but ultimately I’m fine with this.

svcb.go
ameshkov: I was thinking about it, but we'll end up with an empty array in the `s.Hint = nil` case which is a little bit different from the original. Is it okay?

tmthrgd: Yeah that should be fine. It’s rare for such a distinction to matter. It’s much faster to preallocate and avoid append where that’s easy to do.

tmthrgd (approved):
LGTM.

PR #1258, opened 2021-05-05
Check if DNSKEY Zone-Key bit is set on RRSIG Verify

frelon:
As commented in the code, the DNSKEY must only be used to verify an RRSIG if the Zone Key bit is set.


dnssec.go
```diff
@@ -392,6 +392,12 @@ func (rr *RRSIG) Verify(k *DNSKEY, rrset []RR) error {
 	if k.Protocol != 3 {
 		return ErrKey
 	}
+	// RFC 4034 2.1.1 If bit 7 has value 0, then the DNSKEY record holds some
```
miekg, line 395: We might want to document both this and the k.Protocol in the function doc as well

frelon: I added a mention in the function doc as well. :+1:

miekg (approved):

PR #1260, opened 2021-05-07
Remove cifuzz

miekg:
The existing of the file signifies it needs to run, so now it errors
that there is nothing to run (everything is commented out).

Remove the entire file.


tmthrgd (approved):
LGTM.

PR #1268, opened 2021-06-17
small cleans from go report card

miekg:
I went through the list and cleaned things up here and there.


tmthrgd (approved):
LGTM.

PR #1273, opened 2021-06-22
Move makeDataOpt into edns.go

miekg:
Make it more obvious that these two lists (const, and case) need to be
in sync.

Also sort the list to match the const sorting.


tmthrgd (approved):
LGTM. Was going to suggest doing exactly this.

PR #1274, opened 2021-06-23
Fix race condition in ExchangeContext.

### LINKED ISSUE #988: data race found in ExchangeContext()
Pls take a look at the problem below
```
WARNING: DATA RACE
Write at 0x00c0003d6100 by goroutine 35:
  myproject/vendor/github.com/miekg/dns.(*Client).ExchangeContext()
      /go/src/myproject/vendor/github.com/miekg/dns/client.go:501 +0x11b
  myproject/gateway.(*DNSClient).dig()
      /go/src/myproject/gateway/dns_client.go:113 +0x2b8

Previous write at 0x00c0003d6100 by goroutine 33:
  myproject/vendor/github.com/miekg/dns.(*Client).ExchangeContext()
      /go/src/myproject/vendor/github.com/miekg/dns/client.go:501 +0x11b
  myproject/gateway.(*DNSClient).dig()
      /go/src/myproject/gateway/dns_client.go:113 +0x2b8

==================

WARNING: DATA RACE
Read at 0x00c0000726c0 by goroutine 33:
  myproject/vendor/github.com/miekg/dns.(*Client).exchange()
      /go/src/myproject/vendor/github.com/miekg/dns/client.go:401 +0x6d6
  myproject/vendor/github.com/miekg/dns.(*Client).Exchange()
      /go/src/myproject/vendor/github.com/miekg/dns/client.go:137 +0x44f
  myproject/vendor/github.com/miekg/dns.(*Client).ExchangeContext()
      /go/src/myproject/vendor/github.com/miekg/dns/client.go:502 +0x16c
  myproject/gateway.(*DNSClient).dig()
      /go/src/myproject/gateway/dns_client.go:113 +0x2b8

Previous write at 0x00c0000726c0 by goroutine 32:
  myproject/vendor/github.com/miekg/dns.(*Client).ExchangeContext()
      /go/src/myproject/vendor/github.com/miekg/dns/client.go:501 +0xac
  myproject/gateway.(*DNSClient).dig()
      /go/src/myproject/gateway/dns_client.go:113 +0x2b8

```

jsha:
Previously ExchangeContext relied on modifying a client's dialer just
before dialing, which was a race condition. This plumbs through contexts
in various paths so that ExchangeContext no longer has to rely on that
race condition. When a context with a deadline is used, and the client
also has a ReadTimeout or WriteTimeout configured, whichever is soonest
takes precedence. This allows us to implement the non-Context functions
in terms of the Context functions, by passing a context.Background().

Since this relies on tls.Dialer.DialContext, which is new in Go 1.15,
this should not be merged until Go 1.17 is released and the minimum
supported Go version is bumped.

Fixes #988.

### DIFF
--- a/client.go
+++ b/client.go
@@ -82,6 +82,12 @@ func (c *Client) writeTimeout() time.Duration {
 
 // Dial connects to the address on the named network.
 func (c *Client) Dial(address string) (conn *Conn, err error) {
+	return c.DialContext(context.Background(), address)
+}
+
+// DialContext connects to the address on the named network, with a
+// context.Context.
+func (c *Client) DialContext(ctx context.Context, address string) (conn *Conn, err error) {
 	// create a new dialer with the appropriate timeout
 	var d net.Dialer
 	if c.Dialer == nil {
@@ -101,9 +107,13 @@ func (c *Client) Dial(address string) (conn *Conn, err error) {
 	if useTLS {
 		network = strings.TrimSuffix(network, "-tls")
 
-		conn.Conn, err = tls.DialWithDialer(&d, network, address, c.TLSConfig)
+		tlsDialer := tls.Dialer{
+			NetDialer: &d,
+			Config:    c.TLSConfig,
+		}
+		conn.Conn, err = tlsDialer.DialContext(ctx, network, address)
 	} else {
-		conn.Conn, err = d.Dial(network, address)
+		conn.Conn, err = d.DialContext(ctx, network, address)
 	}
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


miekg:
thanks!

Could we split this in a version that can be merged today and one (smaller one) that we can merge after 1.17 is out?

tmthrgd:
Doing this would be good, but it's not clear this is the right way. The problem is the `SingleInflight` grouping. Using just a random context means one timeout and cancellation will apply to *all* currently grouped requests. If that grouping didn't exist, this would be easy.

I think the best fix for this is probably in `exchangeWithConnContext` to only use `ctx` when `!SingleInflight` and just use `context.Background()` when `SingleInflight`. That means that the context is only respected when the caller isn't using `SingleInflight`, but I think that's fine and likely to be a lot less buggy.

miekg:
good point @tmthrgd, that change should be made. We can patch that up in the documenation and I think it makes sense both SingleInflight and canceling of a Context is to prevent work from happening. So ignoring a cancellation in case of a single-inflight seems totally appropriate.

miekg:
@jsha are you able to make these changes?

jsha:
Hi @miekg! I'm afraid I'm on vacation through the end of summer so won't have time soon to work on these changes. I'll pick them up again when I get back; or, of course, if you'd like to incorporate before then please feel free to pick them up. Thanks!

miekg:
[ Quoting ***@***.***> in "Re: [miekg/dns] Fix race condition ..." ]
>Hi @miekg! I'm afraid I'm on vacation through the end of summer so won't have
>time soon to work on these changes. I'll pick them up again when I get back;
>or, of course, if you'd like to incorporate before then please feel free to
>pick them up. Thanks!

good times! Enjoy, I'll pick apart some of this. Thanks!

Issue #436, opened 2017-01-10
Support RFC-6762 / mDNS fields


sayotte:
I don't see it in the API or in the README: does this library directly support encoding/decoding records with RFC-6762 fields (see: https://tools.ietf.org/html/rfc6762#section-19)?

Namely:
* The Cache-Flush bitfield (first bit in Class field of RR header)
* The Unicast-Response-Acceptable bitfield (first bit of Class field in Query header)

And also:
* The [RFC-6763 (DNS-SD) section 6.3](https://tools.ietf.org/html/rfc6763#section-6.3) TXT record format

I created [an mDNS library](https://github.com/sayotte/rawmdns) mostly for fun over the holidays, but now that it's done I worry someone might find it and try to **use** it. The world would be better off if a widely-adopted, community supported library such as this one had the functionality instead.

miekg:
See this thread: https://github.com/miekg/dns/issues/423

No, there aren't any functions directly dealing with these fields and yes, make sense to add them.

shotahino:
@miekg Is this still open and on your list?

miekg:
[ Quoting <notifications@github.com> in "Re: [miekg/dns] Support RFC-6762 / ..." ]
>@miekg Is this still open and on your list?

Support for this is still welcome, noone is working on it though.

Issue #423, opened 2016-12-01
ErrTruncated should not be returned by Msg.Unpack() just because TC=1


markdingo:
Msg.Unpack() returns ErrTruncated when the TC header bit is set.

This is misleading because TC=1 is not an unpack error - it's merely a protocol signal between sender and receiver. Furthermore the error return forces the caller to assume that the message is malformed and cannot be used when in fact the message is perfectly fine.

Since TC is visible via Msg.Truncated it strikes me that Unpack() has no need to return an error at all as callers can determined TC for themselves. But this is an incompatible change so a hack might be to return a unique error for TC=1. Maybe:

ErrTruncatedBitSet    error = &Error{err: "Unpack worked, but found TC=1"}

so clients can distinguish at least. Alternatively use a new error for all the genuinely borken messages:

ErrTruncated            error = &Error{err: "Unpack worked, but found TC=1"}
ErrReallyTruncated  error = &Error{err: "failed to unpack truncated message"}

andrewtj:
I believe this behaviour came about in #281

MarkDAtEmu:
Yes, it does seem related. But it seems to me that there is a conflation between TC=1 and whether a packet is in error or not.

There is no correlation between TC=1 and whether a message is well formed. TC=1 is a common response setting if, eg, DNS RRL uses it extensively as does Bonjour.

I think it's a mistake to assume that a TC=1 message is an error.

miekg:
I think we can distinguish between the two cases. But what is the use? You want to inspect the TC reply for bits that *are* useful? What about a signature in a message and 1/2 RRset? There is no indication you have the full RRset because of TC=1.

Is there RFC text on being able to use TC=1 messages?

MarkDAtEmu:
My particular use-case is Bonjour. RFC6762 has numerous examples where TC is set because Bonjour data can get very large. Generally the sender sends multiple packets with TC=1 on allbut the last message related to the same set of services.

In terms of code, I assumed that an error return from Unpack() means that the message could not be unpacked properly. Perhaps because the payload is corrupted in transit or bits were dropped during transmission.

My original code simply went:

if err := msg.Unpack(buffer); err != nil {
       fmt.Println("Message discarded", err)
       return err
}
.... do something useful with msg
fmt.Println("Message Processed")
return nil

But that code doesn't work if the sender sets TC=1 and in fact there is no code I can write that lets me know that I have a well-formed message that just happens to have TC=1 set.

miekg:
[...]

PR #1275, opened 2021-07-01
Add MDNS class helper functions to RR_Header and Question

nemith:
Add two methods to be used with MDNS (RFC6762) messages for dealing
with extra data encoded in to the class field of queries and resource
record headers.

For queries the top bit of the class field is the unicast-response bit
indicating the querier can accept a unicast response to the queuery (in
addition or in lieu of a multicast response)

For resource records the top bit of the class field is for the
cache-flush bit that indicates the record is unique and should replace
any entries vs appending to the cache.

Fixes some of the discussed parts of #436 and #423

### DIFF
--- a/dns.go
+++ b/dns.go
@@ -75,6 +75,24 @@ type RR_Header struct {
 // Header returns itself. This is here to make RR_Header implements the RR interface.
 func (h *RR_Header) Header() *RR_Header { return h }
 
+// MDNSClass will return an unpacked record class as defined in RFC6762. The
+// lower 15 bits of the Class field is returned along with if the cache-flush
+// bit has been set or not.
+func (h *RR_Header) MDNSClass() (uint16, bool) {
+	return h.Class ^ 1<<15, (h.Class & 1 << 15) != 0
+}
+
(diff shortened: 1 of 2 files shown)


miekg:
I think this warrants a new `mdns` pkg, where these methods can be better named. Also it is just some bit shifting, so those mdns helpers should just do that. Take a h.Class and return the value needed for mdns.

nemith:
> I think this warrants a new mdns pkg, where these methods can be better named
Yeah, i was following a lead set may years ago in a previous issue.  Would this be something you would want as a subpackage under this repo or just separate?

> Also it is just some bit shifting, so those mdns helpers should just do that. Take a h.Class and return the value needed for mdns.

Not 100% sure what you mean here.  Yes i can use bit shifting to return the value instead of a mask.

miekg:
having explicit functions for mdns requires more thought. Can't just add a few functions here and there and call it a day. Would like to see a more thorough explanation on what would be needed for mdns - also think it should be all in a subpackage - if that's impossible it should sit in an altogether dfiferent package

PR #1276, opened 2021-07-02
Set up Dependabot to keep gomod and github-actions up-to-date 

mmorel-35:


miekg:
This is not needed.

/close

PR #1277, opened 2021-07-03
add IP_TRANSPARENT socket option.

mostafanamazy:


miekg:
server.Listener and server..PacketConn are exported so you can just give those your own `conn`.

IOW: this is not needed in this library.

mostafanamazy:
Hello dear Miek
Thank you for your reply to my pull request. I want to use that but I can't detect where ListenAndServe was completed so I can use the server.Listener and server..PacketConn and they have a valid value.
I saw there was a function pointer named NotifyStartedFunc but I can't use that because I have many server of dns and this function had no input for detect which sever started.
Do you have any ideas for help?

PR #1280, opened 2021-07-13
fix Msg.Used TTL must be zero

mimuret:
Prerequisite Section RRset Exist Value Dependent TTL must be zero.

```
   2.4.2 - RRset Exists (Value Dependent)

   A set of RRs with a specified NAME and TYPE exists and has the same
   members with the same RDATAs as the RRset specified here in this
   section.  While RRset ordering is undefined and therefore not
   significant to this comparison, the sets be identical in their
   extent.

   For this prerequisite, a requestor adds to the section an entire
   RRset whose preexistence is required.  NAME and TYPE are that of the
   RRset being denoted.  CLASS is that of the zone.  TTL must be
   specified as zero (0) and is ignored when comparing RRsets for
   identity.
``

### DIFF
--- a/update.go
+++ b/update.go
@@ -32,8 +32,9 @@ func (u *Msg) Used(rr []RR) {
 		u.Answer = make([]RR, 0, len(rr))
 	}
 	for _, r := range rr {
-		r.Header().Class = u.Question[0].Qclass
-		r.Header().Ttl = 0
+		hdr := r.Header()
+		hdr.Class = u.Question[0].Qclass
+		hdr.Ttl = 0
 		u.Answer = append(u.Answer, r)
 	}
 }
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: update.go [hunk 1]
```diff
@@ -33,6 +33,7 @@ func (u *Msg) Used(rr []RR) {
 	}
 	for _, r := range rr {
 		r.Header().Class = u.Question[0].Qclass
+		r.Header().Ttl = 0
```
tmthrgd, line 36: `hdr := r.Header()` and then use the variable.

tmthrgd (changes requested):

tmthrgd (approved):
I haven’t checked any of the newer RFCs but as long as nothing has changed since RFC 2136 this looks correct. Code is fine also. LGTM.

PR #1281, opened 2021-07-14
Fix race condition in ExchangeContext

miekg:
- Fix race condition in ExchangeContext.
- Implement suggestions from review in #1274

### DIFF
--- a/client.go
+++ b/client.go
@@ -82,6 +82,12 @@ func (c *Client) writeTimeout() time.Duration {
 
 // Dial connects to the address on the named network.
 func (c *Client) Dial(address string) (conn *Conn, err error) {
+	return c.DialContext(context.Background(), address)
+}
+
+// DialContext connects to the address on the named network, with a context.Context.
+// For TLS over TCP (DoT) the context isn't used yet. This will be enabled when Go 1.18 is released.
+func (c *Client) DialContext(ctx context.Context, address string) (conn *Conn, err error) {
 	// create a new dialer with the appropriate timeout
 	var d net.Dialer
 	if c.Dialer == nil {
@@ -101,9 +107,17 @@ func (c *Client) Dial(address string) (conn *Conn, err error) {
 	if useTLS {
 		network = strings.TrimSuffix(network, "-tls")
 
+		// TODO(miekg): Enable after Go 1.18 is released, to be able to support two prev. releases.
+		/*
+			tlsDialer := tls.Dialer{
+				NetDialer: &d,
+				Config:    c.TLSConfig,
+			}
+			conn.Conn, err = tlsDialer.DialContext(ctx, network, address)
+		*/
 		conn.Conn, err = tls.DialWithDialer(&d, network, address, c.TLSConfig)
 	} else {
-		conn.Conn, err = d.Dial(network, address)
+		conn.Conn, err = d.DialContext(ctx, network, address)
 	}
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


miekg:
I've cherry picked the commit from #1274 in here and then implemented the suggestions from that review. Added documentation/comments on why the context sometimes isn't used.

client.go
```diff
@@ -435,15 +469,11 @@ func DialTimeoutWithTLS(network, address string, tlsConfig *tls.Config, timeout
...
+	conn, err := c.DialContext(ctx, a)
+	if err != nil {
+		return nil, 0, err
 	}
-	// not passing the context to the underlying calls, as the API does not support
-	// context. For timeouts you should set up Client.Dialer and call Client.Exchange.
-	// TODO(tmthrgd,miekg): this is a race condition.
-	c.Dialer = &net.Dialer{Timeout: timeout}
-	return c.Exchange(m, a)
+	defer conn.Close()
```
miekg, line 476: hmm, is this a separate bug that is fixed here?

client.go
```diff
@@ -101,9 +107,18 @@ func (c *Client) Dial(address string) (conn *Conn, err error) {
 	if useTLS {
 		network = strings.TrimSuffix(network, "-tls")
 
+		/*
+			// Enable after Go 1.18 is released, to be able to support two prev. releases.
```
tmthrgd, line 111: It would be a little clearer to move the single line comment here before the multi-line comment and prefix with the TODO.

client.go
```diff
@@ -139,24 +154,34 @@ func (c *Client) Exchange(m *Msg, address string) (r *Msg, rtt time.Duration, er
...
 //
-//  This allows users of the library to implement their own connection management,
-//  as opposed to Exchange, which will always use new connections and incur the added overhead
-//  that entails when using "tcp" and especially "tcp-tls" clients.
+// This allows users of the library to implement their own connection management,
+// as opposed to Exchange, which will always use new connections and incur the added overhead
+// that entails when using "tcp" and especially "tcp-tls" clients.
+//
+// When the singleflight is set for this client the context is _not_ forwarded to the (shared) exchange, to
+// prevent one cancelation, canceling all outstanding requests.
```
tmthrgd, line 169: Nit: "prevent one cancelation *from* canceling all outstanding requests."

client.go
```diff
@@ -139,24 +154,34 @@ func (c *Client) Exchange(m *Msg, address string) (r *Msg, rtt time.Duration, er
...
-		return c.exchange(m, conn)
+		return c.exchangeContext(ctx, m, conn)
 	}
 
 	q := m.Question[0]
 	key := fmt.Sprintf("%s:%d:%d", q.Name, q.Qtype, q.Qclass)
 	r, rtt, err, shared := c.group.Do(key, func() (*Msg, time.Duration, error) {
-		return c.exchange(m, conn)
+		// When we're doing singleflight we don't want one context cancelation, cancel _all_ outstanding queries.
+		// Hence we ignore the context and use Background()
```
tmthrgd, line 183: Nit: missing full stop.

client.go
```diff
@@ -180,12 +204,22 @@ func (c *Client) exchange(m *Msg, co *Conn) (r *Msg, rtt time.Duration, err erro
...
+	if deadline, ok := ctx.Deadline(); ok {
+		if deadline.Before(writeDeadline) {
+			writeDeadline = deadline
+		}
+		if deadline.Before(readDeadline) {
+			readDeadline = deadline
+		}
+	}
+	co.SetWriteDeadline(writeDeadline)
+	co.SetReadDeadline(readDeadline)
```
tmthrgd, line 218: I think we should rearrange this slightly. I'd add a blank line here before `WriteMsg` and move the `co.TsigSecret, co.TsigProvider = c.TsigSecret, c.TsigProvider` line to directly above `WriteMsg`. And probably move `t := time.Now()` to right after the 'write timeout' comment.

client.go
tmthrgd: If you mean the `Close`, it shouldn't be as `Exchange` previously handled the dialing.

tmthrgd (changes requested):
Mostly minor nits. Also a more descriptive PR title would be good.

miekg:
Thanks for reviewing, addressed all your comments.

tmthrgd (approved):
LGTM.

PR #1290, opened 2021-07-28
Fix go generate missing required go.mod entry

### LINKED ISSUE #1287: go generate is broken
x/tools/packages moved to a new API and we're missing the go.sum field

after fixing _those_:

~~~
 go gen                                                                                                                                                                                               ~dns amtrelay +
go list stderr <<# github.com/miekg/dns
./defaults.go:94:12: cannot use s (type *SOA) as type RR in assignment:
	*SOA does not implement RR (missing isDuplicate method)
./defaults.go:116:20: cannot use t (type *TSIG) as type RR in append:
	*TSIG does not implement RR (missing isDuplicate method)
./defaults.go:139:38: impossible type assertion:
	*TSIG does not implement RR (missing isDuplicate method)
./dnssec.go:645:3: impossible type switch case: r1 (type RR) cannot have dynamic type *NS (missing isDuplicate method)
./dnssec.go:647:3: impossible type switch case: r1 (type RR) cannot have dynamic type *MD (missing isDuplicate method)
./dnssec.go:649:3: impossible type switch case: r1 (type RR) cannot have dynamic type *MF (missing isDuplicate method)
./dnssec.go:651:3: impossible type switch case: r1 (type RR) cannot have dynamic type *CNAME (missing isDuplicate method)
./dnssec.go:653:3: impossible type switch case: r1 (type RR) cannot have dynamic type *SOA (missing isDuplicate method)
./dnssec.go:656:3: impossible type switch case: r1 (type RR) cannot have dynamic type *MB (missing isDuplicate method)
./dnssec.go:658:3: impossible type switch case: r1 (type RR) cannot have dynamic type *MG (missing isDuplicate method)
./dnssec.go:658:3: too many errors
>>
2021/07/16 09:29:48 internal error: nil Pkg importing "context" from "github.com/miekg/dns"
exit status 1
~~~

tmthrgd:
There were no non-excluded files importing golang.org/x/tools so it's require was missing from go.mod. This caused the go run commands invoked by go generate to fail.

tools.go is currently the recommended way to force non-build dependencies to be retained by the go mod commands. If that changes in the future, we can update then. As nothing from the import is actually used, it won't impact anyone's build at all (except requiring an extra download by the go tool which is unavoidable).

Pulling in the @master version of golang.org/x/tools (which doesn't yet have a stable release version) also updates a number of other dependencies, but not quite to latest, so while I'm here just update them all.

Fixes #1287
Closes #1288


tmthrgd:
#1288 actually regressed x/tools back to a version from 2018 (`v0.0.0-20180917221912-90fa682c2a6e`) which caused the `packages.NeedTypes` build failures. This PR brings it up to `@master` and prevents `go mod tidy` from removing the import.

tools.go
```diff
@@ -0,0 +1,3 @@
+package dns
+
+import _ "golang.org/x/tools/go/packages"
```
miekg, line 3: little comment here on why we have this 3 line file?

tmthrgd: Will do.

PR #1291, opened 2021-08-06
Add ReverseAddrIP function

Felixoid:
Hello Meik, thank you for a great project.

To avoid unnecessary conversions from `dns.A -> string -> dns.ReverseAddr(string)` I've implemented ReverseAddrIP function.

Here you can see the benchmark results:

```
go test -bench BenchmarkReverse -benchmem . -benchtime 10000000x
goos: linux
goarch: amd64
pkg: github.com/miekg/dns
cpu: AMD Ryzen 7 4800H with Radeon Graphics         
BenchmarkReverseAddr/IP4-16      	10000000	      180.3 ns/op	     40 B/op	      2 allocs/op
BenchmarkReverseAddr/IP6-16      	10000000	      234.4 ns/op	     96 B/op	      2 allocs/op
BenchmarkReverseAddrIP/IP4-16    	10000000	       96.96 ns/op	     24 B/op	      1 allocs/op
BenchmarkReverseAddrIP/IP6-16    	10000000	      142.1 ns/op	     80 B/op	      1 allocs/op
BenchmarkReverseAddrFromIP/IP4-16         	10000000	      222.3 ns/op	     56 B/op	      3 allocs/op
BenchmarkReverseAddrFromIP/IP6-16         	10000000	      342.1 ns/op	    112 B/op	      3 allocs/op
PASS
ok  	github.com/miekg/dns	15.107s
```

### DIFF
--- a/defaults.go
+++ b/defaults.go
@@ -2,6 +2,7 @@ package dns
 
 import (
 	"errors"
+	"fmt"
 	"net"
 	"strconv"
 	"strings"
@@ -326,13 +327,23 @@ func CanonicalName(s string) string {
 // Copied from the official Go code.
 
 // ReverseAddr returns the in-addr.arpa. or ip6.arpa. hostname of the IP
-// address suitable for reverse DNS (PTR) record lookups or an error if it fails
-// to parse the IP address.
-func ReverseAddr(addr string) (arpa string, err error) {
-	ip := net.ParseIP(addr)
-	if ip == nil {
-		return "", &Error{err: "unrecognized address: " + addr}
+// address suitable for reverse DNS (PTR) record lookups. It accepts string or net.IP
+// as addr parameter. Error is returned if it's unable to parse IP address from string,
+// neither IPv4 nor IPv6 is used, or other but string of net.IP object is used
+func ReverseAddr(addr interface{}) (arpa string, err error) {
+	var ip net.IP
+	switch v := addr.(type) {
+	case net.IP:
+		ip = v
+	case string:
+		ip = net.ParseIP(v)
+		if ip == nil {
+			return "", &Error{err: "unrecognized address: " + v}
+		}
+	default:
+		return "", &Error{err: fmt.Sprintf("the addr type is not string or net.IP: %T", v)}
 	}
+
 	if v4 := ip.To4(); v4 != nil {
 		buf := make([]byte, 0, net.IPv4len*4+len("in-addr.arpa."))
 		// Add it, in reverse, to the buffer
@@ -343,6 +354,9 @@ func ReverseAddr(addr string) (arpa string, err error) {
 		// Append "in-addr.arpa." and return (buf already has the final .)
 		buf = append(buf, "in-addr.arpa."...)
 		return string(buf), nil
+	}
+	if v6 := ip.To16(); v6 == nil {
+		return "", &Error{err: "ip address neither IPv4 nor IPv6: " + ip.String()}
 	}
 	// Must be IPv6
 	buf := make([]byte, 0, net.IPv6len*4+len("ip6.arpa."))
--- a/dns_bench_test.go
+++ b/dns_bench_test.go
@@ -343,7 +343,7 @@ func BenchmarkIdGeneration(b *testing.B) {
 	}
 }
 
-func BenchmarkReverseAddr(b *testing.B) {
+func BenchmarkReverseAddrString(b *testing.B) {
 	b.Run("IP4", func(b *testing.B) {
 		for n := 0; n < b.N; n++ {
 			addr, err := ReverseAddr("192.0.2.1")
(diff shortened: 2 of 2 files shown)


defaults.go
```diff
@@ -333,6 +333,12 @@ func ReverseAddr(addr string) (arpa string, err error) {
 	if ip == nil {
 		return "", &Error{err: "unrecognized address: " + addr}
 	}
+	return ReverseIP(ip), nil
+}
+
+// ReverseIP returns the in-addr.arpa. or ip6.arpa. hostname of the IP
+// address suitable for reverse DNS (PTR) record lookups.
+func ReverseIP(ip net.IP) (arpa string) {
```
tmthrgd, line 341: `(arpa string)` -> `string`

defaults.go
```diff
@@ -333,6 +333,12 @@ func ReverseAddr(addr string) (arpa string, err error) {
 	if ip == nil {
 		return "", &Error{err: "unrecognized address: " + addr}
 	}
+	return ReverseIP(ip), nil
+}
+
+// ReverseIP returns the in-addr.arpa. or ip6.arpa. hostname of the IP
+// address suitable for reverse DNS (PTR) record lookups.
```
tmthrgd, line 340: This should document that `ip` must be either an IPv4 or IPv6 address.

tmthrgd: I’d probably also like it to return an empty string if it’s invalid. I think `ip.To6() == nil` should be the right check.

tmthrgd (changes requested):

tmthrgd:
I’m fine with this, it makes sense to me. Though I think `ReverseAddrIP` might be a better name.

defaults.go
Felixoid: When you say so regarding check for `ip.To16`, I feel like the function should return an error as well

Felixoid:
Thank you for review, I've added the changes

Felixoid:
@tmthrgd and @miekg, anything preventing it from merging?

Felixoid:
Well, I hope you at least have a lot of fun on vacations :+1: :heart:

Felixoid:
@miekg and @tmthrgd, any hopes to get it reviewed?

miekg:
There is also ReverseAddr ? So now ewe have two? The benchmarks show ns/op but you claim memory efficiency, also don't really think that's important...
The commit message are rather sparse and don't have any of the comments you put in the PR text

Felixoid:
Hey Miek.

> There is also ReverseAddr ? So now we have two?

Yes. Somebody may find it useful to have ReverseAddr still. So I don't feel comfortable breaking the backward compartibility.

> The benchmarks show ns/op but you claim memory efficiency, also don't really think that's important...

```
go test -bench BenchmarkReverse -benchmem . -benchtime 10000000x
goos: linux
goarch: amd64
pkg: github.com/miekg/dns
cpu: AMD Ryzen 7 4800H with Radeon Graphics         
BenchmarkReverseAddr/IP4-16      	10000000	      180.3 ns/op	     40 B/op	      2 allocs/op
BenchmarkReverseAddr/IP6-16      	10000000	      234.4 ns/op	     96 B/op	      2 allocs/op
BenchmarkReverseAddrIP/IP4-16    	10000000	       96.96 ns/op	     24 B/op	      1 allocs/op
BenchmarkReverseAddrIP/IP6-16    	10000000	      142.1 ns/op	     80 B/op	      1 allocs/op
BenchmarkReverseAddrFromIP/IP4-16         	10000000	      222.3 ns/op	     56 B/op	      3 allocs/op
BenchmarkReverseAddrFromIP/IP6-16         	10000000	      342.1 ns/op	    112 B/op	      3 allocs/op
PASS
ok  	github.com/miekg/dns	15.107s
```

I'd rather agree with you. But the new function looks useful to me since it avoids the unnecessary converting from `net.IP` address to string and back to `net.IP`. To show what I mean I've added BenchmarkReverseAddrFromIP

> The commit message are rather sparse and don't have any of the comments you put in the PR text

Updated the commit and rebased it on top of the current master :+1:

Felixoid:
Ok, I didn't get the answer, but from the gut feeling, it looks like the current approach is better than adding an extra function.

defaults.go
```diff
@@ -326,13 +327,23 @@ func CanonicalName(s string) string {
...
-// address suitable for reverse DNS (PTR) record lookups or an error if it fails
-// to parse the IP address.
-func ReverseAddr(addr string) (arpa string, err error) {
-	ip := net.ParseIP(addr)
-	if ip == nil {
-		return "", &Error{err: "unrecognized address: " + addr}
+// address suitable for reverse DNS (PTR) record lookups. It accepts string or net.IP
+// as addr parameter. Error is returned if it's unable to parse IP address from string,
+// neither IPv4 nor IPv6 is used, or other but string of net.IP object is used
+func ReverseAddr(addr interface{}) (arpa string, err error) {
```
tmthrgd, line 333: This isn't backwards compatible so we can't make this change.

tmthrgd (changes requested):
This isn't backwards compatible.

PR #1293, opened 2021-09-04
Handle packing of empty RDATA correctly for EDNS0_EXPIRE (Resolves #1292)

### LINKED ISSUE #1292: TSIG fails to verify when EDNS Expire is set
I have set up a Bind server as secondary to my DNS server that runs `miekg/dns` library and I was seeing `bad signature` errors. After some `dig`ging and packet sniffing I found out that the reason the verification fails is because the Bind server is sending an additional option for EDNS Expire.

I have setup a simple playground environment to mimic this behavior:

```
package main

import (
	"flag"
	"fmt"
	"log"
	"os"
	"os/signal"
	"strconv"
	"syscall"

	"github.com/miekg/dns"
)

type mockServer struct {
}

func (ms *mockServer) ServeDNS(writer dns.ResponseWriter, query *dns.Msg) {

	raw, _ := query.Pack()
	err := dns.TsigVerify(raw, "Ch/iN6YuFKTX6xI7Fm00HkcPaD/Pk2ilKc6gOyt8/kE=", "", false)
	if err != nil {
		fmt.Printf("error: %s", err)
	} else {
		fmt.Printf("signature verified")
	}

	err = writer.WriteMsg(query)
	if err != nil {
		return
	}
}

func main() {
	port := flag.Int("port", 8053, "port to run on")

	go func() {
		srv := &dns.Server{Addr: ":" + strconv.Itoa(*port), Net: "udp", Handler: &mockServer{}}
		if err := srv.ListenAndServe(); err != nil {
			log.Fatalf("Failed to set udp listener %s\n", err.Error())
		}
	}()

	go func() {
		srv := &dns.Server{Addr: ":" + strconv.Itoa(*port), Net: "tcp", Handler: &mockServer{}}
		if err := srv.ListenAndServe(); err != nil {
			log.Fatalf("Failed to set tcp listener %s\n", err.Error())
		}
	}()

	sig := make(chan os.Signal)
	signal.Notify(sig, syscall.SIGINT, syscall.SIGTERM)
	s := <-sig
	log.Fatalf("Signal (%v) received, stopping\n", s)
}
```

If you do a DNS query without the EDNS expiry option the query is verified correctly:
```
dig @localhost -p 8053 +nocookie -y hmac-sha256:key2.:Ch/iN6YuFKTX6xI7Fm00HkcPaD/Pk2ilKc6gOyt8/kE= foo.invalid SOA

signature verified
```
but if you do a DNS with the EDNS expiry it fails verifying the signature
```
dig @localhost -p 8053 +nocookie +ednsopt=9 -y hmac-sha256:key2.:Ch/iN6YuFKTX6xI7Fm00HkcPaD/Pk2ilKc6gOyt8/kE= foo.invalid SOA

error: dns: bad signature
```

I tried with other EDNS options and the verification is working fine. Is this a bug or am I missing something regarding the Expiry option?

dmavrommatis:
Issue explained in #1292

### DIFF
--- a/edns.go
+++ b/edns.go
@@ -576,7 +576,7 @@ func (e *EDNS0_N3U) copy() EDNS0 { return &EDNS0_N3U{e.Code, e.AlgCode} }
 // EDNS0_EXPIRE implements the EDNS0 option as described in RFC 7314.
 type EDNS0_EXPIRE struct {
 	Code   uint16  // Always EDNS0EXPIRE
-	Expire []uint8 // can be zero or 4 length
+	Expire []uint8 // either zero or 4 octets
 }
 
 // Option implements the EDNS0 interface.
@@ -590,7 +590,7 @@ func (e *EDNS0_EXPIRE) pack() ([]byte, error) {
 }
 func (e *EDNS0_EXPIRE) unpack(b []byte) error {
 	if len(b) != 0 && len(b) != 4 {
-		return errors.New("dns: expire length mismatch, want 0/4 but got " + strconv.Itoa(len(b)))
+		return ErrBuf
 	}
 	e.Expire = b
 	return nil
@@ -598,11 +598,9 @@ func (e *EDNS0_EXPIRE) unpack(b []byte) error {
 
 func (e *EDNS0_EXPIRE) String() (s string) {
 	if len(e.Expire) == 0 {
-		s = "<omitted>"
-	} else {
-		s = fmt.Sprintf("<%d>", binary.BigEndian.Uint32(e.Expire))
+		return "<MISSING>"
 	}
-	return s
+	return fmt.Sprintf("%d", binary.BigEndian.Uint32(e.Expire))
 }
 
 // The EDNS0_LOCAL option is used for local/experimental purposes. The option
--- a/edns.go
+++ b/edns.go
@@ -575,32 +575,43 @@ func (e *EDNS0_N3U) copy() EDNS0 { return &EDNS0_N3U{e.Code, e.AlgCode} }
 
 // EDNS0_EXPIRE implements the EDNS0 option as described in RFC 7314.
 type EDNS0_EXPIRE struct {
-	Code   uint16  // Always EDNS0EXPIRE
-	Expire []uint8 // either zero or 4 octets
+	Code   uint16 // Always EDNS0EXPIRE
+	Expire uint32
+	Empty  bool
 }
 
 // Option implements the EDNS0 interface.
 func (e *EDNS0_EXPIRE) Option() uint16 { return EDNS0EXPIRE }
-func (e *EDNS0_EXPIRE) copy() EDNS0    { return &EDNS0_EXPIRE{e.Code, e.Expire} }
+func (e *EDNS0_EXPIRE) copy() EDNS0    { return &EDNS0_EXPIRE{e.Code, e.Expire, e.Empty} }
+
 func (e *EDNS0_EXPIRE) pack() ([]byte, error) {
-	if len(e.Expire) != 0 && len(e.Expire) != 4 {
-		return nil, errors.New("dns: expire length is not 0/4")
+	if e.Empty {
+		return []byte{}, nil
 	}
-	return e.Expire, nil
+	b := make([]byte, 4)
+	binary.BigEndian.PutUint32(b, e.Expire)
+	return b, nil
 }
+
 func (e *EDNS0_EXPIRE) unpack(b []byte) error {
-	if len(b) != 0 && len(b) != 4 {
+	if len(b) == 0 {
+		// zero-length EXPIRE query, see RFC 7314 Section 2
+		e.Empty = true
+		return nil
+	}
+	if len(b) < 4 {
 		return ErrBuf
 	}
-	e.Expire = b
+	e.Expire = binary.BigEndian.Uint32(b)
+	e.Empty = false
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


miekg:
Thanks. Could you provide a better description in your commit. uint32 was plain wrong? Where in the RFC do I need to look? Etc.

dmavrommatis:
I've documented the bug and thought process in the issue mentioned. I added the following description in the commit which also mentions the RFC:

As per [RFC7134](https://datatracker.ietf.org/doc/html/rfc7314#section-2) the Expire Option in queries should be zero-length. In the current implementation the field is `uint32` which always instatiates 4bytes for that field when packing to wire format.
For that reason we change the field to `[]uint8` so it can support 0-length and 4-byte
length option data.

edns.go
```diff
@@ -575,33 +575,36 @@ func (e *EDNS0_N3U) copy() EDNS0 { return &EDNS0_N3U{e.Code, e.AlgCode} }
...
-		return ErrBuf
-	}
-	e.Expire = binary.BigEndian.Uint32(b)
+	e.Expire = b
 	return nil
 }
 
+func (e *EDNS0_EXPIRE) String() (s string) {
+	if len(e.Expire) == 0 {
+		s = "<omitted>"
```
miekg, line 601: omitted? This comes from the RFC? also you can return here and drop the else clause

edns.go
```diff
@@ -575,33 +575,36 @@ func (e *EDNS0_N3U) copy() EDNS0 { return &EDNS0_N3U{e.Code, e.AlgCode} }
 
 // EDNS0_EXPIRE implements the EDNS0 option as described in RFC 7314.
 type EDNS0_EXPIRE struct {
-	Code   uint16 // Always EDNS0EXPIRE
-	Expire uint32
+	Code   uint16  // Always EDNS0EXPIRE
+	Expire []uint8 // can be zero or 4 length
```
miekg, line 579: 'either zero or 4 octets'

edns.go
```diff
@@ -575,33 +575,36 @@ func (e *EDNS0_N3U) copy() EDNS0 { return &EDNS0_N3U{e.Code, e.AlgCode} }
...
+	}
+	return e.Expire, nil
 }
-
 func (e *EDNS0_EXPIRE) unpack(b []byte) error {
-	if len(b) == 0 {
-		// zero-length EXPIRE query, see RFC 7314 Section 2
-		return nil
+	if len(b) != 0 && len(b) != 4 {
+		return errors.New("dns: expire length mismatch, want 0/4 but got " + strconv.Itoa(len(b)))
```
miekg, line 593: this returned ErrBuf before

edns.go
dmavrommatis: RFC doesn't specify. I can change it to `<MISSING>` which is what wireshark returns when the value is not set.

edns.go
dmavrommatis: should I set it back to ErrBuf? the explanation does not match the actual error
```
ErrBuf           error = &Error{err: "buffer size too small"}          // ErrBuf indicates that the buffer used is too small for the message.
```

dmavrommatis:
@miekg thanks for the review; I've addressed the comments. Please take a look again!

tmthrgd:
Any code that operates on `EDNS0_EXPIRE` would be broken by this as the type of `Expire` will have changed. Consider code that did something like (nonsense example):

```go
func adjustExpire(opt *EDNS0_EXPIRE) {
	if opt.Expire < 10 {
		opt.Expire = 10
	} else {
		opt.Expire++
	}
}
```

Even code that simply inspects or prints the value of `Expire` will break.

A flag allows us to track whether we're writing zero or four bytes and persist that from unpack to pack without *(hopefully)* breaking any existing code.

dmavrommatis:
> Any code that operates on `EDNS0_EXPIRE` would be broken by this as the type of `Expire` will have changed. Consider code that did something like (nonsense example):
> 
> ```go
> func adjustExpire(opt *EDNS0_EXPIRE) {
> 	if opt.Expire < 10 {
> 		opt.Expire = 10
> 	} else {
> 		opt.Expire++
> 	}
> }
> ```
> 
> Even code that simply inspects or prints the value of `Expire` will break.
> 
> A flag allows us to track whether we're writing zero or four bytes and persist that from unpack to pack without _(hopefully)_ breaking any existing code.

oh gotcha, I will update the PR to avoid this issue

dmavrommatis:
@tmthrgd found some time to make the change. please check to see if the implementation is sufficient.

dmavrommatis:
hey @tmthrgd I think this PR is ready for merging!

edns.go
```diff
@@ -577,14 +577,17 @@ func (e *EDNS0_N3U) copy() EDNS0 { return &EDNS0_N3U{e.Code, e.AlgCode} }
 type EDNS0_EXPIRE struct {
 	Code   uint16 // Always EDNS0EXPIRE
 	Expire uint32
+	Empty  bool
```
miekg, line 580: This should have documentation. Something along the lines, Empty is used to signal an empty Expire option in a backwards compatible way, it's not used on the wire.

dmavrommatis: added the comment as suggested

dmavrommatis:
@tmthrgd @miekg can you check if the PR is complete now?

tmthrgd (approved):
LGTM. As long as we're happy with the `String` output.

miekg (approved):

PR #1294, opened 2021-09-04
Fix un/packing of EDNS0 TCP keepalive extension

rs:
The current code is assuming the OPT code and length are part of the
data passed to the unpacker and that it should be appenedded during
packing while those fields are parsed by the caller.

This change fixes the parsing and make sure a request with a keepalive
won't fail.


miekg:
thanks, can we add some kind of test for this?

rs:
The refactor with multiple returns changed the behavior of my initial PR regarding String. It is thus not correct anymore as you noted.

PR #1297, opened 2021-09-13
Export TsigProvider functions (Resolves #1296)

### LINKED ISSUE #1296: TsigProvider functions are not exported
When you define a custom DNS handler there is no way to specify a custom TsigProvider.

`TsigGenerate` and `TsigVerify` pass hard-coded the default TsigProvider and the `tsigGenerateProvider` and `tsigVerifyProvider` are unexported. 

Then only place I saw that you can define the TsigProvider is in the DNS client which is unrelated to the DNS handler for receiving traffic.

dmavrommatis:
Export `TsigProvider` functions to allow use when writing custom DNS server handlers.

Context in #1296
```
When you define a custom DNS handler there is no way to specify a custom TsigProvider.

TsigGenerate and TsigVerify pass hard-coded the default TsigProvider and the tsigGenerateProvider and tsigVerifyProvider are unexported.

The only place I saw that you can define the TsigProvider is in the DNS client which is unrelated to the DNS handler for receiving traffic.
```

### DIFF
--- a/client.go
+++ b/client.go
@@ -260,7 +260,7 @@ func (co *Conn) ReadMsg() (*Msg, error) {
 	}
 	if t := m.IsTsig(); t != nil {
 		if co.TsigProvider != nil {
-			err = tsigVerifyProvider(p, co.TsigProvider, co.tsigRequestMAC, false)
+			err = TsigVerifyProvider(p, co.TsigProvider, co.tsigRequestMAC, false)
 		} else {
 			if _, ok := co.TsigSecret[t.Hdr.Name]; !ok {
 				return m, ErrSecret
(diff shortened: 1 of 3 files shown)


miekg:
why can't you use co.TsigProvider ?

dmavrommatis:
> why can't you use co.TsigProvider ?

As far as I understand the code this is only when we initiate a connection as a client.

How can I do that if I am the server inside the `ServeDNS` handler function?
```
func (server *DNSServer) ServeDNS(writer dns.ResponseWriter, query *dns.Msg) {
  ...
  reply.SetTsig(tsig.Hdr.Name, tsig.Algorithm, 300, time.Now().Unix())
  data, _, err = dns.TsigGenerate(reply, tsigSecret, tsig.MAC, false)
  writer.Write(data)
}
```

dmavrommatis:
After digging more into the code, the problem seems to be that the server does not support `TsigProvider`. So this means that you have to directly use the `TsigGenerateProvider` function and then with the writer to use `Write` instead of `WriteMsg`.

dmavrommatis:
I'm going to close this PR as this issue is mostly a design issue on TsigProviders for server/xfr rather than implementation

PR #1301, opened 2021-09-27
Added support for ENUM Source-URI Extension

Trial97:
Hi Miek,

Added the ENUM Source-URI Extension described here: https://datatracker.ietf.org/doc/html/draft-kaplan-enum-source-uri-00

Thanks for this great project!


miekg (approved):

sebastic:
Note that option code 4 is reserved for [draft-cheshire-edns0-owner-option](https://datatracker.ietf.org/doc/draft-cheshire-edns0-owner-option/), see:

 https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#dns-parameters-11

Oracle (ab)uses this option code in their Acme Packet [Source URI Information in ENUM Requests](https://docs.oracle.com/cd/E95619_01/html/esbc_ecz810_configuration/GUID-547D2727-D3F9-4F36-AF99-0BED47F506B8.htm#GUID-BE12EB6B-F5FD-4448-80A0-CD7228AB2F2E) implementation.

Heuristics may be needed to support both drafts in the future.

miekg:
[ Quoting ***@***.***> in "Re: [miekg/dns] Added support for E..." ]
>Note that option code 4 is reserved for draft-cheshire-edns0-owner-option, see:
>
>https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#
>dns-parameters-11
>
>Oracle (ab)uses this option code in their Acme Packet Source URI Information in
>ENUM Requests implementation.
>
>Heuristics may be needed to support both drafts in the future.

interesting... that should be sorted at the iana level at some point. We will not add
heuristics to detect things here, only one implementation can claim this code point.


/Miek

--
Miek Gieben

PR #1304, opened 2021-10-06
Add basic support for contexts in handlers

tangrs:
This change adds basic context-aware versions of ServeMux and the Handler types.

The request context pattern in Go is very useful when writing servers, since they can propagate metadata about the request, such as deadlines. I currently have code that implements a custom wrapper around `ResponseWriter` to inject and propagate request contexts through nested ServeMuxes. However this approach is somewhat ugly and I'd much rather see the upstream library have built-in support for contexts.

Please take a look to see if this code is a reasonable approach. Care has been taken to preserve backwards compatibility with existing code.

Alternatives considered:

- Add a `Context` getter to `*Msg` in the same fashion as in `*http.Request`. This doesn't seem like a good idea, since `*Msg` is useful independent of the server, and adding a context field to it seems like it's polluting the API.
- Add a `Context` getter to `ResponseWriter`. This doesn't seem to make much semantic sense (the context is associated with the request, not the response). It also can break compatibility with existing code that implement custom ResponseWriters (since adding a function to the interface would require the user to implement a new method).

### DIFF
--- a/serve_mux.go
+++ b/serve_mux.go
@@ -1,34 +1,34 @@
 package dns
 
 import (
+	"context"
 	"sync"
 )
 
-// ServeMux is an DNS request multiplexer. It matches the zone name of
-// each incoming request against a list of registered patterns add calls
-// the handler for the pattern that most closely matches the zone name.
+// handlerContextFromHandler is used internally to wrap a Handler into a
+// HandlerContext, to aid in maintaining backwards compatibility.
+func handlerContextFromHandler(h Handler) HandlerContext {
+	return HandlerFuncContext(func(_ context.Context, w ResponseWriter, r *Msg) {
+		h.ServeDNS(w, r)
+	})
+}
+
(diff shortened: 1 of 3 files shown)


miekg:
coredns uses this: https://github.com/coredns/coredns/blob/master/core/dnsserver/server.go#L111
seems to work fine

tangrs:
Ah, it looks like CoreDNS is essentially implementing their own ServeMux, so they're not bound to using the Handler interface. The code that I'm working with currently makes heavy use of the built-in ServeMux implementation.

I suppose I could also reimplement a custom, context-aware ServeMux equivalent for my own code, but this doesn't seem great in terms of code re-use. I think it'd be nicer to just support this in the library.

Any thoughts?

miekg:
I think that shows that's doable with the current code. I don't want to carry a whole bunch of ...Context stuff around just because context.Context became a thing 4 (5? 6?) years ago.

PR #1311, opened 2021-10-30
Support TsigProvider for zone transfers

### LINKED ISSUE #1300: XFR is not using the TsigProvider
Even if the `TsigProvider` is set on the connection/transfer struct the read/write functions do not use the provider.

https://github.com/miekg/dns/blob/master/xfr.go#L215-L254

tmthrgd:
This adds `TsigProvider` support to `Transfer`.

I've not added any tests as the existing XFR TSIG code doesn't have any tests and I can't say I understand this code well enough to write any tests myself.

Fixes #1300

/cc @dmavrommatis

### DIFF
--- a/xfr.go
+++ b/xfr.go
@@ -18,6 +18,7 @@ type Transfer struct {
 	ReadTimeout    time.Duration     // net.Conn.SetReadTimeout value for connections, defaults to 2 seconds
 	WriteTimeout   time.Duration     // net.Conn.SetWriteTimeout value for connections, defaults to 2 seconds
 	TsigSecret     map[string]string // Secret(s) for Tsig map[<zonename>]<base64 secret>, zonename must be in canonical form (lowercase, fqdn, see RFC 4034 Section 6.2)
+	TsigProvider   TsigProvider      // An implementation of the TsigProvider interface. If defined it replaces TsigSecret and is used for all TSIG operations.
 	tsigTimersOnly bool
 }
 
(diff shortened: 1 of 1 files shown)


tmthrgd:
@dmavrommatis can you check if this fixes your problem?

dmavrommatis:
I remember when I was looking at this that there were some issues with this approach. 
Without the provider when you create a secret you also pass the `t.Hdr.Name` field to get the algorithm while with the tsigProvider you don't. I can't recall exactly what was the issue though.

tmthrgd:
It's slightly cumbersome, but you can get access to `t.Hdr.Name` from within `TsigProvider`. The first argument `msg` is the exact byte-slice we parsed initially. You could safely do something like this:

```go
m := new(dns.Msg)
m.Unpack(msg)
name := m.IsTsig().Hdr.Name
```

Considering the only way you can get there is to have `Unpack` succeed and `IsTsig()` return non-nil, you don't even need to check the error or return values.

dmavrommatis:
Should be good then. Are you going to do the same for the DNS server to support the TsigProvider? https://github.com/miekg/dns/blob/master/server.go#L715

tmthrgd:
@dmavrommatis That would make sense. I'll do that in another PR.

Fattouche:
What is the status of this being merged? I didn't realize that it already existed and I ended up making my own PR to solve this problem. https://github.com/miekg/dns/pull/1323.

tmthrgd:
I'm withdrawing this in favour of a much nicer implementation.

PR #1312, opened 2021-11-05
add option to some rr for extend

victory460:
Signed-off-by: xuweiwei <xuweiwei_yewu@cmss.chinamobile.com>

**Why is this pull request needed and what does it do?**
In many cases we have some requirement for intelligent resolve, such as by line and weight. So need to extend the original A, AAAA, CNAME records.
For example, it can filter some rrs by the _opt interface{}_ . Following code implement a demo that filter rrs by city name in _coredns_.
```
type RROption struct {
	City   string
	Weight int
}

func autoLineFilter(ctx context.Context, rrs []dns.RR) []dns.RR {
	for index, item := range rrs {
		if a, ok := item.(*dns.A); ok {
			if getCityName := metadata.ValueFunc(ctx, "geoip/city/name"); getCityName != nil {
				if cityName := getCityName(); len(cityName) > 0 {
					aRR := a.opt.(RROption)
					if aRR.City == cityName {
						rrs = rrs[index : index+1]
					} 
				} 
			} 
		}
	}
	return rrs
}
```
This _opt interface{}_ can also be extended to enable more interesting functionality by add additional information to it.

### DIFF
--- a/types.go
+++ b/types.go
@@ -277,6 +277,7 @@ func (*NULL) parse(c *zlexer, origin string) *ParseError {
 type CNAME struct {
 	Hdr    RR_Header
 	Target string `dns:"cdomain-name"`
+	opt    interface{}
 }
 
 func (rr *CNAME) String() string { return rr.Hdr.String() + sprintName(rr.Target) }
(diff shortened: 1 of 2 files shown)


miekg:
If you need to store meta data with an RR, you'll need to wrap it yourself:
~~~
type rrOpt {
         rr dns.RR
         opt interface{}
}
~~~
We cannot increase the memory taken up by each RR to make implementing an edge usecase slightly easier.

PR #1316, opened 2021-11-30
Do not sign BADKEY and BADSIG TSIG error responses

chrisohaver:
Per RFC 8945 5.3.2, responses with BADKEY and BADSIG errors must not be signed.

> When a server detects an error relating to the key or MAC in the
   incoming request, the server SHOULD send back an unsigned error
   message (MAC Size == 0 and empty MAC).  It MUST NOT send back a
   signed error message.

https://datatracker.ietf.org/doc/html/rfc8945#section-5.3.2

Signed-off-by: Chris O'Haver <cohaver@infoblox.com>

### DIFF
--- a/tsig.go
+++ b/tsig.go
@@ -162,20 +162,31 @@ func tsigGenerateProvider(m *Msg, provider TsigProvider, requestMAC string, time
 	if err != nil {
 		return nil, "", err
 	}
-	buf, err := tsigBuffer(mbuf, rr, requestMAC, timersOnly)
-	if err != nil {
-		return nil, "", err
-	}
 
 	t := new(TSIG)
-	// Copy all TSIG fields except MAC and its size, which are filled using the computed digest.
-	*t = *rr
-	mac, err := provider.Generate(buf, rr)
-	if err != nil {
-		return nil, "", err
+
+	if rr.Error == RcodeSuccess || rr.Error == RcodeBadTime {
+		// Only compute response MAC for non-errors and bad time errors (RFC 8945 5.3.2)
+		buf, err := tsigBuffer(mbuf, rr, requestMAC, timersOnly)
+		if err != nil {
+			return nil, "", err
+		}
+		// Copy all TSIG fields except MAC and its size, which are filled using the computed digest.
+		*t = *rr
+		mac, err := provider.Generate(buf, rr)
+		if err != nil {
+			return nil, "", err
+		}
+		t.TimeSigned = rr.TimeSigned
+		t.MAC = hex.EncodeToString(mac)
+		t.MACSize = uint16(len(t.MAC) / 2) // Size is half!
+	} else {
+		// Copy all TSIG fields except MAC, its size and time signed, since we aren't signing.
+		*t = *rr
+		t.TimeSigned = 0
+		t.MAC = ""
+		t.MACSize = 0
 	}
-	t.MAC = hex.EncodeToString(mac)
-	t.MACSize = uint16(len(t.MAC) / 2) // Size is half!
 
 	tbuf := make([]byte, Len(t))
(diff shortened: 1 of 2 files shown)


miekg:
Good catch. The PR description should at least also show up in the commit and commit title could also just be the PR title used here.

tsig.go
```diff
@@ -162,20 +162,30 @@ func tsigGenerateProvider(m *Msg, provider TsigProvider, requestMAC string, time
...
+		}
+
+		mac, err := provider.Generate(buf, rr)
+		if err != nil {
+			return nil, "", err
+		}
+		t.MAC = hex.EncodeToString(mac)
+		t.MACSize = uint16(len(t.MAC) / 2) // Size is half!
+	} else {
+		// for TSIG errors (except bad time errors), respond with an empty MAC
```
miekg, line 184: this can probably be pulled out of the if-statement and set unconditionally above it

tsig.go
```diff
@@ -162,20 +162,30 @@ func tsigGenerateProvider(m *Msg, provider TsigProvider, requestMAC string, time
...
-	}
 
 	t := new(TSIG)
 	// Copy all TSIG fields except MAC and its size, which are filled using the computed digest.
 	*t = *rr
-	mac, err := provider.Generate(buf, rr)
-	if err != nil {
-		return nil, "", err
+
+	if rr.Error == 0 || rr.Error == RcodeBadTime {
```
miekg, line 170: 0? Would a constant be better here?

tsig.go
```diff
@@ -162,20 +162,30 @@ func tsigGenerateProvider(m *Msg, provider TsigProvider, requestMAC string, time
...
+		// Only compute response MAC for non-errors and bad time errors
```
miekg, line 171: add RFC reference here as well

tsig.go
chrisohaver: When testing this I tripped over a bug, which would require us to copy the TSIG struct twice for non-errors if we set unconditionally above it. I think it works better with the `else`... but if you're really opposed to the else, I can look for a way yo refactor the immediate code to accommodate it.

tsig.go
chrisohaver: yes - changed to RcodeSuccess.

chrisohaver:
> Good catch. The PR description should at least also show up in the commit and commit title could also just be the PR title used here.

Commits squashed with a more descriptive commit message. "Per RFC 8945 5.3.2, responses with BADKEY and BADSIG errors must not be signed."

tsig.go
chrisohaver: > I can look for a way yo refactor the immediate code to accommodate 

Done. The `else` is gone.

tsig.go
```diff
@@ -162,20 +162,29 @@ func tsigGenerateProvider(m *Msg, provider TsigProvider, requestMAC string, time
...
+	// Copy all TSIG fields except MAC, its size, and time signed which are filled when signing.
 	*t = *rr
-	mac, err := provider.Generate(buf, rr)
-	if err != nil {
-		return nil, "", err
+	t.TimeSigned = 0
+	t.MAC = ""
+	t.MACSize = 0
+
+	if rr.Error == RcodeSuccess || rr.Error == RcodeBadTime {
```
tmthrgd, line 178: If the RFC only wants us to avoid signing on "an error relating to the key or MAC", wouldn't this make more sense as "rr.Error != ..." for the relevant errors? I presume we're still aloud and want to sign other error values.

chrisohaver: Yeah, that makes better sense if we’re allowed to return other errors.

chrisohaver: Done. Thanks!

tmthrgd (approved):
LGTM.

PR #1317, opened 2021-12-14
Fix edns keepalive

ameshkov:
Added tests for EDNS0_TCP_KEEPALIVE pack/unpack as requested in the original PR.

Closes #1294


miekg:
this succeeds https://github.com/miekg/dns/pull/1294 ? and that can be closed?

ameshkov:
@miekg I think if you merge this one #1294 will be automatically merged as well since this PR contains the commit by @rs

edns.go
```diff
@@ -658,56 +660,40 @@ func (e *EDNS0_LOCAL) unpack(b []byte) error {
 // the TCP connection alive. See RFC 7828.
 type EDNS0_TCP_KEEPALIVE struct {
 	Code    uint16 // Always EDNSTCPKEEPALIVE
-	Length  uint16 // the value 0 if the TIMEOUT is omitted, the value 2 if it is present;
```
tmthrgd, line 661: We can't removed an exported field on an exported struct. That's a no go.

tmthrgd (changes requested):

edns.go
rs: No valid keepalive edns could be parsed or generated with the current code, so it is unlikely somebody reference this field that should not exist in the first place.

ameshkov: @tmthrgd can we keep it but mark deprecated and just not use it?

tmthrgd: I don't necessarily object to that. I'm happy to review a change that does that and see what it's like.

miekg: I don't mind either way, breaking users is a compile time error (good! Not some fuzzy thing failing at runtime) and a fairly easy fix (remove setting the field). OTOH, keeping it (for now) and marking it unused in the docs doesn't really cost us anything

ameshkov: Length is back, marked deprecated now.

edns.go
```diff
@@ -657,57 +659,49 @@ func (e *EDNS0_LOCAL) unpack(b []byte) error {
...
-	}
-	if e.Timeout == 0 && e.Length != 0 {
-		return nil, errors.New("dns: timeout not specified but length is not 0")
-	}
-	b := make([]byte, 4+e.Length)
-	binary.BigEndian.PutUint16(b[0:], e.Code)
-	binary.BigEndian.PutUint16(b[2:], e.Length)
-	if e.Length == 2 {
-		binary.BigEndian.PutUint16(b[4:], e.Timeout)
+func (e *EDNS0_TCP_KEEPALIVE) pack() (b []byte, err error) {
```
tmthrgd, line 677: Remove the named returns and just include a `return b, nil` in the if-statement.

tmthrgd (changes requested):

ameshkov:
Fixed review comments

tmthrgd (approved):
I have no idea if the `String` change is correct or not, but the rest of it LGTM.

miekg:
hadn't spotted the String change, see no real reason do that (also), or is the RFC mandating something?

thanks @tmthrgd

ameshkov:
Restored `String`'s behavior

tmthrgd (approved):
LGTM.

PR #1319, opened 2021-12-24
Add https://github.com/markdingo/autoreverse into list of README.md users

markdingo:
As requested on the README.md page. I've created a pull request to add a new application into the user mix.

It's not clear whether there is any order to this list, so I just added it at the end. Obviously that can be changed as needed.


markdingo:
Thanks. I appreciate the commit.

PR #1322, opened 2021-12-28
Only treat a *net.UnixConn of unixgram as a packet conn

### LINKED ISSUE #1321: not support unix socks
I have a piece of test code：
```go
package main

import (
	"fmt"
	"net"
	"strings"
)

func main() {
	var tcp net.Conn = &net.TCPConn{}
	_, ok := tcp.(net.PacketConn)
	fmt.Println("tcp", ok)

	var unix net.Conn = &net.UnixConn{}
	_, ok = unix.(net.PacketConn)
	fmt.Println("unix", ok)

	var udp net.Conn = &net.UDPConn{}
	_, ok = udp.(net.PacketConn)
	fmt.Println("udp", ok)
}
```
result：
tcp false
unix true
udp true

I want TCP and UNIX results to be consistent，But the end result is that UDP and UNIX are consistent，The problem code is as follows
```go
// ReadMsgHeader reads a DNS message, parses and populates hdr (when hdr is not nil).
// Returns message as a byte slice to be parsed with Msg.Unpack later on.
// Note that error handling on the message body is not possible as only the header is parsed.
func (co *Conn) ReadMsgHeader(hdr *Header) ([]byte, error) {
	var (
		p   []byte
		n   int
		err error
	)

// FIXME   The problem code is as follows
	if _, ok := co.Conn.(net.PacketConn); ok {
		if co.UDPSize > MinMsgSize {
			p = make([]byte, co.UDPSize)
		} else {
			p = make([]byte, MinMsgSize)
		}
		n, err = co.Read(p)
	} else {
		var length uint16
		if err := binary.Read(co.Conn, binary.BigEndian, &length); err != nil {
			return nil, err
		}

		p = make([]byte, length)
	}

	if err != nil {
		return nil, err
	} else if n < headerSize {
		return nil, ErrShortRead
	}

	p = p[:n]
	if hdr != nil {
		dh, _, err := unpackMsgHdr(p, 0)
		if err != nil {
			return nil, err
		}
		*hdr = dh
	}
	return p, err
}
```
```go
// Write implements the net.Conn Write method.
func (co *Conn) Write(p []byte) (int, error) {
	if len(p) > MaxMsgSize {
		return 0, &Error{err: "message too large"}
	}

//   FIXME  The problem code is as follows
	if _, ok := co.Conn.(net.PacketConn); ok {
		return co.Conn.Write(p)
	}

	msg := make([]byte, 2+len(p))
	binary.BigEndian.PutUint16(msg, uint16(len(p)))
	copy(msg[2:], p)
	return co.Conn.Write(msg)
}
```
At present, I use the following methods in my project
```go
conn, err = net.DialTimeout(c.Net, addr, c.Timeout)
if _, ok := conn.(*net.UnixConn); ok {
	conn = &mc{c: conn}   //  Here are my changes
}
return &dns.Conn{
	UDPSize: c.UDPSize,
	Conn:    conn,
}, nil

type mc struct{ c net.Conn }
func (c *mc) Read(b []byte) (n int, err error)   { return c.c.Read(b) }
func (c *mc) Write(b []byte) (n int, err error)  { return c.c.Write(b) }
func (c *mc) Close() error                       { return c.c.Close() }
func (c *mc) LocalAddr() net.Addr                { return c.c.LocalAddr() }
func (c *mc) RemoteAddr() net.Addr               { return c.c.RemoteAddr() }
func (c *mc) SetDeadline(t time.Time) error      { return c.c.SetDeadline(t) }
func (c *mc) SetReadDeadline(t time.Time) error  { return c.c.SetReadDeadline(t) }
func (c *mc) SetWriteDeadline(t time.Time) error { return c.c.SetWriteDeadline(t) }
```

tmthrgd:
Go supports three different types of unix connections `SOCK_STREAM` (unix), `SOCK_DGRAM` (unixgram) and `SOCK_SEQPACKET` (unixpacket) all of which are `*net.UnixConn`s. That means `*net.UnixConn` always implements `net.PacketConn` regardless of the socket type, though certain methods will return errors.

`SOCK_DGRAM` is the only one we want to treat as a packet connection and use the UDP framing for (IIUC). To do this we special case `*net.UnixConn`s and only treat `SOCK_DGRAM` as a `net.PacketConn`.

This will make it possible to properly use the DNS client over unix `SOCK_STREAM` connections (the most common type).

Fixes #1321

### DIFF
--- a/client.go
+++ b/client.go
@@ -18,6 +18,11 @@ const (
 	tcpIdleTimeout time.Duration = 8 * time.Second
 )
 
+func isPacketConn(c net.Conn) bool {
+	_, ok := c.(net.PacketConn)
+	return ok
+}
+
 // A Conn represents a connection to a DNS server.
 type Conn struct {
 	net.Conn                         // a net.Conn holding the connection
@@ -221,7 +226,7 @@ func (c *Client) exchangeContext(ctx context.Context, m *Msg, co *Conn) (r *Msg,
 		return nil, 0, err
 	}
 
-	if _, ok := co.Conn.(net.PacketConn); ok {
+	if isPacketConn(co.Conn) {
 		for {
 			r, err = co.ReadMsg()
 			// Ignore replies with mismatched IDs because they might be
@@ -282,7 +287,7 @@ func (co *Conn) ReadMsgHeader(hdr *Header) ([]byte, error) {
 		err error
 	)
 
-	if _, ok := co.Conn.(net.PacketConn); ok {
+	if isPacketConn(co.Conn) {
 		if co.UDPSize > MinMsgSize {
 			p = make([]byte, co.UDPSize)
 		} else {
@@ -322,7 +327,7 @@ func (co *Conn) Read(p []byte) (n int, err error) {
 		return 0, ErrConnEmpty
 	}
 
-	if _, ok := co.Conn.(net.PacketConn); ok {
+	if isPacketConn(co.Conn) {
 		// UDP connection
 		return co.Conn.Read(p)
 	}
@@ -371,7 +376,7 @@ func (co *Conn) Write(p []byte) (int, error) {
 		return 0, &Error{err: "message too large"}
 	}
 
-	if _, ok := co.Conn.(net.PacketConn); ok {
+	if isPacketConn(co.Conn) {
 		return co.Conn.Write(p)
 	}
 
--- a/client.go
+++ b/client.go
@@ -19,6 +19,10 @@ const (
 )
 
 func isPacketConn(c net.Conn) bool {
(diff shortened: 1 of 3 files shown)


tmthrgd:
@jan-bar Could you please test this and see if it fixes your issue?

miekg (approved):

jan-bar:
> @jan-bar Could you please test this and see if it fixes your issue?

Unfortunately, the server I need to connect to is of the unix type, I tried to use the unixgram type, and the server returned an error  **connect: protocol wrong type for socket**.  But if I modify the `unixgram` in your code to `unix`, it can be used normally. Thank you for your contribution.

PR #1324, opened 2022-01-10
Support custom tsig names inside tsigSecret map

Fattouche:
Previously the tsigSecret variable on the Server object was keyed on tsig name.
This means that if you have multiple zones using the same server, it is not possible
for each of the zones to have the same tsig name. By specifying a TsigKeyNameBuilder
function, the server can now differentiate between zones even if they have the same tsig name
without requiring the client to change anything.

### DIFF
--- a/client.go
+++ b/client.go
@@ -32,11 +32,12 @@ func isPacketConn(c net.Conn) bool {
 
 // A Conn represents a connection to a DNS server.
 type Conn struct {
-	net.Conn                         // a net.Conn holding the connection
-	UDPSize        uint16            // minimum receive buffer for UDP messages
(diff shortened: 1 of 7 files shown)


client.go
```diff
@@ -32,11 +32,12 @@ func isPacketConn(c net.Conn) bool {
 
 // A Conn represents a connection to a DNS server.
 type Conn struct {
```
miekg, line 34: conn is used a lot... it should grow and grow and we've already added a new tsig thing here.

tmthrgd:
Please no. Just use `TsigProvider`. It was added for exactly these kinds of use cases.

PR #1331, opened 2022-01-25
Support TsigProvider for Server and Transfer

### LINKED ISSUE #1300: XFR is not using the TsigProvider

### LINKED ISSUE #1325: TSIG names cannot be reused across different zones.
Due to the nature of the TsigSecret in the Server{} object, it's not possible to re-use TSIG names across different zones. 

For example:
```
TsigSecret = map[string]string{"axfr.": "so6ZGir4GPAqINNh9U5c3A=="}
```
Here the tsig name is `axfr.` which means that if any other zone wanted to use `axfr.` as well, it would be overwrite this one. Alternatively, the server could specify the TsigSecret to use zoneName for the TSIG name, which is better because duplicates would not occur, however this means that every client using that zone would need to have the same TSIG and also wouldn't allow TSIGs to be re-used across multiple zones which again is not great.

tmthrgd:
This adds support for `TsigProvider` to both `Server` and `Transfer`. It also simplifies the implementation for `Client` and `Conn`.

Not sure why I didn't think to do this earlier. A miekg/dns/v2 could just push the `TsigSecret` code out into the callers.

This also has the advantage that it doesn't add a bunch of extra code paths throughout, so it's safer to land.

Fixes #1300
Closes #1323
Fixes #1325

### DIFF
--- a/server.go
+++ b/server.go
@@ -71,12 +71,12 @@ type response struct {
 	tsigTimersOnly bool
 	tsigStatus     error
 	tsigRequestMAC string
-	tsigSecret     map[string]string // the tsig secrets
-	udp            net.PacketConn    // i/o connection if UDP was used
-	tcp            net.Conn          // i/o connection if TCP was used
-	udpSession     *SessionUDP       // oob data to get egress interface right
-	pcSession      net.Addr          // address to use when writing to a generic net.PacketConn
-	writer         Writer            // writer to output the raw DNS bits
+	tsigProvider   TsigProvider
+	udp            net.PacketConn // i/o connection if UDP was used
+	tcp            net.Conn       // i/o connection if TCP was used
+	udpSession     *SessionUDP    // oob data to get egress interface right
+	pcSession      net.Addr       // address to use when writing to a generic net.PacketConn
+	writer         Writer         // writer to output the raw DNS bits
 }
 
 // handleRefused returns a HandlerFunc that returns REFUSED for every request it gets.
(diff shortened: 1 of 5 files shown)


server.go
```diff
@@ -238,6 +240,16 @@ type Server struct {
 	udpPool sync.Pool
 }
 
+func (srv *Server) tsigProvider() TsigProvider {
```
Fattouche, line 243: This is a nice approach 👍

tsig.go
```diff
@@ -74,6 +74,24 @@ func (key tsigHMACProvider) Verify(msg []byte, t *TSIG) error {
 	return nil
 }
 
+type tsigSecretProvider map[string]string
```
Fattouche, line 77: This is good for a default TsigProvider, however it still doesn't solve the problem described in 

https://github.com/miekg/dns/issues/1325#issuecomment-1018712904

We need some way for Generate to be able to unwrap the Message so that we can see the query name. This is specifically useful in situations where multiple tsigs have the same name, but exist under a different zone.

tmthrgd: That can be handled either in a separate PR or externally, but yes we do need to do *something*.

tmthrgd: How common a use case is that do you think?

Fattouche: I would say it's fairly common. I ran into this problem early on with this library when trying to serve multiple zones that had different primaries, therefore different TSIGs on each primary meaning they could not use the TSIG name as the zone Name.

miekg (approved):
lgtm

tsig.go
Fattouche: Hey @tmthrgd have you thought any more about how we can best obtain the query name from the Generate function. I think having the opposite of TsigBuffer available in a function would make this possible. Alternatively we could consider passing in a context? It would be useful to have contexts everywhere anyway so that we could pass data around without needing to alter signatures.

PR #1336, opened 2022-02-02
Client.exchange: Use a minimum bufsize of DefaultMsgSize when EDNS0 is present

sladkani:
Some old peers do not properly honor the advertized EDNS0 bufsize,
sending a larger response.

As a result, the read buffer passed to 'co.Read()' is insufficient to
store the actual response, leading to a message parse error.

Let's be more liberal in the receive path: when EDNS0 is present, have
'co.UDPSize' at least 'DefaultMsgSize' (4096) bytes long, in order to
workaround such situations.

### DIFF
--- a/client.go
+++ b/client.go
@@ -206,6 +206,11 @@ func (c *Client) exchangeContext(ctx context.Context, m *Msg, co *Conn) (r *Msg,
 	// If EDNS0 is used use that for size.
 	if opt != nil && opt.UDPSize() >= MinMsgSize {
 		co.UDPSize = opt.UDPSize()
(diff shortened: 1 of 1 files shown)


miekg:
>Some old peers do not properly honor the advertized EDNS0 bufsize,
>sending a larger response.

those should be fixed.

>Let's be more liberal in the receive path: when EDNS0 is present, have
>'co.UDPSize' at least 'DefaultMsgSize' (4096) bytes long, in order to
>workaround such situations.

this approach (in DNS) has led to more problems than it solved. DNS servers and clients
not honoring the spec have hold the DNS back for decades. We should not do this anymore.

so nack

/close

PR #1338, opened 2022-02-06
Invalid NSEC/3 bitmap on non-zero buffer

rs:
If the PackBuffer is used to encode an NSEC/3 record, the bitmap is
xored with the content of the buffer instead of being zeroed first.

The algorithm has been changed so it is able zero bytes without
losing too much performance (around 2x slower).

### DIFF
--- a/msg_helpers.go
+++ b/msg_helpers.go
@@ -558,31 +558,23 @@ func packDataNsec(bitmap []uint16, msg []byte, off int) (int, error) {
 	if len(bitmap) == 0 {
 		return off, nil
 	}
+	if off > len(msg) {
+		return off, &Error{err: "overflow packing nsec"}
+	}
+	toZero := msg[off:]
+	if maxLen := typeBitMapLen(bitmap); maxLen < len(toZero) {
+		toZero = toZero[:maxLen]
+	}
+	for i := range toZero {
+		toZero[i] = 0
+	}
 	var lastwindow, lastlength uint16
-	for i, t := range bitmap {
+	for _, t := range bitmap {
 		window := t / 256
 		length := (t-window*256)/8 + 1
-		if (window > lastwindow && lastlength != 0) || i == 0 {
-			if i > 0 {
-				// New window, jump to the new offset
-				off += int(lastlength) + 2
-				lastlength = 0
-			}
-
-			// Zero out window bitmap
-			windowLength := length
-			for j := i + 1; j < len(bitmap); j++ {
-				if bitmap[j]>>8 == window {
-					windowLength = bitmap[j]&0xff/8 + 1
-				}
-			}
-			lastOff := off + 1 + int(windowLength)
-			if lastOff > len(msg) {
-				return len(msg), &Error{err: "overflow packing nsec"}
-			}
-			for zoff := off + 1; zoff <= lastOff; zoff++ {
-				msg[zoff] = 0
-			}
+		if window > lastwindow && lastlength != 0 { // New window, jump to the new offset
+			off += int(lastlength) + 2
+			lastlength = 0
 		}
 		if window < lastwindow || length < lastlength {
 			return len(msg), &Error{err: "nsec bits out of order"}
--- a/msg_helpers.go
+++ b/msg_helpers.go
@@ -554,45 +554,52 @@ func typeBitMapLen(bitmap []uint16) int {
 	return l
 }
 
-func packDataNsec(bitmap []uint16, msg []byte, off int) (int, error) {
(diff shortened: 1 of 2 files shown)


tmthrgd:
What’s the actual issue here? Is it observable using the packages public API?

rs:
I creates invalid DNS messages when `PackBuffer` is used with a non-zeroed buffer. The existing tests would actually not pass if the generated content was tested (like it is now added).

This is randomly breaking DNSSEC when using `PackBuffer`.

miekg:
[ Quoting ***@***.***> in "Re: [miekg/dns] Invalid NSEC/3 bitm..." ]
>I creates invalid DNS messages when PackBuffer is used with a non-zeroed
>buffer. The existing tests would actually not pass if the generated content was
>tested (like it is now added).
>
>This is randomly breaking DNSSEC when using PackBuffer.

Think this should added to the function docs and you should clear the buffer, no need to
impose that cost on every user.

rs:
When you reuse buffers, you are not expected to clear them, that would defeat the purpose of reusing buffers. Your own existing tests was expecting a non zero buffer and was failing to generate a valid message.

The performance difference is 10ns, it’s not a big cost. Zeroing the full buffer would cost more.

miekg:
ok,

I don't like the entire re-write of the loop, I could understnad the first iteration. Please see if there is a more minimal change, also use the go std naming scheme, so no underscores like `_off`

rs:
First loop is for window blocks of high order 8 bits and the second is to encode the low order 8 bits bitmaps and the third is to find all additional types fitting in the same bitmap. There is another loop to make sure we zero empty bitmaps before the current one if any. I don't see a more efficient way to do it with less loops. Previous implementation couldn't know which bytes to 0.

I'll rename `_off`.

rs:
Please have a look at the last commit. I renamed a few vars + added some comments for clarity.

rs:
@miekg are you considering merging this fix?

tmthrgd:
Considering we only advance `off` in one place, why can't we just put a zeroing loop there to zero that windows data? Plus one to zero the first window and we're done. I don't think we need to rewrite or overcomplicate this.

rs:
You don't know the number of lower order 8bit bitmap blocks in advance, and the previous algorithm had no way to know if the current block its writing in needs to be zeroed. The previous algorithm also not give you a hint on the number of blocks was skipped and thus need to be zeroed as well.

tmthrgd:
But we can trivially know the maximum `length` for a given window and just opportunistically zero up to that.

rs:
Sure you can, it will be another loop and performance will be probably the same or worth.

rs:
I reverted to the previous algo with window size compute+0 on each window boundaries.

rs:
@miekg are you still considering merging this fix?

tmthrgd (commented):
This is a lot of complexity that I'm not the biggest fan off. What's the performance like of just zeroing the rest of `msg`?

rs:
It depends on the size of `msg`.

tmthrgd:
Even when using the assembly backed `for i := range x { x[i] = 0 }`?

rs:
This is with this construct yes.

tmthrgd:
I just noticed we already have `typeBitMapLen`. I think we should just do something like the following before the main loop:

```go
toZero := msg[off:]
if maxLen := typeBitMapLen(bitmap); maxLen < len(toZero) {
	toZero = toZero[:maxLen]
}
for i := toZero {
	toZero[i] = 0
}
```

rs:
Awersome, I moved to this function. Please review.

msg_helpers.go
```diff
@@ -558,6 +558,16 @@ func packDataNsec(bitmap []uint16, msg []byte, off int) (int, error) {
 	if len(bitmap) == 0 {
 		return off, nil
 	}
+	if off > len(msg) {
```
tmthrgd, line 561: I don't think it's possible for us to ever be called with `off > len(msg)`, so I think this check is unneeded.

tmthrgd (changes requested):
Minor change. Otherwise LGTM.

msg_helpers.go
rs: Famous last word :)

The first unit test was testing this and would panic without this test. The cost is marginal so I would tend leave it.

tmthrgd: It is too. I’m certain we can never hit this condition, but I’ll defer to the unit test.

tmthrgd (approved):
LGTM.

PR #1340, opened 2022-02-10
Add identityHash for algos that don't need hashing

miekg:
This adds hash.go and creates a identityHash that is used for algorithms
that do their own hashing (ED25519) for instance.

This unifies the hash variable naming between dnssec and sig(0) signing
and removes the special casing that existed for ED25519.

This unifies the variable naming between sig(0) and dnssec signing and
verifying.

I didn't want to used crypto.RegisterHash as not to fiddle with the
global namespaces of hashes, so the value of '0' from AlgorithmsToHash
is handled specially in dnssec and sig(0) code.

<strike>Note that ED448 isn't implemented at all.</strike> Is not implemented in the standard libary

### DIFF
--- a/hash.go
+++ b/hash.go
@@ -1,0 +1,31 @@
+package dns
+
+import (
+	"bytes"
+	"crypto"
+	"hash"
+)
+
+// identityHash will not hash, it only buffers the data written into it and returns it as-is.
+type identityHash struct {
+	b *bytes.Buffer
+}
+
+// Implement the hash.Hash interface.
+
+func (i identityHash) Write(b []byte) (int, error) { return i.b.Write(b) }
+func (i identityHash) Size() int                   { return i.b.Len() }
+func (i identityHash) BlockSize() int              { return 1024 }
+func (i identityHash) Reset()                      { i.b.Reset() }
+func (i identityHash) Sum(b []byte) []byte         { return append(b, i.b.Bytes()...) }
+
+func hashFromAlgorithm(alg uint8) (hash.Hash, crypto.Hash, error) {
+	hashnumber, ok := AlgorithmToHash[alg]
+	if !ok {
+		return nil, 0, ErrAlg
+	}
+	if hashnumber == 0 {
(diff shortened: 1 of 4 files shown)


tmthrgd (changes requested):

miekg:
>+func (i identityHash) Write(b []byte) (int, error) { return i.b.Write(b) }
>+func (i identityHash) Size() int                   { return i.b.Len() }
>+func (i identityHash) BlockSize() int              { return 1024 }
>+func (i identityHash) Reset()                      { i.b.Reset() }
>+func (i identityHash) Sum(b []byte) []byte         { i.b.Write(b); return i.b.Bytes() }
>
>This is wrong. Sum should append to b not write it into the buffer. It allows
>the caller to Sum into a preallocated byte-slice.

oh yeah, good catch, will change

hash.go
```diff
@@ -0,0 +1,31 @@
+package dns
+
+import (
+	"bytes"
+	"crypto"
+	"hash"
+)
+
+// identityHash will not hash, it only buffers the data written into it and returns it as-is.
+type identityHash struct {
```
tmthrgd, line 10: Rather than adding this, what about if instead we did something like:
```go
func hashFromAlgorithm(alg uint8, data ...[]byte) (hash.Hash, []byte, error)
```

Then just perform the hashing or appending inside that function?

I don't know if that's nicer or not, but it could be.

miekg: I think that would be shorter (in the end), however there are some shenigans in the sig0 code, where there is a couple of lines of:
~~~
	sigend := offset
	h.Write(buf[sigstart:sigend])
	h.Write(buf[:10])
	h.Write([]byte{
		byte((adc - 1) << 8),
		byte(adc - 1),
	})
	h.Write(buf[12:bodyend])

	hashed := h.Sum(nil)
~~~
and the `hashFromAlgorithm` is some lines back.

So yes, probably a good idea to clean further, but we don't already have the correct bytes in a neat slice when we call this.

PR #1341, opened 2022-02-20
Update SVCB

DesWurstes:
The only breaking change worth mentioning is renaming SVCB_ECHCONFIG to SVCB_ECH. We can avoid this and use its legacy name internally if you wish.

I had chosen ( https://github.com/miekg/dns/pull/1067#issuecomment-699635130 ) to forbid IPv4 addresses in ipv6hint to be consistent with the rest of the library. I don't know why ( https://github.com/MikeBishop/dns-alt-svc/issues/361 ) but the SVCB spec requires otherwise.

@miekg @tmthrgd Would you like to review?

### DIFF
--- a/svcb.go
+++ b/svcb.go
@@ -20,7 +20,7 @@ const (
 	SVCB_NO_DEFAULT_ALPN
 	SVCB_PORT
 	SVCB_IPV4HINT
-	SVCB_ECH
+	SVCB_ECHCONFIG
 	SVCB_IPV6HINT
 
 	svcb_RESERVED SVCBKey = 65535
@@ -32,7 +32,7 @@ var svcbKeyToStringMap = map[SVCBKey]string{
 	SVCB_NO_DEFAULT_ALPN: "no-default-alpn",
 	SVCB_PORT:            "port",
 	SVCB_IPV4HINT:        "ipv4hint",
-	SVCB_ECH:             "ech",
+	SVCB_ECHCONFIG:       "ech",
 	SVCB_IPV6HINT:        "ipv6hint",
 }
 
@@ -188,8 +188,8 @@ func makeSVCBKeyValue(key SVCBKey) SVCBKeyValue {
 		return new(SVCBPort)
 	case SVCB_IPV4HINT:
 		return new(SVCBIPv4Hint)
-	case SVCB_ECH:
-		return new(SVCBECH)
+	case SVCB_ECHCONFIG:
+		return new(SVCBECHConfig)
 	case SVCB_IPV6HINT:
 		return new(SVCBIPv6Hint)
 	case svcb_RESERVED:
@@ -542,42 +542,42 @@ func (s *SVCBIPv4Hint) copy() SVCBKeyValue {
 	}
 }
 
-// SVCBECH pair contains the ECHConfig structure defined in draft-ietf-tls-esni [RFC xxxx].
+// SVCBECHConfig pair contains the ECHConfig structure defined in draft-ietf-tls-esni [RFC xxxx].
 // Basic use pattern for creating an ech option:
 //
 //	h := new(dns.HTTPS)
 //	h.Hdr = dns.RR_Header{Name: ".", Rrtype: dns.TypeHTTPS, Class: dns.ClassINET}
-//	e := new(dns.SVCBECH)
-//	e.ECHConfigList = []byte{0xfe, 0x08, ...}
+//	e := new(dns.SVCBECHConfig)
+//	e.ECH = []byte{0xfe, 0x08, ...}
 //	h.Value = append(h.Value, e)
-type SVCBECH struct {
-	ECHConfigList []byte // includes the redundant length prefix
+type SVCBECHConfig struct {
+	ECH []byte // Specifically ECHConfigList including the redundant length prefix
 }
 
-func (*SVCBECH) Key() SVCBKey     { return SVCB_ECH }
-func (s *SVCBECH) String() string { return toBase64(s.ECHConfigList) }
-func (s *SVCBECH) len() int       { return len(s.ECHConfigList) }
+func (*SVCBECHConfig) Key() SVCBKey     { return SVCB_ECHCONFIG }
+func (s *SVCBECHConfig) String() string { return toBase64(s.ECH) }
+func (s *SVCBECHConfig) len() int       { return len(s.ECH) }
 
-func (s *SVCBECH) pack() ([]byte, error) {
-	return append([]byte(nil), s.ECHConfigList...), nil
+func (s *SVCBECHConfig) pack() ([]byte, error) {
+	return append([]byte(nil), s.ECH...), nil
 }
 
-func (s *SVCBECH) copy() SVCBKeyValue {
-	return &SVCBECH{
-		append([]byte(nil), s.ECHConfigList...),
+func (s *SVCBECHConfig) copy() SVCBKeyValue {
+	return &SVCBECHConfig{
+		append([]byte(nil), s.ECH...),
 	}
 }
 
-func (s *SVCBECH) unpack(b []byte) error {
-	s.ECHConfigList = append([]byte(nil), b...)
+func (s *SVCBECHConfig) unpack(b []byte) error {
+	s.ECH = append([]byte(nil), b...)
 	return nil
 }
-func (s *SVCBECH) parse(b string) error {
+func (s *SVCBECHConfig) parse(b string) error {
 	x, err := fromBase64([]byte(b))
 	if err != nil {
 		return errors.New("dns: svcbech: bad base64 ech")
 	}
-	s.ECHConfigList = x
+	s.ECH = x
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


miekg:
Breaking changes in this record I fine, it's still draft IIRC

> I had chosen ( https://github.com/miekg/dns/pull/1067#issuecomment-699635130 ) to forbid IPv4 addresses in ipv6hint to be consistent with the rest of the library. I don't know why ( https://github.com/MikeBishop/dns-alt-svc/issues/361 ) but the SVCB spec requires otherwise.

That sgtm, internal consistency is a good thing to have

svcb.go
```diff
@@ -13,14 +13,14 @@ import (
 // SVCBKey is the type of the keys used in the SVCB RR.
 type SVCBKey uint16
 
-// Keys defined in draft-ietf-dnsop-svcb-https-01 Section 12.3.2.
+// Keys defined in draft-ietf-dnsop-svcb-https-08 Section 14.3.2.
 const (
 	SVCB_MANDATORY       SVCBKey = 0
```
miekg, line 18:
```suggestion
	SVCB_MANDATORY       SVCBKey = iota
```

svcb.go
```diff
@@ -13,14 +13,14 @@ import (
...
 	SVCB_ALPN            SVCBKey = 1
 	SVCB_NO_DEFAULT_ALPN SVCBKey = 2
 	SVCB_PORT            SVCBKey = 3
 	SVCB_IPV4HINT        SVCBKey = 4
-	SVCB_ECHCONFIG       SVCBKey = 5
+	SVCB_ECH             SVCBKey = 5
 	SVCB_IPV6HINT        SVCBKey = 6
```
miekg, line 24: can remove all the "= N" and let `iota` do its thing

svcb.go
```diff
@@ -13,14 +13,14 @@ import (
...
 	svcb_RESERVED        SVCBKey = 65535
```
miekg, line 25:
```suggestion

	svcb_RESERVED        SVCBKey = 65535
```

svcb.go
miekg: think you can even remove the type in that case

svcb.go
```diff
@@ -200,12 +200,14 @@ func makeSVCBKeyValue(key SVCBKey) SVCBKeyValue {
...
 
-// SVCB RR. See RFC xxxx (https://tools.ietf.org/html/draft-ietf-dnsop-svcb-https-01).
+// SVCB RR. See RFC xxxx (https://tools.ietf.org/html/draft-ietf-dnsop-svcb-https-08).
 type SVCB struct {
 	Hdr      RR_Header
 	Priority uint16
 	Target   string         `dns:"domain-name"`
-	Value    []SVCBKeyValue `dns:"pairs"` // Value must be empty if Priority is zero.
+	Value    []SVCBKeyValue `dns:"pairs"`
+	// If Priority is zero, Value should be empty.
```
miekg, line 209: move this as a comment after Priority

svcb.go
```diff
@@ -200,12 +200,14 @@ func makeSVCBKeyValue(key SVCBKey) SVCBKeyValue {
...
+	// If it is not, when receiving, Value must be considered empty.
```
miekg, line 210: not sure what this adds, expand/delete?

svcb.go
```diff
@@ -235,15 +237,26 @@ type SVCBKeyValue interface {
...
 
 // SVCBMandatory pair adds to required keys that must be interpreted for the RR
-// to be functional.
+// to be functional. If ignored, the whole RRSet must be ignored.
+// "port" and "no-default-alpn" are mandatory by default if present,
+// so they shouldn't be included here.
+//
+// It is incumbent upon the user of this library to reject the RRSet if
+// or avoid constructing such an RRSet that:
+// - "mandatory" is included as one of the keys of mandatory
```
miekg, line 246: if you want this to make a list in godoc you need empty lines before and after. Maybe just make this more prose?

svcb.go
```diff
@@ -235,15 +237,26 @@ type SVCBKeyValue interface {
...
+// - no key is listed multiple times in mandatory
+// - all keys listed in mandatory are present
+// - escape sequences are not used in mandatory
+// - mandatory, when present, lists at least one key
+//
 // Basic use pattern for creating a mandatory option:
 //
 //	s := &dns.SVCB{Hdr: dns.RR_Header{Name: ".", Rrtype: dns.TypeSVCB, Class: dns.ClassINET}}
 //	e := new(dns.SVCBMandatory)
 //	e.Code = []uint16{65403}
```
miekg, line 256: this is just a random number in the "correct" range?

svcb.go
```diff
@@ -302,6 +315,7 @@ func (s *SVCBMandatory) copy() SVCBKeyValue {
 }
 
 // SVCBAlpn pair is used to list supported connection protocols.
+// The user of this library must ensure that at least one protocol is listed when alpn is present.
 // Protocol ids can be found at:
```
miekg, line 319: IDs

svcb.go
```diff
@@ -370,6 +384,7 @@ func (s *SVCBAlpn) copy() SVCBKeyValue {
 }
 
 // SVCBNoDefaultAlpn pair signifies no support for default connection protocols.
+// Should be used in conjunction with alpn.
```
miekg, line 387: 'It should..."

and then "only be used" ?

svcb.go
```diff
@@ -522,42 +537,42 @@ func (s *SVCBIPv4Hint) copy() SVCBKeyValue {
...
-// SVCBECHConfig pair contains the ECHConfig structure defined in draft-ietf-tls-esni [RFC xxxx].
-// Basic use pattern for creating an echconfig option:
+// SVCBECH pair contains the ECHConfig structure defined in draft-ietf-tls-esni [RFC xxxx].
+// Basic use pattern for creating an ech option:
 //
 //	h := new(dns.HTTPS)
 //	h.Hdr = dns.RR_Header{Name: ".", Rrtype: dns.TypeHTTPS, Class: dns.ClassINET}
-//	e := new(dns.SVCBECHConfig)
+//	e := new(dns.SVCBECH)
 //	e.ECH = []byte{0xfe, 0x08, ...}
```
miekg, line 546: doc doesn't match struct members anymore

svcb.go
```diff
@@ -522,42 +537,42 @@ func (s *SVCBIPv4Hint) copy() SVCBKeyValue {
...
 //	h.Value = append(h.Value, e)
-type SVCBECHConfig struct {
-	ECH []byte
+type SVCBECH struct {
+	ECHConfigList []byte // includes the redundant length prefix
```
miekg, line 549: this is too long.  Just `Config`?  What does the draft use?

DesWurstes: The draft uses `ECHConfigList`

svcb.go
DesWurstes: I can't turn this list into a paragraph; it would be unreadable.

DesWurstes:
@miekg  Many thanks for your review. I have addresses your comments except "if you want this to make a list in godoc you need empty lines before and after. Maybe just make this more prose?" because I don't know how to turn this into a paragraph without making it unreadable.

> > I had chosen ( https://github.com/miekg/dns/pull/1067#issuecomment-699635130 ) to forbid IPv4 addresses in ipv6hint to be consistent with the rest of the library. I don't know why ( https://github.com/MikeBishop/dns-alt-svc/issues/361 ) but the SVCB spec requires otherwise.
>
> That sgtm, internal consistency is a good thing to have

On the SVCB side, they decided to update their example, so this library now achieves full compliance and consistency. Hence I updated the relevant code.

svcb.go
```diff
@@ -167,10 +168,10 @@ func (rr *SVCB) parse(c *zlexer, o string) *ParseError {
 		}
 		l, _ = c.Next()
 	}
+
+	// Don't check rr.Priority == 0 && len(xs) > 0
+	// as explained in parse_test.go
```
tmthrgd, line 173: Other way around. Test can reference comment in the code, not the opposite.

tmthrgd (changes requested):
Most of the change is fine, but I really don't think we can or should make backwards incompatible changes. Nothing here was marked as WIP or anything. Considering the name changes don't have any actual impact on the code, I think they should be left as-is regardless of whether that differs from the draft spec or not. I'm fine with the other changes though.

svcb.go
```diff
@@ -13,16 +13,17 @@ import (
 // SVCBKey is the type of the keys used in the SVCB RR.
 type SVCBKey uint16
 
-// Keys defined in draft-ietf-dnsop-svcb-https-01 Section 12.3.2.
```
DesWurstes, line 16: @tmthrgd Does this count?

tmthrgd: No it doesn't.

DesWurstes:
Thanks for the review. I addressed the issues.

tmthrgd (approved):
LGTM.

DesWurstes:
Hello again @miekg . I have been impatiently waiting for Cloudflare to implement ECH over SVCB so it would be great if we could get this merged unless further reviews are planned. Many thanks.

DesWurstes:
👍thank you

PR #1346, opened 2022-03-10
Add context to server handler interfaces

### LINKED ISSUE #1345: Information cannot be passed from the TCP/UDP read decorators to server handler
This allows server operators to pass data from the read decorators to the actual DNS handlers.
    
An example use case where this is needed is when there is any proxying service between eyeball and DNS server. In order
to preserve the client IP, the proxy server might have to use PROXY protocol v2. If you don't have a context you can read and see
the client IPs in the decorators but your handler will not know this information and therefore might not be able to do things like:
    
    1. Block specific IPs
    2. Verify TSIGs based off specific IP
    3. Respond with TSIGs based off source IP
    4. Respond with different answers based off Geo Location
    
Apart from knowing the source IP, there is other situations where you might want to pass information parsed in the read decorators to the handler.

Fattouche:
This allows server operators to pass data from the read decorators to the actual DNS handlers.

An example use case where this is needed is when there is any proxying service between eyeball and DNS server. In order
to preserve the client IP, the proxy server might have to use PROXY protocol v2. If you don't have a context you can read and see the client IPs in the decoraters but your handler will not know this information and therefore might not be able to do things like:

1. Block IPs
2. Verify TSIGs based off specific IP
3. Respond with TSIGs based off source IP
4. Respond with different answers based of IP location

Apart from knowing the source IP, there is other situations where you might want to pass information parsed in the read decorators to the handler.

Fixes #1345

### DIFF
--- a/acceptfunc_test.go
+++ b/acceptfunc_test.go
@@ -1,6 +1,7 @@
 package dns
 
 import (
+	"context"
 	"testing"
 )
 
@@ -28,7 +29,7 @@ func TestAcceptNotify(t *testing.T) {
 	}
 }
 
-func handleNotify(w ResponseWriter, req *Msg) {
+func handleNotify(ctx context.Context, w ResponseWriter, req *Msg) {
 	m := new(Msg)
 	m.SetReply(req)
 	w.WriteMsg(m)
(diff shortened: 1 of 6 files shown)


miekg:
not needed here, coredns shows how this can be easily done.

and indeed, not backwards compatible

Fattouche:
> not needed here, coredns shows how this can be easily done.
> 
> and indeed, not backwards compatible

CoreDNS doesn't appear to add values to the ctx at query time, it just adds some arbitrary key at serveDNS time. Am I missing something?

PR #1348, opened 2022-03-10
Pass DNS Msg via context into TSIG Verify and Generate functions

### LINKED ISSUE #1347: TSIGProvider Generate and Verify functions don't have access to the DNS Message
This causes problems on zone transfers where the server is hosting many different zone names. Passing the DNS Message via the context as a parameter opens up many possibilities for how TSIGs are handled. For example:

1. We could use a sync.RWMutex to ensure that even after the server has been started, TSIGs can be rotated out of band without worrying about data races.
2. We could allow multiple zones with different names to support TSIGs with the same name. 

There is many more possibilities that this could open up. I.E Zone specific TSIG verification/Generation.

Fattouche:
This gives the server operator a lot more flexibility as to how they manage the verification and generation of tsigs.
The test TestServerRoundtripTsigProvider displays some of these possibilities.

For example:

1. Establish a sync.RWMutex on the secret map so that secrets can be dynamically updated even after the server has been started.
2. Establish more granular secret maps. I.E to ensure that many zones can share TSIG names without collisions in the map.

You could do other things like have zone specific verification or generation.

Fixes #1347

### DIFF
--- a/acceptfunc_test.go
+++ b/acceptfunc_test.go
@@ -1,6 +1,7 @@
 package dns
 
 import (
+	"context"
 	"testing"
 )
 
@@ -28,7 +29,7 @@ func TestAcceptNotify(t *testing.T) {
 	}
 }
 
-func handleNotify(w ResponseWriter, req *Msg) {
+func handleNotify(ctx context.Context, w ResponseWriter, req *Msg) {
 	m := new(Msg)
 	m.SetReply(req)
 	w.WriteMsg(m)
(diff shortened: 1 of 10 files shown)


miekg:
context.Context is a good way to cancel functions, using is to smuggle extra parameter into a function is not.

Fattouche:
> This isn't backwards compatible.

I'm trying to solve https://github.com/miekg/dns/pull/1331#discussion_r819703410. Surely there must be a way to get the zoneName inside Generate() function, however it seems that after calling `TsigBuffer` it's non trivial. I don't want to break signatures, but preventing TSIGs from having duplicate names across different zones is far too strict IMO.

PR #1359, opened 2022-04-11
Add SVCB dohpath key

ainar-g:
The parameter is being added in [its own IETF draft][1] and also being used in
the [IETF draft about Discovery of Designated Resolvers][2].

Additionally, the mappings of the numeric key values to strings are exported,
under names consistent with the already existing exported mappings, to make it
easier for the clients of the module to validate and print SVCB keys.

Testing was done by sending SVCB queries for the "_dns.resolver.arpa" domain to
OpenDNS's 146.112.41.2 server.

[1]: https://datatracker.ietf.org/doc/html/draft-ietf-add-svcb-dns-02
[2]: https://datatracker.ietf.org/doc/html/draft-ietf-add-ddr-06.html

### DIFF
--- a/svcb.go
+++ b/svcb.go
@@ -679,8 +679,8 @@ func (s *SVCBIPv6Hint) copy() SVCBKeyValue {
 // See RFC xxxx (https://datatracker.ietf.org/doc/html/draft-ietf-add-svcb-dns-02)
 // and RFC yyyy (https://datatracker.ietf.org/doc/html/draft-ietf-add-ddr-06).
 //
-// Basic use pattern for using the dohpath option together with an alpn
-// option:
+// A basic example of using the dohpath option together with the alpn
+// option to indicate support for DNS over HTTPS on a certain path:
 //
 //	s := new(dns.SVCB)
 //	s.Hdr = dns.RR_Header{Name: ".", Rrtype: dns.TypeSVCB, Class: dns.ClassINET}
@@ -690,35 +690,23 @@ func (s *SVCBIPv6Hint) copy() SVCBKeyValue {
 //	p.Template = "/dns-query{?dns}"
 //	s.Value = append(s.Value, e, p)
 //
-// The parsing currently only validates the length of Template.
-// It doesn't validate that Template is a valid RFC 6570 URI template.
+// The parsing currently doesn't validate that Template is a valid
+// RFC 6570 URI template.
 type SVCBDoHPath struct {
 	Template string
 }
 
-func (*SVCBDoHPath) Key() SVCBKey { return SVCB_DOHPATH }
-func (s *SVCBDoHPath) len() int   { return len(s.Template) }
-
-func (s *SVCBDoHPath) pack() ([]byte, error) {
-	if len(s.Template) > 65535 {
-		return nil, errors.New("dns: svcbdohpath: template too long")
-	}
-	return []byte(s.Template), nil
-}
+func (*SVCBDoHPath) Key() SVCBKey            { return SVCB_DOHPATH }
+func (s *SVCBDoHPath) String() string        { return s.Template }
+func (s *SVCBDoHPath) len() int              { return len(s.Template) }
+func (s *SVCBDoHPath) pack() ([]byte, error) { return []byte(s.Template), nil }
 
 func (s *SVCBDoHPath) unpack(b []byte) error {
 	s.Template = string(b)
 	return nil
 }
 
-func (s *SVCBDoHPath) String() string {
-	return s.Template
-}
-
 func (s *SVCBDoHPath) parse(b string) error {
-	if len(b) > 65535 {
-		return errors.New("dns: svcbdohpath: template too long")
-	}
 	s.Template = b
 	return nil
 }
--- a/svcb.go
+++ b/svcb.go
@@ -13,8 +13,7 @@ import (
 // SVCBKey is the type of the keys used in the SVCB RR.
 type SVCBKey uint16
 
-// Keys defined in draft-ietf-dnsop-svcb-https-08 Section 14.3.2
-// and draft-ietf-add-svcb-dns-02 Section 9.
+// Keys defined in draft-ietf-dnsop-svcb-https-08 Section 14.3.2.
 const (
 	SVCB_MANDATORY SVCBKey = iota
 	SVCB_ALPN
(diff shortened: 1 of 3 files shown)


svcb.go
```diff
@@ -669,6 +677,62 @@ func (s *SVCBIPv6Hint) copy() SVCBKeyValue {
...
+// It doesn't validate that Template is a valid RFC 6570 URI template.
+type SVCBDoHPath struct {
+	Template string
+}
+
+func (*SVCBDoHPath) Key() SVCBKey { return SVCB_DOHPATH }
+func (s *SVCBDoHPath) len() int   { return len(s.Template) }
+
+func (s *SVCBDoHPath) pack() ([]byte, error) {
+	if len(s.Template) > 255 {
```
ainar-g, line 707: Now that I think about it, I'm not actually sure if there is a limit to the length of a URI template.  Perhaps, it should have the usual 65535-octet limit of a SvcParamKey.

svcb.go
```diff
@@ -669,6 +677,62 @@ func (s *SVCBIPv6Hint) copy() SVCBKeyValue {
...
+//	s.Hdr = dns.RR_Header{Name: ".", Rrtype: dns.TypeSVCB, Class: dns.ClassINET}
+//	e := new(dns.SVCBAlpn)
+//	e.Alpn = []string{"h2", "h3"}
+//	p := new(dns.SVCBDoHPath)
+//	p.Template = "/dns-query{?dns}"
+//	s.Value = append(s.Value, e, p)
+//
+// The parsing currently only validates the length of Template.
+// It doesn't validate that Template is a valid RFC 6570 URI template.
+type SVCBDoHPath struct {
```
ainar-g, line 699: “DoH” is spelled with a little “o”, because that's how [RFC 8484] spells it.  Please let me know if you find some other spelling better.

[RFC 8484]: https://datatracker.ietf.org/doc/html/rfc8484

svcb.go
```diff
@@ -22,23 +23,28 @@ const (
 	SVCB_IPV4HINT
 	SVCB_ECHCONFIG
 	SVCB_IPV6HINT
+	SVCB_DOHPATH
```
miekg, line 26:
```suggestion
	SVCB_DOHPATH // draft-ietf-add-svcb-dns-02 Section 9
```

svcb.go
```diff
@@ -13,7 +13,8 @@ import (
 // SVCBKey is the type of the keys used in the SVCB RR.
 type SVCBKey uint16
 
-// Keys defined in draft-ietf-dnsop-svcb-https-08 Section 14.3.2.
+// Keys defined in draft-ietf-dnsop-svcb-https-08 Section 14.3.2
+// and draft-ietf-add-svcb-dns-02 Section 9.
```
miekg, line 17:
```suggestion
```

svcb.go
```diff
@@ -22,23 +23,28 @@ const (
...
 	SVCB_IPV4HINT:        "ipv4hint",
 	SVCB_ECHCONFIG:       "ech",
 	SVCB_IPV6HINT:        "ipv6hint",
+	SVCB_DOHPATH:         "dohpath",
 }
 
-var svcbStringToKeyMap = reverseSVCBKeyMap(svcbKeyToStringMap)
+// StringToSVCBKey is the reverse of SVCBKeyToString, needed for string
+// parsing.
+var StringToSVCBKey = reverseSVCBKey(SVCBKeyToString)
```
miekg, line 45: making this stuff public should probably be done in a new PR, and keep this to just add dohpath.

ainar-g:
@miekg, thanks for the review, ptal!

svcb.go
```diff
@@ -669,6 +673,62 @@ func (s *SVCBIPv6Hint) copy() SVCBKeyValue {
...
+// See RFC xxxx (https://datatracker.ietf.org/doc/html/draft-ietf-add-svcb-dns-02)
+// and RFC yyyy (https://datatracker.ietf.org/doc/html/draft-ietf-add-ddr-06).
+//
+// Basic use pattern for using the dohpath option together with an alpn
+// option:
+//
+//	s := new(dns.SVCB)
+//	s.Hdr = dns.RR_Header{Name: ".", Rrtype: dns.TypeSVCB, Class: dns.ClassINET}
+//	e := new(dns.SVCBAlpn)
+//	e.Alpn = []string{"h2", "h3"}
```
miekg, line 688: I need Alpn because? (haven't read the drafts)

svcb.go
```diff
@@ -669,6 +673,62 @@ func (s *SVCBIPv6Hint) copy() SVCBKeyValue {
...
+// It doesn't validate that Template is a valid RFC 6570 URI template.
+type SVCBDoHPath struct {
+	Template string
+}
+
+func (*SVCBDoHPath) Key() SVCBKey { return SVCB_DOHPATH }
+func (s *SVCBDoHPath) len() int   { return len(s.Template) }
+
+func (s *SVCBDoHPath) pack() ([]byte, error) {
+	if len(s.Template) > 65535 {
```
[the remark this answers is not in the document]
miekg: these functions then also become 1-liners so you can group them above with Key() and len()

svcb.go
ainar-g: The two RFC drafts that define or use `dohpath` ([`draft-ietf-add-svcb-dns`][1] and [draft-ietf-add-ddr-06][2]) both use these two together.  Neither of them says that the `dohpath` parameter cannot be used without the `alpn` one *explicitly,* but the first of the draft reads:

 >  If the "alpn" SvcParamKey indicates support for HTTP, "dohpath" MUST be present \[…\]

And the semantics of using `dohpath` without the `alpn` set to one of the HTTP ones seem to remain unclear.

Since the code illustrates a “basic use pattern”, which seems to basically always use `alpn` and `dohpath` together, I've decided to make the example use those together as well.

[1]: https://datatracker.ietf.org/doc/html/draft-ietf-add-svcb-dns-02
[2]: https://datatracker.ietf.org/doc/html/draft-ietf-add-ddr-06.html

miekg (approved):

tmthrgd:
I have to say I really dislike that we’re pulling in so many draft standards for SVCB. We’ve already seen them change in ways bordering on backwards-incompatible. I’d personally rather wait for these to be actually published as RFCs.

tmthrgd:
I should clarify, I don’t object to implementing drafts where they can be kept as internal implementation details, but I do object when they become part of our public API.

miekg:
[ Quoting ***@***.***> in "Re: [miekg/dns] Add SVCB dohpath ke..." ]
>I should clarify, I don’t object to implementing drafts where they can be kept
>as internal implementation details, but I do object when they become part of
>our public API.

I hear you. We should indeed be careful. OTOH this new RR has gained popularity quickly
and there seems to be some experimentation going on. Waiting for these I-Ds to be RFC can
take multiple years, so we should be a bit lenient here.

My position is: you know you're dealing with a pre-RFC RR here, so it's fine for us to
break things when the actual RFC turns out to be different.

(We could put such a warning in the code, but I don't think we have something prominent
enough where folks would actually read it)

ainar-g:
@miekg, the Go standard library has a few spots where it explicitly says that something is experimental and is likely to change in the future.  For example, package [`syscall/js`][1].  I could send a PR where I go through all exported SVCB identifiers and add a warning regarding the draft status of this functionality.

[1]: https://pkg.go.dev/syscall/js

miekg:
[ Quoting ***@***.***> in "Re: [miekg/dns] Add SVCB dohpath ke..." ]
***@***.***, the Go standard library has a few spots where it explicitly says that
>something is experimental and is likely to change in the future. For example,
>package syscall/js. I could send a PR where I go through all exported SVCB
>identifiers and add a warning regarding the draft status of this functionality.

As ugly as that is, it might "solve" some of Tom's conerns (and mine - frankly)

Issue #1361, opened 2022-04-12
String() in SVCBAlpn conversion doesn't escape properly


shane-kerr:
I _think_ the idea is that `String()` routines should provide something that is presentation format, suitable for DNS zone files.

The `SVCBAlpn` type contains an array of strings. If set via the `unpack()` function (or by direct initialization) it can contain non-printable characters, but the corresponding `String()` function does not attempt to escape them. Such DNS messages are bogus, of course, but such is DNS. The code in the `SVCBLocal` type properly escapes any values.

If this is a bug that makes sense to fix, I can submit a PR.

Note that the code in PR #1359 seems to have a similar issue.

shane-kerr:
Here's a simple program demonstrating the issue:
https://go.dev/play/p/QjQB7IRdRB0

miekg:
yes, Stringyfied value should be suitable for inclusion in zone files, so PR welcome.

thanks!

shane-kerr:
Fixes merged!

PR #1363, opened 2022-04-12
Properly parse alpn values in SVCB

shane-kerr:
PR for #1361 

This code adds more comprehensive parsing and string conversion to the alpn support in the HTTPS/SVCB types.

Before we were just splitting on `,`, but this version actually parses as per the draft, which allows alpn with a `,` in the value. This adds a ton of error cases and a fair bit of code, so I'm hoping the RFC simplifies this to just forbidding `,` in the alpn completely (am I too late to suggest that?). On the plus side, the examples in the tests now actually return correct values.

I used the existing `nextByte()` function in the `types.go` file, since I needed the same functionality. This means that we allow incorrect values like `\456`, but I figured fixing that was out of scope. Is it worth a separate PR? :thinking: Note that the parser for `SVCBLocal` uses `strconv.ParseUint()` and should actually bounds check, so we have different strictness/interpretations about escaped characters already.

The `String()` is also updated to handle outputting non-printable values, and also handles the double-escaping specified in the draft. I decided to output `\\044` instead of `\,` so that naive parsers (like the old one) would at least have a chance to split the output into something like what it should be (since there is no `,` because we use the numeric version). This shouldn't have any impact on the actual meaning, but I did have to update a couple of tests based on this format.

I added a few more tests to sanity check that everything is working as it is supposed to.

I noticed that the "dohpath" support was merged in while I was working on this. This introduces another place where `String()` will possibly produce incorrect output. :disappointed: I'll have a look tomorrow....

### DIFF
--- a/svcb.go
+++ b/svcb.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/binary"
 	"errors"
+	"fmt"
 	"net"
 	"sort"
 	"strconv"
@@ -334,13 +335,56 @@ func (s *SVCBMandatory) copy() SVCBKeyValue {
 //	h.Hdr = dns.RR_Header{Name: ".", Rrtype: dns.TypeHTTPS, Class: dns.ClassINET}
 //	e := new(dns.SVCBAlpn)
 //	e.Alpn = []string{"h2", "http/1.1"}
-//	h.Value = append(o.Value, e)
+//	h.Value = append(h.Value, e)
 type SVCBAlpn struct {
 	Alpn []string
 }
 
-func (*SVCBAlpn) Key() SVCBKey     { return SVCB_ALPN }
-func (s *SVCBAlpn) String() string { return strings.Join(s.Alpn, ",") }
+func (*SVCBAlpn) Key() SVCBKey { return SVCB_ALPN }
+
+func (s *SVCBAlpn) String() string {
+	// An ALPN value is a comma-separated list of values, each of which can be
+	// an arbitrary binary value. In order to allow parsing, the comma and
+	// backslash characters are themselves excaped.
+	//
+	// However, this escaping is done in addition to the normal escaping which
+	// happens in zone files, meaning that these values must be
+	// double-escaped. This looks terrible, so if you see a never-ending
+	// sequence of backslash in a zone file this may be why.
+	//
+	// https://datatracker.ietf.org/doc/html/draft-ietf-dnsop-svcb-https-08#appendix-A.1
+	esc := make([]string, len(s.Alpn))
+	for i, alpn := range s.Alpn {
+		var str strings.Builder
+		str.Grow(4 * len(alpn))
+		for j := 0; j < len(alpn); j++ {
+			e := alpn[j]
+			if ' ' <= e && e <= '~' {
+				switch e {
+				// We escape a few characters which may confuse humans or
+				// parsers.
+				case '"', ';', ' ':
+					str.WriteByte('\\')
+					str.WriteByte(e)
+				// The comma and backslash characters themselves must be
+				// doubly-escaped. We use `\\` for the first backslash and
+				// the escaped numeric value for the other value. We especially
+				// don't want a comma in the output.
+				case ',':
+					str.WriteString(`\\\044`)
+				case '\\':
+					str.WriteString(`\\\092`)
+				default:
+					str.WriteByte(e)
+				}
+			} else {
+				str.WriteString(escapeByte(e))
+			}
+		}
+		esc[i] = str.String()
+	}
+	return strings.Join(esc, ",")
+}
 
 func (s *SVCBAlpn) pack() ([]byte, error) {
 	// Liberally estimate the size of an alpn as 10 octets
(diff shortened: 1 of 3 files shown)


miekg:
I thought it as already an RFC, but still an I-D https://datatracker.ietf.org/doc/html/draft-ietf-dnsop-svcb-https-08 then you should be able to get some changes through, esp. with I tried to code this and it was horrible.

svcb.go
```diff
@@ -334,13 +334,56 @@ func (s *SVCBMandatory) copy() SVCBKeyValue {
...
+				// the escaped numeric value for the other value. We especially
+				// don't want a comma in the output.
+				case ',':
+					str.WriteString(`\\\044`)
+				case '\\':
+					str.WriteString(`\\\092`)
+				default:
+					str.WriteByte(e)
+				}
+			} else {
```
miekg, line 379: reverse this `if`, so the else part is just done +  a `continue` and the current if-body can be pulled to the left

svcb.go
```diff
@@ -375,7 +418,47 @@ func (s *SVCBAlpn) unpack(b []byte) error {
...
+		if q == 0 {
+			return errors.New("dns: svcbalpn: unterminated escape")
+		}
+		p += q
+		// If we find a comma, we have finished reading an alpn.
+		if c == ',' {
+			if len(a) == 0 {
+				return errors.New("dns: svcbalpn: empty protocol identifier")
+			}
+			alpn = append(alpn, string(a))
```
miekg, line 439: why doesn't directly assign to `s.Apln`? To leave it in a semi-sane state if parsing fails? Unsure if we are that careful anywhere else...

svcb_test.go
```diff
@@ -95,6 +96,42 @@ func TestDecodeBadSVCB(t *testing.T) {
...
+		`. 4 IN SVCB 40 four.test. alpn="part1,part2,part3\\,part4\\\\"`:           {"part1", "part2", `part3,part4\`},
+		`. 5 IN SVCB 50 five.test. alpn=part1\,\p\a\r\t2\044part3\092,part4\092\\`: {"part1", "part2", `part3,part4\`},
+	}
+	for s, v := range tests {
+		rr, err := NewRR(s)
+		if err != nil {
+			t.Error("failed to parse RR: ", err)
+			continue
+		}
+		if !reflect.DeepEqual(v, rr.(*SVCB).Value[0].(*SVCBAlpn).Alpn) {
```
miekg, line 129: just range over the slices, no real need to pull in reflect for this

shane-kerr: Okay, I've refactored this. I figured that since this was test code `reflect` would be fine.

svcb.go
shane-kerr: Makes sense, done.

svcb.go
```diff
@@ -334,13 +334,56 @@ func (s *SVCBMandatory) copy() SVCBKeyValue {
...
+				default:
+					str.WriteByte(e)
+				}
+			} else {
+				str.WriteString(escapeByte(e))
+			}
+		}
+		esc[i] = str.String()
+	}
+	return strings.Join(esc, ",")
```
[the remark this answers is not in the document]
shane-kerr: Yes, this is straightforward. I should have done that to begin with, but I started with the code for `SVCBLocal` and didn't really think it through. :joy: Done!

shane-kerr:
As I mentioned on #1366, I think this PR is as reasonable and should be merged. I checked on the IETF dnsop list and apparently this extra escaping is just how it is, because parsing pain is apparently no reason to change a standard. :shrug:

svcb.go
```diff
@@ -334,13 +334,57 @@ func (s *SVCBMandatory) copy() SVCBKeyValue {
...
+		}
+		for j := 0; j < len(alpn); j++ {
+			e := alpn[j]
+			if ' ' > e || e > '~' {
+				str.WriteString(escapeByte(e))
+				continue
+			}
+			switch e {
+			// We escape a few characters which may confuse humans or
+			// parsers.
```
miekg, line 369:
```suggestion
			// We escape a few characters which may confuse humans or parsers.
```

svcb.go
```diff
@@ -334,13 +334,57 @@ func (s *SVCBMandatory) copy() SVCBKeyValue {
...
+	//
+	// However, this escaping is done in addition to the normal escaping which
+	// happens in zone files, meaning that these values must be
+	// double-escaped. This looks terrible, so if you see a never-ending
+	// sequence of backslash in a zone file this may be why.
+	//
+	// https://datatracker.ietf.org/doc/html/draft-ietf-dnsop-svcb-https-08#appendix-A.1
+	var str strings.Builder
+	for i, alpn := range s.Alpn {
+		str.Grow(4*len(alpn) + 1)
```
miekg, line 357: might need a little comment here, why 4?

miekg (commented):
2 nits, lgtm

svcb.go
shane-kerr: Makes sense, I've added a comment here.

miekg (approved):

PR #1364, opened 2022-04-13
Add notes about SVCB draft changes to SVCB-related API

ainar-g:
See discussion in #1359.

### DIFF
--- a/svcb.go
+++ b/svcb.go
@@ -11,9 +11,15 @@ import (
 )
 
 // SVCBKey is the type of the keys used in the SVCB RR.
+//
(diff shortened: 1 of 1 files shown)


ainar-g:
@miekg, done.

miekg (approved):

miekg:
https://datatracker.ietf.org/doc/draft-ietf-dnsop-svcb-https/ is heading for RFC - so the changes going forward will be minimal, would be good to see what needs to be changed and remove the warnings

ainar-g:
@miekg, I'll try to keep up with the drafting process and send a PR when the RFC is out.

Issue #1365, opened 2022-04-14
IPv6 syntax error in example_test: dial udp: address <IPv6 address>:53: too many colons in address


gsejourne:
Hey,

Run into an error trying to use the sample ExampleMX() function in example_test.go: 
https://github.com/miekg/dns/blob/08c2616301e68a33ab37e37a091130695a3630ac/example_test.go#L19

It fails with the following error:
`dial udp: address <my_v6_address>:53: too many colons in address`

This is most likely due to 	`r, _, err := c.Exchange(m, config.Servers[0]+":"+config.Port)`
as my main resolver address is a V6 --
here's the excerpt from my resolv.conf (note: untouched, and OS X populated, so potentially highly reproducible):
`
search home

nameserver v6 address here

nameserver 192.168.1.1
`

Since the c.Exchange() function is trying config.Servers[0], it fails, as it expects the v6 address to be in [].

PR #1367, opened 2022-04-15
#1365: Add test to sanitize IPv6 address in case it's not in brackets

gsejourne:
This is an attempt at fixing issue https://github.com/miekg/dns/issues/1365

Note: not using To4() methods to avoid string/ip conversion, but feel free to fail the request if you'd rather have this.


tmthrgd:
This has been separately fixed by #1368.

PR #1376, opened 2022-05-24
Disallow names that start with '.' in IsDomainName()

shane-kerr:
We noticed that `IsDomainName()` was allowing names that start with at `.`. This is an error, which this PR fixes.

### DIFF
--- a/defaults.go
+++ b/defaults.go
@@ -218,6 +218,11 @@ func IsDomainName(s string) (labels int, ok bool) {
 
 			wasDot = false
 		case '.':
+			if i == 0 && len(s) > 1 {
+				// leading dots are not legal except for the root zone
+				return labels, false
+			}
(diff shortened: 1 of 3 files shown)


tmthrgd:
// XXX: The logic in this function was copied from packDomainName and
	// should be kept in sync with that function.
	
^See the comment at the start of the function.

shane-kerr:
> ```
> // XXX: The logic in this function was copied from packDomainName and
> // should be kept in sync with that function.
> ```
> 
> ^See the comment at the start of the function.

Ah, okay. I've updated `packDomainName()`. I didn't see any relevant tests in `msg_test.go` and I also didn't add any. Hopefully that is okay.

miekg:
thanks both, comment is acted upon. merging

Issue #1370, opened 2022-04-19
Do you have any plans to support DoQ (DNS-over-Quic) in the near future?


viagram:
I hope this is possible!

miekg:
What do you miss/need? According to https://www.ietf.org/archive/id/draft-ietf-dprive-dnsoquic-12.html you can just create a net.Conn with some quic library and send queries.

Server side support will only happen once quic is in the std lib I guess, or we would need a ListenAndServe that takes a `listener`

miekg:
quic-go uses Streams which are not net.Conn`s, and we don't want to pull all of that in this library. Quic/http3 is slowly happening in Go's std lib, until that is done I don't see what can be done in this library.

If you want to play with it yourself, various things can be wrapped in dns.Conn and the server could check if the conn support quic or some such

miekg:
I haven't look too closely, but i would assume it would work like this:

You use a lib that does quic in Go (haven't looked lately, are there any?). That
should give you a net.Conn which can then be used.

There might be some need for helper functions, but we also don't have these for DoH, so I
don't think we should do that for DoQ either.

/Miek

--
Miek Gieben

PR #1377, opened 2022-05-25
QUIC/DoQ support using quic-go

jelu:
# Work In Progress !

Draft implementation of DoQ support using quic-go, this PR serves as discussion forum for if this is the right way forward as the dependencies are quite large.

TODO:
- [ ] Should there be a `ListenAndServeDoQ()`?
- [ ] Add `_test` code
- [ ] Discuss usage of `response.tcp` when using QUIC

\+ all TODO's in the code

Example:
```
cert, _ := tls.LoadX509KeyPair("<cert>", "<key>")
config := tls.Config{
    Certificates: []tls.Certificate{cert},
}
server := &dns.Server{Addr: "<addr>", Net: "doq", TLSConfig: &config, Handler: <handle>})
```

ref #1370

### DIFF
--- a/server.go
+++ b/server.go
@@ -12,6 +12,13 @@ import (
 	"strings"
 	"sync"
 	"time"
+
+	// Dependencies for QUIC:
+	"github.com/lucas-clemente/quic-go"
+	// For enabling QUIC connection tracing:
+	// "github.com/lucas-clemente/quic-go/logging"
+	// "github.com/lucas-clemente/quic-go/qlog"
+	// "os"
 )
 
 // Default maximum number of TCP queries before we close the socket.
@@ -74,6 +81,7 @@ type response struct {
 	tsigProvider   TsigProvider
 	udp            net.PacketConn // i/o connection if UDP was used
 	tcp            net.Conn       // i/o connection if TCP was used
+	doq            quic.Stream    // i/o connection if QUIC was used
 	udpSession     *SessionUDP    // oob data to get egress interface right
 	pcSession      net.Addr       // address to use when writing to a generic net.PacketConn
 	writer         Writer         // writer to output the raw DNS bits
@@ -143,6 +151,9 @@ type Reader interface {
 	// ReadTCP reads a raw message from a TCP connection. Implementations may alter
 	// connection properties, for example the read-deadline.
 	ReadTCP(conn net.Conn, timeout time.Duration) ([]byte, error)
+	// ReadQUIC reads a raw message from a QUIC stream. Implementations may alter
+	// connection properties, for example the read-deadline.
+	ReadQUIC(stream quic.Stream, timeout time.Duration) ([]byte, error)
 	// ReadUDP reads a raw message from a UDP connection. Implementations may alter
 	// connection properties, for example the read-deadline.
 	ReadUDP(conn *net.UDPConn, timeout time.Duration) ([]byte, *SessionUDP, error)
@@ -168,6 +179,10 @@ var _ PacketConnReader = defaultReader{}
 
 func (dr defaultReader) ReadTCP(conn net.Conn, timeout time.Duration) ([]byte, error) {
 	return dr.readTCP(conn, timeout)
+}
+
+func (dr defaultReader) ReadQUIC(stream quic.Stream, timeout time.Duration) ([]byte, error) {
+	return dr.readQUIC(stream, timeout)
 }
 
 func (dr defaultReader) ReadUDP(conn *net.UDPConn, timeout time.Duration) ([]byte, *SessionUDP, error) {
@@ -238,6 +253,8 @@ type Server struct {
 
 	// A pool for UDP message buffers.
 	udpPool sync.Pool
+
+	quicListener quic.Listener
 }
 
 func (srv *Server) tsigProvider() TsigProvider {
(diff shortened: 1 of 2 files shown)


jelu:
@miekg @tmthrgd No hurry to review, think I'll spend some time just looking at the `server.go` code. Thinking maybe a interface for all protocols is better, so any protocol can just be plugged in.

miekg:
thanks for doing this. I'll go over the code and try to ask stupid questions.

Note: I have no intention to merge this, because I'm waiting what Go's std lib will do here. Also `net.Conn` is the defacto std interface we're looking for, so quic should (somehow) adapt to that, instead of the other way around

server.go
```diff
@@ -74,6 +81,7 @@ type response struct {
 	tsigProvider   TsigProvider
 	udp            net.PacketConn // i/o connection if UDP was used
 	tcp            net.Conn       // i/o connection if TCP was used
+	doq            quic.Stream    // i/o connection if QUIC was used
```
miekg, line 84: can't we check if it's a quick stream when we accept a udp conn? And then do a type assertion on the `net.PacketConn` we have if it implements `Stream()` - or whatever the interface is?

server.go
```diff
@@ -238,6 +253,8 @@ type Server struct {
 
 	// A pool for UDP message buffers.
 	udpPool sync.Pool
+
+	quicListener quic.Listener
```
miekg, line 257: same here, i.e. use a net.Listener and check if the type implements quic and then do quic things?

server.go
```diff
@@ -340,6 +357,46 @@ func (srv *Server) ListenAndServe() error {
 		srv.started = true
 		unlock()
 		return srv.serveUDP(u)
+	case "doq", "doq4", "doq6":
+		if srv.TLSConfig == nil || (len(srv.TLSConfig.Certificates) == 0 && srv.TLSConfig.GetCertificate == nil) {
```
miekg, line 361: Does does imply `WithInsecure()` doesn't work?

server.go
```diff
@@ -834,3 +916,169 @@ func (w *response) ConnectionState() *tls.ConnectionState {
 	}
 	return nil
 }
+
+type quicConn struct {
```
miekg, line 920: why is internal house keeping needed? quic-go doesn't do this for us? Or there is no other way to read/write for the correct stream?

server.go
jelu: A `quic.Stream` is not a `net.PacketConn`, and a `quic.Connection` can have many `quic.Stream`s (each stream will carry one query/response)

server.go
jelu: `quic.Listener` doesn't implement `net.Listener`.

server.go
jelu: This was the only way I could wrap a `quic.Connection` and it's streams so I could place it in `srv.conns[]`.

server.go
jelu: I donno :D (disclaimer: I just copied this `if` from the TLS section)

jelu:
> Note: I have no intention to merge this, because I'm waiting what Go's std lib will do here. Also `net.Conn` is the defacto std interface we're looking for, so quic should (somehow) adapt to that, instead of the other way around

Ah OK.

So are you willing to change the internal of `dns.Server` to allow for a more pluggable protocol design?

If so, then I can spend next week hammering out an suggestion.

If not, then I'll solved all of this in another way.

Please let me know ASAP, thanks!

server.go
miekg: yes, but https://github.com/lucas-clemente/quic-go/blob/master/example/echo/echo.go#L39 shows it _is_ a conn first and after that some quic stuff happens, I wonder if we could do the same to avoid much code duplication (this may be impossible)

server.go
miekg: sad - that will imply that a std lib impl will have a hard time as well. Wonder if anyone actually works on that

server.go
miekg: ok, would be nice to have that at some point

server.go
miekg: check

miekg:
honestly don't know - when we added grpc in coredns nothing needed changing this lib, so I don't see yet why quic is somehow special here (yes I understand the streams are fundamentally different than anything before)

Also once we merged we can't easily break folks because of backwards compat and we might also design it towards quic-go while noone knows how the final impl. looks that will hopefully end up in the std lib.

miekg:
ref https://github.com/golang/go/issues/32204

jelu:
The suggestion, if any, would be made in a way that has zero affect on the current API and interfaces. I'm thinking just internal stuff to add more protocol via `dns.Server.Net` in a pluggable way so that I can make (and manage) a "dns-quic-go" module that people can use seamlessly.

Basically use-case would be:
```
import (
  "github.com/miekg/dns"
  "github.com/DNS-OARC/dns-quic-go"
)

func () {
  srv := &dns.Server{Addr: "...", Net: "quic-go", Handler: ...}
  srv.ListenAndServe()
}
```

Of course if there is no interest at all in having such pluggable design then I'll solve it in another way and we can close this PR.

miekg:
[ Quoting ***@***.***> in "Re: [miekg/dns] QUIC/DoQ support us..." ]
>The suggestion, if any, would be made in a way that has zero affect on the
>current API and interfaces. I'm thinking just internal stuff to add more
>protocol via dns.Server.Net in a pluggable way so that I can make (and manage)
>a "dns-quic-go" module that people can use seamlessly.
>
>Basically use-case would be:
>
>import (
>  "github.com/miekg/dns"
>  "github.com/DNS-OARC/dns-quic-go"
>)
>
>func () {
>  srv := &dns.Server{Addr: "...", Net: "quic-go", Handler: ...}
>  srv.ListenAndServe()
>}
>
>Of course if there is no interest at all in having such pluggable design then
>I'll solve it in another way and we can close this PR.

That would be a nice way forward.

Cevaet: a new protocol happens every 5/10 years-ish?? So it may turn out having this
pluggeable doesn't give us that much and we should rip it out again.

jelu:
After looking over the code I can't really see a simple way to make protocols pluggable, they are too tied into a lot of things such as the `Reader` interface which has `ReadTCP()`, `ReadUDP()` etc and none of that is applicable to DoQ.

The design is also around having one listener listening for one connection that delivers DNS messages where QUIC is conn listener -> stream listener -> streams. Sure, that could be solved with channels etc but then it's workarounds which makes the code clunky and slow.

I think a better approach is to make my own module that mimics `dns.Server` as much as possible.

Thanks for the reviews and discussion!

PR #1379, opened 2022-06-09
TSIG Verify/Generate using TsigProvider

jelu:
By exposing `TsigVerifyProvider` and `TsigGenerateProvider` functions I am able to fully implement TSIG support in [DNS-OARC/golang-dns-server-doq](https://github.com/DNS-OARC/golang-dns-server-doq) similar to how it's done in `dns.Server` and without needed to copy almost all of `tsig.go`.

I see no harm in exposing these functions and maybe others will find them useful also, hope it can be merged.

### DIFF
--- a/tsig.go
+++ b/tsig.go
@@ -166,10 +166,12 @@ type timerWireFmt struct {
 // timersOnly is false.
 // If something goes wrong an error is returned, otherwise it is nil.
 func TsigGenerate(m *Msg, secret, requestMAC string, timersOnly bool) ([]byte, string, error) {
-	return tsigGenerateProvider(m, tsigHMACProvider(secret), requestMAC, timersOnly)
+	return TsigGenerateProvider(m, tsigHMACProvider(secret), requestMAC, timersOnly)
 }
 
-func tsigGenerateProvider(m *Msg, provider TsigProvider, requestMAC string, timersOnly bool) ([]byte, string, error) {
+// TsigGenerate fills out the TSIG record attached to the message using
+// a TsigProvider, for more details and return see TsigGenerate.
+func TsigGenerateProvider(m *Msg, provider TsigProvider, requestMAC string, timersOnly bool) ([]byte, string, error) {
 	if m.IsTsig() == nil {
 		panic("dns: TSIG not last RR in additional")
 	}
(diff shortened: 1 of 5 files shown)


miekg:
I can't find a good reason to say 'no', so likely to be merged.

as the function docs are now public, some comments there.

miekg:
I can't find a good reason to say 'no', so likely to be merged.

as the function docs are now public, some comments there.

tsig.go
```diff
@@ -166,10 +166,12 @@ type timerWireFmt struct {
 // timersOnly is false.
 // If something goes wrong an error is returned, otherwise it is nil.
 func TsigGenerate(m *Msg, secret, requestMAC string, timersOnly bool) ([]byte, string, error) {
-	return tsigGenerateProvider(m, tsigHMACProvider(secret), requestMAC, timersOnly)
+	return TsigGenerateProvider(m, tsigHMACProvider(secret), requestMAC, timersOnly)
 }
 
-func tsigGenerateProvider(m *Msg, provider TsigProvider, requestMAC string, timersOnly bool) ([]byte, string, error) {
+// TsigGenerate fills out the TSIG record attached to the message using
+// a TsigProvider, for more details and return see TsigGenerate.
```
miekg, line 173: also unfinished docs

tsig.go
```diff
@@ -166,10 +166,12 @@ type timerWireFmt struct {
 // timersOnly is false.
 // If something goes wrong an error is returned, otherwise it is nil.
 func TsigGenerate(m *Msg, secret, requestMAC string, timersOnly bool) ([]byte, string, error) {
```
miekg, line 168: on the docs - it's weirdly formatted, also Tsig RR -> TSIG RR..

"... called for the first time requestMAC should be set to the empty string and timersOnly should be false.":

would also be nice to tell what subsequent calls should look like,.

The "if something goes wrong an error is returned..." can be removed IMO

tsig.go
```diff
@@ -223,7 +225,9 @@ func TsigVerify(msg []byte, secret, requestMAC string, timersOnly bool) error {
 	return tsigVerify(msg, tsigHMACProvider(secret), requestMAC, timersOnly, uint64(time.Now().Unix()))
 }
 
-func tsigVerifyProvider(msg []byte, provider TsigProvider, requestMAC string, timersOnly bool) error {
+// TsigVerify verifies the TSIG on a message using a TsigProvider, for
+// more details and return see TsigVerify.
```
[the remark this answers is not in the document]
jelu: It says `see TsigVerify`(?)

tsig.go
jelu: same here

tsig.go
jelu: I have made no change in the documentation for that function.

jelu:
The return of the functions are the same, do you really want duplicated documentation?

jelu:
Could anyone please let me know what kind of timeline I should expect for getting this merged and into a release?

Just want to know so I can plan the release of my library that depends on this PR.

jelu:
> The return of the functions are the same, do you really want duplicated documentation?

@miekg Do you really want duplicated documentation?

miekg:
this lgtm although `WithProvider` would be more Go-like and we already have some other 'With...' functions. 

I'll close this and send in a new PR with that name in it *and* also adjust any docs

jelu:
OK.

It's totally okay for me if you push onto this branch also, it should be
open for you.

miekg:
https://github.com/miekg/dns/pull/1382

please double check for anything stupid

miekg:
Ah, must have misread that.

Ok..I'll push through a separate PR to clean that wording a bit.

PR #1382, opened 2022-06-20
Make tsigGenerate/VerifyProvider public

miekg:
Make it public as TsigGenerateWithProvider and update the docs a little.


jelu:
This is only for generation, my PR was for verification also.

@miekg can you make `tsigVerifyProvider` public also?

PR #1383, opened 2022-06-20
NewRR does not remove escape char

TomOnTime:
This PR just demonstrates the bug. It is not intended for merge.

See https://github.com/miekg/dns/issues/1384


miekg:
I'll try to (finally) fix this in https://codeberg.org/miekg/dns

PR #1386, opened 2022-06-28
ReadMsgHeader use bytes pool

IIvyPy:
I find the  makeslice of ReadMsgHeader will cost much GC time. So I suggest to use bytes pool to avoid the GC cost.

### DIFF
--- a/client.go
+++ b/client.go
@@ -319,7 +319,6 @@ func (co *Conn) ReadMsgHeader(hdr *Header) ([]byte, error) {
 		return nil, ErrShortRead
 	}
 
-	p = p[:n]
 	if hdr != nil {
 		dh, _, err := unpackMsgHdr(p, 0)
 		if err != nil {
--- a/client.go
+++ b/client.go
@@ -10,6 +10,7 @@ import (
 	"io"
 	"net"
 	"strings"
+	"sync"
 	"time"
 )
 
@@ -260,6 +261,13 @@ func (c *Client) exchangeContext(ctx context.Context, m *Msg, co *Conn) (r *Msg,
 	return r, rtt, err
 }
 
+// use sync.Pool to avoid gc overhead.
+var udpPool sync.Pool
+
+func init() {
+	udpPool.New = makeUDPBuffer(DefaultMsgSize)
+}
+
 // ReadMsg reads a message from the connection co.
 // If the received message contains a TSIG record the transaction signature
 // is verified. This method always tries to return the message, however if an
@@ -270,6 +278,7 @@ func (co *Conn) ReadMsg() (*Msg, error) {
 	if err != nil {
 		return nil, err
 	}
+	defer udpPool.Put(p)
 
 	m := new(Msg)
 	if err := m.Unpack(p); err != nil {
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -260,6 +261,13 @@ func (c *Client) exchangeContext(ctx context.Context, m *Msg, co *Conn) (r *Msg,
 	return r, rtt, err
 }
 
+// use sync.Pool to avoid gc overhead.
+var udpPool sync.Pool
+
+func init() {
```
miekg, line 267: please, no init functions

client.go
```diff
@@ -260,6 +261,13 @@ func (c *Client) exchangeContext(ctx context.Context, m *Msg, co *Conn) (r *Msg,
 	return r, rtt, err
 }
 
+// use sync.Pool to avoid gc overhead.
```
miekg, line 264: comment doesn't add a lot

client.go
```diff
@@ -260,6 +261,13 @@ func (c *Client) exchangeContext(ctx context.Context, m *Msg, co *Conn) (r *Msg,
...
+var udpPool sync.Pool
+
+func init() {
+	udpPool.New = makeUDPBuffer(DefaultMsgSize)
```
miekg, line 268: DefaultMsgSize is 512, a header is max 12 bytes, is this intentional?

miekg (changes requested):

miekg:
your commit message also don't explain anything about this change, it's just a one-liner

So it's on the client you want to use a buffer just like we do on the server side? Any data available (that should be in the ocmmit msg) on why and how this came about??

client.go
```diff
@@ -295,33 +304,25 @@ func (co *Conn) ReadMsgHeader(hdr *Header) ([]byte, error) {
...
-		}
 		n, err = co.Read(p)
 	} else {
-		var length uint16
-		if err := binary.Read(co.Conn, binary.BigEndian, &length); err != nil {
-			return nil, err
-		}
-
-		p = make([]byte, length)
 		n, err = io.ReadFull(co.Conn, p)
```
tmthrgd, line 311: This is wrong. You're ignoring the length which can be up to 64KiB. This would limit the max message size under TCP to whatever the pool returns.

client.go
```diff
@@ -295,33 +304,25 @@ func (co *Conn) ReadMsgHeader(hdr *Header) ([]byte, error) {
 		err error
 	)
 
+	p = udpPool.Get().([]byte)
 	if isPacketConn(co.Conn) {
-		if co.UDPSize > MinMsgSize {
-			p = make([]byte, co.UDPSize)
-		} else {
-			p = make([]byte, MinMsgSize)
-		}
```
tmthrgd, line 303: This code was important. `UDPSize` can be set larger than `MinMsgSize`. You're not handling that at all.

client.go
```diff
@@ -295,33 +304,25 @@ func (co *Conn) ReadMsgHeader(hdr *Header) ([]byte, error) {
...
 
 	if err != nil {
+		udpPool.Put(p)
 		return nil, err
 	} else if n < headerSize {
+		udpPool.Put(p)
 		return nil, ErrShortRead
 	}
 
-	p = p[:n]
```
tmthrgd, line 321: This slice call is important, we only read n-bytes not the full length of p.

tmthrgd (changes requested):

tmthrgd:
I'm sympathetic to the performance concerns here, but this isn't quite the way to go about it. We'd need a pool on the Client struct itself. But considering there's been no movement here in almost a year, I'm going to go ahead and close this PR. Feel free to re-open and address the issues above.

PR #1395, opened 2022-09-24
Support custom dialers in dns.Client

folbricht:
This introduces a `dns.Dialer` interface in the Client, rather than expecting a `*net.Dialer`. It's to allow passing in custom dialers that establish connections other than direct TCP or UDP. It would let me open connections through various types of proxies.

It's still a draft as I didn't add a test yet, and the comment around how `Timeout` is applied would need to be updated. Just wanted to first see if this is something you'd accept first.

### DIFF
--- a/client.go
+++ b/client.go
@@ -47,12 +47,16 @@ func (co *Conn) tsigProvider() TsigProvider {
 	return tsigSecretProvider(co.TsigSecret)
 }
 
+type Dialer interface {
+	DialContext(ctx context.Context, network, address string) (net.Conn, error)
+}
+
 // A Client defines parameters for a DNS client.
 type Client struct {
 	Net       string      // if "tcp" or "tcp-tls" (DNS over TLS) a TCP query will be initiated, otherwise an UDP one (default is "" for UDP)
 	UDPSize   uint16      // minimum receive buffer for UDP messages
 	TLSConfig *tls.Config // TLS connection configuration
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -47,12 +47,16 @@ func (co *Conn) tsigProvider() TsigProvider {
...
+	DialContext(ctx context.Context, network, address string) (net.Conn, error)
+}
+
 // A Client defines parameters for a DNS client.
 type Client struct {
 	Net       string      // if "tcp" or "tcp-tls" (DNS over TLS) a TCP query will be initiated, otherwise an UDP one (default is "" for UDP)
 	UDPSize   uint16      // minimum receive buffer for UDP messages
 	TLSConfig *tls.Config // TLS connection configuration
-	Dialer    *net.Dialer // a net.Dialer used to set local address, timeouts and more
+	Dialer    Dialer      // Dialer used to set local address, timeouts and more
```
tmthrgd, line 59: This is a breaking change, so we can't make it.

tmthrgd (changes requested):

tmthrgd:
I'm not sure this change really makes sense. If you need to use a custom connection type, like to proxy over other transports, you can already do that by creating a `Conn` yourself:

```go
conn := new(Conn)
conn.Conn = ...
conn.UDPSize = ...
resp, _, err := c.ExchangeWithConn(msg, conn)
...
```

PR #1402, opened 2022-11-17
Add AMTRELAY and IPSECKEY RR types

monoidic:
This PR adds the AMTRELAY and IPSECKEY RR types to this library, and bumps dependencies because `go generate` appears to be broken otherwise.


miekg:
looks good in general, but I've got a bunch of nits and questions.

You can also update the README and add RFC 8777 and 4025

go.mod
```diff
@@ -3,8 +3,8 @@ module github.com/miekg/dns
 go 1.14
 
 require (
-	golang.org/x/net v0.0.0-20210726213435-c6fcb2dbf985
-	golang.org/x/sync v0.0.0-20210220032951-036812b2e83c
-	golang.org/x/sys v0.0.0-20210630005230-0f9fa26af87c
-	golang.org/x/tools v0.1.6-0.20210726203631-07bc1bf47fb2
+	golang.org/x/net v0.2.0
```
miekg, line 6: probably not needed, but adding new RRs something results in a cycle, although pusing these to an actual release might not be bad

types.go
```diff
@@ -159,6 +161,22 @@ const (
 	ZoneMDHashAlgSHA512 = 2
 )
 
+// Used in IPSEC https://datatracker.ietf.org/doc/html/rfc4025#section-2.3
+const (
+	IPSECGatewayNone uint8 = 0
```
miekg, line 166:
```suggestion
	IPSECGatewayNone uint8 = iota
```

types.go
```diff
@@ -159,6 +161,22 @@ const (
...
+	IPSECGatewayIPv4 uint8 = 1
```
miekg, line 167:
```suggestion
	IPSECGatewayIPv4
```

types.go
```diff
@@ -159,6 +161,22 @@ const (
...
+	IPSECGatewayIPv6 uint8 = 2
+	IPSECGatewayHost uint8 = 3
```
miekg, line 169:
```suggestion
	IPSECGatewayIPv6
	IPSECGatewayHost
```

types.go
```diff
@@ -994,6 +1012,72 @@ func (rr *DNSKEY) String() string {
...
+		" " + strconv.Itoa(int(rr.Algorithm)) +
+		" " + gateway +
+		" " + rr.PublicKey
+}
+
+// AMTRELAY RR. See RFC 8777.
+type AMTRELAY struct {
+	Hdr               RR_Header
+	Precedence        uint8
+	DiscoveryOptional bool   `dns:"-"` // in the GatewayType byte
```
miekg, line 1050: this is here because the text presentation has it? Think it can be removed here and deal with this bit when parsing text.

types.go
```diff
@@ -994,6 +1012,72 @@ func (rr *DNSKEY) String() string {
...
+	GatewayType       uint8  `dns:"amtrelaytype"`
+	GatewayAddr       net.IP `dns:"-"`
```
miekg, line 1052: why is this here and set to `dns:"-"` ?

types.go
```diff
@@ -994,6 +1012,72 @@ func (rr *DNSKEY) String() string {
...
+	GatewayHost       string `dns:"amtrelayhost"`
+}
+
+func (rr *AMTRELAY) String() string {
+	var gateway string
+	switch rr.GatewayType {
+	case AMTRELAYIPv4, AMTRELAYIPv6:
+		gateway = rr.GatewayAddr.String()
+	case AMTRELAYHost:
+		gateway = rr.GatewayHost
```
miekg, line 1062: can you just make all of these have a String() so you can just call that when you construct the final string that is returned.

scan_rr.go
```diff
@@ -1216,6 +1216,115 @@ func (rr *DS) parse(c *zlexer, o string) *ParseError      { return rr.parseDS(c,
...
+	rr.GatewayAddr, rr.GatewayHost, err = parseAddrHostUnion(l, o, rr.GatewayType, "AMTRELAY")
+	if err != nil {
+		return err
+	}
+
+	return slurpRemainder(c)
+}
+
+// same constants and parsing between IPSECKEY and AMTRELAY
+func parseAddrHostUnion(l lex, o string, gatewayType uint8, errType string) (addr net.IP, host string, err *ParseError) {
```
miekg, line 1303: just let this return a normal error, can let the caller contruct the ParseError with the correct type information

go.mod
monoidic: Running `go generate` in a freshly clone of the repo yields this error:
```
2022/11/20 12:00:47 internal error: package "context" without types was imported from "github.com/miekg/dns"
exit status 1
```
Bumping the dependencies does fix it on my system, for whatever reason. I have not figured out why that is.

types.go
monoidic: Merging it into `Precedence` would make it a bit more of a hassle for users of the library to modify either `Precedence` or `DiscoveryOptional`, at the benefit of saving one byte per RR. Is this trade-off worth it?

types.go
monoidic: `AMTRELAY` and `IPSECKEY` behave very similarly with regards to the gateway, the value of `GatewayType` defines whether an IP address or hostname will be given as the gateway. Though there should be a comment here remarking on this, the same way there is one in `IPSECKEY`.

types.go
monoidic: These are RFC-defined constants, not locally defined enum values. Is this not the same reason that the RR codes in `types.go` are written out explictly rather than via iota values, despite a sizeable subset of them being continuous values?

go.mod
miekg: saw that also when adding new RRs, some kind of cycle between what you have locally and what's available online, never paid enough attention to it, to figure out what it is

types.go
miekg: fair point, but these are just sequentially assigned (unlike the mess that is RR codes). I would just opt for using less characters here, and the constant name should be such that people use that instead of caring of the actual value

types.go
miekg: I'm not aware of any other RR defined in this lib that does this, so please no.

types.go
miekg: ok, yes and no enums in Go - def. warrants a comment

types.go
monoidic: Fair enough. Already implemented this in the last commit, marking as resolved.

types.go
monoidic: I am not sure if I understood the last comment. Please do not what?
Will merge the two if so desired.

types.go
monoidic: Comment added with last PR, marking as resolved.

scan_rr.go
monoidic: Done with last PR, marking as resolved.

types.go
monoidic: I'm not sure I get what you're implying I should do.

scan_rr.go
```diff
@@ -1216,6 +1217,115 @@ func (rr *DS) parse(c *zlexer, o string) *ParseError      { return rr.parseDS(c,
...
 
+func (rr *IPSECKEY) parse(c *zlexer, o string) *ParseError {
+	for i := 0; i < 3; i++ {
+		l, _ := c.Next()
+		num, err := strconv.ParseUint(l.token, 10, 8)
+		if err != nil || l.err {
+			return &ParseError{"", "bad IPSECKEY value", l}
+		}
+
+		num8 := uint8(num)
```
tmthrgd, line 1228: This doesn't need to be a variable. Just use `uint8(num)` below directly.

scan_rr.go
```diff
@@ -1216,6 +1217,115 @@ func (rr *DS) parse(c *zlexer, o string) *ParseError      { return rr.parseDS(c,
 func (rr *DLV) parse(c *zlexer, o string) *ParseError     { return rr.parseDS(c, o, "DLV") }
 func (rr *CDS) parse(c *zlexer, o string) *ParseError     { return rr.parseDS(c, o, "CDS") }
 
+func (rr *IPSECKEY) parse(c *zlexer, o string) *ParseError {
+	for i := 0; i < 3; i++ {
```
tmthrgd, line 1221: I think this might be cleaner if we did something like:
```go
for _, out := range []*uint8{&rr.Precedence, &rr.GatewayType, &rr.Algorithm} {
```

and then below:
```go
*out = uint8(num)
```

scan_rr.go
```diff
@@ -1216,6 +1217,115 @@ func (rr *DS) parse(c *zlexer, o string) *ParseError      { return rr.parseDS(c,
...
+
+func (rr *AMTRELAY) parse(c *zlexer, o string) *ParseError {
+	for i := 0; i < 3; i++ {
+		l, _ := c.Next()
+		num, err := strconv.ParseUint(l.token, 10, 8)
+		if err != nil || l.err {
+			return &ParseError{"", "bad AMTRELAY value", l}
+		}
+
+		num8 := uint8(num)
```
tmthrgd, line 1270: Same here. Just use `uint8(num)`.

scan_rr.go
```diff
@@ -1216,6 +1217,115 @@ func (rr *DS) parse(c *zlexer, o string) *ParseError      { return rr.parseDS(c,
...
+		switch i {
+		case 0:
+			rr.Precedence = num8
+		case 1:
+			if num8 > 1 {
+				return &ParseError{"", "bad discovery value", l}
+			}
+			rr.DiscoveryOptional = num8 == 1
```
tmthrgd, line 1278: This and the if's can just use `num` itself.

scan_rr.go
tmthrgd: Or it might be better just to spell it out one-by-one honestly.

scan_rr.go
```diff
@@ -1216,6 +1217,115 @@ func (rr *DS) parse(c *zlexer, o string) *ParseError      { return rr.parseDS(c,
...
+	s, pErr := endingToString(c, "bad IPSECKEY PublicKey")
+	if pErr != nil {
+		return pErr
+	}
+	rr.PublicKey = s
+	return slurpRemainder(c)
+}
+
+func (rr *AMTRELAY) parse(c *zlexer, o string) *ParseError {
+	for i := 0; i < 3; i++ {
```
tmthrgd, line 1263: I think this is better just listing each field one-by-one.

scan_rr.go
```diff
@@ -1216,6 +1217,115 @@ func (rr *DS) parse(c *zlexer, o string) *ParseError      { return rr.parseDS(c,
...
+
+// same constants and parsing between IPSECKEY and AMTRELAY
+func parseAddrHostUnion(token, o string, gatewayType uint8) (addr net.IP, host string, err error) {
+	switch gatewayType {
+	case IPSECGatewayNone:
+		if token != "." {
+			return addr, host, errors.New("gateway type none with gateway set")
+		}
+	case IPSECGatewayIPv4, IPSECGatewayIPv6:
+
```
tmthrgd, line 1311: No space here.

scan_rr.go
```diff
@@ -1216,6 +1217,115 @@ func (rr *DS) parse(c *zlexer, o string) *ParseError      { return rr.parseDS(c,
...
+		if addr = net.ParseIP(token); addr == nil {
```
tmthrgd, line 1312: Move the `net.ParseIP` out of the if-statement.

scan_rr.go
```diff
@@ -1216,6 +1217,115 @@ func (rr *DS) parse(c *zlexer, o string) *ParseError      { return rr.parseDS(c,
...
+			return addr, host, errors.New("gateway IP invalid")
+		}
+		if !((gatewayType == IPSECGatewayIPv4 && addr.To4() != nil) || (gatewayType == IPSECGatewayIPv6 && addr.To16() != nil)) {
```
tmthrgd, line 1315: `To16` will always return non-nil for the result of `net.ParseIP`. If you're trying to make sure this isn't an IPv4 mapped address, use `To4() == nil`. Otherwise, just drop the `IPSECGatewayIPv6` check.

types.go
```diff
@@ -159,6 +161,22 @@ const (
...
+const (
+	IPSECGatewayNone uint8 = iota
+	IPSECGatewayIPv4
+	IPSECGatewayIPv6
+	IPSECGatewayHost
+)
+
+// Used in AMTRELAY https://datatracker.ietf.org/doc/html/rfc8777#section-4.2.3
+const (
+	AMTRELAYNone uint8 = IPSECGatewayNone
```
tmthrgd, line 174: Drop the `uint8` type here, it will be inferred.

types.go
```diff
@@ -994,6 +1012,72 @@ func (rr *DNSKEY) String() string {
...
+	case AMTRELAYIPv4, AMTRELAYIPv6:
+		gateway = rr.GatewayAddr.String()
+	case AMTRELAYHost:
+		gateway = rr.GatewayHost
+	case AMTRELAYNone:
+		fallthrough
+	default:
+		gateway = "."
+	}
+	var boolS string
```
tmthrgd, line 1068:
```go
boolS := "0"
if rr.DiscoveryOptional {
	boolS = "1"
}
```

types.go
```diff
@@ -1450,7 +1534,7 @@ func (a *APLPrefix) str() string {
 // equals reports whether two APL prefixes are identical.
 func (a *APLPrefix) equals(b *APLPrefix) bool {
 	return a.Negation == b.Negation &&
-		bytes.Equal(a.Network.IP, b.Network.IP) &&
+		a.Network.IP.Equal(b.Network.IP) &&
```
tmthrgd, line 1537: While this probably makes sense, let's leave unrelated changes for separate PRs.

tmthrgd (changes requested):
Mostly fine with me, just a handful of nits to address.

scan_rr.go
monoidic: Is the new solution sufficient?

miekg:
@tmthrgd you happy with this now? If so, you (or I) can click 'squash and merge'

types.go
tmthrgd: @miekg This still needs a decision.

tmthrgd:
@miekg There's still one un-resolved review comment about `DiscoveryOptional`. I think that either needs to be marked as resolved or changed first.

types.go
miekg: ah, so, yes. Please remove any bits from the RR definition that are only there to facility text parsing. The RR needs to be as close to the wirefmt as it can be.

scan_rr.go
```diff
@@ -1216,6 +1217,121 @@ func (rr *DS) parse(c *zlexer, o string) *ParseError      { return rr.parseDS(c,
...
+	l, _ := c.Next()
+	num, err := strconv.ParseUint(l.token, 10, 8)
+	if err != nil || l.err {
+		return &ParseError{"", "bad AMTRELAY value", l}
+	}
+	rr.Precedence = uint8(num)
+	c.Next() // zBlank
+
+	l, _ = c.Next()
+	num, err = strconv.ParseUint(l.token, 10, 8)
```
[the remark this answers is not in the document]
monoidic: Implemented a variant of this after removing `DiscoveryOptional`.

types.go
monoidic: `DiscoveryOptional` is now merged with `GatewayType`

tmthrgd (approved):
LGTM.

PR #1403, opened 2022-12-05
Add slackhq/nebula to users of the library

jasikpark:
This is currently used to provide DNS server capabilities to the nebula lighthouses.


miekg:
:+1:

PR #1414, opened 2023-01-16
Update actions & add go 1.19.x and 1.20.x to go workflow

GiddyGoatGaming:
Updates actions and updates go to 1.19.x

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go: [ 1.18.x, 1.19.x ]
+        go: [ 1.19.x, 1.20.x ]
     steps:
 
     - name: Set up Go
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go: [ 1.17.x, 1.18.x, 1.19.x ]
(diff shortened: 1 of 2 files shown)


.github/workflows/go.yml
```diff
@@ -7,16 +7,16 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go: [ 1.17.x, 1.18.x ]
+        go: [ 1.17.x, 1.18.x, 1.19.x ]
```
tmthrgd, line 10: @GiddyGoatGaming Could you remove `1.17.x` at the same time please? We only need to test the two most recent supported versions.

GiddyGoatGaming: > @GiddyGoatGaming Could you remove `1.17.x` at the same time please? We only need to test the two most recent supported versions.

Done!

tmthrgd (approved):
LGTM. (@miekg I haven't checked if there are any important changes between the action versions that might need to be considered).

.github/workflows/go.yml
```diff
@@ -7,16 +7,16 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go: [ 1.17.x, 1.18.x ]
+        go: [ 1.18.x, 1.19.x ]
```
tmthrgd, line 10: Go1.20 is now out. Can you change this to 1.19.x and 1.20.x?

tmthrgd (approved):

miekg:
think it's just fine to go to v3

PR #1416, opened 2023-02-04
Goroutine Pool for UDP datagram reads

horahoradev:
This PR puts UDP server datagram reads into a Goroutine pool, and was made to address this issue within CoreDNS: https://github.com/coredns/coredns/issues/5595#issuecomment-1304496589

Here are some benchmarks of CoreDNS with and without my changes, with one server, using DNSperf.
## Without my change:
```
  Queries sent:         13988687
  Queries completed:    13988288 (100.00%)
  Queries lost:         300 (0.00%)
  Queries interrupted:  99 (0.00%)

  Response codes:       NOERROR 13988288 (100.00%)
  Average packet size:  request 32, response 100
  Run time (s):         105.433637
  Queries per second:   132673.863845

  Average Latency (s):  0.000627 (min 0.000023, max 0.022370)
  Latency StdDev (s):   0.000151

CPU utilization ~420%
```

## With my change:
```
  Queries sent:         5735429
  Queries completed:    5735336 (100.00%)
  Queries lost:         0 (0.00%)
  Queries interrupted:  93 (0.00%)

  Response codes:       NOERROR 5735336 (100.00%)
  Average packet size:  request 32, response 100
  Run time (s):         29.624372
  Queries per second:   193601.943697

  Average Latency (s):  0.000483 (min 0.000014, max 0.009152)
  Latency StdDev (s):   0.000336

CPU utilization ~560%
```
So notably we see CPU utilization went up significantly but QPS went up ~50%, and avg latency is down.

This is a pretty stupid change, but it seems pretty impactful. Let me know if there's anything you'd like to see here; from my interpretation, all affected method calls should be thread-safe (but let me know if I've missed anything!).

### DIFF
--- a/server.go
+++ b/server.go
@@ -12,6 +12,9 @@ import (
 	"runtime"
 	"strings"
 	"sync"
+
+	"golang.org/x/sync/errgroup"
+
 	"time"
 )
 
@@ -499,14 +502,10 @@ func (srv *Server) serveUDP(l net.PacketConn) error {
 
 	rtimeout := srv.getReadTimeout()
 
-	errCh := make(chan error, runtime.NumCPU())
-	poolWg := sync.WaitGroup{}
+	g := new(errgroup.Group)
 
 	for i := 0; i < runtime.NumCPU(); i++ {
-		poolWg.Add(1)
-
-		go func() {
-			defer poolWg.Done()
+		g.Go(func() error {
 			// deadline is not used here
 			for srv.isStarted() {
 				var (
@@ -523,13 +522,13 @@ func (srv *Server) serveUDP(l net.PacketConn) error {
 				}
 				if err != nil {
 					if !srv.isStarted() {
-						return
+						return nil
 					}
 					if netErr, ok := err.(net.Error); ok && netErr.Temporary() {
 						continue
 					}
-					errCh <- err
-					return
+
+					return err
 				}
 				if len(m) < headerSize {
 					if cap(m) == srv.UDPSize {
@@ -540,16 +539,16 @@ func (srv *Server) serveUDP(l net.PacketConn) error {
 				wg.Add(1)
 				go srv.serveUDPPacket(&wg, m, l, sUDP, sPC)
 			}
-		}()
+
+			return nil
+		})
 	}
 
-	poolWg.Wait()
+	err := g.Wait()
 	wg.Wait()
-	close(errCh)
 	close(srv.shutdown)
 
-	// TODO
-	return <-errCh
+	return err
 }
 
 // Serve a new TCP connection.
--- a/server.go
+++ b/server.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"io"
 	"net"
+	"runtime"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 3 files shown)


server.go
```diff
@@ -495,45 +496,60 @@ func (srv *Server) serveUDP(l net.PacketConn) error {
...
-		}
-		wg.Add(1)
-		go srv.serveUDPPacket(&wg, m, l, sUDP, sPC)
+		}()
 	}
 
-	return nil
+	poolWg.Wait()
+	wg.Wait()
+	close(errCh)
```
tmthrgd, line 548: There’s no need to close this channel.

server.go
```diff
@@ -495,45 +496,60 @@ func (srv *Server) serveUDP(l net.PacketConn) error {
...
-				continue
-			}
-			return err
-		}
-		if len(m) < headerSize {
-			if cap(m) == srv.UDPSize {
-				srv.udpPool.Put(m[:srv.UDPSize])
+
+	errCh := make(chan error, runtime.NumCPU())
+	poolWg := sync.WaitGroup{}
```
tmthrgd, line 503: This might be simpler with x/sync/errgroup.

tmthrgd (changes requested):
I’d like to see this with errgroup as I think that’s going to be better here.

horahoradev:
Thanks for the quick review! I changed the waitgroup/error channel out for an error group.
I opted not to cause the rest of the Goroutine read loops to terminate if one of them returns an error, but let me know if you'd like to see something different here.

miekg:
@tmthrgd is this good to merge?

tmthrgd:
@miekg I'm not a huge fan. Pools and multiple readers are hard to get right in a general sense. What might improve things for one workload, could easily worsen things in others. In particular I know things can get very counterintuitive with high core count CPUs.

server.go
```diff
@@ -9,8 +9,12 @@ import (
 	"errors"
 	"io"
 	"net"
+	"runtime"
 	"strings"
 	"sync"
+
+	"golang.org/x/sync/errgroup"
```
tmthrgd, line 16: This isn't formatted properly. Stdlib should be grouped up separately from other imports.

server.go
```diff
@@ -495,45 +499,56 @@ func (srv *Server) serveUDP(l net.PacketConn) error {
...
+			for srv.isStarted() {
+				var (
+					m    []byte
+					sPC  net.Addr
+					sUDP *SessionUDP
+					err  error
+				)
+
+				if isUDP {
+					m, sUDP, err = reader.ReadUDP(lUDP, rtimeout)
```
tmthrgd, line 519: There's nothing to actually ensure this call unblocks when the errgroup returns with an error. This could easily end up still reading in one go routine while others have already exited.

server.go
```diff
@@ -495,45 +499,56 @@ func (srv *Server) serveUDP(l net.PacketConn) error {
...
-			if netErr, ok := err.(net.Error); ok && netErr.Temporary() {
-				continue
-			}
-			return err
-		}
-		if len(m) < headerSize {
-			if cap(m) == srv.UDPSize {
-				srv.udpPool.Put(m[:srv.UDPSize])
+
+	g := new(errgroup.Group)
```
tmthrgd, line 505: var g errgroup.Group

server.go
```diff
@@ -495,45 +499,56 @@ func (srv *Server) serveUDP(l net.PacketConn) error {
...
+
+	for i := 0; i < runtime.NumCPU(); i++ {
```
tmthrgd, line 507: I don't know if NumCPU is correct here or not.

tmthrgd (changes requested):

tmthrgd:
Maybe we could define some sort of interface for people who want to implement pools of some kind. I'm not sure.

miekg:
Ack. Thanks for your thoughts.

Tend to agree, so I'll close this PR. As for an interface, nothing forces
you to use all the things in this library, so prolly best to not define
anything for this use case

PR #1426, opened 2023-03-02
client: introduce dialOptions, and new dial option WithLocalAddr

aashh:
`DialTimeout()` etc. are very opinionated about how they construct their `Client`. This makes it a bit cumbersome to introduce small variations, such as a dialer that uses a specific local (source) network address.

Rather than having to duplicate these two functions for just a slight variation, or proposing an incompatible API change, I propose here the introduction of optional `dialOption` parameters to these functions. A first such option is introduced here, `WithLocalAddr()`. This allows constructing a call like:
```
DialTimeout(proto, addr, timeout, WithLocalAddr(local))
```

This change continues forward on the `dialOptions` front to remove duplicate code across the `DialTimeout()` family of functions, introducing `WithTimeout()` and `WithTLSConfig()` as well.

This change would allow avoiding the duplication of code in coredns/coredns#5933, which is the reason for proposing it.

### DIFF
--- a/client.go
+++ b/client.go
@@ -445,28 +445,57 @@ func ExchangeConn(c net.Conn, m *Msg) (r *Msg, err error) {
 	return r, err
 }
 
-// DialTimeout acts like Dial but takes a timeout.
-func DialTimeout(network, address string, timeout time.Duration) (conn *Conn, err error) {
-	client := Client{Net: network, Dialer: &net.Dialer{Timeout: timeout}}
+type dialOption func(*Client)
+
+// WithTimeout returns a function that mutates the dialer timeout of a Client.
+func WithTimeout(timeout time.Duration) dialOption {
+	return func(client *Client) {
+		client.Dialer.Timeout = timeout
+	}
+}
+
+// WithTimeout returns a function that mutates the TLS config of a Client.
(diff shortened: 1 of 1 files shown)


miekg:
you can make a dns.Conn from a net.Conn with `&dns.Conn{Conn:conn}`, so don't see the need to extend this api

aashh:
> you can make a dns.Conn from a net.Conn with `&dns.Conn{Conn:conn}`, so don't see the need to extend this api

It's just about avoiding duplication of code. For example it seems wasteful to duplicate the `if !strings.HasSuffix(network, "-tls")` logic inside of CoreDNS when these helpers already know what to do and just lack some flexibility. But if the answer is to just not use the helpers when they're too opinionated, I can live with that.

tmthrgd:
The Client API is a little rough in a few places, but I don't think we should make this change. The API would be worth re-thinking if we ever proceed with a v2. Also given that DNS requests can be quite performance sensitive, the overhead of this kind of approach might be too much. (As much as I do like this kind of API).

In terms of duplication, I actually think this library probably should have been a little stricter. Things like whether to interpret "tcp4-tls" (and similar) should really be done in higher level libraries, so I think duplication is okay here.

PR #1432, opened 2023-03-07
Introduce a generic slice clone function

### LINKED ISSUE #1381: A.copy does not preserve the nilness of A.IP
In short:

```go
package main

import (
	"fmt"

	"github.com/miekg/dns"
)

func main() {
	a := &dns.A{}
	a2 := dns.Copy(a).(*dns.A)
	fmt.Println(a.A == nil, a2.A == nil)
}
```

```none
true false
```

<https://go.dev/play/p/1i32nD_s_GY>

We've recently stumbled upon this while performing some tests.  We had some code that returned a `net.IP` from `*dns.Msg` data, and another piece of code that assumed that the returned `net.IP` is either `nil` or a valid `net.IP`.  But it seems that if the response contains a `*dns.A` RR with a nil `net.IP`, and that response is then copied, that nilness is lost, and the copy now contains an empty, non-nil `net.IP`.

tmthrgd:
This is nil-preserving unlike the existing code.

We also add some missing []byte slice copies.

Fixes #1381


tmthrgd:
The go.mod / go.sum changes come from bumping the go version and running `go mod tidy`.

miekg (approved):

PR #1433, opened 2023-03-07
Correct TODO in udp_windows.go

tmthrgd:
These TODOs were not correct as x/net/ipv4 and x/net/ipv6 still don't have Windows support.

Updates #738
Updates #765
Updates coredns/coredns#2138
See golang/go#7175
See golang/go#7174

### DIFF
--- a/udp_windows.go
+++ b/udp_windows.go
@@ -1,6 +1,9 @@
 //go:build windows
 // +build windows
 
+// TODO(tmthrgd): Remove this Windows-specific code if go.dev/issue/7175 and
+//   go.dev/issue/7174 are ever fixed.
+
 package dns
 
 import "net"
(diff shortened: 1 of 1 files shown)


francislavoie:
Hopefully you don't mind me asking here, but it's the most recent issue/PR about Windows.

Would this be an explanation for why failures to connect to a DNS server takes significantly longer on Windows?

We're investigating the tests in https://github.com/caddyserver/certmagic/issues/228#issuecomment-1510497224 and the ones that use intentionally incorrect resolver addresses (localhost with a non-standard port) takes until the configured timeout to return, whereas on other platforms it fails fast. We're using `udp.Exchange()`.

tmthrgd:
@francislavoie I don't believe so as this should only really impact sending packets and making sure they go out the right interface. I'd be surprised if it could cause that.

PR #1436, opened 2023-03-12
refact(msg): constant refactoring

isaqueveras:


tmthrgd:
Thanks. What's the motive for this change? It seems to me, this is less clear.

isaqueveras:
Hi @tmthrgd, I added the calculation value. I even left the calculation in the comment on the side.

miekg:
Also don't think this an improvement...

PR #1440, opened 2023-03-15
Updates to LOC type RFC1876

systemcrash:
https://www.rfc-editor.org/rfc/rfc1876

### DIFF
--- a/types.go
+++ b/types.go
@@ -198,7 +198,7 @@ const (
 	_CD = 1 << 4  // checking disabled
 )
 
-// Various constants used in the LOC RR. See RFC 1887.
+// Various constants used in the LOC RR. See RFC 1876.
 const (
 	LOC_EQUATOR       = 1 << 31 // RFC 1876, Section 2.
 	LOC_PRIMEMERIDIAN = 1 << 31 // RFC 1876, Section 2.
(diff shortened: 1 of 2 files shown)


types.go
```diff
@@ -198,7 +198,7 @@ const (
 	_CD = 1 << 4  // checking disabled
 )
 
-// Various constants used in the LOC RR. See RFC 1887.
+// Various constants used in the LOC RR. See RFC 1876.
```
miekg, line 201: fine with this

types_test.go
```diff
@@ -5,37 +5,37 @@ import (
 )
 
 func TestCmToM(t *testing.T) {
-	s := cmToM(0, 0)
+	s := cmToM((0 << 0) + 0)
```
tmthrgd, line 8: `<< 4`

types.go
```diff
@@ -792,7 +792,10 @@ type LOC struct {
 
 // cmToM takes a cm value expressed in RFC 1876 SIZE mantissa/exponent
 // format and returns a string in m (two decimals for the cm).
-func cmToM(m, e uint8) string {
+func cmToM(x uint8) string {
+	var m = x & 0xf0 >> 4
+	var e = x & 0x0f
```
tmthrgd, line 797: `:=` rather than `var ... =`

tmthrgd (changes requested):
Minor nits.

tmthrgd (approved):
LGTM.

types_test.go
systemcrash: seemed a waste of cpu cycles :)

PR #1442, opened 2023-03-17
Fix some lint checks by (go) staticcheck ./...

systemcrash:
Minimally invasive, identical behaviour. 

Some warn about deprecation.

There are a few more which require some more thought.


dnssec.go
```diff
@@ -128,9 +128,9 @@ type dnskeyWireFmt struct {
 	/* Nothing is left out */
 }
 
-func divRoundUp(a, b int) int {
-	return (a + b - 1) / b
-}
+// func divRoundUp(a, b int) int {
```
tmthrgd, line 131: I think it's safe to delete this, it was used only for DSA I believe.

dnssec.go
```diff
@@ -417,11 +417,11 @@ func (rr *RRSIG) Verify(k *DNSKEY, rrset []RR) error {
...
 	}
 
-	sigbuf := rr.sigBuf()           // Get the binary signature data
-	if rr.Algorithm == PRIVATEDNS { // PRIVATEOID
-		// TODO(miek)
-		// remove the domain name and assume its ours?
-	}
+	sigbuf := rr.sigBuf() // Get the binary signature data
+	// if rr.Algorithm == PRIVATEDNS { // PRIVATEOID
+	// 	// TODO(miek)
```
tmthrgd, line 422: I think pull this TODO above the if. Something like:

// TODO...
// if ...

msg.go
```diff
@@ -870,11 +870,11 @@ func (dns *Msg) unpack(dh Header, msg []byte, off int) (err error) {
 		dns.Rcode |= opt.ExtendedRcode()
 	}
 
-	if off != len(msg) {
-		// TODO(miek) make this an error?
-		// use PackOpt to let people tell how detailed the error reporting should be?
-		// println("dns: extra bytes in dns packet", off, "<", len(msg))
-	}
+	// if off != len(msg) {
+	// 	// TODO(miek) make this an error?
```
tmthrgd, line 874: Same here.

msg_helpers_test.go
```diff
@@ -434,7 +434,7 @@ func TestUnpackDataAplPrefix(t *testing.T) {
 			if got.Negation != tt.negation {
 				t.Errorf("expected negation %v, got %v", tt.negation, got.Negation)
 			}
-			if !bytes.Equal(got.Network.IP, tt.ip) {
+			if !net.IP.Equal(got.Network.IP, tt.ip) {
```
tmthrgd, line 437: tt.ip.Equal(...)

msg_helpers_test.go
```diff
@@ -583,7 +583,7 @@ func TestUnpackDataApl(t *testing.T) {
 		if got[i].Negation != exp.Negation {
 			t.Errorf("[%d] expected negation %v, got %v", i, exp.Negation, got[i].Negation)
 		}
-		if !bytes.Equal(got[i].Network.IP, exp.Network.IP) {
+		if !net.IP.Equal(got[i].Network.IP, exp.Network.IP) {
```
tmthrgd, line 586: Same here. exp.Network.IP.Equal(...)

parse_test.go
```diff
@@ -225,6 +225,7 @@ func GenerateTXT(r *rand.Rand, size int) []byte {
 	return rd
 }
 
+/*
```
tmthrgd, line 228: I think we can just delete this.

msg.go
```diff
@@ -860,7 +860,7 @@ func (dns *Msg) unpack(dh Header, msg []byte, off int) (err error) {
 	// The header counts might have been wrong so we need to update it
 	dh.Nscount = uint16(len(dns.Ns))
 	if err == nil {
-		dns.Extra, off, err = unpackRRslice(int(dh.Arcount), msg, off)
+		dns.Extra, _, err = unpackRRslice(int(dh.Arcount), msg, off)
```
tmthrgd, line 863: I'm a bit torn on this. Yea it's not read, but it breaks symmetry with the line above.

tmthrgd (changes requested):
Thanks. A few changes though.

tmthrgd:
Would you mind adding the entire output of staticcheck as comment just for reference?

systemcrash:
Before PR
```text
clientconfig_test.go:4:2: "io/ioutil" has been deprecated since Go 1.19: As of Go 1.16, the same functionality is now provided by package io or package os, and those implementations should be preferred in new code. See the specific function documentation for details.  (SA1019)
dnssec.go:131:6: func divRoundUp is unused (U1000)
dnssec.go:421:2: empty branch (SA9003)
generate_test.go:5:2: "io/ioutil" has been deprecated since Go 1.19: As of Go 1.16, the same functionality is now provided by package io or package os, and those implementations should be preferred in new code. See the specific function documentation for details.  (SA1019)
msg.go:873:2: empty branch (SA9003)
msg_helpers_test.go:437:8: use net.IP.Equal to compare net.IPs, not bytes.Equal (SA1021)
msg_helpers_test.go:586:7: use net.IP.Equal to compare net.IPs, not bytes.Equal (SA1021)
parse_test.go:234:6: func testTXTRRQuick is unused (U1000)
scan_test.go:5:2: "io/ioutil" has been deprecated since Go 1.19: As of Go 1.16, the same functionality is now provided by package io or package os, and those implementations should be preferred in new code. See the specific function documentation for details.  (SA1019)
server.go:462:44: neterr.Temporary has been deprecated since Go 1.18 because it shouldn't be used: Temporary errors are not well-defined. Most "temporary" errors are timeouts, and the few exceptions are surprising. Do not use this method.  (SA1019)
server.go:521:44: netErr.Temporary has been deprecated since Go 1.18 because it shouldn't be used: Temporary errors are not well-defined. Most "temporary" errors are timeouts, and the few exceptions are surprising. Do not use this method.  (SA1019)
server.go:528:21: argument should be pointer-like to avoid allocations (SA6002)
server.go:640:20: argument should be pointer-like to avoid allocations (SA6002)
server.go:656:19: argument should be pointer-like to avoid allocations (SA6002)
server.go:697:19: argument should be pointer-like to avoid allocations (SA6002)
server.go:715:19: argument should be pointer-like to avoid allocations (SA6002)
```

After
```text
server.go:462:44: neterr.Temporary has been deprecated since Go 1.18 because it shouldn't be used: Temporary errors are not well-defined. Most "temporary" errors are timeouts, and the few exceptions are surprising. Do not use this method.  (SA1019)
server.go:521:44: netErr.Temporary has been deprecated since Go 1.18 because it shouldn't be used: Temporary errors are not well-defined. Most "temporary" errors are timeouts, and the few exceptions are surprising. Do not use this method.  (SA1019)
server.go:528:21: argument should be pointer-like to avoid allocations (SA6002)
server.go:640:20: argument should be pointer-like to avoid allocations (SA6002)
server.go:656:19: argument should be pointer-like to avoid allocations (SA6002)
server.go:697:19: argument should be pointer-like to avoid allocations (SA6002)
server.go:715:19: argument should be pointer-like to avoid allocations (SA6002)
```

msg_helpers_test.go
systemcrash: function signature changes as a result..

msg.go
tmthrgd: Yeah I think back out this change and then LGTM.

tmthrgd (changes requested):
One last nit.

systemcrash:
> One last nit.

`msg.go:863:3: this value of off is never used (SA4006)`

msg.go
systemcrash: `msg.go:863:3: this value of off is never used (SA4006)`

tmthrgd:
Okay that was missed out of the comment above. Yeah sure add the `off` -> `_` change back in.

Quick thing though, please don't force push as it messes up our ability to review PRs. Just add new commits to the branch.

tmthrgd (approved):
LGTM.

Issue #1441, opened 2023-03-17
Test failure on macOS. 


systemcrash:
When I run `go test` I get
```text
--- FAIL: TestIsPacketConn (0.00s)
    client_test.go:74: unable to run test server: listen unixpacket /....../unixpacket.sock: socket: protocol not supported
FAIL
exit status 1
FAIL	github.com/miekg/dns	1.215s
```

Normal for macos? It does say 'unix' there, but...

`go version go1.19.4 darwin/amd64`

miekg:
closed via #1443

PR #1443, opened 2023-03-18
tests: skip unix socket tests for non *nix OSes

pmalek:
To make the full test suite pass on e.g. darwin, skip tests that try to create local unix server for protocols not supported on non *nix OSes.

This might not be the full, comprehensive list but it will do for darwin and windows.

Feel free to suggest a more robust check. OSes might be consulted with output of `go tool dist list`


miekg:
this shouldn't be wholesale disabled like this.

Can't think of a reason why this wouldn't work on macos

pmalek:
> this shouldn't be wholesale disabled like this.
> 
> Can't think of a reason why this wouldn't work on macos

I've updated the description with errors that are returned.

pmalek:
Alright so there's 2 problems:

One which is socket file path length limitations:

- https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_un.h.html#tag_13_67_04
- https://unix.stackexchange.com/questions/367008/why-is-socket-path-length-limited-to-a-hundred-chars

The error is coming from here: https://github.com/golang/go/blob/go1.20.2/src/syscall/syscall_bsd.go#L191

I've added a func that creates temp dir for tests which seems to work for those tests that were initially failing on darwin. It barelly fits the 104 character limit now.

```
/var/folders/0m/_63w01516tgf3cftmp9h7ylm0000gn/T/TestIsPacketConn-Unix_Seqpacket2194565336/unixpacket.sock
```

vs (before)

```
/var/folders/0m/_63w01516tgf3cftmp9h7ylm0000gn/T/TestIsPacketConnUnix_Seqpacket1787766192/001/unixpacket.sock
```

---

Second problem is with darwin not supporting sockets with domain: `syscall.AF_UNIX` `1`, sotype `syscall.SOCK_SEQPACKET` = `5`, and proto `0`. Not sure why that's the case, https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/socket.2.html doesn't mention anything about it not being supported.

tmthrgd:
I'd rather detect the actual error and skip on that than try to guess compatibility.

pmalek:
> I'd rather detect the actual error and skip on that than try to guess compatibility.

👍 Makes sense. I've updated the PR, PTAL @tmthrgd

client_test.go
```diff
@@ -6,106 +6,136 @@ import (
...
-		t.Error("Unix stream connection should not be a packet conn")
-	}
-	if isPacketConn(struct{ *net.UnixConn }{c.(*net.UnixConn)}) {
-		t.Error("Unix stream connection (wrapped type) should not be a packet conn")
-	}
+	t.Run("Unix Seqpacket", func(t *testing.T) {
+		shutChan, addrstr, err := RunLocalUnixSeqPacketServer(filepath.Join(tempDir(t), "unixpacket.sock"))
+		if err != nil {
+			var sysErr syscall.Errno
+			if errors.As(err, &sysErr) && sysErr == syscall.EPROTONOSUPPORT {
```
tmthrgd, line 99: This can just be `errors.Is`. Is `EPROTONOSUPPORT` actually defined on all systems? Perhaps the new `errors.ErrUnsupported` is more appropriate once go1.21 lands.

tmthrgd (changes requested):

client_test.go
pmalek: Yup, you're right. Changed

systemcrash:
Fixes #1441 ?

pmalek:
> Fixes #1441 ?

Yup

miekg:
isn't the only fix needed here that `filepath.Join(tempDir(t), "unixpacket.sock")` returns a short enough path for macos?

can that be the only change, instead of rewriting the entire file? basically only tempdir, but with a better name (shortUnixSockPath or whatever), with comments that mac os is doing dumb here

pmalek:
> isn't the only fix needed here that `filepath.Join(tempDir(t), "unixpacket.sock")` returns a short enough path for macos?
> 
> can that be the only change, instead of rewriting the entire file? basically only tempdir, but with a better name (shortUnixSockPath or whatever), with comments that mac os is doing dumb here

I can extract a func like `tempFile` with appropriate build tags for `darwin` and `!darwin` to ensure both work and tests pass but e.g. this

```
		shutChan, addrstr, err := RunLocalUnixSeqPacketServer(filepath.Join(tempDir(t), "unixpacket.sock"))
		if err != nil {
			if errors.Is(err, syscall.EPROTONOSUPPORT) {
				t.Skip("unix seqpacket not supported on this OS")
			}
		}
```

wouldn't work without a subtest enclosing this, i.e. the unsupported error will skip the whole test without the subtest.

I've pushed a change that adds those 2 functions for darwin and non darwin. PTAL

miekg (approved):

PR #1446, opened 2023-03-29
Buffer bump

systemcrash:
Also version bump

See individual commit messages


tmthrgd:
Do people really care about GPG/PGP these days?

version.go
```diff
@@ -3,7 +3,7 @@ package dns
 import "fmt"
 
 // Version is current version of this library.
-var Version = v{1, 1, 53}
```
miekg, line 6: without saying yes or no to this pr, please revert this version bump from this pr

miekg:
To add to @tmthrgd reply: this is with older RSA keys or something? Messsages this big will also lead to fragmentation problems on the wire

version.go
systemcrash: Sure.

miekg:
[ Quoting ***@***.***> in "Re: [miekg/dns] Buffer bump (PR #14..." ]
>    Do people really care about GPG/PGP these days?
>
>case in point.
>
>    To add to @tmthrgd reply: this is with older RSA keys or something?
>    Messsages this big will also lead to fragmentation problems on the wire
>
>messages this small evidently lead to problems in this lib before it even gets
>on the wire.
>
>your point is moot: DOT, DOH, TCP based protos resolve this. These problems are
>discussed in the RFC.

just upping the value also isn't a real solution. Add a space (as bind also randomly does)
should (i think) also work, as then you get multiple tokens.

all the newer based protos are mostly for the last mile, so of very much limited value

systemcrash:
> just upping the value also isn't a real solution. Add a space (as bind also randomly does) should (i think) also work, as then you get multiple tokens. all the newer based protos are mostly for the last mile, so of very much limited value

It's low hanging fruit - the RFC states just a single line of base64 without elaborating further.

systemcrash:
OK - how about this? Just reduce the overall buffer size to 512 bytes, and make it something we can grow by 512 bytes every time we run out. It also fixes me problem, and obsoletes length test cases (or at least reverses the fail logic).

miekg:
> OK - how about this? Just reduce the overall buffer size to 512 bytes, 

if that's easy to do, then yes

systemcrash:
Already done. Sitting ready in this PR.

miekg (approved):

PR #1450, opened 2023-04-09
Define MaxGoroutines to limit the number goroutines created by the server

yaa110:
A DNS flood attack can simply cause an OOM due to creation of unlimited goroutines by the server.

This attack can be simulated using [dns-flood](https://github.com/nickwinn/dns-flood):

```bash
clone https://github.com/nickwinn/dns-flood.git
cd dns-flood
gcc -o dns-flood dnsflood.c
./dns-flood <domain> <server-ip> -r -i 0 -p <server-port>
```

### DIFF
--- a/server.go
+++ b/server.go
@@ -12,6 +12,8 @@ import (
 	"strings"
 	"sync"
 	"time"
+
+	"golang.org/x/sync/semaphore"
 )
 
 // Default maximum number of TCP queries before we close the socket.
@@ -77,6 +79,11 @@ type response struct {
 	udpSession     *SessionUDP    // oob data to get egress interface right
 	pcSession      net.Addr       // address to use when writing to a generic net.PacketConn
 	writer         Writer         // writer to output the raw DNS bits
+}
+
+type limiter struct {
+	ctx context.Context
+	sem *semaphore.Weighted
 }
 
 // handleRefused returns a HandlerFunc that returns REFUSED for every request it gets.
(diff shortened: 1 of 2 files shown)


miekg:
I believe the official `net/http` package has similar issues and the upstream stance there is: this is solveable in lower levels. The same should apply here.

yaa110:
standard http package provides a [`Serve`](https://pkg.go.dev/net/http#Serve) method to accept a custom listener such as [`netutil.LimitListener`](https://pkg.go.dev/golang.org/x/net/netutil?utm_source=godoc#LimitListener) to limit the number of concurrent connections. However, this package lacks such a functionality.

miekg:
Thanks, hadn't heard of LimitListener before. That seems like a better approach than just capping the number of goroutines, although that also uses a plain number as a cutoff. 
But you can use that pkg to set you're own listener in a Server struct.

tmthrgd:
You can achieve this with a custom
Reader and net.Listener I believe. Though it's a valid problem, that ideally we'd have some solution for.

tmthrgd:
I think we need to consider both this and #1416 together.

semihalev:
I tested the `dnsflood` on `sdns` on  one of live recursive resolver server. I have query based ratelimit on sdns. My goroutines was stable.  This is not security issue on this library! 

The Grafana test results: https://snapshots.raintank.io/dashboard/snapshot/NCQmjYhF2lHD36p2XFEEpN9g4VEkHDJ3

miekg:
closing this per last comment

PR #1453, opened 2023-04-18
Improve IsFQDN performance

### LINKED ISSUE #1452: IsFqdn performance issues
I have recently added the kubernetes node-cache and noticed a quite high CPU load. The cached node was actually higher than any of the load on the dns server before adding the caching layer.

So after a bit of pprof- and perf-ing around, I've noticed, that the `IsFqdn` call seems to be excessively expensive. Is this something that can be easily optimized? I'm not too deep into Go and the associated logic here, so feedback is welcome :D 

image (11)

tmthrgd:
While this code may be slightly less clear, it's significantly faster and this function seems to be a hot path for certain workloads.

This function still has significant overhead for names that have an escape sequence immediately before the final dot. While an important edge case to handle, these should be vanishingly rare with most callers noticing the speedup. This change even offers a rather decent performance improvement for these names anyway.

I tried splitting `IsFQDN` into a fast and slow path for better inlining, but couldn't get it below the inlining threshold no matter what I tried, so I have left it as one function.

```
name                 old time/op    new time/op    delta
IsFQDN/no_dot-12       5.86ns ± 2%    1.48ns ± 3%  -74.71%  (p=0.000 n=10+10)
IsFQDN/unescaped-12    8.73ns ± 2%    1.57ns ± 1%  -81.98%  (p=0.000 n=9+8)
IsFQDN/escaped-12      27.4ns ± 2%    23.8ns ± 2%  -13.19%  (p=0.000 n=10+10)
FQDN/is_fqdn-12        8.36ns ± 1%    1.80ns ± 2%  -78.50%  (p=0.000 n=9+10)
FQDN/not_fqdn-12       36.8ns ±15%    33.4ns ±12%   -9.25%  (p=0.035 n=10+10)
```

Fixes #1452

### DIFF
--- a/defaults.go
+++ b/defaults.go
@@ -285,7 +285,7 @@ func IsFqdn(s string) bool {
 	}
 
 	// Otherwise we have to check if the dot is escaped or not by checking if
-	// the escape sequences before the dot is odd or even.
+	// there are an odd or even number of escape sequences before the dot.
 	i := strings.LastIndexFunc(s, func(r rune) bool {
 		return r != '\\'
 	})
--- a/defaults.go
+++ b/defaults.go
@@ -272,18 +272,24 @@ func IsMsg(buf []byte) error {
 
 // IsFqdn checks if a domain name is fully qualified.
 func IsFqdn(s string) bool {
-	s2 := strings.TrimSuffix(s, ".")
-	if s == s2 {
+	// Check for (and remove) a trailing dot, returning if there isn't one.
+	if s == "" || s[len(s)-1] != '.' {
 		return false
 	}
+	s = s[:len(s)-1]
 
-	i := strings.LastIndexFunc(s2, func(r rune) bool {
+	// If we don't have an escape sequence before the final dot, we know it's
+	// fully qualified and can return here.
+	if s == "" || s[len(s)-1] != '\\' {
+		return true
+	}
+
+	// Otherwise we have to check if the dot is escaped or not by checking if
+	// the escape sequences before the dot is odd or even.
+	i := strings.LastIndexFunc(s, func(r rune) bool {
 		return r != '\\'
 	})
-
-	// Test whether we have an even number of escape sequences before
-	// the dot or none.
-	return (len(s2)-i)%2 != 0
+	return (len(s)-i)%2 != 0
 }
 
 // IsRRset checks if a set of RRs is a valid RRset as defined by RFC 2181.
--- a/dns_test.go
+++ b/dns_test.go
@@ -314,3 +314,49 @@ func TestTKEY(t *testing.T) {
 		t.Fatalf("unable to parse TKEY string: %s", newError)
 	}
 }
+
+var (
+	sinkBool   bool
+	sinkString string
+)
+
+func BenchmarkIsFQDN(b *testing.B) {
+	b.Run("no_dot", func(b *testing.B) {
+		var r bool
+		for n := 0; n < b.N; n++ {
+			r = IsFqdn("www.google.com")
+		}
+		sinkBool = r
+	})
+	b.Run("unescaped", func(b *testing.B) {
+		var r bool
+		for n := 0; n < b.N; n++ {
+			r = IsFqdn("www.google.com.")
+		}
+		sinkBool = r
+	})
+	b.Run("escaped", func(b *testing.B) {
+		var r bool
+		for n := 0; n < b.N; n++ {
+			r = IsFqdn(`www.google.com\\\\\\\\.`)
+		}
+		sinkBool = r
+	})
+}
+
+func BenchmarkFQDN(b *testing.B) {
+	b.Run("is_fqdn", func(b *testing.B) {
+		var r string
+		for n := 0; n < b.N; n++ {
+			r = Fqdn("www.google.com.")
+		}
+		sinkString = r
+	})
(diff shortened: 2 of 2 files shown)


miekg:
lgtm

defaults.go
```diff
@@ -272,18 +272,24 @@ func IsMsg(buf []byte) error {
...
+	// Otherwise we have to check if the dot is escaped or not by checking if
+	// the escape sequences before the dot is odd or even.
+	i := strings.LastIndexFunc(s, func(r rune) bool {
 		return r != '\\'
 	})
-
-	// Test whether we have an even number of escape sequences before
-	// the dot or none.
-	return (len(s2)-i)%2 != 0
+	return (len(s)-i)%2 != 0
```
kerolasa, line 292: Changing modulo operation to bit wise and should make function to run a tiny bit quicker.

```
return ((len(s)-i)&1 == 1)
```

defaults.go
```diff
@@ -272,18 +272,24 @@ func IsMsg(buf []byte) error {
 
 // IsFqdn checks if a domain name is fully qualified.
 func IsFqdn(s string) bool {
-	s2 := strings.TrimSuffix(s, ".")
-	if s == s2 {
+	// Check for (and remove) a trailing dot, returning if there isn't one.
+	if s == "" || s[len(s)-1] != '.' {
 		return false
 	}
+	s = s[:len(s)-1]
```
miekg, line 279: isn't this an allocation we can do without?

defaults.go
tmthrgd: The compiler does exactly this optimisation by default. No need to be explicit.

defaults.go
tmthrgd: No allocation, just changing the string length. I left the benchmark out, but this function doesn't allocate at all.

lobeck:
I'm sorry, but I can't provide traces for the moment. I've let it run in one of the environments and didn't see much of a change. But we've found the root cause for the load, so currently we're busy addressing this first.

defaults.go
```diff
@@ -272,18 +272,24 @@ func IsMsg(buf []byte) error {
 
 // IsFqdn checks if a domain name is fully qualified.
 func IsFqdn(s string) bool {
-	s2 := strings.TrimSuffix(s, ".")
-	if s == s2 {
+	// Check for (and remove) a trailing dot, returning if there isn't one.
```
gibson042, line 275:
```suggestion
	// Check for (and remove) a trailing dot, rejecting if there isn't one.
```

defaults.go
```diff
@@ -272,18 +272,24 @@ func IsMsg(buf []byte) error {
...
+	if s == "" || s[len(s)-1] != '.' {
 		return false
 	}
+	s = s[:len(s)-1]
 
-	i := strings.LastIndexFunc(s2, func(r rune) bool {
+	// If we don't have an escape sequence before the final dot, we know it's
+	// fully qualified and can return here.
+	if s == "" || s[len(s)-1] != '\\' {
```
gibson042, line 283: It might be faster still if a new slice is only constructed in the slow path.
```go
// IsFqdn checks if a domain name is fully qualified
// (specifically, if it ends with an unescaped dot).
func IsFqdn(s string) bool {
	// If there's no trailing dot, the input is definitely not fully qualified.
	if s == "" || s[len(s)-1] != '.' {
		return false
	}

	// If there's no backslash before the final dot,
	// it's definitely not escaped so assume the input is fully qualified.
	if len(s) == 1 || s[len(s)-2] != '\\' {
		return true
	}

	// Otherwise the dot is escaped iff the count of backslashes before the
	// one we already found is even (including if there are none).
	prefix := s[:len(s)-2]
	i := strings.LastIndexFunc(prefix, func(r rune) bool {
		return r != '\\'
	})
	return (len(prefix)-i)%2 == 0
}
```

tmthrgd: It actually ends up slower on my system:
```
name                 old time/op    new time/op    delta
IsFQDN/no_dot-12       1.48ns ± 3%    1.54ns ± 1%  +3.90%  (p=0.000 n=10+9)
IsFQDN/unescaped-12    1.57ns ± 1%    1.66ns ± 3%  +5.37%  (p=0.000 n=8+9)
IsFQDN/escaped-12      23.8ns ± 2%    24.4ns ± 2%  +2.58%  (p=0.000 n=10+10)
FQDN/is_fqdn-12        1.80ns ± 2%    1.87ns ± 3%  +4.09%  (p=0.000 n=10+9)
FQDN/not_fqdn-12       33.4ns ±12%    32.5ns ±13%    ~     (p=0.315 n=10+10)
```

defaults.go
tmthrgd: I think return is clearer, we're not validating or otherwise approving the input, only answering a question regarding it.

defaults.go
gibson042: Ok. My other thought is special-casing `if len(s) <= 1 || s[len(s)-1] != '.' { return s == "." }` to avoid a comparison in the nontrivial-dot-terminated-string case, but it's probably best to avoid drowning in micro-optimization possibilities.

miekg:
think this can go in regardless? Perf improvements look impressive enough

tmthrgd:
> think this can go in regardless? Perf improvements look impressive enough

Agreed. Won't be ground breaking, but definitely a nice improvement.

PR #1454, opened 2023-04-18
Remove SingleInflight support from Client

### LINKED ISSUE #1449: singleInFlight can cause all future DNS requests to fail
Currently singleInFlight only takes into account the message and not the server address. I see the appeal because servers should return the same answer so why not group them. However, if a server is down then this has the unintentional consequence of causing all future DNS requests to fail.

<details>
  <summary>Contrived example code</summary>

```
package main

import (
	"log"
	"time"

	"github.com/miekg/dns"
)

func main() {
	server := &dns.Server{
		Addr: ":9999",
		Net:  "udp",
		Handler: dns.HandlerFunc(func(w dns.ResponseWriter, r *dns.Msg) {
			m := new(dns.Msg)
			m.SetRcode(r, dns.RcodeSuccess)
			w.WriteMsg(m)
		}),
	}

	cfg := dns.ClientConfig{
		Servers: []string{
			"127.0.0.1:9999",
			"127.0.0.1:100",
		},
		Timeout: 1,
	}

	// let's pretend that the first server is temporarily down for an upgrade or
	// restart by starting the ListenAndServer purposefully late...

	newMsg := func() *dns.Msg {
		m := new(dns.Msg)
		m.SetQuestion("test.test.", dns.TypeA)
		return m
	}

	c := new(dns.Client)
	c.SingleInflight = true
	timeout := time.Duration(cfg.Timeout) * time.Second
	c.DialTimeout = timeout
	c.ReadTimeout = timeout
	c.WriteTimeout = timeout

	query := func() {
		for _, srv := range cfg.Servers {
			_, _, err := c.Exchange(newMsg(), srv)
			if err == nil {
				log.Printf("success from %v\n", srv)
				return
			}
			log.Printf("error from %v: %v\n", srv, err)
		}
	}

	go query()

	go func() {
		panic(server.ListenAndServe())
	}()

	for range time.Tick(500 * time.Millisecond) {
		go query()
	}
}
```
</details>

When running the above code you'll see:
```
2023/04/08 21:14:32 error from 127.0.0.1:9999: read udp 127.0.0.1:52618->127.0.0.1:9999: i/o timeout
2023/04/08 21:14:32 error from 127.0.0.1:9999: read udp 127.0.0.1:52618->127.0.0.1:9999: i/o timeout
2023/04/08 21:14:32 error from 127.0.0.1:9999: read udp 127.0.0.1:52618->127.0.0.1:9999: i/o timeout
2023/04/08 21:14:33 error from 127.0.0.1:100: read udp 127.0.0.1:52621->127.0.0.1:100: i/o timeout
2023/04/08 21:14:33 error from 127.0.0.1:100: read udp 127.0.0.1:52621->127.0.0.1:100: i/o timeout
2023/04/08 21:14:33 error from 127.0.0.1:9999: read udp 127.0.0.1:52621->127.0.0.1:100: i/o timeout
2023/04/08 21:14:33 error from 127.0.0.1:100: read udp 127.0.0.1:52621->127.0.0.1:100: i/o timeout
2023/04/08 21:14:33 error from 127.0.0.1:9999: read udp 127.0.0.1:52621->127.0.0.1:100: i/o timeout
2023/04/08 21:14:34 error from 127.0.0.1:9999: read udp 127.0.0.1:60130->127.0.0.1:100: i/o timeout
2023/04/08 21:14:34 error from 127.0.0.1:100: read udp 127.0.0.1:60130->127.0.0.1:100: i/o timeout
2023/04/08 21:14:34 error from 127.0.0.1:100: read udp 127.0.0.1:60130->127.0.0.1:100: i/o timeout
2023/04/08 21:14:34 error from 127.0.0.1:9999: read udp 127.0.0.1:60130->127.0.0.1:100: i/o timeout
2023/04/08 21:14:35 error from 127.0.0.1:9999: read udp 127.0.0.1:60134->127.0.0.1:100: i/o timeout
2023/04/08 21:14:35 error from 127.0.0.1:100: read udp 127.0.0.1:60134->127.0.0.1:100: i/o timeout
2023/04/08 21:14:35 error from 127.0.0.1:100: read udp 127.0.0.1:60134->127.0.0.1:100: i/o timeout
2023/04/08 21:14:35 error from 127.0.0.1:9999: read udp 127.0.0.1:60134->127.0.0.1:100: i/o timeout
```

Every query will forever fail despite only the second server being "down". This is because when it tries the first server a request to the second server is still timing out and then when it times out that goroutine tries the second server. The only way to get out of this is to wait for the second server to start working again or kill the program.

A simple fix would be to include the server's address in the group key. Alternatively, `exchangeWithConnContext` could check to see if the error returned from `Do` was for a different endpoint and immediately try again if so? But that could cause the `Exchange` function to take up to 2*Timeout which isn't ideal.

Finally, when `SingleInflight` is true, `Exchange` still calls `Dial` even if it will end up re-using an existing answer and just throwing away the connection without even using it. Any thoughts on tidying that up at the same time?

tmthrgd:
Callers should instead implement their own in flight query caching.

Closes #1449


client.go
```diff
@@ -56,14 +55,20 @@ type Client struct {
...
+	WriteTimeout time.Duration     // net.Conn.SetWriteTimeout value for connections, defaults to 2 seconds - overridden by Timeout when that value is non-zero
+	TsigSecret   map[string]string // secret(s) for Tsig map[<zonename>]<base64 secret>, zonename must be in canonical form (lowercase, fqdn, see RFC 4034 Section 6.2)
+	TsigProvider TsigProvider      // An implementation of the TsigProvider interface. If defined it replaces TsigSecret and is used for all TSIG operations.
+
+	// SingleInflight previously serialised multiple concurrent queries for the
+	// same Qname, Qtype and Qclass to ensure only one would be in flight at a
+	// time.
+	//
+	// Deprecated: This is a no-op. Callers should implement their own in flight
+	// query caching.
```
miekg, line 70: *maybe* link to the issue for some more background on why?

tmthrgd: Will go doc properly linkify it in the form I've added? Do you know?

miekg: lol, we have both forms already:
~~~
reverse.go:	// Preserve previous NOTIMP typo, see github.com/miekg/dns/issues/733.
dns_test.go:	tkey.Hdr.Class = ClassINET // https://github.com/miekg/dns/issues/577
~~~
except these are not part of the public docs.

I thought you can make pkg.go.dev look at a specific commit but my go gets fail with unknown revision, either way, folks can find the issue regardless of this being linkified

jsha:
It's worth noting that if someone was relying on this for birthday attack prevention (https://datatracker.ietf.org/doc/html/rfc5452#section-5), they will no longer be protected, starting from v1.1.54, unless they implement their own protection. Probably instead of "Callers should implement their own in flight query caching if needed" the comment should warn about birthday attacks and encourage callers more strongly to implement their own.

PR #1456, opened 2023-04-27
Allow RRSIG algorithm mnemonics

miekg:
Java outputs these *and* the RFC says we should parse them, so parse
them. We'll never output them though. Throwback to the "be lenient to
what you accept, but strict with what you output". Anyhow the diff is
tiny and it helps interop.

Fixes: #1447

### DIFF
--- a/parse_test.go
+++ b/parse_test.go
@@ -1975,3 +1975,22 @@ func TestParseOPENPGPKEY(t *testing.T) {
 		}
 	}
 }
+
+func TestParseRRSIGAlgNames(t *testing.T) {
+	tests := map[string]bool{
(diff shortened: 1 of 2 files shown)


parse_test.go
```diff
@@ -1975,3 +1975,22 @@ func TestParseOPENPGPKEY(t *testing.T) {
 		}
 	}
 }
+
+func TestParseRRSIGAlgNames(t *testing.T) {
+	tests := map[string]bool{
+		`miek.nl.  IN RRSIG SOA RSASHA1 2 43200 20140210031301 20140111031301 12051 miek.nl. MVZUyrYwq0iZhMFDDnVXD2Bvu7pcBoc=`:         true,
```
evansrg, line 1981: Check the parsed values match expected values perhaps?

PR #1457, opened 2023-05-02
Use correct string format IPv4 encoded as IPv6 for for AAAA records

Chreo:
In order to parse an IPv4 address encoded as IPv6, it needs to have the prefix of "::ffff:", i.e. ::ffff:192.0.2.1

This allows export of parsed IPv4 encoded IPv6 AAAA records to be scanned back to a valid AAAA record.
Without it the following code fails as the address is "plain" dotted IPv4-format:
```
rr, err := dns.NewRR(". 1 IN AAAA ::ffff:192.0.2.1")
if err != nil {
    log.Fatalf("parse failure: %w", err)
}

_, err = dns.NewRR(rr.String())
if err != nil {
    log.Fatalf("parse failure: %w", err)
}
```

This comes from the std libs net.IP String() function which outputs IPv4-in-IPV6 as standard dotted-IPv4. net/netip.Addr type, in contrast, differs here in that the String() func outputs it with the proper prefix "::ffff:" Ref: https://pkg.go.dev/net/netip#Addr.String

The other attempt to fix this https://github.com/miekg/dns/issues/1107 and the code in PR https://github.com/miekg/dns/pull/1108 has worse performance than this PR, due to the use of fmt.Sprintf() which is not needed

### DIFF
--- a/parse_test.go
+++ b/parse_test.go
@@ -1686,6 +1686,33 @@ func TestNULLRecord(t *testing.T) {
 		t.Fatalf("Expected packet to contain NULL record")
 	}
 }
+func TestAAAAParsing(t *testing.T) {
+	tests := []string{
+		"2001:db8::1",
+		"1::1",
+		"2001:db81:d2b4:b6ba:50db:49cc:a8d1:5bb1",
+		"::ffff:192.0.2.0",
+	}
+
+	rrPrefix := ".\t1\tIN\tAAAA\t"
+
+	for num, tc := range tests {
+		t.Run(fmt.Sprintf("Test %d", num), func(t *testing.T) {
+			rr, err := NewRR(rrPrefix + tc)
+			if err != nil {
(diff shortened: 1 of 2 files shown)


miekg:
I'm sympathetic to merging this as anything we output should be parsed back in.

Can you add a (small) test for this as well?

Chreo:
Thanks and sorry for the late response. I've added a test with a few IPv6 addresses that should parse to a proper AAAA record, be able to export in presentation format and the result should be able to parsed again into a new AAAA record.
It may be worth noting that any viable net.IP address string is able to be used to parse into an AAAA record, including "::".

janik-cloudflare:
Hi @miekg, just wondering if you've had a chance to look at the updated version yet. We'd love to get this pulled in!

miekg (approved):

PR #1458, opened 2023-05-08
Expose number of active TCP connections

cesarkuroiwa:
New func (*Server) ActiveConns to expose to the application the number of active TCP connections in use at any given moment

### DIFF
--- a/server.go
+++ b/server.go
@@ -429,6 +429,11 @@ func (srv *Server) ShutdownContext(ctx context.Context) error {
 	return ctxErr
 }
 
+// ActiveConns returns the number for active TCP connections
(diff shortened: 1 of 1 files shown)


miekg:
I don't think this should be added.

There are other ways to extract this (and much more data) from the OS level.

This only exposed a tiny bit of information.

Coredns shows how to do this without any changes to this library.

/Miek

cesarkuroiwa:
> I don't think this should be added. There are other ways to extract this (and much more data) from the OS level. This only exposed a tiny bit of information. Coredns shows how to do this without any changes to this library. /Miek

Where can I find it?

PR #1459, opened 2023-05-10
Make ExchangeWithConnContext public

andrew-d:
If we want to use a custom mechanism of obtaining a Conn that doesn't match the net.Dialer type, but retain the timeout behaviour from ExchangeContext, there was previously no way to accomplish this.

This PR makes the underlying ExchangeWithConnContext function public, which allows this behaviour.

Now that #1454 is merged, there is no longer any interaction between the context provided and the singleflight behaviour, so I removed the comment from ExchangeWithConn.

### DIFF
--- a/client.go
+++ b/client.go
@@ -183,14 +183,13 @@ func (c *Client) Exchange(m *Msg, address string) (r *Msg, rtt time.Duration, er
 // This allows users of the library to implement their own connection management,
 // as opposed to Exchange, which will always use new connections and incur the added overhead
 // that entails when using "tcp" and especially "tcp-tls" clients.
-//
-// When the singleflight is set for this client the context is _not_ forwarded to the (shared) exchange, to
-// prevent one cancellation from canceling all outstanding requests.
 func (c *Client) ExchangeWithConn(m *Msg, conn *Conn) (r *Msg, rtt time.Duration, err error) {
-	return c.exchangeWithConnContext(context.Background(), m, conn)
+	return c.ExchangeWithConnContext(context.Background(), m, conn)
 }
 
-func (c *Client) exchangeWithConnContext(ctx context.Context, m *Msg, co *Conn) (r *Msg, rtt time.Duration, err error) {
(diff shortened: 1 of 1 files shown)


andrew-d:
@miekg @tmthrgd gentle ping on this; anything I can do to help move this forward? Happy to e.g. write additional tests.

miekg:
> If we want to use a custom mechanism of obtaining a Conn that doesn't match the net.Dialer type, but retain the timeout behaviour from ExchangeContext, there was previously no way to accomplish this.

I don't understand... you mean `doesn't match the net.Conn interface` or some such? As `exchangeWithConnContext` has the identical paramaters (apart from the context) I don't see how this would help

andrew-d:
> > If we want to use a custom mechanism of obtaining a Conn that doesn't match the net.Dialer type, but retain the timeout behaviour from ExchangeContext, there was previously no way to accomplish this.
> 
> I don't understand... you mean `doesn't match the net.Conn interface` or some such? As `exchangeWithConnContext` has the identical paramaters (apart from the context) I don't see how this would help

Concretely: if I need to do something to obtain a `*dns.Conn` other than using a `*net.Dialer` (through `Client.Dialer`), there's no way to both do that and add a timeout to the request. I can do something like:

```go
netConn, err := myDialFunc(/* args */)
// check err
conn := &dns.Conn{
	Conn: netConn,
}

var c dns.Client
resp, _, err := c.ExchangeWithConn(msg, conn)
```

However, there's no way to add a timeout to the `ExchangeWithConn` function. `ExchangeContext` doesn't let me provide a `*dns.Conn`, and the `Dialer` field of `Client` is of concrete type `*net.Dialer` so I can't override it with my custom dial logic.

Exposing `ExchangeWithConnContext` would allow me to pass a `context.Context` with an appropriate deadline and thus add a timeout to the query.

miekg:
Ah I see.
Also for symmetry it would be nice to have both those *Context function to
be public.

Cheers

andrew-d:
Thank you!

PR #1464, opened 2023-06-05
Use ContextDialer interface instead of concrete net.Dialer

ydnar:
This PR changes the concrete `net.Dialer` in `dns.Client` to a `ContextDialer` interface, which allows the caller to use an alternative `net` package or method to open sockets, such as Wasm or `wasip1` with extensions on Go 1.21.

`ContextDialer` is a simple one-method interface:

```go
// ContextDialer is any type that implements DialContext, like net.Dialer.
type ContextDialer interface {
	DialContext(ctx context.Context, network, address string) (net.Conn, error)
}
```

A couple caveats:

1. This somewhat duplicates functionality in `(*Client).ExchangeWithConn`. Desirable?
1. If `TLS` is desired and `Client.Dialer` is set to a `ContextDialer` that’s not a concrete `net.Dialer`, then the `tls.Dialer` created in `(*Client).DialContext` will default to the zero value. However, the caller could implement TLS support with a custom dialer by setting `Client.Dialer` to a `tls.Dialer`.

### DIFF
--- a/client.go
+++ b/client.go
@@ -51,6 +51,8 @@ type ContextDialer interface {
 	DialContext(ctx context.Context, network, address string) (net.Conn, error)
 }
 
+var _ ContextDialer = (*net.Dialer)(nil)
+
 // A Client defines parameters for a DNS client.
 type Client struct {
 	Net       string        // if "tcp" or "tcp-tls" (DNS over TLS) a TCP query will be initiated, otherwise an UDP one (default is "" for UDP)


miekg:
can't see how this change is backwards compatible?

ydnar:
> can't see how this change is backwards compatible?

It's backwards compatible if and only if the caller assigns, but does not read Dialer.

I could search open-source dependents for any instances of a caller dereferencing Dialer.

What compatibility guarantee(s) for this package exist?

PR #1465, opened 2023-06-07
Create types for Overflow errors

SriHarsha001:
The upstream message helper /msg_helpers.go that return overflow errors was not using any of the exported defined errors from the package - https://github.com/miekg/dns/blob/master/msg.go#L51-L71

This came up in PR review here - https://github.com/coredns/coredns/pull/6003/files#r1154665103
"To accurately identify the error by type rather than string compares"

So in this PR, created types for Overflow errors (msg.go) and used them in (msg_helpers.go).


miekg:
as said in the issue, error *types* should have been used in the beginning.

this pr is complete overkill

PR #1466, opened 2023-06-13
fix: CVE-2022-41723 CVE-2022-41717

testwill:


miekg:
any downstream consumer needs to do this by themself

See https://github.com/miekg/dns/pull/1049#issuecomment-565718842

PR #1469, opened 2023-06-16
Codeql api version bump v2

bduggan-ns1:
Codeql api version bump v2


bduggan-ns1:
aborting. This change will cause unintended changes

Issue #1434, opened 2023-03-12
CanonicalName case folding of non-US-ASCII letters


zeroizerz:
https://www.rfc-editor.org/rfc/rfc4034#section-6.2:

> all uppercase US-ASCII letters in the owner name of the RR are replaced by the corresponding lowercase US-ASCII letters

Currently the CanonicalName function is implemented using strings.ToLower, which lowercases both US-ASCII and non-US-ASCII letters. But only US-ASCII letters should be lowercased according to the RFC.

This is further corroborated by https://www.rfc-editor.org/rfc/rfc4343#section-3:

> Note that the DNS case insensitivity does NOT correspond to the case folding specified in [ISO-8859-1] or [ISO-8859-2].  For example, the octets 0xDD (\221) and 0xFD (\253) do NOT match, although in other contexts, where they are interpreted as the upper- and lower-case version of "Y" with an acute accent, they might.

Accordingly, `dns.CanonicalName("ÉxamplE.com")` should output `Éxample.com.`, but currently outputs `éxample.com.`.

PR #1470, opened 2023-07-02
CanonicalName should casefold non-US-ASCII chars

retornam:
According to Section 6.2 [1] of RFC 4034,all uppercase US-ASCII letters in the owner name of the RR are replaced by the corresponding lowercase US-ASCII letters.

This updates CanonicalName to conform to the RFC.

[1] https://www.rfc-editor.org/rfc/rfc4034#section-6.2

Fixes miekg/dns#1434

### DIFF
--- a/defaults.go
+++ b/defaults.go
@@ -5,6 +5,7 @@ import (
 	"net"
 	"strconv"
 	"strings"
+	"unicode"
 )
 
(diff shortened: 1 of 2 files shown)


miekg:
seems like the right thing to do.

PR #1475, opened 2023-07-20
Create an overflow error type that implements error, and raise the error for overflow cases

SriHarsha001:
The upstream message helper /msg_helpers.go that return overflow errors was not using any of the exported defined errors from the package - https://github.com/miekg/dns/blob/master/msg.go#L51-L71

This came up in PR review here - https://github.com/coredns/coredns/pull/6003/files#r1154665103
"To accurately identify the error by type rather than string compares"

So in this PR, create an overflow error type that implements error (msg.go), and raise the error for Overflow errors (msg_helpers.go).

### DIFF
--- a/msg.go
+++ b/msg.go
@@ -72,14 +72,15 @@ var (
 )
 
 type OverflowError struct {
-	err string
+	err *Error
+}
+
+func (e *OverflowError) Unwrap() error {
+	return e.err
 }
 
 func (e *OverflowError) Error() string {
-	if e == nil {
-		return "dns: <nil>"
-	}
-	return "dns: " + e.err
+	return e.err.Error()
 }
 
 // Id by default returns a 16-bit random number to be used as a message id. The
--- a/msg.go
+++ b/msg.go
@@ -71,6 +71,17 @@ var (
 	ErrTime          error = &Error{err: "bad time"}      // ErrTime indicates a timing error in TSIG authentication.
 )
 
+type OverflowError struct {
+	err string
+}
+
+func (e *OverflowError) Error() string {
+	if e == nil {
+		return "dns: <nil>"
+	}
+	return "dns: " + e.err
+}
+
 // Id by default returns a 16-bit random number to be used as a message id. The
 // number is drawn from a cryptographically secure random number generator.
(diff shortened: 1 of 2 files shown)


msg.go
```diff
@@ -71,6 +71,17 @@ var (
 	ErrTime          error = &Error{err: "bad time"}      // ErrTime indicates a timing error in TSIG authentication.
 )
 
+type OverflowError struct {
+	err string
```
jameshartig, line 75: I think this should be *Error and then OverflowError should implement the Unwrap method. Otherwise all of the errors you changed will be a breaking change since they won't be *Error anymore.

SriHarsha001: @jameshartig - I have implemented the changes. Could you please take another look ?

SriHarsha001: @tmthrgd - Would you mind reviewing this PR please ?

SriHarsha001: As far as I know, I have addressed all feedback and this is ready for review.

SriHarsha001: Hi Team, I kindly request a review on this PR. This is blocking another PR for coreDNS.

SriHarsha001: @tmthrgd - I kindly request you to review this PR.

SriHarsha001: @jameshartig @tmthrgd - I apologize for multiple follow up. I kindly request you to review this PR when you get a moment.

SriHarsha001: Hi Team, please review this PR when you get a chance.

tmthrgd (commented):
I don't think changing the returned error type is backward compatible. Nor do I really like having a type that just wraps another error like that.

What we could do though is add a sentinel error `var ErrOverflow = errors.New("...")` and then implement an `Is(error) bool` method on `*Error`.

It can either use `strings.HasPrefix` or some kind of unexpired field to determine whether it's an overflow error.

miekg:
Having type errors would have been a great idea, but that notion only came later in the Go eco system.

Checking for the string 'overflow' seems just as robust as checking for a type - and as said, we know this is part of the API

Also this seems incomplete because bad rdlength errors are _not_ overflow errors

PR #1479, opened 2023-09-12
Use proper section names when producing dig-like output for UPDATEs

dnschecktool:
When `Opcode` == `OpcodeUpdate`, `String()` should use section names as described in [RFC 2136 Section 2](https://datatracker.ietf.org/doc/html/rfc2136#section-2):

Query -> Zone
Answer -> Prerequisite
Authority -> Update


tmthrgd:
That's annoying, but I suppose if this is the correct output we should do it.

tmthrgd (approved):

Issue #1421, opened 2023-02-16
Structured DNS errors draft


miekg:
https://datatracker.ietf.org/doc/draft-ietf-dnsop-structured-dns-error/

looks worthwhile and destined to become an RFC (i think).

miekg:
and RFC: https://www.rfc-editor.org/rfc/rfc8914.html

hope to have an PR somewhere today

miekg:
already done

PR #1481, opened 2023-09-15
EDE is already implemented

miekg:
This adds the RFC to the README.md.

Closes: #1421

    will not be merged.


miekg:
and even that was already done *And* in the correct place --> more coffee

PR #1484, opened 2023-09-20
Make SetReply a Reflexive Operation

askurydzin:
Allow SetReply to be called against the request message, ie:

```
r.SetRcode(r, dns.RcodeNameError)
w.WriteMsg(r)
```

If not, r.SetRcode(r, dns.RcodeNameError) will reset Question section to empty.


tmthrgd (approved):
This seems sensible to me.

Issue #1460, opened 2023-05-16
`(*dns.OPT).SetDo` ignores several arguments


EugeneOne1:
The documentation of [`(*dns.OPT).SetDo`](https://pkg.go.dev/github.com/miekg/dns#OPT.SetDo) states:

 >  It is possible to pass 2 or more arguments. Any arguments after the 1st is silently ignored.

However, passing 2 or more arguments (as well as passing no arguments) sets the DO bit to `true` regardless of the 1st argument's value.  I believe, this is due to incorrect conditional statement:

https://github.com/miekg/dns/blob/730c265d18c3f51b447f6a4c61fe3cde336585d8/edns.go#L190

Which should actually check if the length of slice **greater or** equal to `1`.

Alternatively, the documentation should state that all the arguments are ignored in case there are 2 or more of them.

EugeneOne1:
@miekg, sorry if I've misunderstood, but is this no longer an issue?  I can't find any changes to the documentation in the last 7 years.

PR #1488, opened 2023-10-07
The small `(*dns.OPT).SetDo` documentation tweak

EugeneOne1:
As I identified in #1460, there is a mistake in the documentation appeared, that I would like to address through the following pull request.  I've tried to change as few symbols, as possible, to keep the documentation comprehensive and concise.

Best regards, @EugeneOne1.


miekg:
lgtm. It's more to the point. 

thanks

PR #1491, opened 2023-10-16
Remove comment about checking zero bit.

tangrs:
The check was removed in https://github.com/miekg/dns/issues/975.


tmthrgd (approved):
Good catch. LGTM.

PR #1500, opened 2023-11-05
Expose destination ip to ResponseWriter

### LINKED ISSUE #1411: Obtain the public interface the client is connected to
Is there a way I could get the public interface (my server has and uses multiple) that was used for the query? I have this handler but all I get printed out from it (as the server ip) is `::53`:

```
func (h *Handler) ServeDNS(w dns.ResponseWriter, req *dns.Msg) {
	h.server.logger.Debugf("client ip: %v", w.RemoteAddr())
	h.server.logger.Debugf("server ip: %v", w.LocalAddr())

	...
}
```

I noticed that `PacketConn` is being set with `FlagDst` and `FlagInterface` so I'm out of ideas as to why nothing comes out for the local address that could help me. Any ideas?

hamidrezakp:
This PR will add a `DestinationIP` field to `ResponseWriter` interface.
It will allow downstream applications to have the ability to make decisions based on the destination of the request; Which is useful in situations that we have DNS server listening on more than one IP address.

Fixes #1411

PS: I'm new to go, so if there is somethings that are done badly, please let me know. Thanks.

### DIFF
--- a/server.go
+++ b/server.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"io"
 	"net"
+	"net/netip"
 	"strings"
 	"sync"
 	"time"
@@ -44,6 +45,8 @@ type ResponseWriter interface {
 	LocalAddr() net.Addr
 	// RemoteAddr returns the net.Addr of the client that sent the current request.
 	RemoteAddr() net.Addr
+	// DestinationIP returns the net.IP that client is connected to.
+	DestinationIP() net.IP
 	// WriteMsg writes a reply back to the client.
 	WriteMsg(*Msg) error
 	// Write writes a raw buffer back to the client.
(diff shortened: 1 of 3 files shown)


hamidrezakp:
> I don't think this is going to fly - if the only demonstrated use case is a log line.


Consider this situation:
Server listens to `10.0.0.1` and `192.0.0.1` and i want to respond to each one differently. currently i have no way to access the destination ip in `udp` mode.

> Maybe a `LocalAddr()` override could be made for the various protocols, one that takes oob into consideration for UDP, and even then I wonder if that would help.

Well i think we can use `LocalAddr()` for this purpose too, but it will be a breaking change imo.

miekg:
yeah, `LocalAddr` is also wrong too. Anyway apart from consuming net.Conns not much else should be added to this lib. I think if anything `sessionUDP`'s member could be public. 

otoh if you listen on the on all interfaces (`::53`) this is what you get with udp

hamidrezakp:
> yeah, `LocalAddr` is also wrong too. Anyway apart from consuming net.Conns not much else should be added to this lib.

I couldn't find a good way to take `net.IP` out of `net.Addr`, so i went with `netip` package.
If there is a way to do that, i would be so happy to remove unnecessary imports.




> otoh if you listen on the on all interfaces (`::53`) this is what you get with udp

Exactly and this is the main reason of this PR.

miekg:
I'm not sure what a good fix would look like, preferable a lot smaller than this PR,

also as said before the only use case that came up here is a log line, I can still see a broader need, but don't feel this pr is the correct way to solve it

PR #1507, opened 2023-11-06
Use golang.org/x/crypto/cryptobyte to unpack DNS records

tmthrgd:
This substantially simplifies the machinery and state tracking around DNS record unpacking. Gone are the `off int` arguments! Gone are the manual slice length checks!

For the most part care has been taken to maintain existing behaviour, including quirks, with one of the few notable intentional change being to `UnpackDomainName` where we now strictly enforce that pointers point backwards into the message. We also fix a bug here or there (see `unpackQuestion`) and are more strict in our unpacking of certain EDNS0 options.


msg_helpers.go
```diff
@@ -8,19 +8,35 @@ import (
 	"net"
 	"sort"
 	"strings"
+
+	"golang.org/x/crypto/cryptobyte"
 )
 
+type dnsString struct {
```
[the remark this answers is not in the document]
tmthrgd: It actually ends up way nicer if I get rid of `dnsString` entirely so I'll update this PR.

msg.go
```diff
@@ -48,27 +48,29 @@ const (
...
+	ErrPrivKey              error = &Error{err: "bad private key"}
+	ErrRcode                error = &Error{err: "bad rcode"}
+	ErrRdata                error = &Error{err: "bad rdata"}
+	ErrRRset                error = &Error{err: "bad rrset"}
+	ErrSecret               error = &Error{err: "no secrets defined"}
+	ErrShortRead            error = &Error{err: "short read"}
+	ErrSig                  error = &Error{err: "bad signature"} // ErrSig indicates that a signature can not be cryptographically validated.
+	ErrSoa                  error = &Error{err: "no SOA"}        // ErrSOA indicates that no SOA RR was seen when doing zone transfers.
+	ErrTime                 error = &Error{err: "bad time"}      // ErrTime indicates a timing error in TSIG authentication.
+	errUnpackOverflow             = &Error{err: "overflow unpacking integer"}
```
tmthrgd, line 72: @miekg We're somewhat inconsistent in what errors we return. This is most similar to what we currently do, but realistically this error and `ErrBuf` have the same meaning. Are we better just using that one?

tmthrgd:
> holy moly this cleans stuff up.

You know I thought it would, hence the PR, but even I was surprised just how much it improves things.

msg.go
```diff
@@ -603,71 +605,91 @@ func packRR(rr RR, msg []byte, off int, compression compressionMap, compress boo
...
+	if off < 0 || off > len(msg) {
+		return nil, off, &Error{err: "bad off"}
 	}
 
-	return UnpackRRWithHeader(h, msg, off)
+	s := cryptobyte.String(msg[off:])
+	rr, err = unpackRR(&s, msg)
+	if err != nil {
+		// Keep existing behaviour of returning len(msg) here on error.
+		return rr, len(msg), err
```
tmthrgd, line 616: @miekg Do we want to keep this behaviour? It's not actually documented here unlike on `UnpackDomainName`.

msg.go
```diff
@@ -603,71 +605,91 @@ func packRR(rr RR, msg []byte, off int, compression compressionMap, compress boo
...
 func UnpackRRWithHeader(h RR_Header, msg []byte, off int) (rr RR, off1 int, err error) {
+	if off < 0 || off > len(msg) {
+		return &h, off, &Error{err: "bad off"}
+	}
+
+	s := cryptobyte.String(msg[off:])
+	rr, err = unpackRRWithHeader(h, &s, msg)
+	if err != nil {
+		// Keep existing behaviour of returning len(msg) here on error.
+		return rr, len(msg), err
```
tmthrgd, line 632: @miekg Ditto here.

msg.go
miekg: yes, agree. Think ErrBuf is "better" here, more detail could help the caller, but in the end the unpack failed which is what anyone should care about

msg.go
miekg: make it return `off` here too? Seems more useful than `len(msg)` in any case.

svcb_test.go
```diff
@@ -70,10 +72,12 @@ func TestDecodeBadSVCB(t *testing.T) {
 			key:  SVCB_ALPN,
 			data: []byte{3, 0, 0}, // There aren't three octets after 3
 		},
-		{
+		// The caller is responsible for ensuring the buffer is empty after
+		// unpacking, see unpackDataSVCB.
+		/*{
```
miekg, line 77: why this change?

svcb.go
```diff
@@ -667,8 +661,9 @@ func (s *SVCBECHConfig) copy() SVCBKeyValue {
 	return &SVCBECHConfig{cloneSlice(s.ECH)}
 }
 
-func (s *SVCBECHConfig) unpack(b []byte) error {
-	s.ECH = cloneSlice(b)
+func (s *SVCBECHConfig) unpack(b *cryptobyte.String) error {
+	s.ECH = cloneSlice(*b)
+	b.Skip(len(*b))
```
miekg, line 666: slightly concerning that a b.Skip is needed here. Should there be a cloneSliceAndSkip function?

i.e. it could be nice if these become 1-liners? :
~~~ go
func (s *SVCBECHConfig) unpack(b *cryptobyte.String) error { s.ECH = clonseSliceAndSkip(*b); return nil}
~~~

svcb.go
```diff
@@ -531,11 +525,10 @@ func (*SVCBPort) len() int             { return 2 }
 func (s *SVCBPort) String() string     { return strconv.FormatUint(uint64(s.Port), 10) }
 func (s *SVCBPort) copy() SVCBKeyValue { return &SVCBPort{s.Port} }
 
-func (s *SVCBPort) unpack(b []byte) error {
-	if len(b) != 2 {
-		return errors.New("dns: svcbport: port length is not exactly 2 octets")
+func (s *SVCBPort) unpack(b *cryptobyte.String) error {
+	if !b.ReadUint16(&s.Port) {
+		return errUnpackOverflow
```
miekg, line 530: regarding this errors: I think the est. pattern in this lib is: "bad.. ", with maybe slightly more context than just 'bad SVCB'

miekg (commented):
Would be good to run the coredns testsuite with this version.

svcb_test.go
tmthrgd: That's because I moved the error condition out of `(*SVCB).unpack`, I've moved it back so I can restore this test case again.

tmthrgd:
> Would be good to run the coredns testsuite with this version.

Yep that's actually my hold up at the moment. I've been debugging and working through a few scattered CoreDNS test failures.

tmthrgd:
> regarding this errors: I think the est. pattern in this lib is: "bad.. ", with maybe slightly more context than just 'bad SVCB'

Yeah I've gone back and forth on the errors and ended up with something quite different. There may be compatibility concerns, but ultimately I think we need to be far more semantic with our unpacking errors. I've gone and distinguished between failures to unpack the message framing and failures to unpack the record data specifically. Once we're in a length-delimited section, it doesn't make sense to use the same error anymore as we're not dealing with potential truncation but rather just corrupt marshalling.

tmthrgd:
I've left quite a few TODOs and comments where I've discovered incorrect or problematic behaviour. I've successfully run the CoreDNS test suite against this change now as well. Figuring out why it was failing was quite difficult and led down the RFC rabbit hole and exposed some bugs in this library. I'll aim to address all the TODOs in future PRs.

client.go
```diff
@@ -291,22 +292,20 @@ func (co *Conn) ReadMsgHeader(hdr *Header) ([]byte, error) {
...
-	p = p[:n]
-	if hdr != nil {
-		dh, _, err := unpackMsgHdr(p, 0)
-		if err != nil {
-			return nil, err
-		}
-		*hdr = dh
+	s := cryptobyte.String(p)
+	if hdr != nil && !hdr.unpack(&s) {
+		panic("dns: internal error: failed to unpack message header")
```
miekg, line 305: i don't think we should introduce panics

I believe the prev. code was here to at least know something about the mangled packet, i.e the `*hdr = dh`

edns.go
```diff
@@ -214,9 +216,8 @@ type EDNS0 interface {
 	Option() uint16
 	// pack returns the bytes of the option data.
 	pack() ([]byte, error)
-	// unpack sets the data as found in the buffer. Is also sets
-	// the length of the slice as the length of the option data.
-	unpack([]byte) error
+	// unpack sets the data as found in the buffer.
+	unpack([]byte) bool
```
miekg, line 220: i don't know about this.... two things
1. usually booleans turn into something that have more state
2. regardless of how we error, they (in some cases) do case some useful hints 

And maybe (3): should we really attempt this in this PR?

dns_test.go
```diff
@@ -168,6 +170,16 @@ func TestNoRdataPack(t *testing.T) {
...
 
 func TestNoRdataUnpack(t *testing.T) {
+	// TODO(tmthrgd): This test isn't correct and we sometimes don't correctly
+	// pack the messages. It currently succeeds as we explicitly handle certain
+	// broken records in their unpack method, but we shouldn't be. See the TODO
+	// in msg_generate.go. In fact we don't actually produce RRs with an
+	// RDLENGTH of zero, but rather one that just happens to contain mostly
+	// nothing. Further this isn't even how we ourselves construct dynamic
+	// update messages, in update.go we use ANY for when producing an update
+	// message. This correctly produces an record with a type, class, TTL, etc.
```
miekg, line 180: OMFG, yes of course., I can't gather this from your comment (And I'm too lazy to look), but are you saying that `example.org IN A` is made via ANY and fiddled with to make into an A?

if the `type == ` line a few lines below is indeed proof that this was always hacky

msg.go
```diff
@@ -62,13 +65,24 @@ var (
...
 	ErrRRset         error = &Error{err: "bad rrset"}
 	ErrSecret        error = &Error{err: "no secrets defined"}
 	ErrShortRead     error = &Error{err: "short read"}
 	ErrSig           error = &Error{err: "bad signature"} // ErrSig indicates that a signature can not be cryptographically validated.
 	ErrSoa           error = &Error{err: "no SOA"}        // ErrSOA indicates that no SOA RR was seen when doing zone transfers.
 	ErrTime          error = &Error{err: "bad time"}      // ErrTime indicates a timing error in TSIG authentication.
+
+	// TODO(tmthrgd): Stop returning and deprecate ErrRdata. It's not well
+	// defined.
+	ErrRdata error = &Error{err: "bad rdata"}
```
miekg, line 77: agrreed

msg.go
```diff
@@ -62,13 +65,24 @@ var (
...
+
+	// For CoreDNS, we need to use an error that either matches
+	// errors.Is(_, ErrBuf) or contains the substring "overflow".
+	//
+	// TODO(tmthrgd): Update CoreDNS and use "truncated message".
+	errTruncatedMessage = &Error{err: "overflow unpacking truncated message"}
```
miekg, line 83: a properly truncated message should unpack just fine.
unsure why TC bit messages deserve special treatment?

msg.go
```diff
@@ -62,13 +65,24 @@ var (
...
+	errUnpackOverflow   = &Error{err: "overflow unpacking data"}
```
miekg, line 84: data -> rdata?

msg.go
```diff
@@ -62,13 +65,24 @@ var (
...
+	errTrailingRData    = &Error{err: "trailing record data"}
```
miekg, line 85: also data -> rdata?

msg.go
```diff
@@ -360,94 +376,116 @@ func isRootLabel(s string, bs []byte, off, end int) bool {
...
+			if ptrs == 0 {
+				*s = cs
 			}
-			if ptr++; ptr > maxCompressionPointers {
-				return "", lenmsg, &Error{err: "too many compression pointers"}
+			// Don't follow too many pointers in case there is a loop.
+			//
+			// TODO(tmthrgd): This check is no longer required to ensure
+			// loop free behaviour. Should we remove it?
+			if ptrs++; ptrs > maxCompressionPointers {
```
miekg, line 467: technically there could still be a lot of pointers backwards? `maxCompressionsPointers` now has a value of 126; that could still be reach, although not via a loop.

Think it can go, even with a mesg with a lot of pointers backwards, we should not introduce an arbitrary limit

msg_test.go
```diff
@@ -122,41 +123,39 @@ func TestUnPackExtendedRcode(t *testing.T) {
 }
 
 func TestUnpackDomainName(t *testing.T) {
+	// manyPointers unpacks perfectly validly to "example.com." after following
+	// 200 pointers. This trips our maxCompressionPointers check as we refuse
```
miekg, line 127: yeah, see prev. comment, who are we to impose limits like this.

client.go
tmthrgd: So this panic actually can't ever happen (hence the "internal error" part). `Header.unpack` only ever returns false if p is shorter than 12 bytes, which is *exactly* what we check above with `if len(p) < headerSize`. I can document that.

dns_test.go
tmthrgd: Yeah basically that. `update.go` uses the `ANY` struct as an explicitly generic RR that has no RDATA. So that already works today, but we also try to handle the zero value of the different RRs, and that doesn't actually work.

msg.go
tmthrgd: Yeah it's a weird one. Pointing to another pointer would never be generated by a sensible non-buggy implementation. It doesn't even save you a byte. There's annoyingly no RFC that prohibits it though, so it is very much a judgement call. The backward check prevents a loop so we can remove this, it just opens the door to being forced to do more work by an attacker. I can remove it in a follow up PR.

msg_test.go
tmthrgd: Will remove in a future PR.

msg.go
tmthrgd: Done.

msg.go
tmthrgd: Well this one gets returned for both msg header, question headers, resource headers _and_ resource data. I'm not sure we can be more specific here.

msg.go
tmthrgd: This isn't for a properly framed messaged with or without the TC bit set, this is for a message that is say randomly cut in half. We return this if we're say unpacking a resource, but the message buffer stops half way through the RDATA. This already fails, but with a less obvious error message. I'm open to suggestions on a different message.

client.go
miekg: short comment can't hurt

dns_test.go
miekg: Ok. I'll take a look separately from this pr to see what can be done here.  It would be nice if the following can be used to make an empty rdata RR:
* NewRR
* &dns.A{}
* and/or setting rdlength to 0

It think it's currently _assumed_ this sorta works today

msg.go
miekg: pointer to self is then still checked?

miekg:
think this could go in and the fix some fallout and the dyn. update stuff?

(pointer to self should still be checked I guess, but can also be done in followup)

miekg:
I haven't heard from @tmthrgd in ages. closing this....

miekg:
I'm going to pick this up and use it in miekg/dnsv2 - from scratch rewrite, sans the tests. Will copy a lot from dns, but I'm starting from scratch

PR #1511, opened 2023-11-13
Swap closing order in `inAxfr` and `inIxfr`

noxer:
This PR swaps the order in which the connections and channels are closed in `inAxfr` and `inIxfr`. This makes the order more logical (consumers of the channel can be sure that the connection has been closed when the channel is closed).

The motivation for changing this is a heisenbug we had in our test pipeline. After closing the channel, the test code checked if the connection had been closed as well. Due to the nature of go routines, the test would randomly fail, depending on the scheduler. Swapping the order eliminates the problem, putting both deferred calls into the same function may even give (a very slight) performance improvement and it makes the order of operations more predictable and logical.

If you have any questions or suggestions I'm happy to incorporate them.

-- Tim

### DIFF
--- a/xfr.go
+++ b/xfr.go
@@ -81,6 +81,9 @@ func (t *Transfer) In(q *Msg, a string) (env chan *Envelope, err error) {
 func (t *Transfer) inAxfr(q *Msg, c chan *Envelope) {
 	first := true
 	defer func() {
+		// First close the connection, then the channel. This allows functions blocked on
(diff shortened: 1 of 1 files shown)


miekg:
thanks, can you add a comment in the code (at both places) where you briefly summarize why this is done in this (new) order? thanks!

noxer:
@miekg thank you for your feedback. I've added the comment.

PR #1526, opened 2024-01-09
Allow use of fs.FS for $INCLUDE and wrap errors

dpifke:
This adds ZoneParser.SetIncludeAllowedFS, to specify an fs.FS when enabling support for $INCLUDE, for reading included files from somewhere other than the local filesystem.

I've also modified ParseError to support wrapping another error, such as errors encountered while opening the $INCLUDE target.  This allows for much more robust handling, using errors.Is() instead of testing for particular strings (which may not be identical between fs.FS implementations).

ParseError was being constructed in a lot of places using positional instead of named members.  Updating ParseError initialization after the new member field was added makes this change seem a lot larger than it actually is.

The changes here should be completely backwards compatible.  The ParseError change should be invisible to anyone not trying to unwrap it, and ZoneParser will continue to use os.Open if the existing SetIncludeAllowed method is called instead of the new SetIncludeAllowedFS method.

### DIFF
--- a/scan.go
+++ b/scan.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"io/fs"
 	"os"
+	"path"
 	"path/filepath"
 	"strconv"
 	"strings"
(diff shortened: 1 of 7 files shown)


dpifke:
fs.FS was added in Go 1.21, which I think meets the "two latest major versions" criteria (latest being 1.22).  Let me know if you're concerned about keeping support for go 1.20 and below, and if so I can hide the new functionality behind a build tag.  (My plan to do so would be move SetIncludeAllowedFS into a separate file and shadow the fs.FS interface.)

Error wrapping was added in Go 1.13, so I don't anticipate any version concerns there.

dpifke:
(Deleted previous comment where I completely got wrong when fs.FS was introduced.  Sorry for the noise.)

fs.FS was introduced in Go 1.16: https://go.dev/doc/go1.16#fs

Error wrapping was introduced in Go 1.13: https://go.dev/doc/go1.13#error_wrapping

miekg:
brilliant idea! 

Code looks good, bit messy indeed due to the error changes. Some comments incomding

scan.go
```diff
@@ -220,6 +228,12 @@ func (zp *ZoneParser) SetIncludeAllowed(v bool) {
 	zp.includeAllowed = v
 }
 
+// SetIncludeAllowedFS controls whether $INCLUDE directives are allowed, and
+// specifies an [fs.FS] from which to open and read files.
+func (zp *ZoneParser) SetIncludeAllowedFS(v bool, fsys fs.FS) {
```
miekg, line 233: On the fence if we need a whole new method for this. Should we give SetIncludeAllowed an optional fs.FS parameter instead?

dpifke: Shipping separate methods as a non-breaking change, and then consolidating them if/when you release a v2 is an option.

miekg: ack. Think my main objection is the overlap in functionality. Maybe a better idea is to have a seperate method to set the underlying FS? `SetIncludeFS` that just sets the FS

Issue #1528, opened 2024-01-17
Trailing backslash results in non-FQDN targets


janik-cloudflare:
Calling `NewRR(". 1 IN CNAME example.com")` normally turns `example.com` into an FQDN, `example.com.`, before storing it in the RR's `Target` variable.

However, `NewRR(". 1 IN CNAME example.com\\")` with a trailing backslash results in `example.com\.`, which is not an FQDN since the last dot is escaped.

It seems to me like the parser should either add another dot in these cases, or (perhaps better) reject the trailing backslash with an error.

```
package main

import (
	"fmt"

	"github.com/miekg/dns"
)

func test(content string) {
	rr, err := dns.NewRR(content)
	if err != nil {
		panic(err)
	}

	target := rr.(*dns.CNAME).Target
	fmt.Printf("target %s is FQDN: %t\n", target, dns.IsFqdn(target))
}

func main() {
	test(`. 1 IN CNAME example.com`)
	test(`. 1 IN CNAME example.com\`)
}
```

```
$ go run ./...
target example.com. is FQDN: true
target example.com\. is FQDN: false
```

This may be related to #1384.

miekg:
but `\\` is legal, no?
There is def. a problem here, but if we are going to reject `\\` there is bound to be more stuff that will be deemed `illegal` in the future.
Is blindly `Fqdn`-ing the real problem here? Or is doing `Fqdn` in the first place the actual, actual problem?

janik-cloudflare:
An even number of slashes definitely seems legal.

Digging a bit, it looks like `CNAME.parse()` receives an `example.com\` token. Maybe `(*zlexer).Next()` should return an error when `escape` is still `true` when returning? I need to look more into how (and if) this would work.

FQDN-ing seems fine to me since `CNAME.parse()` calls `toAbsoluteName` based on the provided `origin`, right?

miekg:
(on my phone)

Returning an error when escape is true looks like a good thing to do.
Unsure if there is RFC text on that corner case.

gibson042:
It's not a corner case in the RFC; [CNAME RDATA format](https://datatracker.ietf.org/doc/html/rfc1035#section-3.3.1) is just a \<domain-name> and per [RFC 1035 section 5.1](https://datatracker.ietf.org/doc/html/rfc1035#section-5.1):
> The labels in the domain name are expressed as character strings and
separated by dots.  Quoting conventions allow arbitrary characters to be
[...]

PR #1532, opened 2024-01-21
IsDomainName: check for escape as last character

miekg:
Keep track if the escape, if still true when returning isDomainName
should return false.

TODO:
- Should still be done in packDomainName as well.
- And that should be tested
- Some tests now fail

There are multiple other places that supposedly also check for this, but
they are not called in the parsing.

Fixes: #1528

### DIFF
--- a/defaults.go
+++ b/defaults.go
@@ -198,10 +198,12 @@ func IsDomainName(s string) (labels int, ok bool) {
 		off    int
 		begin  int
 		wasDot bool
+		escape bool
 	)
 	for i := 0; i < len(s); i++ {
 		switch s[i] {
 		case '\\':
+			escape = !escape
 			if off+1 > lenmsg {
 				return labels, false
 			}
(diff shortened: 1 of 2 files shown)


miekg:
testing in my tree, yields:
~~~
--- FAIL: TestUpdate2NoRdataUnpack (0.00s)
    update2_test.go:72: failed to unpack RR with zero rdata: SOA: dns: overflow unpacking uint32
    update2_test.go:72: failed to unpack RR with zero rdata: TSIG: dns: overflow unpacking uint16
    update2_test.go:72: failed to unpack RR with zero rdata: TKEY: dns: overflow unpacking uint16
FAIL
~~~

janik-cloudflare (approved):
This is awesome, thank you so much!

janik-cloudflare:
I've applied this to our DNS API repository which has a fairly extensive test suite, and everything looks good!

janik-cloudflare:
> Should still be done in packDomainName as well.

Maybe also on the lexer level so that `. 1 IN TXT hello\` is also rejected? (I'd be happy to work on a PR for that, but it might take another two weeks or so.)

miekg:
yes, please do. I've merged this for now, but there is indeed some followup work to be done

PR #1533, opened 2024-01-31
Add option to do a zone transfer via TLS

cesarkuroiwa:
Add a new helper func InTLS to start a zone transfer using a TLS connection

### DIFF
--- a/xfr.go
+++ b/xfr.go
@@ -1,6 +1,7 @@
 package dns
 
 import (
+	"crypto/tls"
 	"fmt"
 	"time"
 )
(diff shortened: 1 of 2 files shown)


miekg:
[ Quoting ***@***.***> in "Re: [miekg/dns] New func InTLS: sta..." ]
>@cesarkuroiwa requested your review on: #1533 New func InTLS: start a zone
>transfer via TLS as a code owner.

Thanks, this is nice to have, but I feel this should not be a new function, but be folded
into the current In function. I.e. add a *tls.Config to *Transfer structure and then
rework In() a bit.

Regards,
Miek

cesarkuroiwa:
> [ Quoting ***@***.***> in "Re: [miekg/dns] New func InTLS: sta..." ]
> @cesarkuroiwa requested your review on: #1533 New func InTLS: start a zone transfer via TLS as a code owner.
> Thanks, this is nice to have, but I feel this should not be a new function, but be folded into the current In function. I.e. add a *tls.Config to *Transfer structure and then rework In() a bit. Regards, Miek

Actually I kind of thought of something like that too. Maybe add a *tls.Config to Transfer, and check if it's not nil, then call DialTimeoutTLS

Would that be better?

miekg:
[ Quoting ***@***.***> in "Re: [miekg/dns] New func InTLS: sta..." ]
>    [ Quoting @.***> in "Re: [miekg/dns] New func InTLS: sta..." ]
>    @cesarkuroiwa requested your review on: #1533 New func InTLS: start a zone
>    transfer via TLS as a code owner.
>    Thanks, this is nice to have, but I feel this should not be a new function,
>    but be folded into the current In function. I.e. add a *tls.Config to
>    *Transfer structure and then rework In() a bit. Regards, Miek
>
>Actually I kind of thought of something like that too. Maybe add a *tls.Config
>to Transfer, and check if it's not nil, then call DialTimeoutTLS
>
>Would that be better?

yep.

With quic support (whenever that lands in Go), there is probably another way to connect,
so doing it (sort of) transparantly in In() has my preference. We don't have
transport-specific functions anywhere else .

Cheers,
Miek

Issue #1384, opened 2022-06-20
Parsing TXT records does not properly de-escape quotes


TomOnTime:
When calling NewRR() on a TXT record, an escaped quote isn't properly de-escaped.

A line such as:

```
example.com. IN TXT "inner\"quote"
```

It gets parsed to: `inner\"quote`
I was expecting: `inner"quote`

Maybe I'm misreading the RFC?  It just seems that without removing the `\` there's no way to know what the `\` means in the resulting string.

I've created https://github.com/miekg/dns/pull/1383 to demonstrate the issue.

tlimoncelli:
"you lost the fact that it was quoted" -- I would imagine that the fact that it is ASCII 34 means it was quoted.  

Can the string be formed any other way?

miekg:
trouble parsing what you mean here. I still believe current behavior is correct, as calling this should give the same output:

NewRR -> String() -> NewRR on that string -> String() etc.

janik-cloudflare:
We've just run into a related issue. In this comment, `X` represents a string containing the character `a` 253 times. Then:

- a TXT with zone file content `"X0123456789"` parses into slice `X01` + `23456789` (as expected)
- a TXT with zone file content `"X\;0123456789"` parses into slice `X\;` + `0123456789` (this seems like a bug: the backslash and semicolon are each counted as a byte, and the string split earlier than necessary, even though this escape sequence would only produce a single byte in wire format)

The `String()` function removes unnecessary backslashes, so parsing and then calling `String()` on something like `"X\;a" "test"` turns it into `"X;" "a" "test"`, adding an unnecessary string. In my opinion, the way to go would be to remove unnecessary backslashes immediately upon parsing instead of in `String()`. This seems more logical and consistent to me: a string in a parsed dns.TXT would then really only contain the contents of that string, without additional zone file syntax. AAAA addresses, for example, are also stored in canonical form (e.g., unnecessary explicit zeros are removed during parsing) so there should be no need to keep literal backslashes in TXT records, right?

(For simplicity, all strings in this comment are raw/literal strings, without Go syntax. In other words, every backslash above is a literal backslash and every quote is a literal quote, such as a quote in a zone file.)

weppos:
I think we also just run into a related scenario.

```
// Convenient function to take a RDATA string,
// and return a corresponding RR entry that has all the RDATA parts decomposed.
func parseRR(rtype string, rdata string) (dns.RR, error) {
[...]

PR #1540, opened 2024-03-01
Fix counting of escape sequences when splitting TXT strings

janik-cloudflare:
`endingToTxtSlice`, used by TXT, SPF and a few other types, parses a string such as `"hello world"` from an RR's content in a zone file. These strings are limited to 255 characters, and `endingToTxtSlice` automatically splits them if they're longer than that. However, it didn't count the length correctly: escape sequences such as `\\` or `\123` were counted as multiple characters (2 and 4 respectively in these examples), but they should only count as one character because they represent a single byte in wire format (which is where this 255 character limit comes from). This commit fixes that.

### DIFF
--- a/parse_test.go
+++ b/parse_test.go
@@ -1098,18 +1098,41 @@ func TestTXT(t *testing.T) {
 		}
 	}
 
-	// Test TXT record with chunk larger than 255 bytes, they should be split up, by the parser
-	s := ""
-	for i := 0; i < 255; i++ {
-		s += "a"
-	}
-	s += "b"
-	rr, err = NewRR(`test.local. 60 IN TXT "` + s + `"`)
+	// Test TXT record with string larger than 255 bytes that should be split
+	// up by the parser. Add some escape sequences too to ensure their length
+	// is counted correctly.
+	s := `"\;\\\120` + strings.Repeat("a", 255) + `b"`
+	rr, err = NewRR(`test.local. 60 IN TXT ` + s)
 	if err != nil {
 		t.Error("failed to parse empty-string TXT record", err)
 	}
-	if rr.(*TXT).Txt[1] != "b" {
-		t.Errorf("Txt should have two chunk, last one my be 'b', but is %s", rr.(*TXT).Txt[1])
+	if rr.(*TXT).Txt[1] != "aaab" {
+		t.Errorf("Txt should have two strings, last one must be 'aaab', but is %s", rr.(*TXT).Txt[1])
+	}
+	rrContent := strings.Replace(rr.String(), rr.Header().String(), "", 1)
+	expectedRRContent := `";\\x` + strings.Repeat("a", 252) + `" "aaab"`
+	if expectedRRContent != rrContent {
+		t.Errorf("Expected TXT RR content to be %#q but got %#q", expectedRRContent, rrContent)
+	}
+
+	// Test TXT record that is already split up into strings of len <= 255.
+	s = fmt.Sprintf(
+		"%q %q %q %q %q %q",
+		strings.Repeat(`a`, 255),
+		strings.Repeat("b", 255),
+		strings.Repeat("c", 255),
+		strings.Repeat("d", 0),
+		strings.Repeat("e", 1),
+		strings.Repeat("f", 123),
+	)
+	rr, err = NewRR(`test.local. 60 IN TXT ` + s)
+	if err != nil {
+		t.Error("failed to parse empty-string TXT record", err)
+	}
+	rrContent = strings.Replace(rr.String(), rr.Header().String(), "", 1)
+	expectedRRContent = s // same as input
+	if expectedRRContent != rrContent {
+		t.Errorf("Expected TXT RR content to be %#q but got %#q", expectedRRContent, rrContent)
 	}
 }
 
--- a/scan_rr.go
+++ b/scan_rr.go
@@ -51,25 +51,21 @@ func endingToTxtSlice(c *zlexer, errstr string) ([]string, *ParseError) {
 		switch l.value {
 		case zString:
 			empty = false
-			if len(l.token) > 255 {
-				// split up tokens that are larger than 255 into 255-chunks
-				sx := []string{}
-				p, i := 0, 255
-				for {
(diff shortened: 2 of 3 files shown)


janik-cloudflare:
A more detailed explanation of the problem: https://github.com/miekg/dns/issues/1384#issuecomment-1946017184

janik-cloudflare:
Apologies, found a bug! Fix and more tests coming today.

miekg:
I think this is OK (my gawd... txt records)... I see a `escapedStringOffset` without looking too closely, don't we already have a bunch of those type of functions? Is a new one really needed or can something older be re-used?

janik-cloudflare:
Thanks for looking! I had indeed found one, `escapedNameLen`, but, while it's similar, it's also different enough to make it difficult to reuse for our purposes. Looking at callers of `isDDD` again now, I also don't see anything else that could easily be reused.

janik-cloudflare:
FYI: we've been running this in production for 3+ weeks with no issues

catenacyber:
Seems to have triggered https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=68124 cc @miekg

miekg:
hmm, thanks.

After some clicking this is the testcase:
~~~~
 hinFo ÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿ\      
 ~~~

janik-cloudflare:
Yikes, thanks @catenacyber. I don't have access to the report but I'll try to figure out what's happening here.

janik-cloudflare:
@miekg I'm using slightly modified versions of `Fuzz` and `FuzzNewRR` (surely the latter is the problematic one), but I'm not able to replicate the issue. I've tried some variations of the test case, such as removing the spaces at the start of each of the two lines, removing the six spaces at the end of the first line, in case those were added by mistake, but no luck. I also tried the second line with and without a trailing `\n`. Parsing fails if I remove the initial spaces, but I imagine the report isn't about a simple error (fuzz function returning code 0) but about a panic.

I think I need some help here. If the test case is available as a binary file that'd be great. Confirmation that it is indeed FuzzNewRR that's failing would be helpful too.

catenacyber:
Here is the base64-encoded file
```
IGhpbkZvIP//////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////XA==
```

FuzzNewRR is the target

```

panic: runtime error: slice bounds out of range [:256] with length 255
--
  |  
  | goroutine 17 [running, locked to thread]:
  | github.com/miekg/dns.endingToTxtSlice(0x0?, {0x755ec9, 0x10})
  | github.com/miekg/dns/scan_rr.go:60 +0xc78
  | github.com/miekg/dns.(*HINFO).parse(0x10c0000dc000, 0x10c00007c6c0?, {0x0?, 0x0?})
  | github.com/miekg/dns/scan_rr.go:201 +0x4b
  | github.com/miekg/dns.(*ZoneParser).Next(0x10c0000d2090)
  | github.com/miekg/dns/scan.go:672 +0x29b7
  | github.com/miekg/dns.ReadRR({0x810638?, 0x10c0000ce000?}, {0x0?, 0x797180?})
  | github.com/miekg/dns/scan.go:128 +0xa6
  | github.com/miekg/dns.NewRR({0x10c0000c8000, 0x106})
  | github.com/miekg/dns/scan.go:113 +0x13f
  | github.com/miekg/dns.FuzzNewRR({0x612000000940?, 0x0?, 0x5cc301?})
  | github.com/miekg/dns/fuzz.go:29 +0xdb
  | main.LLVMFuzzerTestOneInput(...)
  | ./main.1703470873.go:21
```

janik-cloudflare:
Thanks a ton! The problem seems to be within `escapedStringOffset`, in particular the `return i + 1` within `if offset >= byteOffset`. The function is getting confused by the trailing slash and returning an OOB offset, which should never be allowed to happen.

It should be easy enough to fix by adding a bounds check there but I'll see if I can come up with something better that's easier to follow.

janik-cloudflare:
Haven't tested it, but I think `escapedNameLen` might suffer from a similar issue where an invalid message could be produced if the last character is a backslash. That one is probably less serious because I doubt it would panic and, regardless, a backslash shouldn't be able to get there in the first place unless someone is modifying the structs directly. Perhaps another argument for not trying to preserve zone file syntax such as escape sequences in parsed data structures though?

janik-cloudflare:
Here's my proposed fix: #1557

miekg:
Looks ok. Not near a computer for a couple of days..

Are you willing to accept an invite to become maintainer and able to merge?

/Miek

janik-cloudflare:
Thank you Miek! It would be an honor!

miekg:
Ack. Ack

On mobile, but will push buttons next week.
Thanks all, appreciate all the help

miekg:
invite finally sent

PR #1548, opened 2024-03-27
feat: add the ability to inject a custom dialcontext method

dpeckett:
Add the ability to use a custom DialContext function, this allows bypassing the host networking stack (the original inspiration for this was a userspace vpn project I'm working on).


miekg:
I don't see why this is needed when you can just use a `net.Conn` in most places?

PR #1555, opened 2024-04-17
Update nsecx.go

ComputerScienceMasterStudent:
use sha3

    will not be merged.


ComputerScienceMasterStudent:
use sha3 instead of sha1

miekg:
do you know what you are changing and why? As this PR is BS, I will also close your  other PR

PR #1556, opened 2024-04-17
Update server.go

ComputerScienceMasterStudent:
use TLS12 as a minimal version

    will not be merged.


ComputerScienceMasterStudent:
use TLS12 as minimal version

PR #1557, opened 2024-04-19
Fix possible out-of-bounds read in endingToTxtSlice

janik-cloudflare:
Also, reorganize the function a bit to make it more readable. Details here: https://github.com/miekg/dns/pull/1540#issuecomment-2065854334 (and in the commit messages)


catenacyber:
Seems fixed indeed :-)

PR #1567, opened 2024-05-18
Add RFC 9540 to SVCBs to indicate Oblivious HTTP is available

steffsas:
Hi folks!

The IETF just recently published a new [RFC 9540](https://datatracker.ietf.org/doc/rfc9540/). It introduces a new SVCB parameter `ohttp` to a service binding DNS RR to indicate [Oblivious HTTP](https://datatracker.ietf.org/doc/rfc9458/) is available on the target host. Note that the value is always empty. It can be either used in the HTTPS SVCB or in the SVCBs for DDR (Disovery of Designated Resolvers) (see [RFC 9462](https://www.rfc-editor.org/rfc/rfc9462.html)). 

I've added the new SVCB type and added also the tests from the different RFCs.

### DIFF
--- a/svcb.go
+++ b/svcb.go
@@ -14,7 +14,7 @@ import (
 // SVCBKey is the type of the keys used in the SVCB RR.
 type SVCBKey uint16
 
-// Keys defined in draft-ietf-dnsop-svcb-https-08 Section 14.3.2.
+// Keys defined in rfc9460
 const (
 	SVCB_MANDATORY SVCBKey = iota
 	SVCB_ALPN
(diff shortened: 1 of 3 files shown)


svcb.go
```diff
@@ -771,8 +775,8 @@ func (s *SVCBIPv6Hint) copy() SVCBKeyValue {
 // SVCBDoHPath pair is used to indicate the URI template that the
 // clients may use to construct a DNS over HTTPS URI.
 //
-// See RFC xxxx (https://datatracker.ietf.org/doc/html/draft-ietf-add-svcb-dns-02)
-// and RFC yyyy (https://datatracker.ietf.org/doc/html/draft-ietf-add-ddr-06).
+// See RFC 9461 (https://datatracker.ietf.org/doc/html/rfc9461)
```
miekg, line 778: this RFC number should also be added to the README

62 too probably, and rfc9460 too?

miekg:
lgtm, could you do the small favor and update the README with these RFC nums too? ta

miekg:
doing it myself

PR #1568, opened 2024-06-01
Add a hook to catch invalid messages

bemasc:
Currently there are hooks for reading messages off the wire (DecorateReader), checking if they comply with policy (MsgAcceptFunc), and generating responses (Handler).  However, there is no hook that notifies the server when a message is dropped or rejected due to a syntax error.  That makes it hard to monitor these packets without repeating the parsing process.

This PR adds a hook for notifications about invalid packets.

### DIFF
--- a/server.go
+++ b/server.go
@@ -188,6 +188,14 @@ type DecorateReader func(Reader) Reader
 // Implementations should never return a nil Writer.
 type DecorateWriter func(Writer) Writer
 
+// InvalidMsgFunc is a listener hook for observing incoming messages that were discarded
+// because they could not be parsed.
+// Every message that is read by a Reader will eventually be provided to the Handler,
+// rejected (or ignored) by the MsgAcceptFunc, or passed to this function.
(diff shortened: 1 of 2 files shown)


server.go
```diff
@@ -233,6 +241,8 @@ type Server struct {
 	// AcceptMsgFunc will check the incoming message and will reject it early in the process.
 	// By default DefaultMsgAcceptFunc will be used.
 	MsgAcceptFunc MsgAcceptFunc
+	// InvalidMsgFunc is optional, will be called if a message is received but cannot be parsed.
+	InvalidMsgFunc InvalidMsgFunc
```
miekg, line 245: MsgInvalidFunc?

to be in sync with `MsgAcceptFunc`

miekg:
excellent idea. Small comment and making the names more consistent

server.go
bemasc: Works for me.  Done.

PR #1579, opened 2024-07-04
optimize performance

cedar-gao:

    will not be merged.


miekg:
Should this be extended further? I.e. why only optimize this function?

Where is the performance data for you claim?

PR #1581, opened 2024-07-09
Add ExposesUnderlyingConns interface

### LINKED ISSUE #1487: Optional interface for a ResponseWriter to expose Conns
Hello, and thanks for a great library!

It would be useful to me if I can get access to [the underlying connections](https://github.com/miekg/dns/blob/v1.1.56/server.go#L75-L76) in a `response` object. I only want to do some read-only introspection (I'm aware that actually reading/writing from the conns would be rather dangerous).

It is fairly trivial to do this in a way that won't break backwards compatibility by introducing an optional interface that can be satisfied with one or two functions (depending on if it's one function for `tcp` and one for `udp` or if the same function returns both). I've tested this locally and it's a very small change that works well.

Would this be something that would be accepted in a PR? If not, is there another way to get the connection related to an incoming message?

Thanks!

johanbrandhorst:
The interface can be used by DNS handlers to access the underlying TCP or UDP connection. This can be useful if more information is needed about the request, such as to what port it was made.

Fixes #1487

### DIFF
--- a/server.go
+++ b/server.go
@@ -79,6 +79,26 @@ type response struct {
 	writer         Writer         // writer to output the raw DNS bits
 }
 
+// IncomingConn exposes the underlying TCP connection, if it is used.
+// For UDP connections, it returns nil.
+func (r *response) IncomingConn() net.Conn {
+	return r.tcp
+}
+
+// IncomingPacketConn exposes the underlying UDP connection, if it is used.
+// For TCP connections, it returns nil.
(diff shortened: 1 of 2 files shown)


miekg:
Not sure this warrants an expansion of the API, and i never needed it while writing coredns

johanbrandhorst:
Understood, I will close this and maintain my own fork.

Issue #1578, opened 2024-06-26
Use of []byte in sync.Pool


johanbrandhorst:
Hi!

`staticcheck` flags the use of `sync.Pool` in `server.go` as incorrect, as it's passing `[]byte` instead of `*[]byte`, which means every instance ends up allocating a slice header. See https://staticcheck.dev/docs/checks/#SA6002 and https://go-review.googlesource.com/c/go/+/24371 for more information, discussion and examples.

To fix it, I suggest changing instances of `Put` to pass a reference to the slice, and to ensure that the same alloc that was got is put back.

miekg:
ah TIL.  Care to make the PR too?

johanbrandhorst:
I looked into this and I'm concerned that the existing use of pooling is buggy. If a user uses a custom `Reader` via `srv.DecorateReader`, we could end up endlessly putting slices into the pool without ever taking them out.

With regards to fixing the issue, I'm not sure it's going to be possible to do without breaking changes, since we'd need `Reader.ReadUDP` to return `*[]byte` instead of `[]byte` to avoid allocating a slice header on every call to that function. I think we could close this issue and perhaps turn off pooling if a user uses `DecorateReader` to avoid the memory leak.

miekg:
Documentation issue then? (And a thing to turn off pooling?)

johanbrandhorst:
Yeah, sounds like there's not much to do except warn users about the risks of using a custom Reader.

miekg:
care to work your thoughts into a documentation PR then?

PR #1582, opened 2024-07-09
Fix (easy) linter warnings

johanbrandhorst:
### [Remove use of deprecated net.Error.Temporary](https://github.com/miekg/dns/commit/625f7e9694edc0a1bb47f658c8a144bc9760f261)

net.Error.Temporary has been deprecated since Go 1.18. There
has been some discussion around what to use in server accept loops
instead [1], but the suggestion seems to be that it may be a mistake
to have any sort of retries in place [2].

This PR removes it, which may expose some users to errors that were
previously swallowed and retried, but at the expense of leaking file
descriptors and the like.

1: https://groups.google.com/g/golang-nuts/c/-JcZzOkyqYI/m/xwaZzjCgAwAJ?pli=1
2: https://groups.google.com/g/golang-nuts/c/-JcZzOkyqYI/m/vNNiVn_LAwAJ

## Discussion

I could add a staticcheck CI job if preferred. I intend to submit a fix for #1578 separately.

### DIFF
--- a/scan_rr.go
+++ b/scan_rr.go
@@ -1212,7 +1212,7 @@ func (rr *SSHFP) parse(c *zlexer, o string) *ParseError {
 	return nil
 }
 
-func (rr *DNSKEY) parseDNSKEY(c *zlexer, o, typ string) *ParseError {
+func (rr *DNSKEY) parseDNSKEY(c *zlexer, _, typ string) *ParseError {
 	l, _ := c.Next()
 	i, e := strconv.ParseUint(l.token, 10, 16)
 	if e != nil || l.err {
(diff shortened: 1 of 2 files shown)


scan_rr.go
```diff
@@ -1430,7 +1430,7 @@ func (rr *GPOS) parse(c *zlexer, o string) *ParseError {
 	return slurpRemainder(c)
 }
 
-func (rr *DS) parseDS(c *zlexer, o, typ string) *ParseError {
+func (rr *DS) parseDS(c *zlexer, _, typ string) *ParseError {
```
miekg, line 1433: a lot of functions here don't use the `o string` parameters, I left them all in that make the funcs mostly look the same.

I'll cherry pick the server.go change, that looks like a good one

miekg:
closed in favor of #1589

PR #1589, opened 2024-08-03
Remove use of deprecated net.Error.Temporary

miekg:
net.Error.Temporary has been deprecated since Go 1.18. There has been some discussion around what to use in server accept loops instead [1], but the suggestion seems to be that it may be a mistake to have any sort of retries in place [2].

This PR removes it, which may expose some users to errors that were previously swallowed and retried, but at the expense of leaking file descriptors and the like.

1: https://groups.google.com/g/golang-nuts/c/-JcZzOkyqYI/m/xwaZzjCgAwAJ?pli=1
2: https://groups.google.com/g/golang-nuts/c/-JcZzOkyqYI/m/vNNiVn_LAwAJ

    will not be merged.

### DIFF
--- a/server.go
+++ b/server.go
@@ -476,9 +476,6 @@ func (srv *Server) serveTCP(l net.Listener) error {
 			if !srv.isStarted() {
 				return nil
 			}
-			if neterr, ok := err.(net.Error); ok && neterr.Temporary() {
-				continue
-			}
 			return err
 		}
 		srv.lock.Lock()
@@ -534,9 +531,6 @@ func (srv *Server) serveUDP(l net.PacketConn) error {
 		if err != nil {
 			if !srv.isStarted() {
 				return nil
-			}
-			if netErr, ok := err.(net.Error); ok && netErr.Temporary() {
-				continue
 			}
 			return err
 		}


johanbrandhorst:
I think this change unfortunately interacts badly with the default read timeout used in UDP servers (not TCP servers, in my testing). The default read timeout is 2 seconds, and it seems that any UDP server started via `ListenAndServe` or `ActivateAndServer` dies after 2 seconds of no traffic. It seems the TCP server is staying open, not sure why. Presumably this was already happening before, but we were just automatically retrying the error and so it was hidden from us.

Is timing out UDP reads repeatedly while serving intentional? Digging around in the history, it seems the `Temporary()` check for UDP was introduced to fix a different bug (https://github.com/miekg/dns/pull/621), but I don't see it acknowledged that this will cause us to loop endlessly (albeit not too busily).

Maybe we want to revert this until we have a better understanding of the impact, because on `master` right now any UDP servers just stop immediately once the ReadTimeout is hit.

miekg:
let's revert then. Would be nice if some kind of unit test would have caught this though

PR #1618, opened 2024-11-17
fix: input validation for RRSIG.Verify()

### LINKED ISSUE #1617: DNSSEC RRSIG Verify Function Fails to Enforce RFC4035 Constraints
### Description

The `Verify` function does not fully comply with **RFC 4035 Section 5.3.1**, leading to improper validation of RRSIG records. Specifically, the following constraints are not enforced:

1. **Owner Name Mismatch**:
   - The RRSIG owner name (`rr.Hdr.Name`) must match the RRset owner name. Currently, this validation is missing.
   
2. **Signer Name Validation**:
   - The RRSIG `SignerName` must correspond to the zone containing the RRset. The current implementation only checks against the `DNSKEY` name, which is insufficient.

3. **Labels Field Validation**:
   - The number of labels in the RRset owner name must be greater than or equal to the `Labels` field in the RRSIG. This validation is absent.

---

### Steps to Reproduce

Modify `TestSignVerify` in `dnssec_test.go` as one of follows, then run the test:
   1. Change `soa.Hdr.Name` to `example.com.`.
   2. Change `sig.Hdr.Name` to `example.com.`.
   3. Change `soa.Hdr.Name` to `nl.`

### Expected Behavior

The test should fail due to:
1. Owner name mismatch.
2. Invalid `SignerName`.
3. Labels field mismatch.

### Actual Behavior

The test passes incorrectly, indicating the constraints are not being enforced.

---

### References

- **RFC 4035 Section 5.3.1**: [Link](https://www.rfc-editor.org/rfc/rfc4035#section-5.3.1)

developStorm:
Maybe not a perfect fix, but this attempts to resolve #1617.

Also introduces tests to ensure compliance with RFC 4035.

### DIFF
--- a/dnssec.go
+++ b/dnssec.go
@@ -376,7 +376,7 @@ func (rr *RRSIG) Verify(k *DNSKEY, rrset []RR) error {
 	}
 
 	signerName := CanonicalName(rr.SignerName)
-	if !strings.EqualFold(signerName, k.Hdr.Name) {
+	if !equal(signerName, k.Hdr.Name) {
 		return ErrKey
 	}
 
@@ -400,8 +400,8 @@ func (rr *RRSIG) Verify(k *DNSKEY, rrset []RR) error {
 	if h0 := rrset[0].Header(); h0.Class != rr.Hdr.Class ||
 		h0.Rrtype != rr.TypeCovered ||
 		uint8(CountLabel(h0.Name)) < rr.Labels ||
-		!strings.EqualFold(h0.Name, rr.Hdr.Name) ||
-		!strings.HasSuffix(h0.Name, signerName) {
+		!equal(h0.Name, rr.Hdr.Name) ||
+		!strings.HasSuffix(CanonicalName(h0.Name), signerName) {
 
 		return ErrRRset
 	}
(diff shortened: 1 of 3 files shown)


dnssec.go
```diff
@@ -370,9 +374,12 @@ func (rr *RRSIG) Verify(k *DNSKEY, rrset []RR) error {
 	if rr.Algorithm != k.Algorithm {
 		return ErrKey
 	}
-	if !strings.EqualFold(rr.SignerName, k.Hdr.Name) {
+
+	signerName := CanonicalName(rr.SignerName)
+	if !strings.EqualFold(signerName, k.Hdr.Name) {
```
miekg, line 379: i dont think equalFold is correct for use in DNS....

developStorm: Hi，thanks for reviewing! This function use is identical to what's being used in master now: https://github.com/miekg/dns/blob/b77d1ed8e9282cadf21c4124f53a660fed55c8ca/dnssec.go#L373

What's the correct version you would suggest? Would `==` do? I can make the patch.

monoidic: @developStorm 
 [`equal`](https://github.com/miekg/dns/blob/b77d1ed8e9282cadf21c4124f53a660fed55c8ca/labels.go#L190) should be what you want.

developStorm: @monoidic Thanks! 7b480dc01dc7756cfa85848bf8290b47ebc6bba0 replaces all occurrences of EqualFold with equal.

PR #1623, opened 2024-12-06
Allow SO_REUSEADDR and SO_REUSEPORT to be set simultaneously

rtpt-erikgeiser:
Currently it is impossible to set both `SO_REUSEADDR` and `SO_REUSEPORT` simultaneously. If a user tries to set both, there is no error and none of the options are actually set which can be quite confusing.

This PR not only fixes this but adds a test for all combinations of the socket options. In contrast to the existing tests for `SO_REUSEADDR` and  `SO_REUSEPORT`, this test does not check that the options do what they are supposed to do (this is the kernels responsibility) but it tests whether or not the options are actually set as requested by the user (which is the library's responsibility).

### DIFF
--- a/listen_no_socket_options.go
+++ b/listen_no_socket_options.go
@@ -3,9 +3,15 @@
 
 package dns
 
-import "net"
+import (
+	"fmt"
+	"net"
+)
 
-const supportsReusePort = false
+const (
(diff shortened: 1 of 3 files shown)


miekg:
looks like a nice addition thanks!

Issue #1601, opened 2024-09-03
Creating an RR instance with empty RDATA


mattias-p:
I have a domain name and an rrtype and I want to construct a corresponding fully functional dns.RR instance with empty RDATA. This is required for dns.Msg.RemoveRRset(), and I also need to perform some other processing on these records. However, I couldn’t find any documentation on the intended way to do this.

I tried using a plain *dns.RR_Header. This mostly works since *dns.RR_Header formally implements dns.RR, but its dns.RR.copy implementation just returns nil, so dns.Copy doesn't work with such instances.

I also tried using dns.NewRR(fmt.Sprintf("%s %d IN %s", domain, ttl, dns.TypeToString[rrtype])). This also mostly works but for TXT records it seems dns.PackRR needs a larger buffer than what's reported by dns.Len.

I had a look at the source code for dns.RemoveRRset and realized that wrapping the dns.RR_Header in a dns.ANY might be a solution. So far I haven't stumbled on any surprising behaviors with this solution, but I'm not confident about its robustness either.

What is the recommended way to construct a dns.RR instance with empty RDATA?

miekg:
also unsure and there are conflicting requirements for normal use and dynamic update use, and this is one of them, also see #1627, how people consider it a bug.

Wrapping an ANY is interesting and might be a nice why to do thing in this lib too, officially for dynamic update stuff, as (ab)using the normal record types has always been a hit and miss, esp when considering parsing from text.

miekg:
also see https://github.com/miekg/dns/pull/1642

PR #1627, opened 2025-01-24
Return error for empty target

### LINKED ISSUE #1626: Bad target domain parsing.
Here are some bad RRs with invalid (empty) target (please note the space before the `;`)

```zone
bad.example.org.        CNAME ; bad cname
bad.example.org.        HTTPS 10 ; bad https
bad.example.org.        MX 10 ; bad mx
bad.example.org.        SRV 1 0 80 ; bad srv
```

This will be parsed to the following result without error, which treat the empty blank as the `\010.`.

```zone
bad.example.org.        CNAME   \010.
bad.example.org.        HTTPS   10 \010.
bad.example.org.        MX      10 \010.
bad.example.org.        SRV     1 0 80 \010. 
```

However, if we trim the space before the `;`

```zone
bad.example.org.        CNAME; bad cname
bad.example.org.        HTTPS 10; bad https
bad.example.org.        MX 10; bad mx
bad.example.org.        SRV 1 0 80; bad srv
```

The parser will trigger the following errors as exptected

```
dns: not a TTL: "CNAME" at line: 1:23
dns: bad SVCB Target: " " at line: 1:23
dns: bad MX Mx: " " at line: 1:20
dns: bad SRV Target: " " at line: 1:21
```

I propose to trigger error for any case that contains empty target.

taoso:
try to fix #1626

### DIFF
--- a/scan.go
+++ b/scan.go
@@ -1307,10 +1307,6 @@ func stringToCm(token string) (e, m uint8, ok bool) {
 }
 
 func toAbsoluteName(name, origin string) (absolute string, ok bool) {
-	if name == "\n" {
-		return "", false
-	}
-
 	// check for an explicit origin reference
 	if name == "@" {
 		// require a nonempty origin
@@ -1318,6 +1314,10 @@ func toAbsoluteName(name, origin string) (absolute string, ok bool) {
 			return "", false
 		}
 		return origin, true
+	}
+
+	if name == "\n" {
+		return "", false
 	}
 
 	// require a valid domain name
--- a/scan.go
+++ b/scan.go
@@ -1307,6 +1307,10 @@ func stringToCm(token string) (e, m uint8, ok bool) {
 }
 
 func toAbsoluteName(name, origin string) (absolute string, ok bool) {
+	if name == "\n" {
+		return "", false
+	}
+
 	// check for an explicit origin reference
 	if name == "@" {
 		// require a nonempty origin
--- a/scan_test.go
+++ b/scan_test.go
@@ -226,6 +226,22 @@ func TestZoneParserAddressAAAA(t *testing.T) {
 	}
 }
 
+func TestZoneParserTargetBad(t *testing.T) {
+	records := []string{
+		"bad.example.org. CNAME ; bad cname",
+		"bad.example.org. HTTPS 10 ; bad https",
+		"bad.example.org. MX 10 ; bad mx",
+		"bad.example.org. SRV 1 0 80 ; bad srv",
+	}
+
+	for _, record := range records {
+		const expect = "bad "
+		if got, err := NewRR(record); err == nil || !strings.Contains(err.Error(), expect) {
+			t.Errorf("NewRR(%v) = %v, want err to contain %q", record, got, expect)
+		}
+	}
+}
+
 func TestZoneParserAddressBad(t *testing.T) {
 	records := []string{
 		"1.bad.example.org. 600 IN A ::1",


miekg:
thanks this looks a lot better

defaults.go
```diff
@@ -185,6 +185,10 @@ func IsDomainName(s string) (labels int, ok bool) {
 
 	const lenmsg = 256
 
+	if s == "\n" {
```
miekg, line 188: technically dns is 8-bit clean, and \n is sort of a valid character in a dns name, so this corner case must be dealt with in the zone parser and not here

taoso: @miekg after some dig into the parser code, it seems impossible to deal this case in the zone parser.

Because all the rr depends on the `toAbsoluteName()` need to deal this case. But different rr has different rdata format.

For example,
- the CNAME has the rdata format of `example.org`.
- the MX has the rdata format of `50 mail`
- the SVR has the rdata format of `0 5 5060 sipserver.example.com.`
- the HTTPS has the rdata format of `HTTPS 1 . port=8002`

You cannot detect the bad target without knowing the rdata format.

Fortunately, all target in the rdata need to use the `toAbsoluteName()` method to uniform the domain. It's seems appropriate to add the check logic in the `toAbsoluteName()` method.

taoso: > the fact nothing is specified until you see a comment token.

This is only true for the CNAME case. For the rr with rdate containing both target and other information like MX or SVR, it fails.

For example, in `bad.example.org. MX 10 ; bad mx`, the target is after the priority of 10. If we want to deal this case, it seems we need to add some logic in the `case zExpectRdata:` branch. However,  in the MX example, its rdata is not empty, but still invalid. The parser main loop cannot detect the invalid target because it does not know the rdata format.

If checking target is not accepted in the `toAbsoluteName()` method, it seems no choose by do the validation in the `parse()` method of every related rr separately. There are more that 30 different rr types need to be updated.

miekg: `toAbsolute` is def. the wrong place, I also see you point about the number of RR types.

taoso: If the `toAbsoluteName` is the wrong place, would you kindly like to make any suggestion on which part could the check logic be placed? Or each RR that requires domain data should check in their own `parse()` method?

miekg: I think the latter, unless something more general could be find in the parser itself - but I doubt it, because it ( think it only lexes) doesn't have this information. Maybe something more generic can be found about the RRs having this problem, i.e. why does `A` not run into this if `CNAME` does

miekg:
~~~
	_, err := NewRR("miek.nl. IN CNAME	")
	t.Error(err)
	_, err = NewRR("miek.nl. IN A	")
	t.Error(err)
~~~
gives:

~~~
=== RUN   TestZoneParseWrongComment
    scan_test.go:473: <nil>
    scan_test.go:475: dns: bad A A: "\n" at line: 1:14
 ~~~

and that error is from scan_rr.go. Guess we can look at whitespace just there, because this is def. not something that you need or want

miekg:
hm,, it looks like everything is indeed eaten except the newline (this makes sense as whitespace in this section of the RR should be insignificant). 

This check would work, but needs a bit comment, like:

~~~
In the case of 'example.org. IN CNAME    ; hello', the Target name consists out of a single newline (all other space is discarded). Even though DNS is 8-bit clean and \n could/should be valid, this is clearly not what the writer intended, so we check this here. Not that [IsDomainName] does not have this check.
~~~

scan_test.go
```diff
@@ -226,6 +226,22 @@ func TestZoneParserAddressAAAA(t *testing.T) {
 	}
 }
 
+func TestZoneParserTargetBad(t *testing.T) {
+	records := []string{
+		"bad.example.org. CNAME ; bad cname",
+		"bad.example.org. HTTPS 10 ; bad https",
```
miekg, line 232:
```suggestion
		"bad.example.org. CNAME"",
```

scan.go
```diff
@@ -1307,6 +1307,10 @@ func stringToCm(token string) (e, m uint8, ok bool) {
 }
 
 func toAbsoluteName(name, origin string) (absolute string, ok bool) {
+	if name == "\n" {
```
miekg, line 1310: and move this to line 1322, the `@` is more important then this corner case and also happens more often

scan_test.go
taoso: This may not impossible because the current implementation support RR with empty RDATA to create zero RR value.

miekg: yeah, but not all I think? I get an error with `A` records here e.g. So I guess that is sorta half baked, and could be dropped. Downside is that you sometimes do want those records, and a neat way of making them would be nice.

taoso: The `_, err = NewRR("miek.nl. IN A	")` triggers error because it has the dedicate check logic in the `parse` funciton.

https://github.com/miekg/dns/blob/8d0c4128deeb82bf7b296852e35bd6e198a82a40/scan_rr.go#L104-L106

For all RR that need domain_name target, e.g. CNAME/MX/SRV/CNAME, we have two choose to fix this issue:

- add dedicated domain check logic in their own `parse` function
- add dedicated domain check to  the `toAbsoluteName` function which is used by all the related RR

BTW, this check is not CNAME only. It may not appropriate to only mention CNAME in your proposed comment

```
In the case of 'example.org. IN CNAME    ; hello', the Target name
consists out of a single newline (all other space is discarded).
Even though DNS is 8-bit clean and \n could/should be valid, this is
clearly not what the writer intended, so we check this here.
Not that [IsDomainName] does not have this check.
```

miekg: > The _, err = NewRR("miek.nl. IN A	") triggers error because it has the dedicate check logic in the parse funciton.

yep meaning this functionality was incomplete at best. Just need to think of a better way of doing this, also see #1601

miekg (approved):

PR #1641, opened 2025-03-18
Add RESINFO rr

frcroth:
Adds the RESINFO rr type from [RFC 9606](https://www.rfc-editor.org/rfc/rfc9606.html) (DNS Resolver Information).

RESINFO uses the same format as TXT.

### DIFF
--- a/scan_rr.go
+++ b/scan_rr.go
@@ -1620,6 +1620,16 @@ func (rr *NINFO) parse(c *zlexer, o string) *ParseError {
 	return nil
 }
 
+// Uses the same format as TXT
+func (rr *RESINFO) parse(c *zlexer, o string) *ParseError {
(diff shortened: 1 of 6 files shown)


scan_rr.go
```diff
@@ -1620,6 +1620,16 @@ func (rr *NINFO) parse(c *zlexer, o string) *ParseError {
 	return nil
 }
 
+// Uses the same format as TXT
+func (rr *RESINFO) parse(c *zlexer, o string) *ParseError {
+	s, e := endingToTxtSlice(c, "bad RESINFO ")
```
miekg, line 1625:
```suggestion
	s, e := endingToTxtSlice(c, "bad RESINFO Resinfo")
```

miekg:
thanks!

Issue #941, opened 2019-03-18
macos: UDP server rebinds to *:port after first query


vavrusa:
I was looking at an odd bug report and it looks like the PacketConn server always rebinds to any address after first query. Here's a simple program:

```go
package main

import (
	"net"

	"github.com/miekg/dns"
)

func main() {
	pc, err := net.ListenPacket("udp", "127.0.0.1:1053")
	if err != nil {
		panic(err)
	}

	server := &dns.Server{PacketConn: pc}
	server.ActivateAndServe()
}
```

When I run it it binds to 127.0.0.1:1053 correctly:

```bash
$ go run bug.go &
$ lsof -p 10509 -n | grep UDP
bug     10509 vavrusa    3u    IPv4 0x52525d6e93c51cd9      0t0         UDP 127.0.0.1:remote-as
```

Cool. When I query it it answers, but the binding changes to *:1053:

```bash
$ kdig @127.0.0.1 -p 1053 +short
$ lsof -p 10509 -n | grep UDP
bug     10509 vavrusa    3u    IPv4 0x52525d6e93c51cd9      0t0         UDP *:remote-as
```

It also answers on any interface now:

```bash
$ kdig @192.168.86.110 -p 1053
;; ->>HEADER<<- opcode: QUERY; status: SERVFAIL; id: 16510
;; Flags: qr rd; QUERY: 1; ANSWER: 0; AUTHORITY: 0; ADDITIONAL: 0

;; QUESTION SECTION:
;; .                   		IN	NS

;; Received 17 B
;; Time 2019-03-17 20:03:06 PDT
;; From 192.168.86.110@1053(UDP) in 2.8 ms
```

I traced the problem to https://github.com/miekg/dns/blob/master/server.go#L326 which seems to have been added in https://github.com/miekg/dns/pull/594. The problematic part is https://github.com/miekg/dns/blob/master/udp.go#L60:

```go
// Try setting the flags for both families and ignore the errors unless they
// both error.
err6 := ipv6.NewPacketConn(conn).SetControlMessage(ipv6.FlagDst|ipv6.FlagInterface, true)
err4 := ipv4.NewPacketConn(conn).SetControlMessage(ipv4.FlagDst|ipv4.FlagInterface, true)
```

It does what it says (converts it into a raw socket). I'm not sure why was it added in the first place, but it is probably not what you want, as it opens up a DNS server using this library to any interface without user's control. Like CoreDNS:

```bash
$ cat Corefile
.:1053 {
[...]

PR #1643, opened 2025-03-24
Avoid sendmsg control messages on darwin

### LINKED ISSUE #724: UDP DNS: Bind to specific IP is not retained
On MAC, we used the code 
      

>         dns.HandleFunc(".", handleDNSRequest)
>         defer dns.HandleRemove(".")
>         // start serverUDP
>         serverUDP := &dns.Server{Addr: "127.0.0.1:53", Net: "udp"}
>         go serverUDP.ListenAndServe()
>         defer serverUDP.Shutdown()

It bound properly on specific IP and PORT as shown below:

udp4       0      0  127.0.0.1.53           *.*  

But as soon as we write DNS  response from handler function like

               
>         msg.Authoritative = true
>         domain := msg.Question[0].Name
>         msg.Answer = append(msg.Answer, &dns.A{
>             Hdr: dns.RR_Header{Name: domain, Rrtype: dns.TypeA, Class: dns.ClassINET, Ttl: 60},
>             A:   net.ParseIP("1.2.3.4"),
>         })
>         w.WriteMsg(&msg)

Binding changes to as shown below:

udp4       0      0  *.53                   *.*

This impacts other DNS server which is running on another interface.

I debugged the UDP DNS code and figured out that after the below line is executed the binding changes

> conn.WriteMsgUDP(b, oob, session.raddr)

Please let me know any more information is required such that this great code can be useful in case we wanted to bind to specific IP address and retain the binding

cc: @fastest963

stek29:
sendmsg with PKTINFO is broken on darwin, so don't use it

udp_windows is renamed to udp_no_control since it's no longer exlusive to windows, and is used on darwin too

fixes #724 and #941

### DIFF
--- a/udp.go
+++ b/udp.go
@@ -1,5 +1,5 @@
-//go:build !windows
-// +build !windows
+//go:build !windows && !darwin
+// +build !windows,!darwin
 
 package dns
 
--- a/udp_no_control.go
+++ b/udp_no_control.go
@@ -1,8 +1,10 @@
-//go:build windows
-// +build windows
+//go:build windows || darwin
+// +build windows darwin
 
 // TODO(tmthrgd): Remove this Windows-specific code if go.dev/issue/7175 and
 //   go.dev/issue/7174 are ever fixed.
+
+// NOTICE(stek29): darwin supports PKTINFO in sendmsg, but it unbinds sockets, see https://github.com/miekg/dns/issues/724
 
 package dns
 


stek29:
this is an issue with XNU itself and not with Go

_sidenote_: does using sendmsg make any sense if the socket is already bound to a non- 0.0.0.0 address? maybe it's better to avoid sendmsg completely if socket is bound to an IP?

stek29:
@miekg @tmthrgd PTAL: I've been successfully running coredns with this patch on macOS for some time with no issues

I highly doubt that macOS go servers are used with non-trivial routing setups where PKTINFO is necessary, but I'm willing to implement an adaptive approach where PKTINFO is unused only if there's a non-ANY binding on the socket if necessary

miekg:
thanks for this, I have no way of knowing if this is the correct approach, but happy to try it. We have some old bugs of weird things on macos that noone understood/understands

stek29:
@miekg could you please make a new release with this fix, so the fix gets into coredns?

PR #1647, opened 2025-04-09
Upgrade all deps

miekg:

    will not be merged.


miekg:
putting `go 1.22.0` back in go.mod make go build complain... don't understand. I think we're now bumping the minimum build, but also don't now how to prevent this and still upgrade the deps.

PR #1649, opened 2025-04-23
Fix logic in xfr ReadMsg + add test

ddevine-NS1:
Hi! 

This fixes a possible bug where a xfr client can skip tsig validation if the server doesn't respond with a tsig. 

Say I setup an xfr client with a tsig provider. If it connects to a server which doesn't implement tsig, but that server can perform a transfer without adding the tsig to the response then the current logic will skip validation. 

`m.IsTsig() && t.tsigProvider()` is going to look for the tsig in the message, see none and consider the transfer valid, but I think the client should be checking the signature since the tsig provider is configured on the client.

### DIFF
--- a/xfr.go
+++ b/xfr.go
@@ -251,10 +251,13 @@ func (t *Transfer) ReadMsg() (*Msg, error) {
 	if err := m.Unpack(p); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


miekg:
your rational makes sense

PR #1652, opened 2025-05-30
Fix Msg.Len calculation when an RR contains a padded base64 string

brianshea2:
The `RR.len` functions use `base64.StdEncoding.DecodedLen` to obtain the length of a base64 string field.

Per `base64.StdEncoding.DecodedLen`'s documentation:
> DecodedLen returns the maximum length in bytes of the decoded data

Unfortunately, that function does not account for padding characters.

This PR replaces the use of that function with one that calculates the exact length of the decoded data, and adds some tests to prove that.

### DIFF
--- a/length_test.go
+++ b/length_test.go
@@ -33,9 +33,12 @@ func TestMsgCompressLength(t *testing.T) {
 	name1 := "12345678901234567890123456789012345.12345678.123."
 	rrA := testRR(name1 + " 3600 IN A 192.0.2.1")
 	rrMx := testRR(name1 + " 3600 IN MX 10 " + name1)
+	rrSig := testRR(name1 + " 3600 IN RRSIG A 13 3 3600 20151021000000 20151020000000 12345 12345678.123. 6AF6rwhRrx9cPGptdWkkCZLWwpff+WQ2HvHw8ahhkcyagHD4udi5WQrmOsCg28sAtbd2brTUyOwsrlGjpLHJGA==")
 	tests := []*Msg{
 		makeMsg(name1, []RR{rrA}, nil, nil),
-		makeMsg(name1, []RR{rrMx, rrMx}, nil, nil)}
+		makeMsg(name1, []RR{rrMx, rrMx}, nil, nil),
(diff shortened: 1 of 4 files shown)


miekg:
So in Go's base64 decoding, the decodedLen disregards padding, but when decoding it is not? Isn't that a bug in Go?

miekg:
ok, fair, but Len() should be used as an indication on how large the buffer should be,  if this makes it slightly larger, no harm is done. So I thing carrying all the new (test) code is not worth it

brianshea2:
> ok, fair, but Len() should be used as an indication on how large the buffer should be, if this makes it slightly larger, no harm is done. So I thing carrying all the new (test) code is not worth it

`Msg.Len` can have more use than only an indication on how large a buffer to hold the packed msg should be. I came across this while attempting to get the current size of a response to calculate the amount of edns0 padding would be needed to grow the response to a certain size. In that case, the current inaccuracy of `Msg.Len` causes harm to that calculation, not adding enough padding.

monoidic:
There are a number of other things that inflate the length for some RR types.
I pushed the below into the "fuzz" branch of my fork [here](/monoidic/dns/tree/fuzz), along with some fuzzing code.
Probably need to review some things there, but it reduces some of the `dns.Len()` inflation.

A notable one is `(*NSEC3).len()` using `1 + len(rr.NextDomain) + 1` instead of `base32HexNoPadEncoding.DecodedLen(len(rr.NextDomain))`.
`(*APLPrefix).len()` does not fully account for the omitted zero bytes, e.g the subnet `10.0.0.0/16` would account for 1 byte, not 2, as is currently predicted.
A variety of escaped string fields use `len(s)` instead of `escapedNameLen(s)`
`IPSECKEY` and `AMTRELAY` use `len(s)` instead of `domainNameLen(s)` for getting the length of the gateway (this one's on me...)
I'm not sure if an empty type bitmap is allowed in `CSYNC` or any other RR type that uses NSEC-style bitmaps. Though if it is, then the length for that would be estimated as 2 in `typeBitMapLen` rather than 0.

The RFC for APL doesn't say anything about the expected handling of APL RRs with host bits set, sadly.

PR #1654, opened 2025-06-12
client: enforce TSIG RR on non-FORMERR signed query responses

cpu:
If a client exchanges a `Msg` that is TSIG signed, a response that isn't a `FORMERR` response MUST include a TSIG RR. If it doesn't, surface `ErrNoSig`. Previously these circumstances would generate no error.

[RFC 8945 §5.2](https://www.rfc-editor.org/rfc/rfc8945.html#name-server-processing-of-reques) "Server Processing of Request" says:

> If the TSIG RR cannot be interpreted, the server MUST regard the message as corrupt and return a FORMERR to the server. Otherwise, the server is REQUIRED to return a TSIG RR in the response.

(Note that this text should say "return a FORMERR to the **client**", since its written from the perspective of a server. See RFC 8945 [errata EID7983](https://www.rfc-editor.org/errata/eid7983) where this error in the text is marked a verified errata)

This fix is similar in spirit to https://github.com/miekg/dns/pull/1649, but that fix was isolated to the `Transfer` type used for AXFR, and wasn't sufficient to surface an error for (as an example) a TSIG authenticated update `Msg` that generated a response missing the TSIG RR.

### DIFF
--- a/client.go
+++ b/client.go
@@ -258,7 +258,17 @@ func (co *Conn) ReadMsg() (*Msg, error) {
 		// to use an erroneous message
 		return m, err
 	}
-	if t := m.IsTsig(); t != nil {
+	t := m.IsTsig()
+	// Servers MUST reply with a TSIG RR in the response unless the request
+	// TSIG RR couldn't be interpreted, in which case it must reply FORMERR.
+	// See RFC 8945 §5.2 "Server Processing of Request".
+	//
+	// If we generated a TSIG signature for the request, and we find no TSIG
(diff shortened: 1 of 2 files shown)


cpu:
> Do you still have time to fix this?

Thanks for the ping.

With some reflection, probably not. The contract engagement I was working that pushed me towards proposing this change has since ended and while I think the change is sound I wouldn't want to be on the hook for any additional follow-up. I'm going to close this and anyone with interest can pick up the work themselves.

PR #1655, opened 2025-06-24
Golangci fixes

kerolasa:
* Minor golanci code improvements
 
    Tidy up the following.
 
        labels.go:70:3: QF1006: could lift into loop condition (staticcheck)
        scan_rr.go:522:2: QF1003: could use tagged switch on l.value (staticcheck)
        scan_rr.go:541:2: QF1003: could use tagged switch on l.value (staticcheck)
        scan_rr.go:560:2: QF1003: could use tagged switch on l.value (staticcheck)
        scan_rr.go:1306:14: QF1001: could apply De Morgan's law (staticcheck)
 
    Signed-off-by: Sami Kerola <kerolasa@cloudflare.com>

* Avoid couple allocations when reading packages
    
    Inspired by golangci-lint that gives few 'SA6002: argument should be
    pointer-like to avoid allocations' outputs. Remaining warnings are about
    slices used via indexes, and such things cannot be addressed via pointer.
    
    Signed-off-by: Sami Kerola <kerolasa@cloudflare.com>

* Avoid using deprecated socket error check
    
    Advised by golangci-lint:
    
        SA1019: neterr.Temporary has been deprecated since Go 1.18 because
        it shouldn't be used: Temporary errors are not well-defined. Most
        "temporary" errors are timeouts, and the few exceptions are
        surprising. Do not use this method.
    
    Signed-off-by: Sami Kerola <kerolasa@cloudflare.com>

### DIFF
--- a/labels.go
+++ b/labels.go
@@ -66,10 +66,7 @@ func CompareDomainName(s1, s2 string) (n int) {
 	} else {
 		return
 	}
-	for {
-		if i1 < 0 || i2 < 0 {
-			break
-		}
+	for i1 < 0 || i2 < 0 {
 		if equal(s1[l1[i1]:l1[j1]], s2[l2[i2]:l2[j2]]) {
 			n++
 		} else {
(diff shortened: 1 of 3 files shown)


miekg:
I think we've been bitten before with the whole temp error stuff. For the other "fixes", yeah sure, but current code is fine

PR #1656, opened 2025-07-04
No longer accept short messages

monoidic:
At the moment, the code for unpacking RRs (mostly in the autogenerated code) is peppered with chunks of `off == len(msg)` checks which return with a nil error.
As far as I understand, this is a workaround for parsing messages with an rdlength of zero. This mostly works due to the `noRdata` check in `UnpackRRWithHeader`.
Removing these checks reveals that `TestNoRdataUnpack` does not actually produce zero-rdlength messages for a good number of rrtypes, so the `noRdata` early return does not trigger. These `off == len(msg)` clauses peppered everywhere cause parsing to  instead succeed at the boundaries of arbitrary elements of a resource record, with the zero values assigned to the rest of the messages.
This means that malformed RRs which are too short, but happen to end at the right byte boundary, will be considered valid, with any remaining fields being kept as the zero value.
In fact, even a nil buffer would be parsed as a "valid" RR by `dns.UnpackRR(nil, 0)`, producing this RR:
```
	0	CLASS0	TYPE0	\# 0 
```

This PR removes all of these checks while keeping functionality for real zero-rdlength RRs.
The malformed messages described above will no longer be considered valid.
`TestNoRdataUnpack` has been updated to manually adjust the RRs to produce actual zero-rdlength messages.

### DIFF
--- a/dns_test.go
+++ b/dns_test.go
@@ -2,6 +2,7 @@ package dns
 
 import (
 	"bytes"
+	"encoding/binary"
 	"encoding/hex"
 	"net"
 	"testing"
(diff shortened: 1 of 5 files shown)


miekg:
ohh, this is nice, and removes a whole bunch of (generated) code. IIRC this was added because of actual things going awry when  parsing binary from the wire - but I dont thing we have a test for this :-(

To put it another way if you can show me a binary unpack test that would hit this `if` and it would still work with your PR it will certainly be merged. Might be fiddly to actually make that sort of packet though...

miekg:
https://codeberg.org/miekg/dns used @tmthrgd `cryptobyte.String` which neatly solves this.

I'm not going to merge such an invasive PR here

PR #1659, opened 2025-07-22
Don't cache IdleTimeout for TCP connections

Kentzo:
In DNS Stateful Operations, after a session is established, timeout is a dynamic value negotiated between the server and client.

    will not be merged.

### DIFF
--- a/server.go
+++ b/server.go
@@ -572,11 +572,6 @@ func (srv *Server) serveTCPConn(wg *sync.WaitGroup, rw net.Conn) {
 		reader = srv.DecorateReader(reader)
 	}
 
-	idleTimeout := tcpIdleTimeout
-	if srv.IdleTimeout != nil {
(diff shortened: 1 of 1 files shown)


ccoVeille (approved):
This seems a good idea

Kentzo:
It's not for the library itself but rather for the server / client software that that use it.

The new dynamic value can be communicated via the [edns-tcp-keepalive EDNS0 option ](https://datatracker.ietf.org/doc/html/rfc7828) or with the [Keepalive DSO TLV](https://datatracker.ietf.org/doc/html/rfc8490#section-7.1) (which I'm currently working on) thus the software needs means to adjust it for existing TCP connections.

miekg:
as there is no clear path on how to update that value from a message received I don't see what this helps.

From small query perfing there is an impact in doing this 5kqps less

PR #1667, opened 2025-08-08
Allow providing invalid counts to message

xoudini:
Hi! We're working on a protocol testing tool, and figured it would be good to allow providing invalid data in messages to verify that servers are behaving sanely in such situations. Most fields in the header can already be set to an incorrect value, but the RR counts are automatically determined, and can't be "faked" currently.

Let me know if this is something you'd consider merging! It should be fully backwards compatible as far as I can tell, and doesn't change default behaviour.

### DIFF
--- a/msg.go
+++ b/msg.go
@@ -92,17 +92,18 @@ func id() uint16 {
 
 // MsgHdr is a a manually-unpacked version of (id, bits).
 type MsgHdr struct {
-	Id                 uint16
-	Response           bool
-	Opcode             int
-	Authoritative      bool
-	Truncated          bool
-	RecursionDesired   bool
-	RecursionAvailable bool
-	Zero               bool
-	AuthenticatedData  bool
-	CheckingDisabled   bool
-	Rcode              int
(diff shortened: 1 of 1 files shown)


miekg:
Interesting and definitely something that should be possible.

However I am experimenting in github.com/miekg/dnsv2 for a new and improved
API (its in the its compile, but does not work yet). I'll work this into
that version with some other ideas I have.

Regards

xoudini:
> However I am experimenting in github.com/miekg/dnsv2 for a new and improved API (its in the its compile, but does not work yet). I'll work this into that version with some other ideas I have.

Nice! What's the status and/or timeline on the v2 library? I can of course offer some assistance on that as well, if needed!

> Interesting and definitely something that should be possible.

In the meantime, would it be possible to merge this into the current library? Just let me know if any edits are needed to the current implementation.

miekg:
https://codeberg.org/miekg/dns/src/branch/main/dns.go#L174 will have a Data field that contains the Packed msg, you can there just set the counts

PR #1668, opened 2025-08-11
Support RFC 8490 DNS Stateful Operations (DSO) and RFC 8765 DNS Push Notifications

Kentzo:
Preliminary placeholder PR in light of [@miekg's work on the dnsv2 package](url).

---

This PR adds support for RFC 8490 DNS Stateful Operations and RFC 8765 DNS Push Notifications.

The changes are fully backward compatible and do not affect existing code. Enabling DSO requires an explicit opt-in by setting `Server.DSOHandler`.

Changes to the existing codebase:
- `MsgAcceptAction` is extended with `MsgAbort`. Default action is to set socket's linger to 0 and close.
- `Msg` errors early during packing and unpacking non-empty messages when opcode is Stateful. This is backward compatible because it couldn't represent non-empty DSO messages anyway.
- Indirection was added to `Server` to route OpcodeStateful messages through the DSO dispatcher.

---

1. I anticipate that DSO traffic is going to be low volume, so I used sync.RWMutex to synchronize access to DSO state. But it's feasible to rewrite this with atomics / trylocks.
2. Perhaps it's better to substitute the `DSOHandler` field with a runtime check of whether `Handler` implements `DSOResponseWriter`.
3. Since DSOHandler is opt-in, I'd prefer to change `serveDSOWithHeader(...)` to `go serveDSOWithHeader(...)` with a caveat that the user must ensure that their `DecorateWriter`, if any,  is safe to be called concurrently. It's my understanding that both Go's `TCPConn` and `TLSConn` can properly handle concurrent writes. Perhaps the only change needed for `response` is to make `closed` and, possibly, `hijacked` into `atomic.Bool`.

### DIFF
--- a/acceptfunc.go
+++ b/acceptfunc.go
@@ -19,6 +19,8 @@ type MsgAcceptFunc func(dh Header) MsgAcceptAction
 // * has more than 2 RRs in the Additional section
 var DefaultMsgAcceptFunc MsgAcceptFunc = defaultMsgAcceptFunc
 
+var DefaultDSOMsgAcceptFunc MsgAcceptFunc = defaultDSOMsgAcceptFunc
+
 // MsgAcceptAction represents the action to be taken.
 type MsgAcceptAction int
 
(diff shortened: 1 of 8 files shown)


Kentzo:
Hmm, code duplication was not among the issues I noticed when writing this. Which could be deduplicated?

miekg:
State mgmt should probably be discussed in an issue or even in a discussion
before any changes.

Farhadsakhodi (commented):
Farhadsakhodi

PR #1669, opened 2025-08-12
Support query pipelining in TCP

Kentzo:
Query pipelining is opt-in by setting `Server.PipelineTCP`.

When unset, the only effect of this changes is an extra function call to read values of `hijacked` and `closed`.

When set, `Server.serveDNS` is called concurrently for each incoming request and additional synchronization is used:
- `hijacked` and `closed` are atomics
- Hijacking handler may need to know when Server is done reading from the connection.`PipelineResponseWriter.WaitHijacked` assists with that.
- Each Handler.ServeDNS receives a shallow copy of response, `pipelinedResponse`, that shares common state but has a copy of tsig data


server.go
```diff
@@ -581,26 +616,43 @@ func (srv *Server) serveTCPConn(wg *sync.WaitGroup, rw net.Conn) {
...
 			// TODO(tmthrgd): handle error
 			break
 		}
-		srv.serveDNS(m, w)
-		if w.closed {
+
+		if srv.TCPPipeline {
+			pipelineWg.Add(1)
+			// Per-request copy of the response.
+			pipelineW := *w
```
Kentzo, line 632: Should I use a function here?

miekg:
Making hijack atomically settable is a good idea.
Note that in https://codeberg.org/miekg/dns/ the tsig stuff is out of the server, so it will be easier.

and pipelining should be a default, I think.

I'm also puzzled by the change, just another waitgroup? How does that then pipeline?

Kentzo:
> I'm also puzzled by the change, just another waitgroup? How does that then pipeline?

When you hijack a connection, you want to abandon just that connection, so a separate WaitGroup is used within of them. Note that it's only waited on if the connection was not hijacked.

Without a separate WaitGroup, `Server.serveTCPConn` would be stuck waiting for `Handler.ServeDNS` from the abandoned connections.

Query pipelining happens here:

```
		if srv.TCPPipeline {
			pipelineWg.Add(1)
			// Per-request copy of response.
			pipelineW := *w
			go func() { // <---
				defer pipelineWg.Done()
				srv.serveDNS(m, &pipelineW)
			}()
		} else {
			srv.serveDNS(m, w)
		}
```

I.e. each query received over the same TCP connection gets its own call to `Handler.ServeDNS`

miekg:
Ah fair. I see it now. That's probably something I've missed in
codeberg.org/miekg/DNS too

Thanks

PR #1670, opened 2025-08-12
Use netip.ParseAddr instead of net.ParseIP for clean IPv4-mapped IPv6 handling

fcelda:
Hello. This PR cleans up a few quirks related to IPv4-mapped IPv6 address handling. It doesn't add any new functionality and the change should be backward compatible (at least, no tests broke).

TLDR: The PR replaces net.ParseIP and net.ParseCIDR with netip.ParseAddr and netip.ParsePrefix.

The old address parsing interface has some implicit behavior regarding IPv4-mapped IPv6 handling. Namely, any parsed IPv4 address is stored as IPv4-mapped IPv6 internally. At the same time, parsing explicit IPv4-mapped IPv6 address and converting it back to string will return the address in the IPv4 format. Essentially, `net.ParseIP(in).String() == in` is not always true.

This probably doesn't matter for most applications, however in case of a DNS library, there are a plenty of places where we need to distinguish IPv4 and IPv6 explicitly. The address parsing interface in the netip package works without the magical conversions and therefore seems to be better suited for our use case.

I've replaced the use of the old functions everywhere in the code for consistency except for unit tests where I think it doesn't matter. I've also added a few more tests. Please, let me know what you think and if I missed something else worth updating.

### DIFF
--- a/msg_helpers_test.go
+++ b/msg_helpers_test.go
@@ -265,6 +265,13 @@ func TestPackDataAplPrefix(t *testing.T) {
 			net.CIDRMask(22, 32),
 			[]byte{0x00, 0x01, 0x16, 0x03, 198, 51, 100}, // 1:198.51.100.0/22
 		},
+		{
+			"ipv4-mapped ipv6",
+			false,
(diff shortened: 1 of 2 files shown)


miekg:
will not be merged here, but see https://codeberg.org/miekg/dns/issues/5 to do it properly there.

PR #1672, opened 2025-08-13
Support DNS messages with RFC 8490 DSO-Data

Kentzo:
In this PR I used custom, non-`RR`, type to directly represent DSO data in `Msg`.


msg.go
```diff
@@ -952,14 +967,29 @@ func (dns *Msg) String() string {
...
 func (dns *Msg) isCompressible() bool {
-	// If we only have one question, there is nothing we can ever compress.
-	return len(dns.Question) > 1 || len(dns.Answer) > 0 ||
-		len(dns.Ns) > 0 || len(dns.Extra) > 0
+	switch {
+	case dns.MsgHdr.Opcode == OpcodeStateful:
+		if isDSOCompressible(dns) {
+			return true
+		}
+		fallthrough
```
Kentzo, line 987: This is done in spirit that if user wants to pack an invalid DSO message (with both DSO-Data and RR-data) we should not disallow compression.

miekg:
> isDSOCompressible(dns)

DSO has compression??

(still 1200+ lines ...*if* inclined to review, I'm spending less then an hour on this lib per week, so we're talking early next year)

Kentzo:
You want this checks directly in dns.is Compressible?

miekg:
Nothing of this stuff will be merged as-is.

We can have a discussion on this - should have happened first - but then
still I don't think such a niche DNS feature should warrant such a code
increase in this lib.

/Miek

Kentzo:
*please see edit to the message above.

> still I don't think such a niche DNS feature should warrant such a code
increase in this lib.

The readme says support of rfc 8490, hence my interest to actually help implementing it. I understand if you don’t want, but then I’m confused by the claim.

> We can have a discussion on this - should have happened first

I wanted to have full working implementation (via CoreDNS plugin) and intended to present my work only then. I decided to publish this (and previous related one) PR after learning of your work on dnsv2 just to give you an idea what might be needed for proper DSO support, as claimed in readme.

By no means the PR is final and I’m happy to rearrange as necessary.

However, since you have little to no time to dig into RFCs and read other implementations,  perhaps it’s more efficient to do back-and-forth via PRs? Easier to discuss code at hand than abstract ideas.

PR #1676, opened 2025-09-19
Add Msg.MarshalJSON, Msg.UnmarshalJSON and tests with full coverage

linkdata:
MarshalJSON generates JSON using an explicit, human-readable JSON schema.
Full test coverage is provided.

Schema overview (stable keys; rdata keys per RR type listed below):

```
	{
	  "id": 1234,
	  "msgHdr": {"qr":true, "opcode":"QUERY", "aa":false, "tc":false, "rd":true,
	              "ra":true, "z":0, "ad":false, "cd":false, "rcode":"NOERROR"},
	  "question": [{"name":"example.com.", "qtype":"A", "qclass":"IN"}],
	  "answer":  [ RRJSON, ... ],
	  "ns":      [ RRJSON, ... ],
	  "extra":   [ RRJSON, ... ]
	}
```

RRJSON (common fields) + per-type data (examples):

```json
	{
	  "name":"example.com.", "type":"A", "class":"IN", "ttl":300,
	  "data": { "a":"93.184.216.34" }
	}
```

```
AAAA: {"aaaa":"2001:db8::1"}
CNAME: {"target":"alias.example."}
NS: {"ns":"ns1.example."}
PTR: {"ptr":"host.example."}
TXT: {"txt":["chunk1","chunk2"]}
MX: {"preference":10, "mx":"mail.example."}
SRV: {"priority":0, "weight":5, "port":443, "target":"svc.example."}
SOA: {"ns":"ns1.", "mbox":"hostmaster.", "serial":1, "refresh":7200, "retry":900, "expire":1209600, "minttl":300}
CAA: {"flag":0, "tag":"issue", "value":"letsencrypt.org"}
NAPTR: {"order":100, "preference":50, "flags":"s", "services":"SIP+D2U", "regexp":"", "replacement":"_sip._udp.example."}
DS: {"key_tag":12345, "algorithm":8, "digest_type":2, "digest":"...hex..."}
DNSKEY: {"flags":257, "protocol":3, "algorithm":8, "public_key":"base64..."}
RRSIG: {"type_covered":"A", "algorithm":8, "labels":2, "original_ttl":300,
	"expiration": 1735689600, "inception": 1733097600, "key_tag":12345,
	"signer_name":"example.", "signature":"base64..."}
TLSA: {"usage":3, "selector":1, "matching_type":1, "cert_data":"hex or base64"}
```

Notes
  - Type/class use standard mnemonics (e.g., "A", "AAAA", "IN").
  - Unknown/less-common RR types are round-tripped via a best-effort map in "data"; if a
    type is not implemented below, Marshal will include {"raw": "<presentation>"} and
    Unmarshal will parse it using NewRR.
  - Times in RRSIG use UNIX seconds per miekg/dns conventions.

### DIFF
--- a/json.go
+++ b/json.go
@@ -12,11 +12,11 @@ var _ json.Marshaler = &Msg{}
 var _ json.Unmarshaler = &Msg{}
 
 var (
-	ErrEmptyInput         = errors.New("dnsjson: empty input")
-	ErrInvalidJSON        = errors.New("dnsjson: invalid JSON")
-	ErrInvalidMessage     = errors.New("dnsjson: invalid message")
-	ErrQuestionQType      = errors.New("dnsjson: question qtype")
-	ErrQuestionQClass     = errors.New("dnsjson: question qclass")
+	ErrEmptyInput         = errors.New("empty input")
+	ErrInvalidJSON        = errors.New("invalid JSON")
+	ErrInvalidMessage     = errors.New("invalid message")
+	ErrQuestionQType      = errors.New("question qtype")
+	ErrQuestionQClass     = errors.New("question qclass")
 	ErrAnswerSection      = errors.New("answer")
 	ErrNsSection          = errors.New("ns")
 	ErrExtraSection       = errors.New("extra")
--- a/json.go
+++ b/json.go
@@ -1,8 +1,11 @@
 package dns
 
 import (
+	"encoding/hex"
 	"encoding/json"
 	"errors"
+	"math"
+	"net"
 	"net/netip"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


rdrozhdzh:
How RRJSON will look for OPT record?

linkdata:
> How RRJSON will look for OPT record?

*facepalm* Be right back... :)

miekg:
If this isn't https://www.rfc-editor.org/rfc/rfc8427.html then it is probably not worth adding.
If it _is_, then why should it be added here and not in a seperate repo, dnsjson or some
such? I'm also pushing back on large new additions as per the first line of the readme.

/Miek

linkdata:
Yes, https://github.com/linkdata/dnsjson also exists. I simply offer it as a MR because the JSON functionality has been requested before. Feel free to close the MR if you don't think it's appropriate.

PR #1678, opened 2025-09-28
server: Backport ResponseWriter.Conn()

Kentzo:
Access to the underlying connection is still useful to determine whether non-hijacked ResponseWriter is functional.


Kentzo:
Useless.

PR #1684, opened 2025-10-24
Make OPT.String() more dig-like, Fix MBZ display

brianshea2:
- Minor changes to how the OPT String() function generates its output to better match `dig`
  - colon after "version"
  - comma vs. semicolon
  - flags are space separated (no comma)
- The CO flag _could_ be set without the DO flag
- Fix getting of the Z value

### DIFF
--- a/edns.go
+++ b/edns.go
@@ -80,18 +80,17 @@ type OPT struct {
 }
 
 func (rr *OPT) String() string {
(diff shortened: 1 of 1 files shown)


miekg:
i think moving further away from the dig-format would be better, https://codeberg.org/miekg/dns does that.

no need to do it here.

PR #1687, opened 2025-11-12
provide a way to set socket options

balboah:
## What

- ~~Allow 3rd parties to control socket options~~
- Provide a way to retrieve the OOB destination information
- Expose net.Conn, to be able to read unix.TCP_INFO

## How

- ~~Created a new listener param to set Control of net.ListenConfig~~
- ~~Exposed internal OOB state with a new `PacketConnState`~~
- Introduce parts of v2 dns that exposes a Session and Conn with the help of `Conner` interface

## Why

I need this change to allow my CoreDNS plugin to:
- ~~set the freebind socket option~~
- read the destination IP from OOB information when using freebind sockopts
- provide TCP RTT stats on linux by reading GetsockoptTCPInfo

If this gets accepted, I'll follow up with a PR in the CoreDNS repo for the socket control options

### DIFF
--- a/server.go
+++ b/server.go
@@ -59,6 +59,12 @@ type ResponseWriter interface {
 	Hijack()
 }
 
+// Conner exposes the underlying connection and UDP session metadata when available.
+type Conner interface {
+	Conn() net.Conn
+	Session() *Session
+}
+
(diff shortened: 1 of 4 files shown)


miekg:
this also won't happen, use and fix dns v2

PR #1690, opened 2025-12-02
MsgInvalidFunc: Make DefaultMsgInvalidFunc a variable.

Kentzo:
For parity with [DefaultMsgAcceptFunc](https://github.com/miekg/dns/blob/b39ef963336bc62108809fa7f08b734dd643b6fd/acceptfunc.go#L20) that is a variable.


miekg:
don't see why not

PR #1693, opened 2025-12-11
Fix EDNS0 flags and MBZ in String output

### LINKED ISSUE #1691: Msg.String incorrect output related to CO flag
When the [CO flag](https://pkg.go.dev/github.com/miekg/dns#OPT.Co) is set:

- Msg.String does not include the "co" flag unless DO is also set
- Msg.String includes an invalid Z value due to the mask not being updated when CO was implemented

Related PR: #1684

brianshea2:
Fixes #1691


miekg:
thank you

PR #1696, opened 2026-01-06
server: add MaxConcurrent to limit concurrent queries

cangming:
## Summary

  This PR adds the `MaxConcurrent` configuration option to the `Server` struct to prevent goroutine explosion under heavy load by limiting the number of concurrent query handlers.

  Add MaxConcurrent field to Server to prevent goroutine explosion under heavy load. When set to a positive value, it limits the number of concurrent query handlers for both TCP and UDP.

  When the limit is reached, new connections/packets are dropped and the optional OnDropped callback is invoked with the client address. MaxConcurrent=0 (default) means unlimited, maintaining backward compatibility.

  ## Motivation

  ### Problem

  Under heavy load conditions, DNS servers can experience a goroutine explosion where incoming requests create goroutines faster than the Go scheduler can process them. This occurs because:

  1. **Fast Request Arrival**: When a large number of DNS queries arrive simultaneously (e.g., DDoS attacks, traffic spikes, or slow handler logic), the server rapidly spawns new goroutines to handle each request
  2. **Scheduler Lag**: The Go scheduler may not immediately execute existing goroutines, causing them to queue up in the runnable state
  3. **Memory Exhaustion**: Each goroutine consumes memory (minimum ~2KB stack), and the accumulation can lead to excessive memory usage

  **Real-world Impact**: In production environments, we have observed scenarios where **tens of thousands of goroutines accumulate within seconds**, causing:
  - Memory consumption spikes (potentially reaching GBs)
  - Increased GC pressure and latency
  - Potential OOM kills in containerized environments
  - Server instability or crashes

  ### Solution

  This PR introduces `MaxConcurrent` to provide backpressure by:
  - Limiting concurrent handler goroutines to a configurable threshold
  - Dropping excess requests when the limit is reached (fail-fast)
  - Preventing unbounded goroutine growth and memory exhaustion
  - Maintaining predictable resource usage under extreme load

  The semaphore-based approach ensures that the server remains responsive and stable even when flooded with requests, gracefully shedding load instead of crashing.

  ### Use Cases

  This feature is particularly valuable for:
  - **Public DNS servers** exposed to the internet
  - **High-traffic DNS infrastructure** serving thousands of QPS
  - **Resource-constrained environments** (containers with memory limits)
  - **DDoS mitigation** scenarios requiring load shedding

  ## Implementation Details

  Implementation uses a buffered channel as a semaphore, acquired before spawning handler goroutines and released when handlers complete. This approach:
  - Applies to both TCP connections and UDP packets
  - When the limit is reached:
    - **TCP**: Incoming connections are closed immediately
    - **UDP**: Incoming packets are discarded
    - The `OnDropped` callback (if set) is invoked with the client address

  ## Changes

  ### New Fields
  - **`MaxConcurrent int`**: Maximum number of concurrent query handlers
    - `0` (default): Unlimited concurrency (backward compatible)
    - `> 0`: Limits the number of concurrent handlers
  - **`OnDropped func(net.Addr)`**: Optional callback invoked when a connection/packet is dropped due to the concurrency limit

  ### Example Usage
  ```go
  server := &dns.Server{
      Addr:          ":53",
      Net:           "udp",
      MaxConcurrent: 1000,
      OnDropped: func(addr net.Addr) {
          log.Printf("Dropped request from %s due to concurrency limit", addr)
      },
  }
 ```

  ## Testing

  Added comprehensive test coverage:
  - ✅ TestMaxConcurrentTCP: Verifies TCP concurrency limiting
  - ✅ TestMaxConcurrentUDP: Verifies UDP concurrency limiting
  - ✅ TestMaxConcurrentZero: Verifies unlimited mode (backward compatibility)
  - ✅ TestMaxConcurrentRelease: Verifies proper semaphore slot release
  - ✅ TestMaxConcurrentOnDroppedCallback: Verifies callback functionality

  All tests pass in Docker environment (golang:1.24).

  ## Backward Compatibility

  ✅ Fully backward compatible: The default value of MaxConcurrent=0 means unlimited concurrency, preserving existing behavior.

### DIFF
--- a/server.go
+++ b/server.go
@@ -246,6 +246,12 @@ type Server struct {
 	MsgAcceptFunc MsgAcceptFunc
 	// MsgInvalidFunc is optional, will be called if a message is received but cannot be parsed.
 	MsgInvalidFunc MsgInvalidFunc
+	// MaxConcurrent is the maximum number of concurrent queries.
+	// A value of 0 means no limit.
+	MaxConcurrent int
+	// OnDropped is called when a connection or packet is dropped due to the MaxConcurrent limit.
+	// The remote address of the client is passed to the function.
+	OnDropped func(net.Addr)
 
 	// Shutdown handling
 	lock     sync.RWMutex
@@ -473,6 +479,11 @@ func (srv *Server) serveTCP(l net.Listener) error {
 		close(srv.shutdown)
 	}()
 
+	var sem chan struct{}
+	if srv.MaxConcurrent > 0 {
+		sem = make(chan struct{}, srv.MaxConcurrent)
+	}
+
 	for srv.isStarted() {
 		rw, err := l.Accept()
 		if err != nil {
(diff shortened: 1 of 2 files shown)


cangming:
This PR applies to DNS v1 only. DNS v2's pipelined TCP request handling requires a different concurrency control approach and is out of scope for this change. Since v1 remains the primary version in production use, I'm addressing the goroutine explosion issue in v1 first.

PR #1705, opened 2026-02-13
Reduce allocations in pack/unpack hot path

alexekdahl:
Packing, measuring, or unpacking messages in tight loops, 
like during mDNS service discovery, was creating 
heap allocations on every call that quickly turn into garbage. 

This change adds sync.Pools for compression maps in 
Pack and Len and for the scratch buffer in UnpackDomainName, 
pre-allocates RR and Question slices, and switches string concatenation in 
String methods to strings.Builder. 

Together, these changes remove most temporary allocations 
on the core pack/unpack path without affecting the public API or wire format.

| Benchmark                              | Before (B/op) | Before (allocs) | After (B/op) | After (allocs) |
|----------------------------------------|---------------|-----------------|--------------|----------------|
| PackMsg (compressed)                    | 256           | 2               | 0            | 0              |
| PackMsgMassive (compressed, 130 RRs)   | 20,296        | 14              | 6,787        | 1              |
| MsgLength (compressed)                  | 256           | 2               | 0            | 0              |
| UnpackMsg                               | 568           | 11              | 568          | 11             |
| UnpackDomainName                        | 64            | 1               | 64           | 1              |

### DIFF
--- a/msg.go
+++ b/msg.go
@@ -17,6 +17,7 @@ import (
 	"math/big"
 	"strconv"
 	"strings"
+	"sync"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


alexekdahl:
I didn’t notice that there was a v2 on Codeberg. I’ll try that out first.

miekg:
thanks for your efforts, but doing this in v2 is even *more* appreciated.

Having touched this code there... it is tricky and I believe it can be way faster, see https://codeberg.org/miekg/dns/issues/705

I'm not going to merge this here, because a) tricky code and I *DO NOT* want to deal with any fallout months from now b) v2 exists and much bigger/better changes are allowed there.

PR #1716, opened 2026-04-29
TXT: Fix incorrect estimated length of empty dns:"txt" fields

Kentzo:
As is `dns.Len` can incorrectly estimate length of RRs with dns:"txt" fields, such as TXT, because it does not take into account that `packTxt` writes `\x00' byte for empty strings:

https://github.com/miekg/dns/blob/d1539a788a12830620381c4cc6617762994f3fa1/msg.go#L455-L460

### DIFF
--- a/msg.go
+++ b/msg.go
@@ -457,7 +457,7 @@ func packTxt(txt []string, msg []byte, offset int) (int, error) {
 			return offset, ErrBuf
 		}
 		msg[offset] = 0
-		return offset, nil
+		return offset+1, nil
 	}
 	var err error
 	for _, s := range txt {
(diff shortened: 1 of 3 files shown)


Kentzo:
[RFC 1035, Section 3.3.14](https://datatracker.ietf.org/doc/html/rfc1035#section-3.3.14): TXT-DATA: One or more character-strings.

OTOH parsers probably don't care as they should not read more than RDLENGTH bytes anyway.

To work around `dns.Len` underestimating length of buffer to store empty TXT-like record I use the following:

```go
func rrLen(rr dns.RR) (l int) {
	l = dns.Len(rr)
	switch rr.Header().Rrtype {
	case dns.TypeTXT, dns.TypeSPF, dns.TypeAVC, dns.TypeNINFO, dns.TypeRESINFO:
		l += 1
	}
	return l
}
```

PR #1719, opened 2026-05-13
msg: return *ANY for §2.5.2 delete-RRset records on unpack

johanix:
## Bug

RFC 2136 §2.5.2 "delete RRset" placeholders (`CLASS=ANY`, `Rdlength=0`) currently unpack to a typed RR with zero-valued fields. Per-type `pack()` methods then re-serialize the fixed-size scalar fields (e.g. `*DS.pack` writes 4 zero bytes), so unpack→pack is not byte-identical and SIG(0) verification of inbound UPDATE messages fails with `"dns: bad signature"`.

## Fix

`UnpackRRWithHeader` now returns `*ANY` for `CLASS=ANY` + `Rdlength=0` (excluding `TYPE=ANY`, already handled via `TypeToRR[TypeANY]`). Matches what `Msg.RemoveRRset` produces and makes the round-trip symmetric. No changes to any per-type `pack()` method.

## Tests

New `update_delete_test.go` covers wire-format, pack/unpack stability, and SIG(0)-signed sign→pack→unpack→verify round-trips for both §2.5.2 (delete-RRset) and §2.5.4 (delete-RR) variants. `TestRemoveRRsetDSSIG0RoundTrip` exercises the previously-broken DS case; the existing `TestRemoveRRsetSIG0RoundTrip` uses `TypeA` and was bug-immune.

## Compatibility

No API changes. No new dependencies. Builds on supported Go versions. The fix only changes the unpack result for messages produced by `Msg.RemoveRRset` — those were already round-trip-broken, so no working caller's behavior changes.

### DIFF
--- a/msg.go
+++ b/msg.go
@@ -617,6 +617,18 @@ func UnpackRR(msg []byte, off int) (rr RR, off1 int, err error) {
 // UnpackRRWithHeader unpacks the record type specific payload given an existing
 // RR_Header.
 func UnpackRRWithHeader(h RR_Header, msg []byte, off int) (rr RR, off1 int, err error) {
+	// RFC 2136 §2.5.2 "delete RRset" placeholders carry CLASS=ANY with no
+	// rdata (Rdlength=0). The wire form has zero rdata bytes, but most
+	// per-type pack() methods unconditionally serialize their fixed-size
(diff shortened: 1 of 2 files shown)


miekg:
I would rather like to learn if this is _also_ a problem in codeberg.org/miekg/dns and if so, if there is a better fix than a `if` in the super hot path

